PR #16, opened 2025-12-05
Maestro CLI for Playbooks

pedramamini:
- **feat: Add CLI tool and Process Monitor improvements**
- **fix: Improve GitHub CLI UX in worktree configuration**
- **# CHANGES**
- **# CHANGES**
- **I'd be happy to help you create a clean update summary for your GitHub project! However, I don't see any input provided after "INPUT:" in your message.**
- **I'd be happy to help you create a clean update summary for your GitHub project! However, I notice that you haven't provided the actual input content after "INPUT:" at the end of your message.**
- **# CHANGES**

### DIFF
--- a/src/renderer/components/AchievementCard.tsx
+++ b/src/renderer/components/AchievementCard.tsx
@@ -369,100 +369,6 @@ export function AchievementCard({ theme, autoRunStats, globalStats, onEscapeWith
     return count.toString();
   };
 
-  // Draw the maestro conductor silhouette
-  const drawMaestroSilhouette = (ctx: CanvasRenderingContext2D, x: number, y: number, size: number) => {
-    ctx.save();
-    ctx.translate(x, y);
-    const scale = size / 512;
-    ctx.scale(scale, scale);
-
-    ctx.fillStyle = '#FFFFFF';
-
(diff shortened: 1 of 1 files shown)


pedramamini:
@claude why are we doing this?

Command Injection Risk in agent-spawner.ts:61-62

we have the exact path from the settings, either auto detected or manually overridden:

image

so we should be using that exact path.
---
next, on Inefficient Polling
ProcessMonitor.tsx:152: Polling every 2 seconds for process updates is inefficient. Consider using IPC events or increasing the interval.
this is only happening when the system process viewer is up no? in which case, no worries
---

next, on Large Dependency
@yao-pkg/pkg adds 200+ npm packages. Consider alternatives like esbuild for smaller bundle sizes.
why did we add this in the first place? why do we need it?

pedramamini:
@claude when you say "Verdict: This is the right choice for shipping a standalone CLI binary with the desktop app. The 200+ packages are only in your node_modules during development/build - end users never see them."  i am seeing the maestro.app is now >1GB whereas before it was <500MB where's all the bloat coming from if not this package? what is esbuild? should we switch to that?

PR #64, opened 2025-12-22
feat: worktree auto-discovery and custom Claude command support

petersilberman:
Worktree Features:
- Auto-detect directories containing git worktrees and create grouped agents
- Periodic scanner (30s) discovers new worktrees in watched directories
- Track removed worktree paths to prevent re-discovery
- Bulk delete worktree groups with all agents (trash icon on 🌳 groups)
- Wire SessionList delete through App.tsx for proper cleanup

Custom Claude Commands:
- Discover custom commands from .claude/commands/ directories
- Merge custom commands with built-in agent commands
- Run both discovery methods in parallel for faster loading

### DIFF
--- a/src/main/ipc/handlers/git.ts
+++ b/src/main/ipc/handlers/git.ts
@@ -574,5 +574,142 @@ export function registerGitHandlers(): void {
     }
   ));
 
+  // List all worktrees for a git repository
+  ipcMain.handle('git:listWorktrees', createIpcHandler(
+    handlerOpts('listWorktrees'),
+    async (cwd: string) => {
+      // Run git worktree list --porcelain for machine-readable output
+      const result = await execFileNoThrow('git', ['worktree', 'list', '--porcelain'], cwd);
+      if (result.exitCode !== 0) {
+        // Not a git repo or no worktree support
+        return { worktrees: [] };
+      }
+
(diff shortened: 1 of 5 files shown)


pedramamini:
@petersilberman can you drop some screenshots of what this looks like? are you making an agent per work tree so you can work on the same project, in parallel, on different branches?

pedramamini:
@petersilberman this is awesome. i have an idea for improved UX though. how about when you add a directory that is a Git repository, you get an additional box to specify the path where worktrees are kept or are to be kept. then we add the ability to create / remove work trees that will appear as items underneath the parent agent. the CLI args, the ENV vars, the nudge message, the auto run doc folder will all be inherited from the parent agent, though it can be customized from there.

then i'll drop worktree component from auto run, since it'll be easy to just launch an auto run now from one of your work trees. we can extend the right-click menu for the work trees to include a "create PR" then let the user choose the branch to PR too from there. thoughts on this experience? 

i'll also refactor this code "Periodic scanner (30s) discovers new worktrees in watched directories" to use the cross-platform file system watch module i've got in other places.

pedramamini:
I'm merging this in and then going through some refactoring. Going to tap you on the shoulder to test the beta.

PR #78, opened 2025-12-23
fix: resolve copy-on-drag not working in Auto Run dialog

oliveiraantoniocc:
## Summary

Fixes the drag-and-drop copy feature (hold Cmd/Ctrl while dragging) in the Auto Run dialog's Documents panel, which was not working reliably.

## Problem

The copy-on-drag functionality was broken due to several interacting issues:

1. **Race condition in handleDragLeave**: The handler was clearing \`dropTargetIndex\` before the drop could occur, causing the drop handler to find null state
2. **Unreliable browser drop event**: The browser's \`drop\` event doesn't always fire reliably (e.g., when cursor is slightly outside container boundaries)
3. **Incorrect position validation**: \`handleDragOver\` wasn't properly validating positions for move vs copy mode

## Solution

- **Remove state clearing from handleDragLeave**: Cleanup now only happens in \`handleDrop\` and \`handleDragEnd\`
- **Add \`dropPerformedRef\`**: Prevents double execution if both \`handleDrop\` and \`handleDragEnd\` fire
- **Add fallback logic in handleDragEnd**: Performs the operation if \`handleDrop\` never fired (browser quirk workaround)
- **Improve handleDragOver**: Properly validates positions - always shows indicator for copy mode, only shows for move mode if position would actually change

## Testing

1. Open Auto Run dialog with multiple documents
2. Drag a document while holding Cmd (macOS) or Ctrl (Windows)
3. Verify the green \"+\" icon appears during copy drag
4. Drop the document at a new position
5. Verify a copy is created at the drop location
6. Verify regular drag (without modifier key) still moves documents correctly

### DIFF
--- a/src/renderer/components/DocumentsPanel.tsx
+++ b/src/renderer/components/DocumentsPanel.tsx
@@ -536,6 +536,7 @@ export function DocumentsPanel({
   const draggedIdRef = useRef(draggedId);
   const dropTargetIndexRef = useRef(dropTargetIndex);
   const isCopyDragRef = useRef(isCopyDrag);
+  const dropPerformedRef = useRef(false); // Prevents double execution if both handleDrop and handleDragEnd fire
   draggedIdRef.current = draggedId;
   dropTargetIndexRef.current = dropTargetIndex;
   isCopyDragRef.current = isCopyDrag;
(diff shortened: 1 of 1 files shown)


pedramamini:
@oliveiraantoniocc just FYI there was still one nuance left that i took care of, the original doc has to get the reset state toggled on as well.

image

all good now

PR #98, opened 2025-12-26
feat: adds the ability to close multiple tabs at once

denniswebb:
Just a minor quality-of-life addition. I found myself wanting to close other tabs.

Added logic so it only closes "visible" tabs. If you have "Unreads Only" active, then it will not close the hidden tabs that are considered "Read".

Far right tab doesn't show "Close Tabs to the Right"
image

Center tab shows all options
image

Left-most tab doesn't show "Close Tabs to the Left"
image

CMD+K works also and shows keyboard shortcuts
image

### DIFF
--- a/src/renderer/components/TabBar.tsx
+++ b/src/renderer/components/TabBar.tsx
@@ -670,6 +670,16 @@ export function TabBar({
   // Can always close tabs - closing the last one creates a fresh new tab
   const canClose = true;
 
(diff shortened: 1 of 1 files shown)


pedramamini:
woooot love it! thank you for the contribution!

pedramamini:
Nice you even got this: "Visible Tabs Only - When "Unreads Only" filter is active, close operations only affect visible tabs (hidden read tabs are preserved)"

PR #119, opened 2025-12-30
fix: detect Node version manager paths in process spawning

t1mmen:
## The Problem

Tried launching an agent and got this:

CleanShot 2025-12-29 at 20 00 14@2x

`env: node: No such file or directory` - exit code 127.

Took a while to figure out what was happening. Turns out macOS GUI apps (Electron included) don't inherit your shell's PATH. They're launched by launchd with a bare-bones environment: `/usr/bin:/bin:/usr/sbin:/sbin`. My node lives in `~/.nvm/versions/node/v22.21.0/bin` - nowhere close.

## The Fix

Detect version manager installations at spawn time and prepend their bin paths. Filesystem checks only, no shell spawning.

Supports:
- **nvm** - resolves default alias from `~/.nvm`
- **fnm** - `~/Library/Application Support/fnm` (macOS) or `~/.local/share/fnm` (Linux)
- **volta** - `~/.volta/bin`
- **mise** - `~/.local/share/mise/shims`
- **asdf** - `~/.asdf/shims`
- **n** - `/usr/local/bin`

---

PS: Thank you for building this, looks awesome, looking forward to trying it now that it works for me :)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "maestro",
-  "version": "0.13.1",
+  "version": "0.14.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "maestro",
-      "version": "0.13.1",
+      "version": "0.14.0",
       "hasInstallScript": true,
       "license": "AGPL 3.0",
       "dependencies": {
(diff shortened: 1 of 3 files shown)


pedramamini:
robustness for the win! thank you for these contributions! will go into 0.14.0 that i'm working on getting out the door now.

PR #140, opened 2026-01-02
ios simulation

mattjay:


mattjay:
apologies - was testing the "create PR" button for another work tree and didn't realize this would open back to upstream.

PR #152, opened 2026-01-06
Honor user's configured default shell instead of hardcoding zsh

### LINKED ISSUE #151: Maestro does not honor shell selection in MacOS
Maestro hardcodes `zsh` as the default interactive shell, even when `$SHELL` and `dscl . -read /Users/$USER UserShell` both agree that the user's configured shell is bash.

Per https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html, `$SHELL` reflects "the user's preferred command language interpreter". Maestro should honor this configuration.

charles-dyfis-net:
Fixes #151.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -174,7 +174,23 @@ const store = new Store<MaestroSettings>({
     fontFamily: 'Roboto Mono, Menlo, "Courier New", monospace',
     customFonts: [],
     logLevel: 'info',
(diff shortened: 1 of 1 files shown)


pedramamini:
Awesome! Thanks for the contribution. I just pushed out 0.14.1 but will get this in for 0.14.2!

Issue #185, opened 2026-01-12
Would it be possible to add Bluesky to the list of items possible to register when signing in to the leaderboard?


jlengrand:
Bluesky is used a lot as well in the Tech community, and I believe adding it should be a rather small effort? 

I can see if Claude generates a good PR for the client side, but you'll need to make some changes on the server side probably.

pedramamini:
You're the man! Thanks for the excellent PR. I'm adding server side support and will get this into the next release!

PR #186, opened 2026-01-12
Adding Bluesky to the list of social networks on the client side

jlengrand:
See #185


jlengrand:
image

pedramamini:
Nice add. I need to add server side support as well. Will get this in for the next release!

PR #189, opened 2026-01-13
minor fixes so Electron plays nicer on Windows

matthargett:
* give the main window a name that Windows understands 
  * so that multiple instance aren't launched
  * taskbar grouping works
 
* eliminate the superfluous toolbar menu space. minor tweaks to align with Electron best practices on Windows from Slack, Discord, VS Code, etc. 
 
maybe next: inlining the hamburger menu into the window chrome requires deeper changes, I can do that next if reclaiming the ~100px height is useful (or we just like the modern look).

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -682,7 +682,14 @@ function createWindow() {
     minWidth: 1000,
     minHeight: 600,
     backgroundColor: '#0b0b0d',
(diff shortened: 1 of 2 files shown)


pedramamini:
@matthargett sorry it's taken forever to get to this PR. I'm getting Maestro for Windows setup this weekend and can get hands on testing.

PR #196, opened 2026-01-15
perf: Buffered process data events to reduce IPC frequency

reachrazamair:


pedramamini:
| Priority | Issue | Status in PR |
|----------|-------|--------------|
| 🔴 Must Fix | Timer not cleared in `kill()` method | **Not addressed** |
| 🔴 Must Fix | Timer not cleared in `killAll()` method | **Not addressed** |
| 🟡 Should Fix | No test coverage for buffering behavior | **Not addressed** |
| 🟡 Should Fix | No error handling in `flushDataBuffer()` emit | **Not addressed** |

### Detailed Issues

**1. Missing Timer Cleanup in `kill()`** (Lines ~1960-1989)

The `kill()` method doesn't clear the `dataBufferTimeout` before deleting the process:

```typescript
kill(sessionId: string): boolean {
  const process = this.processes.get(sessionId);
  if (!process) return false;

  try {
    // ⚠️ Missing: clearTimeout(process.dataBufferTimeout)
    if (process.isTerminal && process.ptyProcess) {
      process.ptyProcess.kill();
    } else if (process.childProcess) {
      process.childProcess.kill('SIGTERM');
    }
    this.processes.delete(sessionId);  // Timer callback will fire but process gone
    return true;
  }
}
```

**Impact:** When the timeout fires after process deletion, `flushDataBuffer()` will safely exit (process not found), but this leaves a dangling timer reference. Not critical, but sloppy.

### Recommendations

**Before Merge:**
1. Add timer cleanup in `kill()` and `killAll()`
2. Add unit tests for buffering behavior
3. Consider splitting formatting changes into separate PR (or using eslint/prettier autofix)

**Nice to Have:**
- Add warning log when `emitDataBuffered` is called for non-existent process
- Wrap emit in try-catch in `flushDataBuffer()`
- Add JSDoc documenting the buffering behavior

reachrazamair:
**Buffer flush on process exit** - Already implemented in PTY exit handler (line 627) and child process exit handler (line 1488)

**Timeout cleanup in kill()** - Already implemented (lines 1977-1980), with flush before kill

**Max buffer size check** - Already using 8KB limit (line 1791) which forces immediate flush for large chunks. This is more conservative than the suggested 500KB and provides better latency.

**Try-catch in flushDataBuffer()** - Already wrapped with error logging (lines 1816-1827)

**Test indentation** - Prettier normalized to project settings

The buffering reduces IPC events significantly especially with 3+ agents are running, while maintaining < 50ms latency for output display.

PR #225, opened 2026-01-22
Add Factory.ai droid support

VVX7:
I hacked droid support together this morning after posting this https://github.com/pedramamini/Maestro/issues/223 and realizing I can in fact just vibe code it hope it doesn't fail is spectacular ways.

I'll be using it pretty heavily so I'll follow up with bugs or whatever as I find them.

### DIFF
--- a/src/main/parsers/factory-droid-output-parser.ts
+++ b/src/main/parsers/factory-droid-output-parser.ts
@@ -1,0 +1,337 @@
+/**
+ * Factory Droid Output Parser
+ *
+ * Parses JSON output from Factory Droid CLI (`droid exec -o stream-json`).
+ *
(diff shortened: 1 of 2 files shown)


pedramamini:
Wow incredible! Thanks for the contribution. I'm merging your branch now into an RC branch I am testing. Will see how it looks though I won't be able to test it directly.

VVX7:
Droid lets you use a providers API token so you should be able to drop that in and use claude or whatever, though it will count against your API usage rather than pro/max plan.

In the meantime let me know if there's something that needs testing.

pedramamini:
Oh wow thanks for the heads up. In that case I will give it a try as well.

PR #273, opened 2026-02-01
fix(build): extract inline CSS to fix Vite html-inline-proxy error

greg12579:
## Summary
- Fixes Vite production build failure caused by inline `<style>` blocks in HTML files
- Extracts inline CSS to separate files that Vite can process correctly
- Adds `src/renderer/splash.css` for desktop splash screen styles
- Adds `src/web/critical.css` for web interface loading styles

## Problem
Vite's `html-inline-proxy` plugin fails during production builds when HTML files contain inline `<style>` blocks, resulting in build errors.

## Solution
Move the inline CSS to external stylesheet files and reference them via `<link>` tags. This maintains the same styling behavior while allowing Vite to process the files correctly.

## Test plan
- [x] Run `npm run build` - production build completes successfully
- [ ] Verify splash screen displays correctly on desktop
- [ ] Verify loading spinner displays correctly on web interface


pedramamini:
Hi @greg12579, thank you for taking the time to contribute to Maestro!

After reviewing this PR, we tested the current production build on the `0.15.0-rc` branch and it completes successfully with the inline `<style>` blocks. We weren't able to reproduce the `html-inline-proxy` error you mentioned.

The inline CSS in these files is intentionally placed there as **critical CSS** to prevent Flash of Unstyled Content (FOUC). The comment in the existing code notes this explicitly:

```html
<!-- Critical CSS for splash screen - inline to prevent FOUC -->
```

Moving these styles to external files would introduce a potential race condition where the browser might render the HTML before fetching and parsing the external CSS, causing a flash of unstyled or incorrectly styled content during app startup.

Since the build works as-is and the proposed change could introduce a UX regression, we're going to close this PR for now.

**However**, if you're experiencing an actual build failure that we haven't been able to reproduce, we'd love to hear more details:
- The exact error message you're seeing
- Your Vite version (`npm list vite`)
- Your Node.js version (`node --version`)
- Any local configuration differences

Please feel free to reopen this PR with additional context, and we'll be happy to investigate further. Thanks again for your interest in improving Maestro!

greg12579:
Did you test on windows?

PR #274, opened 2026-02-01
fix(windows): use stdin for long prompts to avoid command line limit

greg12579:
## Summary
- Fixes agent crash on Windows when prompts exceed ~8191 character command line limit
- Prompts over 7000 characters are now passed via stdin instead of CLI argument
- Maintains compatibility with Claude Code's `--print` mode which accepts stdin input

## Problem
On Windows, the command line has a length limit of ~8191 characters. When the wizard builds up conversation history, the prompt can exceed this limit, causing the agent process to fail immediately with exit code 1 and no output.

**Symptoms:**
- "Agent exited with code 1" error on second or subsequent wizard messages
- Process spawns but exits within ~30ms
- No error output from the agent itself

## Solution
Detect when a prompt exceeds 7000 characters (leaving margin for command and args) on Windows, and instead of passing it as a CLI argument, write it to the process's stdin. Claude Code's `--print` mode accepts prompts via stdin, so this maintains compatibility.

## Test plan
- [x] First wizard message works (prompt under limit)
- [x] Second wizard message works (prompt over limit, uses stdin)
- [ ] Verify other agents still work normally
- [ ] Test on macOS/Linux (should be unaffected)

### DIFF
--- a/src/main/process-manager/spawners/ChildProcessSpawner.ts
+++ b/src/main/process-manager/spawners/ChildProcessSpawner.ts
@@ -68,6 +68,11 @@ export class ChildProcessSpawner {
 		const hasImages = images && images.length > 0;
 		const capabilities = getAgentCapabilities(toolType);
 
+		// Windows command line limit is ~8191 chars. Use stdin for long prompts to avoid hitting this.
+		const WINDOWS_CMD_LIMIT = 7000; // Leave margin for command and args
+		const useStdinForPrompt =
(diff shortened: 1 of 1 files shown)


pedramamini:
Thanks for the PR! Much appreciated. We've been rapid firing making changes in this area and it seems we addressed the issue before we could get to your PR. Closing it out but hope you keep helping out because we do need it. Are you on the Discord? We have a Windows specific channel and `chrisyy` on Discord is a good person to collaborate with.

PR #302, opened 2026-02-05
Fix Windows wizard failure by sending prompt via stdin for stream-json mode

### LINKED ISSUE #301: 0,15,0 RC  fails when asking for my idea in wizard mode.
[wizard-debug-2026-02-05T03-01-19-598Z.json](https://github.com/user-attachments/files/25085949/wizard-debug-2026-02-05T03-01-19-598Z.json)

dmaynor:
## Summary
- Fix Windows wizard failure caused by command line length exceeding cmd.exe's ~8191 character limit
- When using `stream-json` output mode, the prompt is now sent only via stdin (not duplicated as CLI arg)

## Root Cause
On Windows with `stream-json` mode, the prompt was being added both as a CLI argument AND sent via stdin. This caused the command line to exceed Windows' limit, resulting in immediate exit code 1.

## Test Plan
- [x] Verified build succeeds
- [x] Verified 18,179 tests pass (18 failures are pre-existing Windows path separator issues)
- [ ] Manual test on Windows with wizard mode

Fixes #301

### DIFF
--- a/src/main/process-manager/spawners/ChildProcessSpawner.ts
+++ b/src/main/process-manager/spawners/ChildProcessSpawner.ts
@@ -73,7 +73,11 @@ export class ChildProcessSpawner {
 
 		// Check if prompt will be sent via stdin instead of command line
 		// This is critical for SSH remote execution to avoid shell escaping issues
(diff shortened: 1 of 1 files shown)


pedramamini:
Thanks for the fix!

PR #328, opened 2026-02-08
Windows Shell Handling and Group Chat Fixes

chr1syy:
This pull request introduces comprehensive improvements to Windows shell handling and process spawning logic, especially for group chat participants and moderator agents. The main focus is to ensure robust, PowerShell-based execution on Windows to avoid command line length issues and shell parsing problems, and to improve test coverage for these scenarios. It also adds extensive tests for new and existing shell utilities.

**Windows Shell Handling and Process Spawning:**

* Refactored group chat participant and moderator agent spawning to use PowerShell (via `getWindowsShellForAgentExecution`) instead of `cmd.exe` on Windows, preventing command line length errors and improving reliability. This includes support for custom shell paths and fallback logic. [[1]](diffhunk://#diff-bd75d59204561f69f90d53c84e8364a59c19eac8fdeaff2f2ec8d9e54ca6df18R31-R50) [[2]](diffhunk://#diff-bd75d59204561f69f90d53c84e8364a59c19eac8fdeaff2f2ec8d9e54ca6df18R196-R197) [[3]](diffhunk://#diff-bd75d59204561f69f90d53c84e8364a59c19eac8fdeaff2f2ec8d9e54ca6df18R224-R237) [[4]](diffhunk://#diff-bd75d59204561f69f90d53c84e8364a59c19eac8fdeaff2f2ec8d9e54ca6df18R252-R254) [[5]](diffhunk://#diff-b668781c3ce607142560e8df19d6130fc3d64eb5f0166a733c8194bf63bb4613R34-R39) [[6]](diffhunk://#diff-3026933d97a1c2fc41388d510ab07ac51d7de42d7fe125f35462beba100edb10R44) [[7]](diffhunk://#diff-3026933d97a1c2fc41388d510ab07ac51d7de42d7fe125f35462beba100edb10R91-R93)
* Added support for sending prompts via stdin as raw text (`sendPromptViaStdinRaw`) on Windows to avoid PowerShell command parsing issues, and updated process manager interfaces accordingly. [[1]](diffhunk://#diff-bd75d59204561f69f90d53c84e8364a59c19eac8fdeaff2f2ec8d9e54ca6df18R224-R237) [[2]](diffhunk://#diff-bd75d59204561f69f90d53c84e8364a59c19eac8fdeaff2f2ec8d9e54ca6df18R252-R254) [[3]](diffhunk://#diff-b668781c3ce607142560e8df19d6130fc3d64eb5f0166a733c8194bf63bb4613R34-R39)

**Testing Improvements:**

* Added extensive tests for Windows stdin handling in `inlineWizardConversation`, covering different agent types, platform detection, and shell argument handling.
* Added and expanded tests for shell escaping utilities, including `getWindowsShellForAgentExecution` and `escapePowerShellPromptContent`, ensuring correct behavior for various shell and prompt scenarios. [[1]](diffhunk://#diff-ebe0b2deb92c18a9f6758492046204116ef2b7eb817082235c6efcb25e12ffb5R9-R10) [[2]](diffhunk://#diff-ebe0b2deb92c18a9f6758492046204116ef2b7eb817082235c6efcb25e12ffb5L101-R105) [[3]](diffhunk://#diff-ebe0b2deb92c18a9f6758492046204116ef2b7eb817082235c6efcb25e12ffb5R142-R392)

**Bug Fixes and Clarifications:**

* Fixed and clarified test logic for `ChildProcessSpawner` to ensure `isStreamJsonMode` is not enabled when `sendPromptViaStdinRaw` is true, reflecting the intended behavior for raw stdin input. [[1]](diffhunk://#diff-f664ca02985257116dfd0ad632ee5e27a88a584ae846df118fd661016b4edf7bL189-R193) [[2]](diffhunk://#diff-f664ca02985257116dfd0ad632ee5e27a88a584ae846df118fd661016b4edf7bL201-R203)

These changes significantly improve reliability and maintainability of process spawning and shell interaction logic, especially on Windows platforms.

### DIFF
--- a/src/__tests__/main/group-chat/group-chat-config.test.ts
+++ b/src/__tests__/main/group-chat/group-chat-config.test.ts
@@ -1,0 +1,241 @@
+/**
+ * Tests for src/main/group-chat/group-chat-config.ts
+ *
+ * Tests the getWindowsSpawnConfig helper which consolidates Windows shell
+ * and stdin mode configuration for group chat agent execution.
+ */
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+
+// Mock the agent capabilities before importing
+vi.mock('../../../main/agents', () => ({
+	getAgentCapabilities: vi.fn(),
+}));
+
+// Mock the shell escape utility
+vi.mock('../../../main/process-manager/utils/shellEscape', () => ({
+	getWindowsShellForAgentExecution: vi.fn(() => ({
+		shell: 'powershell.exe',
+		useShell: true,
+		source: 'powershell-default',
+	})),
+}));
+
+import {
+	getWindowsSpawnConfig,
+	setGetCustomShellPathCallback,
+	getCustomShellPath,
+	type SpawnSshConfig,
+} from '../../../main/group-chat/group-chat-config';
+import { getAgentCapabilities } from '../../../main/agents';
+import { getWindowsShellForAgentExecution } from '../../../main/process-manager/utils/shellEscape';
+
+describe('group-chat-config', () => {
+	let originalPlatform: NodeJS.Platform;
+
+	beforeEach(() => {
+		vi.clearAllMocks();
+		// Save original platform
+		originalPlatform = process.platform;
+	});
+
+	afterEach(() => {
+		// Restore original platform
+		Object.defineProperty(process, 'platform', { value: originalPlatform });
+		// Clear the callback
+		setGetCustomShellPathCallback(() => undefined);
+	});
+
+	describe('getCustomShellPath', () => {
+		it('should return undefined when no callback is registered', () => {
+			setGetCustomShellPathCallback(() => undefined);
+			expect(getCustomShellPath()).toBeUndefined();
+		});
+
+		it('should return the value from the registered callback', () => {
+			setGetCustomShellPathCallback(() => 'C:\\Custom\\PowerShell.exe');
+			expect(getCustomShellPath()).toBe('C:\\Custom\\PowerShell.exe');
+		});
+	});
+
+	describe('getWindowsSpawnConfig', () => {
+		describe('on Windows', () => {
+			beforeEach(() => {
+				// Mock Windows platform
+				Object.defineProperty(process, 'platform', { value: 'win32' });
+			});
+
+			it('should return shell config for stream-json agent on Windows', () => {
+				vi.mocked(getAgentCapabilities).mockReturnValue({
+					supportsStreamJsonInput: true,
+				} as any);
+
+				const result = getWindowsSpawnConfig('claude-code');
+
+				expect(result.shell).toBe('powershell.exe');
+				expect(result.runInShell).toBe(true);
+				expect(result.sendPromptViaStdin).toBe(true);
+				expect(result.sendPromptViaStdinRaw).toBe(false);
+			});
+
+			it('should return shell config for non-stream-json agent on Windows', () => {
+				vi.mocked(getAgentCapabilities).mockReturnValue({
+					supportsStreamJsonInput: false,
+				} as any);
+
+				const result = getWindowsSpawnConfig('opencode');
+
+				expect(result.shell).toBe('powershell.exe');
+				expect(result.runInShell).toBe(true);
+				expect(result.sendPromptViaStdin).toBe(false);
+				expect(result.sendPromptViaStdinRaw).toBe(true);
+			});
+
+			it('should NOT apply Windows config when SSH is enabled', () => {
+				vi.mocked(getAgentCapabilities).mockReturnValue({
+					supportsStreamJsonInput: true,
+				} as any);
+
+				const sshConfig: SpawnSshConfig = {
+					enabled: true,
+					remoteId: 'my-remote',
+				};
+
+				const result = getWindowsSpawnConfig('claude-code', sshConfig);
+
+				expect(result.shell).toBeUndefined();
+				expect(result.runInShell).toBe(false);
+				expect(result.sendPromptViaStdin).toBe(false);
+				expect(result.sendPromptViaStdinRaw).toBe(false);
+				// Should NOT call getWindowsShellForAgentExecution when SSH is enabled
+				expect(getWindowsShellForAgentExecution).not.toHaveBeenCalled();
+			});
+
+			it('should apply Windows config when SSH is disabled', () => {
+				vi.mocked(getAgentCapabilities).mockReturnValue({
+					supportsStreamJsonInput: true,
+				} as any);
+
+				const sshConfig: SpawnSshConfig = {
+					enabled: false,
+					remoteId: null,
+				};
+
+				const result = getWindowsSpawnConfig('claude-code', sshConfig);
+
+				expect(result.shell).toBe('powershell.exe');
+				expect(result.runInShell).toBe(true);
+				expect(getWindowsShellForAgentExecution).toHaveBeenCalled();
+			});
+
+			it('should apply Windows config when sshConfig is undefined', () => {
+				vi.mocked(getAgentCapabilities).mockReturnValue({
+					supportsStreamJsonInput: false,
+				} as any);
+
+				const result = getWindowsSpawnConfig('opencode', undefined);
+
+				expect(result.shell).toBe('powershell.exe');
+				expect(result.runInShell).toBe(true);
+				expect(result.sendPromptViaStdinRaw).toBe(true);
+			});
+
+			it('should use custom shell path from callback', () => {
+				setGetCustomShellPathCallback(() => 'C:\\Custom\\pwsh.exe');
+				vi.mocked(getWindowsShellForAgentExecution).mockReturnValue({
+					shell: 'C:\\Custom\\pwsh.exe',
+					useShell: true,
+					source: 'custom',
+				});
+				vi.mocked(getAgentCapabilities).mockReturnValue({
+					supportsStreamJsonInput: true,
+				} as any);
+
+				const result = getWindowsSpawnConfig('claude-code');
+
+				expect(getWindowsShellForAgentExecution).toHaveBeenCalledWith({
+					customShellPath: 'C:\\Custom\\pwsh.exe',
+				});
+				expect(result.shell).toBe('C:\\Custom\\pwsh.exe');
+			});
+		});
+
+		describe('on non-Windows platforms', () => {
+			beforeEach(() => {
+				// Mock Linux platform
+				Object.defineProperty(process, 'platform', { value: 'linux' });
+			});
+
+			it('should return no-op config on Linux', () => {
+				vi.mocked(getAgentCapabilities).mockReturnValue({
+					supportsStreamJsonInput: true,
+				} as any);
+
+				const result = getWindowsSpawnConfig('claude-code');
+
+				expect(result.shell).toBeUndefined();
+				expect(result.runInShell).toBe(false);
+				expect(result.sendPromptViaStdin).toBe(false);
+				expect(result.sendPromptViaStdinRaw).toBe(false);
+				// Should NOT call getWindowsShellForAgentExecution on non-Windows
+				expect(getWindowsShellForAgentExecution).not.toHaveBeenCalled();
+			});
+
+			it('should return no-op config on macOS', () => {
+				Object.defineProperty(process, 'platform', { value: 'darwin' });
+				vi.mocked(getAgentCapabilities).mockReturnValue({
+					supportsStreamJsonInput: false,
+				} as any);
+
+				const result = getWindowsSpawnConfig('opencode');
+
+				expect(result.shell).toBeUndefined();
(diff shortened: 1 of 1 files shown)


src/main/process-manager/utils/shellEscape.ts
```diff
@@ -119,3 +119,120 @@ export function escapeArgsForShell(args: string[], shell?: string): string[] {
...
+
+/**
+ * Configuration options for Windows shell selection.
+ */
+export interface WindowsShellConfig {
+	/**
+	 * User-configured custom shell path (takes highest priority).
+	 * This is typically from settingsStore.get('customShellPath').
+	 */
+	customShellPath?: string;
```
pedramamini, line 131: **Concern: No-op function** — `escapePowerShellPromptContent` returns its input unchanged. The comment explains why no escaping is needed (stdin is treated as literal text), which makes sense. But if no escaping is needed, should this function and its call site in `ChildProcessSpawner.ts` (along with the conditional check + debug logging there) be removed entirely?

Keeping a no-op function with 19 lines of conditional + logging around it is confusing for future maintainers. The 145 lines of tests that all assert `input === output` compound this.

Alternative: if this is intentionally a placeholder for future escaping needs, a single-line comment at the call site would be clearer than a full function + test suite.

src/renderer/services/inlineWizardConversation.ts
```diff
@@ -588,22 +588,26 @@ export async function sendWizardMessage(
 		// Build args for the agent
 		const argsForSpawn = agent ? buildArgsForAgent(agent) : [];
 
-		// On Windows, use sendPromptViaStdin to bypass cmd.exe ~8KB command line length limit
+		// On Windows, use stdin to bypass cmd.exe ~8KB command line length limit
 		// Note: Use navigator.platform in renderer (process.platform is not available in browser context)
 		const isWindows = navigator.platform.toLowerCase().includes('win');
-		const sendViaStdin = isWindows;
+		// Agents that support --input-format stream-json use sendPromptViaStdin (JSON format)
```
pedramamini, line 595: **Concern: Hardcoded agent IDs instead of capabilities** — This checks `session.agentType === 'claude-code' || session.agentType === 'codex'` but the project already has `agent.capabilities.supportsStreamJsonInput` in the agent capabilities system (`agent-capabilities.ts`).

The same pattern appears in `useInputProcessing.ts`. Adding a new stream-json agent would require updating both renderer files instead of just the capabilities config.

Could this use the capabilities system instead? The agent info is already fetched a few lines above.

src/main/group-chat/group-chat-router.ts
```diff
@@ -512,8 +527,24 @@ ${message}`;
...
+					// causing: "Die Befehlszeile ist zu lang" (command line too long) errors.
+					// ALSO: Send prompt via stdin to avoid PowerShell parsing issues with
+					// markdown content (e.g., bullet points like "- If you..." get parsed as operators).
+					const shellConfig = getWindowsShellForAgentExecution({
+						customShellPath: getCustomShellPathCallback?.(),
+					});
+					spawnShell = shellConfig.shell;
+					spawnRunInShell = shellConfig.useShell;
+					console.log(`[GroupChat:Debug] Windows shell config: ${shellConfig.shell} (source: ${shellConfig.source})`);
 				}
```
pedramamini, line 524: **Concern: Raw stdin for ALL agents, including stream-json agents** — This unconditionally sets `sendPromptViaStdinRaw = true` for all agents on Windows. But Claude Code and Codex support `sendPromptViaStdin` (JSON stream-json mode), not raw mode.

Since the PR correctly removed `sendPromptViaStdinRaw` from the `isStreamJsonMode` detection in `ChildProcessSpawner.ts`, using raw mode for stream-json agents means:
- `isStreamJsonMode` won't be triggered
- The output parser won't parse JSON output
- Claude Code responses in group chat will display as raw JSON text on Windows

Should this check the agent type (or capabilities) and use `sendPromptViaStdin` for stream-json agents, matching the renderer logic?

Same pattern repeats at lines ~681, ~1233, ~1365.

src/main/process-manager/utils/shellEscape.ts
```diff
@@ -119,3 +119,120 @@ export function escapeArgsForShell(args: string[], shell?: string): string[] {
...
+ * actual command code being executed, not in text data being read from stdin.
+ *
+ * @param content - The prompt content to escape
+ * @returns The content unchanged (no escaping needed for stdin)
+ */
+export function escapePowerShellPromptContent(content: string): string {
+	// PowerShell stdin treats input as literal text, so no escaping is needed.
+	// Returning the content as-is ensures markdown formatting and other content
+	// is preserved exactly as intended.
+	return content;
```
pedramamini: **Minor concern: `.includes('cmd')` may be too broad** — This would reject a legitimate shell at a path like `C:\Users\commander\bash.exe` or `C:\Program Files\tcmd\shell.exe`.

Consider a more precise check:
```typescript
const basename = shellLower.split(/[\\\/]/).pop() || '';
const isCmdExe = basename === 'cmd' || basename === 'cmd.exe';
```

src/main/group-chat/group-chat-agent.ts
```diff
@@ -28,13 +28,26 @@ import {
...
+// Module-level callback for getting custom shell path from settings
+let getCustomShellPathCallback: (() => string | undefined) | null = null;
+
+/**
+ * Sets the callback for getting the custom shell path from settings.
+ * This is used on Windows to prefer PowerShell over cmd.exe to avoid command line length limits.
+ * Called from index.ts during initialization.
+ */
+export function setGetCustomShellPathCallback(callback: () => string | undefined): void {
+	getCustomShellPathCallback = callback;
```
pedramamini: **Nit: Duplicate pattern** — This exact module-level callback + setter is duplicated in `group-chat-router.ts` (lines ~91-99). Both are wired separately in `index.ts`. Could this be consolidated — either a shared utility or passed through function parameters?

src/main/process-listeners/exit-listener.ts
```diff
@@ -144,34 +144,12 @@ export function setupExitListener(
...
-										routeErr
-									);
-									logger.error(
-										'[GroupChat] Failed to route moderator response',
-										'ProcessListener',
-										{ error: String(routeErr) }
-									);
-								});
-						}
+						// Do NOT attempt to route the response if chat load fails
```
pedramamini: **Question: Is losing the fallback routing intentional?** — The old code attempted to parse the buffered output and route the moderator response even when `loadGroupChat` failed. The comment says this could cause duplicate responses, which is a valid concern.

But if `loadGroupChat` fails for a transient reason (e.g., disk I/O hiccup), the moderator's entire response is now silently dropped and the group chat stalls. Is there a middle ground — e.g., logging the parsed text for debugging, or retrying the chat load once?

chr1syy:
> ## PR Review — Windows Shell Handling & Group Chat Fixes
> Overall the approach is sound — the `sendPromptViaStdin` (JSON) vs `sendPromptViaStdinRaw` (raw text) distinction and the centralized `getWindowsShellForAgentExecution` are good ideas. A few concerns below that I'd like your feedback on before merge. See inline comments for specifics.
> 
> **High-priority:**
> 
> 1. `escapePowerShellPromptContent` is a no-op — should it be removed?
> 2. Hardcoded agent IDs in renderer instead of capabilities system
> 3. Group chat always uses raw stdin mode, even for stream-json agents
> 
> **Medium-priority:** 4. `.includes('cmd')` shell check may be too broad 5. Duplicate callback pattern across group-chat-agent and group-chat-router 6. Exit listener fallback routing removal

All review concerns got addressed - ready for 2nd review

pedramamini (commented):
## PR Re-Review — Windows Shell Handling & Group Chat Fixes

@chr1syy Thanks for addressing all 6 previous review concerns in `a85be1f1`. Here's the status on those plus new findings.

### Previous Feedback Status

| # | Concern | Status |
|---|---------|--------|
| 1 | No-op `escapePowerShellPromptContent` | ✅ Removed |
| 2 | Hardcoded agent IDs → capabilities | ✅ Now uses `supportsStreamJsonInput` |
| 3 | Raw stdin for ALL agents in group chat | ✅ Correctly splits `sendPromptViaStdin` vs `sendPromptViaStdinRaw` |
| 4 | `.includes('cmd')` too broad | ✅ Basename-only check with test |
| 5 | Duplicate callback pattern | ✅ `group-chat-config.ts` consolidates |
| 6 | Exit listener fallback routing | ✅ Retry + log-and-drop |

### 🚨 CRITICAL: `sendPromptViaStdinRaw` code path is dead

**File:** `ChildProcessSpawner.ts:451`

The PR removes `!!config.sendPromptViaStdinRaw` from `isStreamJsonMode` (correct semantically — raw text ≠ JSON). But the raw stdin write is nested under `isStreamJsonMode`:

```typescript
} else if (isStreamJsonMode && prompt) {   // <-- FALSE when sendPromptViaStdinRaw=true
    if (config.sendPromptViaStdinRaw) {     // <-- NEVER REACHED
        childProcess.stdin?.write(prompt);
        childProcess.stdin?.end();
    }
}
```

When `sendPromptViaStdinRaw=true` and `sendPromptViaStdin=false`, `isStreamJsonMode` is `false`, so the write never executes. The prompt is silently dropped. **All non-stream-json agents (OpenCode, Gemini CLI, Qwen3 Coder) on Windows get no input via group chat or standard chat.**

**Fix:** Promote `sendPromptViaStdinRaw` to its own top-level branch:

```typescript
if (config.sshStdinScript) {
    // SSH script path...
} else if (config.sendPromptViaStdinRaw && prompt) {
    // Raw stdin path (non-stream-json agents on Windows)
    childProcess.stdin?.write(prompt);
    childProcess.stdin?.end();
} else if (isStreamJsonMode && prompt) {
    // Stream-json path...
} else if (isBatchMode) {
    childProcess.stdin?.end();
}
```

### HIGH: Windows shell config duplicated 5 times

The same ~15-line shell config + stdin flag block is copy-pasted across `group-chat-router.ts` (4x) and `group-chat-agent.ts` (1x). Consider extracting to `group-chat-config.ts`:

```typescript
export function getWindowsSpawnConfig(agentId: string, sshConfig?: SshRemoteConfig) {
    if (process.platform !== 'win32' || sshConfig?.enabled) {
        return { sendPromptViaStdin: false, sendPromptViaStdinRaw: false };
    }
    const shellConfig = getWindowsShellForAgentExecution({ customShellPath: getCustomShellPath() });
    const capabilities = getAgentCapabilities(agentId);
    return {
        shell: shellConfig.shell,
        runInShell: shellConfig.useShell,
        sendPromptViaStdin: capabilities.supportsStreamJsonInput,
        sendPromptViaStdinRaw: !capabilities.supportsStreamJsonInput,
    };
}
```

### HIGH: Inconsistent SSH exclusion

| Spawn location | SSH exclusion for shell config | SSH exclusion for stdin flags |
|---|---|---|
| Moderator (~522) | `else if` after SSH block ✅ | `!chat.moderatorConfig?.sshRemoteConfig` ✅ |
| Participant (~913) | `else if` after SSH block ✅ | `!matchingSession?.sshRemoteConfig` ✅ |
| **Synthesis (~1224)** | **None** ❌ | **None** ❌ |
| **Recovery (~1401)** | **None** ❌ | **None** ❌ |

Synthesis and recovery spawns apply Windows shell config unconditionally on win32, even when the moderator runs on a remote Linux host via SSH. The helper function above would fix this consistently.

### MEDIUM: `spawnModeratorSynthesis` early returns leave UI stuck

Lines 1060-1096 have four early returns that log errors but never emit `groupChatEmitters.emitStateChange?.(groupChatId, 'idle')` or remove the power management block reason. The UI can appear permanently stuck in "agent working" state.

### MEDIUM: No PowerShell existence fallback

`getWindowsShellForAgentExecution` defaults to `powershell.exe` with no fallback to `process.env.ComSpec` if PowerShell is not installed (rare but possible on Server Core).

### What's Good

- `getWindowsShellForAgentExecution` is clean, well-tested (13 tests), with proper basename cmd.exe detection
- Capabilities-based stdin mode selection is the right architecture
- Exit listener retry + log-and-drop is better than the old risky fallback
- `windowsWarningShownRef` properly prevents modal re-showing
- IPC types properly updated in `global.d.ts`

### Recommendation

**Please fix the critical `sendPromptViaStdinRaw` dead code path before merge** — it silently breaks Windows support for all non-stream-json agents, which is the core use case of this PR. The other issues are important but can be addressed in follow-up.

chr1syy:
All issues should have been adressed.

pedramamini:
@chr1syy This has been merged into `0.15.0-polish` via local merge with conflict resolution (3 files had context drift: `group-chat-router.ts`, `process.ts`, `exit-listener.ts`). Since the merge was done locally rather than through the GitHub merge button, the PR shows as closed rather than merged — but all your commits and LoC credit carry through as-is in the git history. Closing this PR now. Thanks for the contribution! 🎉

PR #337, opened 2026-02-11
fix: correct regex for UUID tab IDs in WebSocket session output broad…

EEmayank:
### Problem

The web UI (mobile/desktop) does not receive real-time chat updates via WebSocket. Users must refresh the page to see new AI responses. The `session_output` messages are never delivered to connected web clients.

### Root Cause

Process session IDs use the format `{sessionId}-ai-{tabId}`, where both parts are UUIDs containing dashes:

```
51cee651-6629-4de8-abdd-1c1540555f2d-ai-73aaeb23-6673-45a4-8fdf-c769802f79bb
```

The regex `REGEX_AI_SUFFIX = /-ai-[^-]+$/` uses `[^-]+` which only matches characters **without dashes**. Since UUID tab IDs contain dashes, the regex only captures the last segment (`c769802f79bb`), producing an incorrect `baseSessionId`:

```
Expected: 51cee651-6629-4de8-abdd-1c1540555f2d
Actual:   51cee651-6629-4de8-abdd-1c1540555f2d-ai-73aaeb23-6673-45a4-8fdf
```

This mismatched `baseSessionId` never equals the client's `subscribedSessionId`, so `broadcastToSession()` silently drops every `session_output` message.

### Why Tests Didn't Catch It

All existing tests used simple non-UUID tab IDs like `tab1` or `mytab` (no dashes), so the old regex worked fine in tests but failed in production.

### Fix

Updated the regex to explicitly match UUID format (`[a-f0-9]{8}-[a-f0-9]{4}-...`), consistent with the pattern already used for `REGEX_PARTICIPANT_UUID` in the same file.

### Changes

| File | What changed |
|------|-------------|
| `src/main/constants.ts` | `REGEX_AI_SUFFIX` and `REGEX_AI_TAB_ID` updated from `/-ai-[^-]+$/` to full UUID pattern |
| `src/main/process-listeners/exit-listener.ts` | Same fix for a hardcoded copy of the regex used in `session_exit` broadcasts |
| `src/main/process-listeners/__tests__/data-listener.test.ts` | Tests now use realistic UUID session/tab IDs instead of `tab1` |
| `src/main/process-listeners/__tests__/exit-listener.test.ts` | Mock patterns updated to match new regex |
| `src/main/process-listeners/__tests__/session-id-listener.test.ts` | Mock patterns updated to match new regex |
| `src/main/process-listeners/__tests__/usage-listener.test.ts` | Mock patterns updated to match new regex |

### Before / After

```
Before: /-ai-[^-]+$/           → matches "-c769802f79bb" only (WRONG)
After:  /-ai-[a-f0-9]{8}-...$/ → matches "-ai-73aaeb23-6673-45a4-8fdf-c769802f79bb" (CORRECT)
```

### DIFF
--- a/src/main/constants.ts
+++ b/src/main/constants.ts
@@ -23,8 +23,9 @@ export const REGEX_PARTICIPANT_TIMESTAMP = /^group-chat-(.+)-participant-(.+)-(\
 export const REGEX_PARTICIPANT_FALLBACK = /^group-chat-(.+)-participant-([^-]+)-/;
 
 // Web broadcast session ID patterns
-export const REGEX_AI_SUFFIX = /-ai-[^-]+$/;
-export const REGEX_AI_TAB_ID = /-ai-([^-]+)$/;
+// Tab IDs are UUIDs (e.g., 73aaeb23-6673-45a4-8fdf-c769802f79bb) so we must match the full UUID after -ai-
+export const REGEX_AI_SUFFIX = /-ai-[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$/i;
+export const REGEX_AI_TAB_ID = /-ai-([a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12})$/i;
 
 // Auto Run session ID patterns (batch and synopsis operations)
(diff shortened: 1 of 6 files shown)


pedramamini:
Thanks for the solid fix and excellent write-up @EEmayank! Merged.

This will ship in 0.15.0 — we're updating the RC tonight if you want to give it a spin.

Issue #346, opened 2026-02-11
FR: support for custom SSH config such as control items


rnwolfe:
I have a catch all SSH config (though, in this context it is really just relevant to one remote session I am using for Maestro). This is not just an optimization, rather a requirement in case as I have to touch a physical security key for every ssh connection. If they share a control session, I touch it once and everything afterward doesn't need it. Given the way remote SSH agent works, it is **very** chatty and I have to **constantly** tap the key. So it is impossible to use.

```
Match all
  # Timeout failed connections after 120 secs
  ServerAliveInterval 30
  ServerAliveCountMax 4
  ConnectTimeout 60
  ConnectionAttempts 2
  ControlPath ~/.ssh/controlmasters/%r@%h:%p
  ControlMaster auto
  ControlPersist 10m
  Compression yes
  CheckHostIP no
```

Maestro supports custom SSH configs, but I see that these specific ones are not accepted:

https://github.com/pedramamini/Maestro/blob/473cff1fedc5359a184663eebcc58af0873afa1f/src/main/ssh-remote-manager.ts#L71-L73

I'm not sure the background on why these were done, but the comments suggest there was some error. I've been using it with these lines removed via local changes and have had no issues. Curious if I can get more context - if so, I am happy to make appropriate changes myself and give a PR.

pedramamini:
Hmmm I wonder why we did this. Let me dig. Thank you for your excellent and informative issues. We'd love to have you on the Discord. It's also possible that these were attempts to address a bug, that weren't the root-cause, and then never removed. If you would, PR your fix, that way if we can clear that there is no regression, we'll just take your PR and get you the contribution credit (which is important to me).

rnwolfe:
Looking at it further. Claude had this to say:

> The ControlMaster, ControlPath, and ControlPersist options are being forcibly overridden to disable connection multiplexing. The comment mentions preventing "UNKNOWN
>   port -1" errors when multiple agents connect to the same server.
> 
>   Let me analyze this:
> 
>   1. The edge case concern: The "UNKNOWN port -1" error can occur when multiple SSH processes race to establish or use a control socket simultaneously. However, this is typically a transient issue that users
>   with ControlMaster configured have already worked around (e.g., proper ControlPath patterns with %h, %p, %r).
>   2. The problem: By forcibly disabling ControlMaster, users with hardware security keys (YubiKey, etc.) must touch the key for every SSH connection Maestro makes, which is frustrating and unnecessary.
[...]

PR #350, opened 2026-02-12
fix: stop overriding SSH ControlMaster/ControlPath/ControlPersist configs

rnwolfe:
Maestro was forcibly disabling SSH connection multiplexing, requiring a hardware security key touch (or passphrase) for every SSH connection. Users who configure ControlMaster in ~/.ssh/config should have that respected. Updated docs with connection multiplexing guidance.

### DIFF
--- a/docs/ssh-remote-execution.md
+++ b/docs/ssh-remote-execution.md
@@ -235,6 +235,14 @@ This is especially useful for:
 - **Import from SSH config** — Use the dropdown when adding remotes to import from `~/.ssh/config`; saves time and keeps configuration consistent
 - **Bastion hosts** — Use `ProxyJump` in your SSH config for multi-hop connections; Maestro inherits this automatically
 - **Key management** — Use `ssh-agent` to avoid passphrase prompts
+- **Connection multiplexing** — Maestro respects `ControlMaster`, `ControlPath`, and `ControlPersist` from your `~/.ssh/config`. This is highly recommended if you use hardware security keys (e.g., YubiKey) to avoid repeated touches per connection. Example config:
+  ```
+  Host dev-server
+      ControlMaster auto
+      ControlPath ~/.ssh/sockets/%r@%h-%p
+      ControlPersist 600
+  ```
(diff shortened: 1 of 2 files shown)


pedramamini:
Thanks for the clean PR and the thorough investigation in #346, @rnwolfe — really appreciate the effort tracing back through the original commits and providing context. This is exactly the kind of contribution that makes Maestro better.

Merged! We're rebuilding the RC tonight, so if you get a chance to kick the tires on it we'd appreciate the validation from your hardware key setup.

Closing #346 as resolved via this PR.

rnwolfe:
I pulled the RC and ran fresh - works like a charm! 🎉

There is a nuance, which is not really Maestro specific, but could probably be handled a bit more gracefully if desired depending on the context when Maestro is launched:

1. Scenario 1: there is an existing session to the target remote SSH host open, which has established a control session. When Maestro opens, no security key touch/passphrase required. Works exactly as you would want!
2. Scenario 2: there is **no** existing session to the target remote SSH host open, so there is no established control session. When Maestro opens, it simultaneously opens multiple SSH connections (seems ~3 or so). Since they open at the same time, they all establish a new control session and require a security key touch/passphrase. Afterwards, no further touches/passphrases are needed. 

The issue in scenario 2 happens even if I quickly open multiple SSH sessions without Maestro in the mix and works itself out once they are all open (just like with Maestro). So, it is more of a minor UX thing. It could likely be fixed by giving a minor delay between first and subsequent SSH connections within Maestro, but also may not be worth the effort.

@pedramamini if you think this is worth it, let me know (or open an issue and assign to me) and I am happy to give it a go. 

The other quality-related item is that the code handling custom SSH options, etc. is clearly spread through multiple locations and could likely be refactored to a central location so that an update to the core logic handles all of these cases in the future. This could be worth it as remote ssh functionality may get more usage. 

Same as above, if you think this is worth it right now, let me know (or open an issue and assign to me), and I can take a stab.

Cheers! Also, I haven't said this already, but Maestro is great and thanks for taking the lead!

PR #384, opened 2026-02-15
feat: fuzzy substring search with ranked scoring for slash commands

felipeggv:
## Summary

Replaces the prefix-only slash command filter with ranked substring matching. Users can now find commands by typing any part of the command name or description — similar to VS Code's Command Palette.

## Problem

The slash command dropdown only matched commands that **started with** the typed text. With growing numbers of custom commands (Spec-Kit, OpenSpec, user-defined), users had to remember exact command prefixes to find anything.

**Example:** typing `/prd` returned nothing, even though `/GLOBAL:agents:prd-creator` exists.

## Solution

Three-tier ranked scoring system applied to both `InputArea.tsx` (dropdown render) and `App.tsx` (keyboard handler):

| Score | Match Type | Example (input: `hist`) |
|-------|-----------|------------------------|
| **3** | Prefix on command name | `/history` → appears first |
| **2** | Substring in command name | `/command-history-viewer` → appears second |
| **1** | Substring in description | `/clear` (if description mentions "history") → appears last |

Results are sorted descending by score. The leading `/` is stripped from both input and command name before comparison to avoid false negatives.

## Before / After

| User types | Before | After |
|------------|--------|-------|
| `/prd` | No results | `/GLOBAL:agents:prd-creator` (score 2) |
| `/synopsis` | No results | `/history` (description contains "synopsis", score 1) |
| `/ard` | No results | `/wizard` (name contains "ard", score 2) |
| `/hist` | `/history` only | `/history` (score 3, prefix) + description matches (score 1) |
| `/clear` | `/clear` only | `/clear` (score 3, prefix) — same behavior, backwards compatible |

## How This Helps Users

- **Discoverability** — Find commands by what they *do*, not just their exact name
- **Partial recall** — Don't need to remember full command prefixes
- **Scales with custom commands** — As users add Spec-Kit, OpenSpec, and custom slash commands, search stays useful
- **No noise** — Ranked scoring ensures prefix matches always appear above description-only matches

## Files Changed

| File | Change |
|------|--------|
| `src/renderer/components/InputArea.tsx` | Dropdown filter: `startsWith` → scored `includes` on name + description |
| `src/renderer/App.tsx` | Keyboard handler filter: same scoring logic |

## Implementation Details

```typescript
// Scoring logic (same in both files)
const cmdName = cmd.command.toLowerCase().replace(/^\//, '');
if (cmdName.startsWith(searchTerm))  → score 3 (prefix match)
if (cmdName.includes(searchTerm))    → score 2 (substring in name)
if (description.includes(searchTerm)) → score 1 (substring in description)
// No match → excluded from results
```

- `searchTerm` strips the leading `/` from user input
- `cmd.description` is guarded with `&&` (some custom commands may not have descriptions)
- `useMemo` dependency updated from `inputValueLower` to `searchTerm` in InputArea

## Test Plan

- [ ] Type `/` — all commands appear (empty search term = score 0, all included)
- [ ] Type `/prd` — `prd-creator` command appears
- [ ] Type `/synopsis` — `/history` appears (description match)
- [ ] Type `/hist` — `/history` appears **first** (prefix match before description matches)
- [ ] Type `/s` — prefix matches (e.g., `/status`) appear above description-only matches
- [ ] Arrow keys navigate filtered list correctly
- [ ] Enter/Tab selects the highlighted command
- [ ] Escape closes the dropdown
- [ ] Terminal-only commands don't appear in AI mode (mode filtering unchanged)
- [ ] `npm run lint` — zero new errors
- [ ] `npm test -- --run` — all tests pass (5 pre-existing failures in `symphony.test.ts`)
- [ ] `npm run build` — clean production build

### DIFF
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -9364,16 +9364,19 @@ You are taking over this conversation. Based on the context above, provide a bri
 		// Handle slash command autocomplete
 		if (slashCommandOpen) {
 			const isTerminalMode = activeSession?.inputMode === 'terminal';
-			const filteredCommands = allSlashCommands.filter((cmd) => {
-				// Check if command is only available in terminal mode
-				if ('terminalOnly' in cmd && cmd.terminalOnly && !isTerminalMode) return false;
-				// Check if command is only available in AI mode
-				if ('aiOnly' in cmd && cmd.aiOnly && isTerminalMode) return false;
-				// Check if command matches input (strip leading / from both for substring matching)
-				const searchTerm = inputValue.toLowerCase().replace(/^\//, '');
+			const searchTerm = inputValue.toLowerCase().replace(/^\//, '');
+			const scored: { cmd: typeof allSlashCommands[number]; score: number }[] = [];
+			for (const cmd of allSlashCommands) {
+				if ('terminalOnly' in cmd && cmd.terminalOnly && !isTerminalMode) continue;
+				if ('aiOnly' in cmd && cmd.aiOnly && isTerminalMode) continue;
+				if (!searchTerm) { scored.push({ cmd, score: 0 }); continue; }
 				const cmdName = cmd.command.toLowerCase().replace(/^\//, '');
-				return cmdName.includes(searchTerm) || (cmd.description && cmd.description.toLowerCase().includes(searchTerm));
-			});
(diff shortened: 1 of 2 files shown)


pedramamini:
@coderabbit review

felipeggv:
Superseded by #467 (feat/fuzzy-search) which includes the same fuzzy matching improvements plus additional fixes.

PR #407, opened 2026-02-18
Codex integration fixes + agent version detection

openasocket:
## Summary

- Fix Codex CLI integration end-to-end (8 bugs across parser, exit handler, stall detection, UX)
- Add dual-format parser supporting both old (v0.41.0-era) and current Codex CLI JSONL output formats
- Add agent version detection infrastructure with spawn-time warnings for unsupported versions
- Display detected agent versions across 3 UX surfaces (Edit Modal, Wizard, AgentCard)

## Motivation

Maestro's Codex parser was originally built against Codex CLI v0.41.0, which used a flat item-envelope JSONL format (`{ type: 'item.completed', item: {...} }`). Subsequent Codex CLI releases (current: v0.103.0) changed to a nested msg-envelope format (`{ id, msg: { type: 'agent_message', ... } }`). Users who upgraded Codex CLI got **zero parsed output with no explanation** because Maestro's parser only understood the old format.

Beyond the format mismatch, the initial Codex integration had several bugs that caused:
- Reasoning text leaking into user-visible results on process exit
- Tool names lost between `tool_call` and `tool_result` events (Codex emits these as separate events)
- `streamedText` clobbered by reasoning, breaking result extraction
- Auto Run sessions hanging indefinitely with no stall detection
- `showThinking` defaulting to `'off'` for Codex, hiding the primary visibility into agent activity
- Parser singleton state leaking across sessions

This PR fixes all of the above and adds version detection so Maestro can warn users before they hit these silent failures.

### Codex Integration Fixes (8 commits)

| Commit | Fix |
|--------|-----|
| `0cf3fc4f` | Separate `codexPendingResult` from `streamedText` to prevent reasoning text from clobbering the actual result |
| `3f1f3ca8` | Default `showThinking` to `'on'` for Codex across 7 tab creation sites; add length bypass for block-style reasoning |
| `e07454d3` | Fix Auto Run synopsis extraction for short Codex responses; add result capture logging |
| `d1c83270` | Carry tool names across `tool_call` → `tool_result` events; truncate large tool outputs (10K limit) |
| `d23ece56` | Add `modelArgs` for Codex (`-m` flag); fix parser singleton config staleness with `refreshConfig()` |
| `6079a161` | Bridge IPC gap for Auto Run stall detection — track activity from thinking/tool events, add 5-min inactivity timeout |
| `b51f7e5c` | Fix ExitHandler reasoning leak (toolType-aware branching), singleton state leak (`lastToolName` reset), dead stall detection |
[...]

PR #406, opened 2026-02-17
fix: keep AI auto-scroll synced with streaming updates

rnwolfe:
Fix for: https://github.com/RunMaestro/Maestro/issues/402

Disclaimer: pushing for tracking, has not yet been fully tested (had to step away)

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced terminal output auto-scroll behavior to better handle dynamic log updates and streaming content.
  * Improved scroll position management for real-time log monitoring.

PR #390, opened 2026-02-16
feat: add auto-scroll toggle for AI output

openasocket:
### Feature Overview

Adds configurable auto-scroll behavior for AI mode output. When enabled, the terminal output automatically scrolls to the bottom as new AI content arrives — matching terminal mode behavior. The feature includes smart pause/resume: scrolling up pauses auto-scroll so you can read content, and scrolling back to bottom resumes it.

### Controls

| Method | Action |
|--------|--------|
| **Settings Modal** | General tab → "Auto-scroll AI Output" checkbox |
| **Inline Button** | Floating toggle button (bottom-left of AI output) |
| **Keyboard Shortcut** | `Alt+Cmd+S` (works even with modals open) |

### Behavior

- **Auto-scroll ON** → New AI output scrolls viewport to bottom instantly
- **Scroll up** → Auto-scroll pauses, button icon dims to indicate paused state, new message indicator appears
- **Scroll back to bottom** → Auto-scroll resumes automatically, button returns to active state
- **Click button while paused** → Resumes auto-scroll without toggling the setting off
- **Click button while active** → Disables auto-scroll entirely
- **Default: OFF** → Preserves existing behavior where users manually scroll

### Visual Feedback

The inline toggle button reflects three states:
- **Active** (accent color, ↓ icon): Auto-scroll is on and scrolling
- **Paused** (dim, scroll icon): Auto-scroll is on but user scrolled up — tooltip says "paused (click to resume)"
- **Off** (dim, scroll icon): Auto-scroll is disabled — tooltip says "click to enable"

### Technical Details

- Setting persisted via electron-store as `autoScrollAiMode` (boolean)
- Uses `behavior: auto` (instant) instead of `smooth` to prevent jitter during rapid streaming
- Disables browser native `overflow-anchor` CSS on the scroll container when auto-scroll is off/paused, preventing the browser from pinning the viewport to bottom
- `autoScrollPaused` is React state (not a ref) so the toggle button re-renders to reflect pause state
- Settings Modal receives `autoScrollAiMode`/`setAutoScrollAiMode` as props from App (not from its own `useSettings()` instance) to ensure changes take immediate effect
- Keyboard shortcut registered as system utility shortcut so it works even when modals are open

### Files Changed

| File | Change |
|------|--------|
| `useSettings.ts` | `autoScrollAiMode` state, persistence, loading |
| `App.tsx` | Extract setting, pass to keyboard context, MainPanel props, and SettingsModal |
[...]

PR #410, opened 2026-02-18
fix: auto-scroll follows thinking streams + Codex tool display

### LINKED ISSUE #402: AI chat auto-scroll does not follow streaming updates when Thinking/Tool Calls are visible
### Summary
When AI chat is configured to show thinking/tool-call output, the viewport can stop following the live conversation even though auto-scroll is enabled.

### Environment
- App: Maestro (renderer chat view)
- Component: `src/renderer/components/TerminalOutput.tsx`
- Mode: AI mode with thinking/tool-call visibility enabled

### Steps to Reproduce
1. Enable AI chat display mode that shows thinking/tool calls.
2. Start a prompt that produces streaming thinking/tool activity.
3. Keep auto-scroll enabled and observe output as chunks stream.

### Expected Behavior
The chat viewport stays pinned to the latest output while streaming is active.

### Actual Behavior
The viewport intermittently stops following updates during streaming, requiring manual scroll-to-bottom.

openasocket:
## Summary

- Replace `filteredLogs.length` auto-scroll dependency with `MutationObserver` to detect in-place content changes (thinking stream text growth, tool status updates)
- Add programmatic scroll guard to prevent the scroll handler from falsely pausing auto-scroll during effect-driven `scrollTo()` calls
- Fix Codex tool execution display: carry tool name from `tool_call` to `tool_result`, handle array-typed `command` args, add Codex-specific arg extraction, truncate large outputs

### Auto-scroll stops following thinking/tool streams (Fixes #402, supersedes #406)

The auto-scroll feature ([PR #390](https://github.com/RunMaestro/Maestro/pull/390)) depends on `filteredLogs.length` to trigger scroll-to-bottom. However, thinking chunks update the **last log entry's text in-place** without changing the array length:

```typescript
// useAgentListeners.ts — thinking chunk handler
logs: [...tab.logs.slice(0, -1), { ...lastLog, text: combinedText }]
//                                               ^^^^^^^^^^^^^^^^^^
//                                               text grows, length stays the same
```

Timeline of the bug:
```
T1: First thinking chunk → new log added → length 1→2 → scroll fires ✓
T2: Second thinking chunk → text updated → length still 2 → NO scroll ✗
T3: Third thinking chunk → text updated → length still 2 → NO scroll ✗
T4: Tool event → new log added → length 2→3 → scroll fires ✓
T5: More thinking → text updated → length still 3 → NO scroll ✗
```

[PR #406](https://github.com/RunMaestro/Maestro/pull/406) proposed a content-fingerprint memo (`lastVisibleLogActivityKey`) that creates a composite key from the last log's metadata. This works but couples the scroll trigger to React's data model. Our approach uses a **MutationObserver** on the scroll container instead, which:

1. **Catches all content changes** — new DOM nodes, text mutations — regardless of how the React data model produces them
2. **Is framework-agnostic** — doesn't need to know about `filteredLogs`, `collapsedLogs`, or log data shapes
3. **Naturally throttles** — MutationObserver batches mutations and delivers them asynchronously, providing built-in coalescing
4. **Has zero React overhead** — no additional memos, no dependency array churn, no re-renders

### Programmatic scroll falsely pauses auto-scroll (secondary race condition)

When the auto-scroll effect calls `scrollTo({ behavior: 'auto' })`, the instant scroll fires the `onScroll` handler synchronously. If `scrollHeight` just grew from new content but the scroll position hasn't propagated yet, the handler computes `atBottom = false` and sets `autoScrollPaused = true` — pausing auto-scroll during a *programmatic* scroll. This creates intermittent freezes where the viewport stops following output mid-stream.

Fixed with a ref-based guard (`isProgrammaticScrollRef`) set before `scrollTo()` and cleared via `queueMicrotask`. The scroll handler checks this flag before pausing.

### Codex tool executions show no command details

Codex bash tool entries render as just `bash ●` / `bash ✓` with no information about what command was executed. Three sub-bugs:

**Missing tool name on completion:** Codex emits `tool_call` and `tool_result` as separate `item.completed` JSONL events. The parser transforms `tool_result` without a `toolName` field (only `tool_call` items include `item.tool`). StdoutHandler's gate (`event.toolName` check) silently drops the event — the renderer never receives "completed" status.

**Array commands not rendered:** Codex tool_call items use `args: { command: ['ls', '-la'] }` (string array). The renderer's `safeStr(toolInput.command)` returns `null` for non-strings, so no command detail is displayed. This bug is a **new finding** not addressed in [PR #407](https://github.com/RunMaestro/Maestro/pull/407)'s tool fix (commit `451f8871`), which added `lastToolName` tracking and Codex arg patterns but didn't handle the array `command` format.

**No output truncation:** Large tool outputs (e.g., `cat` of a big file) create oversized log entries with no limit.

### Commit 1: `638db6ff` — MutationObserver auto-scroll

Replace the `filteredLogs.length` dependency in the auto-scroll `useEffect` with a `MutationObserver` on `scrollContainerRef`:

```typescript
const observer = new MutationObserver(() => {
    if (shouldAutoScroll()) scrollToBottom();
});
observer.observe(container, {
    childList: true,    // New/removed DOM nodes (new log entries)
    subtree: true,      // Watch all descendants
    characterData: true, // Text node mutations (thinking stream growth)
});
```

The observer is set up and torn down inside the effect, keyed on `[session.inputMode, autoScrollAiMode, autoScrollPaused]`. The `filteredLogs.length` dependency is completely removed from this effect.

**Intentionally kept:** The "detect new messages" effect and `handleScrollInner` still use `filteredLogs.length` for unread badge counting and read state tracking — in-place text updates should NOT increment unread counts.

### Commit 2: `35b86947` — Programmatic scroll guard

- Add `isProgrammaticScrollRef` (boolean ref)
- Set `true` before `scrollTo()`, clear via `queueMicrotask` after
- Check `!isProgrammaticScrollRef.current` in `handleScrollInner` before setting `autoScrollPaused = true`

Why `queueMicrotask` not `setTimeout(0)`: microtasks run after the current task's synchronous work but before the next macro task. Since `behavior: 'auto'` scrolls synchronously, the `onScroll` event fires in the same task. The microtask clears the flag before any subsequent user scroll could fire.

### Commit 3: `fdf8329f` — Tests

4 new tests in `auto-scroll.test.tsx`:
- Thinking text growth triggers auto-scroll (same array length)
- Unread badge does NOT increment on in-place text updates
- Tool event appends trigger auto-scroll
- Genuine user scroll-up still pauses auto-scroll (guard doesn't interfere)

Includes a class-based `MockMutationObserver` for JSDOM compatibility.

### Commit 4: `c276056c` — Codex tool display

**Parser** (`codex-output-parser.ts`):
- Add `lastToolName` field: stored on `tool_call`, carried to `tool_result`, cleared after use
- Add `MAX_TOOL_OUTPUT_LENGTH` (10K chars) truncation in `decodeToolOutput`

**Renderer** (`TerminalOutput.tsx`):
- Add `safeCommand()` helper: handles both string and string-array `command` values (joins arrays with spaces)
- Add `truncate()` helper for length-limited display
- Add Codex-specific arg patterns: `path`, `cmd`, `code`, `content`
- Change `safeStr(toolInput.command)` → `safeCommand(toolInput.command)`

**Tests** (`codex-output-parser.test.ts`):
- Tool name carryover from `tool_call` to `tool_result`
- `lastToolName` reset after consumption (no stale names)
- Output truncation at 10K chars with marker

## Files Changed (4 files, +375 / -29)

| File | Changes |
|------|---------|
| `src/renderer/components/TerminalOutput.tsx` | MutationObserver effect, `isProgrammaticScrollRef` guard, `safeCommand()` helper, Codex arg extraction |
| `src/main/parsers/codex-output-parser.ts` | `lastToolName` tracking, output truncation |
| `src/__tests__/renderer/components/auto-scroll.test.tsx` | 4 new tests + MutationObserver mock |
| `src/__tests__/main/parsers/codex-output-parser.test.ts` | 3 new tests for tool name carryover + truncation |

## Related

- Fixes #402
- Supersedes #406 (closed — content-fingerprint approach replaced with MutationObserver)
- Extends #390 (original auto-scroll feature, merged)
- Overlaps with #407 (Codex tool fixes — `lastToolName` + truncation exist there too, but array `command` handling is new here)

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run test` — 468 test files, 19,984 tests pass
- [x] New auto-scroll tests: thinking stream growth, unread badge isolation, tool event scroll, user scroll-up guard
- [x] New Codex parser tests: tool name carryover, stale name reset, output truncation
- [ ] Manual: enable thinking display → start streaming prompt → verify viewport stays pinned during thinking text growth
- [ ] Manual: scroll up during streaming → verify auto-scroll pauses → scroll back down → verify it resumes
- [ ] Manual: Codex agent → run a command → verify tool entries show command details (not just "bash ●")

## Summary by CodeRabbit

* **New Features**
  * Tool context is preserved across consecutive tool events so displayed tool names remain consistent.
  * Terminal displays additional tool details and trims very long tool content for readability.

* **Bug Fixes**
  * Improved auto-scroll: programmatic scrolling no longer interrupts auto-scroll; user scrolls still pause it.
  * Unread badge/alignment updated to reflect actual scroll-bottom state.

* **Tests**
  * Added coverage for tool name carryover, output truncation, and mutation-driven scroll behavior (includes test polyfill).

### DIFF
--- a/src/renderer/components/TerminalOutput.tsx
+++ b/src/renderer/components/TerminalOutput.tsx
@@ -1500,6 +1500,13 @@ export const TerminalOutput = memo(
 							top: scrollContainerRef.current.scrollHeight,
 							behavior: 'auto',
 						});
+						// Fallback: if scrollTo is a no-op (already at bottom), the browser
+						// won't fire a scroll event, so the handler never consumes the guard.
+						// Clear it after 32ms (2x the 16ms throttle window) to prevent a
+						// stale true from eating the next genuine user scroll-up.
+						setTimeout(() => {
+							isProgrammaticScrollRef.current = false;
+						}, 32);
 					}
 				});
 			};


openasocket:
## Addressed Review Findings — `ce7b086d`

Both actionable items from the CodeRabbit review have been fixed:

### 1. Sentry reporting for `Buffer.from()` decode errors
**File:** `src/main/parsers/codex-output-parser.ts`

The `catch` block in `decodeToolOutput` previously swallowed errors silently. Now imports `captureException` from `../utils/sentry` and reports the error with context (`operation`, `outputType`, `outputLength`) before falling back to `output.toString()`. This follows the project's error handling guidelines — recoverable errors are handled gracefully but still reported to Sentry for visibility.

### 2. Throttle race condition on `isProgrammaticScrollRef`
**File:** `src/renderer/components/TerminalOutput.tsx`

**The bug:** `scrollToBottom()` set `isProgrammaticScrollRef = true`, called `scrollTo()`, then cleared the flag via `queueMicrotask` (~0ms). But `handleScroll` is wrapped in `useThrottledCallback(fn, 16)` — a leading-edge throttle that defers trailing calls via `setTimeout(fn, delay - elapsed)`. When a scroll event fell into the throttle window, the deferred handler fired ~16ms later, **after** the microtask had already cleared the guard, causing a false auto-scroll pause.

**The fix:** Removed the `queueMicrotask` clear entirely. Instead, the guard flag is now **consumed** (checked and cleared) inside `handleScrollInner` itself:

```typescript
} else if (autoScrollAiMode) {
    if (isProgrammaticScrollRef.current) {
        // Consume the guard — this was our own scrollTo()
        isProgrammaticScrollRef.current = false;
    } else {
        // Genuine user scroll-up
        setAutoScrollPaused(true);
    }
}
```

This eliminates the timing dependency between microtasks and the throttle window. The flag persists until the scroll handler actually processes it, regardless of when the throttled invocation fires.

**Tests:** All 65 tests pass (auto-scroll + codex parser), TypeScript compiles clean.

openasocket:
## Addressed Review Round 2 — `a4b74d0e`

### Fixed: Stale guard on no-op scroll (Finding #4)

When `scrollTo()` is called but the container is already at the bottom, the browser doesn't fire a `scroll` event. The `isProgrammaticScrollRef` guard stays `true` indefinitely — the next genuine user scroll-up gets silently consumed instead of pausing auto-scroll.

**Fix:** Added a 32ms `setTimeout` fallback in `scrollToBottom` that clears the flag if the throttled handler hasn't consumed it. 32ms is 2x the 16ms throttle window, guaranteeing the handler has had its chance. If the handler *does* consume it first, the timeout harmlessly sets `false` again (idempotent).

### Skipped: tool_call_id mapping (Finding #3)

The suggestion to replace `lastToolName` with an ID-based `Map<string, string>` keyed by `tool_call_id` is over-engineered for our use case. Codex CLI outputs sequential JSONL — `tool_call` always immediately precedes its `tool_result` in the stdout stream. Out-of-order events can't happen with CLI stdout. The existing test at lines 427-443 validates the intended behavior: an orphaned `tool_result` correctly gets `undefined` as its `toolName`.

**Tests:** All 19 auto-scroll tests pass, TypeScript compiles clean.

PR #411, opened 2026-02-18
Fixed worktree duplication on creation

chr1syy:
This pull request introduces more robust handling of Git worktree session creation to avoid duplicate sessions due to path normalization and race conditions. 

**Worktree session handling improvements:**

* Adds a `normalizePath` utility to consistently compare file paths by converting backslashes to slashes and removing redundant/trailing slashes. This prevents duplicate Git worktree sessions when paths differ only in formatting. [[1]](diffhunk://#diff-2f96bcb53726d77aee2f676a75fe6734df8d8c79962c10fe2b517f116a254d1aR83-R87) [[2]](diffhunk://#diff-2f96bcb53726d77aee2f676a75fe6734df8d8c79962c10fe2b517f116a254d1aL315-R324) [[3]](diffhunk://#diff-2f96bcb53726d77aee2f676a75fe6734df8d8c79962c10fe2b517f116a254d1aL608-R629)

* Ensures that paths are marked as "recently created" before starting asynchronous worktree creation, and that these marks are removed if creation fails. This prevents race conditions where the file watcher could create duplicate sessions for the same path. [[1]](diffhunk://#diff-2f96bcb53726d77aee2f676a75fe6734df8d8c79962c10fe2b517f116a254d1aR422-R428) [[2]](diffhunk://#diff-2f96bcb53726d77aee2f676a75fe6734df8d8c79962c10fe2b517f116a254d1aR439-R440) [[3]](diffhunk://#diff-2f96bcb53726d77aee2f676a75fe6734df8d8c79962c10fe2b517f116a254d1aL440-L443) [[4]](diffhunk://#diff-2f96bcb53726d77aee2f676a75fe6734df8d8c79962c10fe2b517f116a254d1aR506-R512) [[5]](diffhunk://#diff-2f96bcb53726d77aee2f676a75fe6734df8d8c79962c10fe2b517f116a254d1aR522-R523) [[6]](diffhunk://#diff-2f96bcb53726d77aee2f676a75fe6734df8d8c79962c10fe2b517f116a254d1aL518-L521)


pedramamini (commented):
## Review Feedback

Two separate concerns bundled here. The worktree fix is solid; the platform detection needs rework.

---

### Feature 1: Platform-aware "Reveal in Finder/Explorer" — Needs Rework

**Over-engineered for a static value.** `process.platform` never changes at runtime. Adding async IPC handlers (`os:getPlatform`, `os:getArch`), a new hook with caching/error-handling, and `useEffect` + `useState` for something that could be a synchronous constant is unnecessary complexity.

Specific issues:

1. **`'unknown'` flash on first render.** The hook returns `'unknown'` before the IPC resolves. Users briefly see the wrong label (or no label match) until the async call completes. A synchronous value from preload avoids this entirely.

2. **Missing Linux.** `revealLabel` only handles `win32` → "Explorer" with fallback to "Finder". Linux users see "Reveal in Finder" which is incorrect — should be "Reveal in File Manager" or "Show in Files".

3. **Duplicated `useMemo`.** The same `revealLabel` memo is copy-pasted into both `FileExplorerPanel.tsx` and `TabBar.tsx`. Should be a single shared utility.

4. **Redundant type exports.** `OsApi` (interface) and `OsApiType` (ReturnType) export the same shape. Pick one.

**Suggested approach:** Expose `process.platform` as a synchronous property in preload (e.g., `window.maestro.platform`), then use a simple utility function like `getRevealLabel(platform: string): string` in a shared location. No hook needed.

---

### Feature 2: Worktree path normalization + race fix — LGTM

This is the real value in the PR:

- `normalizePath()` correctly centralizes the scattered `.replace(/\/+$/, '')` calls and adds backslash handling for Windows
- Moving `recentlyCreatedWorktreePathsRef.add()` **before** `git.worktreeSetup()` correctly closes the chokidar race window — good catch
- Cleanup on failure (removing from ref if creation fails) is handled correctly
- All path comparisons now consistently use `normalizePath()`

**Minor nit:** `normalizePath` is local to `useWorktreeHandlers.ts`. If path normalization is needed elsewhere later, consider extracting to a shared util. Fine as-is for now.

---

### Recommendation

Split into two PRs:
1. **Worktree path normalization** — merge as-is
2. **Platform-aware labels** — rework with synchronous preload value + shared utility + Linux support

PR #428, opened 2026-02-20
fix: replace JS resize detection with CSS container queries for header

### LINKED ISSUE #420: [BUG]  Overlapping elements on chat window top bar
Elements are overlapping on the toolbar above chat windows when the window is resized 

Image

chr1syy:
The header bar elements were overlapping on smaller screens and not properly restoring when the window was resized larger. The JavaScript- based ResizeObserver with threshold detection was unreliable.

This commit replaces the JS approach with CSS container queries:
- Add header-container class with container-type: inline-size
- Use @container queries to progressively hide elements at breakpoints
- Remove panelWidth state, ResizeObserver, and related variables
- Add CSS classes to header elements for container query targeting
- Support header-auto-mode modifier for when AUTO button is visible

Benefits:
- Instant response to size changes (no state update delays)
- No bugs with threshold detection or resize event batching
- Reduced React re-renders (elements always in DOM, CSS controls visibility)
- Simpler code (~100 lines of JS removed)

Fixes #420

Session: 64f29ed8-30fe-4b59-a4d3-f6bf87e2d3d6

## Summary by CodeRabbit

* **Refactor**
  * Header responsiveness moved from JavaScript to CSS container-query-like behavior for smoother, more reliable resizing and truncation.
  * Git/status and context displays now render consistently while their visibility is controlled by responsive styles rather than runtime width checks.

* **Style**
  * New CSS rules introduce progressive breakpoints and an auto mode for the header, plus improved reduced-motion handling for header animations.

### DIFF
--- a/src/renderer/components/GitStatusWidget.tsx
+++ b/src/renderer/components/GitStatusWidget.tsx
@@ -11,8 +11,6 @@ interface GitStatusWidgetProps {
 	theme: Theme;
 	onViewDiff: () => void;
 	onViewLog?: () => void;
-	/** Use compact mode - just show file count without breakdown */
-	compact?: boolean;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


pedramamini:
@chr1syy what do you think of the 🦎 comments above?

chr1syy:
@pedramamini worth keeping the git compact behaviour, looking into it

PR #461, opened 2026-02-25
fix: Files panel hangs on "Loading remote files..." indefinitely

croakingtoad:
## Problem

The Files right panel in SSH remote sessions gets permanently stuck showing "Loading remote files..." with a spinner and a stale progress count (e.g., "25 files in 7 folders"), but never actually renders the file tree. This happens especially after extended Maestro usage (hours), making the Files panel effectively non-functional for remote sessions.

image

## Root Cause Analysis

I traced the issue through three interacting code paths:

### 1. No timeout on SSH commands — the primary cause

`loadFileTree()` (`fileExplorer.ts`) walks the remote directory tree by making **one SSH call per directory** via `window.maestro.fs.readDir`. Each call flows through:

```
loadFileTreeRecursive (per directory, sequential for-of loop)
  → IPC: fs:readDir
    → readDirRemote (remote-fs.ts)
      → execRemoteCommand
        → execFileNoThrow('ssh', args)  ← NO TIMEOUT
```

`execFileNoThrow` (`execFile.ts`) calls Node's `execFile` without a `timeout` option. If any single SSH command hangs (stale ControlMaster socket, keepalive failure, network degradation after hours of use), the promise never resolves. Since directories are processed sequentially in a `for...of` loop, one hung SSH call blocks the entire tree walk forever.

The progress callback fires after each directory completes, which is why users see "25 files in 7 folders" — that's the count from the last directory that succeeded before the hang. `fileTreeLoading` stays `true` permanently, so the UI never transitions from the spinner to the file list.

The overall `Promise.all([loadFileTree, directorySize])` in `useFileTreeManagement.ts` also cannot resolve because even if `directorySize` (a single `du`/`find` command) completes, `Promise.all` waits for both.

### 2. Stale closure in async callbacks

The `useEffect` that triggers the initial file tree load captures `activeSessionId` from its closure. The `.then()` and `.catch()` callbacks use this captured value to identify which session to update. If the user switches sessions while loading is in progress, the callbacks fire with the old `activeSessionId` — which can cause the completion state to target the wrong session, leaving the original session's `fileTreeLoading` permanently stuck at `true`.

### 3. Timed-out SSH commands weren't retried

Even after adding timeouts, Node kills the process with an `ETIMEDOUT` error code. This pattern wasn't in the `RECOVERABLE_SSH_ERRORS` list, so timed-out commands would fail without retry rather than being retried with the existing exponential backoff logic.

### Fix 1: Add 30s timeout to SSH commands

**Files:** `src/main/utils/execFile.ts`, `src/main/utils/remote-fs.ts`

- Added `timeout` option to `ExecOptions` interface and pass it through to Node's `execFile`
- Set `SSH_COMMAND_TIMEOUT_MS = 30000` (30 seconds) for all SSH remote filesystem commands
- This ensures hung SSH connections are killed and surfaced as errors rather than blocking forever

### Fix 2: Fix stale closure in async callbacks

**File:** `src/renderer/hooks/git/useFileTreeManagement.ts`

- Capture `session.id` at the start of the effect into a local `const sessionId`
- Use `sessionId` (stable, bound at effect start) instead of `activeSessionId` (from closure, can change) in all `.then()`, `.catch()`, and progress callbacks
- This ensures the loading completion always targets the correct session regardless of whether the user switched sessions during the load

### Fix 3: Add ETIMEDOUT to recoverable SSH errors

**File:** `src/main/utils/remote-fs.ts`

- Added `/ETIMEDOUT/i` pattern to `RECOVERABLE_SSH_ERRORS`
- When a timeout kills an SSH process, the existing retry logic now kicks in with exponential backoff (up to 3 retries)
- If all retries fail, the error propagates to the `.catch()` handler which sets `fileTreeError` and schedules a 20s retry — so the user sees an error message instead of an infinite spinner

## Testing

- All 68 existing tests pass (`execFile.test.ts`: 23 tests, `remote-fs.test.ts`: 45 tests)
- No new TypeScript errors introduced (pre-existing errors in `src/prompts/index.ts` are unrelated)

## User Impact

Before: Files panel shows "Loading remote files..." forever after SSH connection degrades (common after hours of use). Only workaround was restarting Maestro.

After: If an SSH command hangs, it times out after 30s, retries up to 3 times with backoff, and either succeeds on retry or shows a clear error with automatic retry after 20s. Session switching during loading no longer corrupts loading state.

## Summary by CodeRabbit

* **Bug Fixes**
  * SSH commands now timeout after 30 seconds and are retried when timeouts occur.
  * File-tree loading fixes prevent stale session state and ensure progress/errors are tied to the correct load session.
  * Recursive directory reads now skip unreadable subfolders instead of aborting the entire tree load.

* **New Features**
  * Optional configurable process timeouts: external commands can be killed after a set time, preserve partial output, and report timeout status.

### DIFF
--- a/src/main/utils/execFile.ts
+++ b/src/main/utils/execFile.ts
@@ -6,6 +6,8 @@ const execFileAsync = promisify(execFile);
 
 export interface ExecOptions {
 	input?: string; // Content to write to stdin
+	/** Timeout in milliseconds. If the process exceeds this, it is killed and an error is returned. */
+	timeout?: number;
 }
 
 // Maximum buffer size for command output (10MB)
@@ -90,11 +92,14 @@ export async function execFileNoThrow(
 	// Handle backward compatibility: options can be env (old signature) or ExecOptions (new)
 	let env: NodeJS.ProcessEnv | undefined;
 	let input: string | undefined;
+	let timeout: number | undefined;
 
 	if (options) {
-		if ('input' in options) {
+		if ('input' in options || 'timeout' in options) {
 			// New signature with ExecOptions
-			input = options.input;
+			const execOpts = options as ExecOptions;
+			input = execOpts.input;
+			timeout = execOpts.timeout;
 		} else {
 			// Old signature with just env
 			env = options as NodeJS.ProcessEnv;
@@ -103,7 +108,7 @@ export async function execFileNoThrow(
 
 	// If input is provided, use spawn instead of execFile to write to stdin
 	if (input !== undefined) {
-		return execFileWithInput(command, args, cwd, input);
+		return execFileWithInput(command, args, cwd, input, timeout);
 	}
 
 	try {
@@ -118,6 +123,7 @@ export async function execFileNoThrow(
 			encoding: 'utf8',
 			maxBuffer: EXEC_MAX_BUFFER,
 			shell: useShell,
+			timeout,
 		});
 
 		return {
@@ -144,7 +150,8 @@ async function execFileWithInput(
 	command: string,
 	args: string[],
 	cwd: string | undefined,
-	input: string
+	input: string,
+	timeout?: number
 ): Promise<ExecResult> {
 	return new Promise((resolve) => {
 		const isWindows = process.platform === 'win32';
@@ -158,6 +165,16 @@ async function execFileWithInput(
 
 		let stdout = '';
 		let stderr = '';
+		let killed = false;
+
+		// spawn() doesn't support timeout natively, so implement it manually
+		let timer: ReturnType<typeof setTimeout> | undefined;
+		if (timeout && timeout > 0) {
+			timer = setTimeout(() => {
+				killed = true;
+				child.kill();
+			}, timeout);
+		}
 
 		child.stdout?.on('data', (data) => {
 			stdout += data.toString();
@@ -168,14 +185,16 @@ async function execFileWithInput(
 		});
 
 		child.on('close', (code) => {
+			if (timer) clearTimeout(timer);
 			resolve({
 				stdout,
-				stderr,
-				exitCode: code ?? 1,
+				stderr: killed ? `${stderr}\nETIMEDOUT: process timed out after ${timeout}ms` : stderr,
+				exitCode: killed ? 'ETIMEDOUT' : (code ?? 1),
 			});
 		});
 
 		child.on('error', (err) => {
+			if (timer) clearTimeout(timer);
 			resolve({
 				stdout: '',
 				stderr: err.message,
--- a/src/main/utils/remote-fs.ts
+++ b/src/main/utils/remote-fs.ts
@@ -68,7 +68,7 @@ export interface RemoteFsDeps {
  */
 const defaultDeps: RemoteFsDeps = {
 	execSsh: (command: string, args: string[]): Promise<ExecResult> => {
-		return execFileNoThrow(command, args);
+		return execFileNoThrow(command, args, undefined, { timeout: SSH_COMMAND_TIMEOUT_MS });
 	},
 	buildSshArgs: (config: SshRemoteConfig): string[] => {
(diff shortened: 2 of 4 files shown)


croakingtoad:
I haven't been able to test this fix as I'm on Windows and the idea of having to build the executable gives me a headache. :)

croakingtoad:
**Update:** Fixed a gap in the initial implementation where `execFileWithInput` (the `spawn`-based code path used when stdin input is provided) did not support the new `timeout` parameter — timeout was silently ignored when both `input` and `timeout` were specified.

Since `spawn()` doesn't have a built-in `timeout` option like `execFile`, I implemented it manually with `setTimeout` + `child.kill()`. When the timeout fires, the process is killed and the result includes an `ETIMEDOUT` exit code and descriptive stderr message, consistent with how Node's `execFile` timeout behaves on the `execFileAsync` path.

The timer is also properly cleaned up in both the `close` and `error` handlers to prevent leaks.

croakingtoad:
**Update:** Addressed both CodeRabbit findings:

**1. ETIMEDOUT retry not firing (remote-fs.ts)** — Good catch. `execFileAsync` puts `'ETIMEDOUT'` in `error.code` which surfaces as `result.exitCode`, not in stderr. The `/ETIMEDOUT/i` regex only checked stderr/stdout, so it never matched for Node.js-triggered timeouts. Fixed by adding an explicit `result.exitCode === 'ETIMEDOUT'` check in `execRemoteCommand` alongside the existing stderr pattern matching:

```ts
const isNodeTimeout = result.exitCode === 'ETIMEDOUT';
if ((isRecoverableSshError(combinedOutput) || isNodeTimeout) && attempt < maxRetries) {
```

I've kept the `/ETIMEDOUT/i` regex in the patterns list as well — it covers the case where the SSH client itself writes "ETIMEDOUT" to stderr (socket-level timeout), which is a distinct scenario from Node killing the process.

**2. Stale closure in stats migration effect (useFileTreeManagement.ts)** — Also valid. The stats migration `useEffect` (line ~487-531) had the same `activeSessionId` stale closure issue in its `.then()` callback. Applied the same fix: capture `session.id` into a local `const sessionId` and use it in the async callbacks.

croakingtoad:
**Update:** Fixed a fourth issue discovered during testing — `directorySize` failure was killing the entire file tree load.

**Problem:** `loadFileTree` and `directorySize` were coupled in a `Promise.all`. If the `du` SSH command failed (timeout, permission error, etc.), `Promise.all` rejected immediately and the `.catch()` handler set `fileTreeError` — preventing the file tree from ever showing, even when `loadFileTree` would have succeeded.

This bug was always present but invisible: previously `loadFileTree` hung forever so `Promise.all` never settled, meaning `directorySize` failures were silently swallowed. The timeout fix unmasked it.

**Fix:** Decoupled the two operations. `directorySize` errors are now caught independently and logged as a non-fatal warning. The file tree loads and displays regardless of whether stats (`du`) succeed. If stats fail, `fileTreeStats` is simply `undefined` and the size/count display is omitted — the file list still renders normally.

croakingtoad:
**Update:** Fixed a fifth issue discovered during testing — a single unreadable subdirectory was killing the entire file tree walk.

**Problem:** `loadFileTreeRecursive` (`fileExplorer.ts`) processes directories sequentially with `await`. If any child directory `readDir` failed (paths with spaces over SSH, permission denied, broken symlinks, etc.), the error propagated all the way up via `throw error`, aborting the entire tree walk. In testing, a directory named `Dr. Sue Johnson` triggered this over SSH.

**Fix:** Wrapped the recursive child directory call in a try/catch. If a subdirectory can't be read, it's shown as an empty folder and the walk continues with the remaining directories. The root directory failure still propagates (no tree to show at all), but child directory failures are now gracefully handled.

**New file changed:** `src/renderer/utils/fileExplorer.ts`

PR #489, opened 2026-03-01
fix: add OPENCODE_CLI_RUN_MODE env var to prevent Auto Run hang with oh-my-opencode plugin

### LINKED ISSUE #338: Auto Run: agent hangs silently with no progress indication or timeout
## Problem

Auto Run sessions can hang indefinitely with no visible indication to the user that something has gone wrong. Multiple users have reported this:

- Agent process remains alive but produces no output for hours
- Progress percentage never updates, leaving users unsure if it's working or stuck
- No timeout or stall detection at the playbook execution level
- No thought stream / activity log to help diagnose *why* a hang occurred

### User Reports

> "Some of the playbooks would run for hours with no indication, the percentage never seemed to change" — @waynez

> "Sometimes my auto runs will get stuck. Recently it was because the node path was not available to the agent - just kept updating the playbook with that note" — @calebpeffer

> "The claude process was still there. Was stuck for like 6 hours doing nothing. I'm not sure what the hell happened." — @pedramamini

## Root Causes (Known/Suspected)

1. **Agent environment issues** — Missing `node` in PATH causes the agent to loop writing notes instead of executing tasks
2. **Agent process stall** — Process alive but not producing output (unknown root cause, needs investigation)
3. **No stall detection at Auto Run level** — The existing stall detection (`stall-count` in batch processor) only tracks task-level progress, not agent-level activity

### 1. Enhanced Auto Run Logging / Thought Stream
Add a visible activity log or "thought stream" to Auto Run that shows:
- Agent spawn/exit events with timestamps
- Last output received from agent (truncated)  
- Current task being processed
- Time elapsed on current task

This gives users visibility into what's happening and helps diagnose hangs.

### 2. Agent-Level Heartbeat / Activity Timeout
Detect when a spawned agent process has produced no output for an extended period:
- Track last output timestamp per agent during Auto Run
- After configurable threshold (e.g., 10 minutes of silence), surface a warning
- After extended threshold (e.g., 30 minutes), offer to kill and retry or skip

### 3. Better Progress Indication
- Show per-task elapsed time in the Auto Run UI
- Show agent activity indicator (last output timestamp, "active" vs "idle")
- Surface the current task description being worked on

## Related

- Existing stall detection: `src/renderer/hooks/batch/useBatchProcessor.ts` (tracks document-level stalling, not agent-level)
- Process management: `src/main/process-manager.ts`
- Batch processor: `src/renderer/hooks/batch/useBatchProcessor.ts`, `useDocumentProcessor.ts`

Chocothin:
## Summary

Fixes #338 — addresses the root cause of OpenCode agent hanging silently during Auto Run when the [oh-my-opencode](https://github.com/nicepkg/oh-my-opencode) plugin is installed.

## Problem

When oh-my-opencode plugin is present, it checks `process.env.OPENCODE_CLI_RUN_MODE` to determine question tool permissions:

```javascript
// oh-my-opencode/dist/index.js:75452-75453
const isCliRunMode = process.env.OPENCODE_CLI_RUN_MODE === "true";
const questionPermission = isCliRunMode ? "deny" : "allow";
```

Without `OPENCODE_CLI_RUN_MODE` set, the plugin overrides Maestro's `question: "deny"` config (set via `OPENCODE_CONFIG_CONTENT`) with `question: "allow"` at runtime. This causes the agent to prompt for stdin input during Auto Run, hanging indefinitely with no progress indication or timeout.

## Fix

Add `OPENCODE_CLI_RUN_MODE: 'true'` to OpenCode's `defaultEnvVars` in agent definitions. This signals to the plugin that OpenCode is running in headless/batch mode, preserving the `question: "deny"` permission.

**Changes:** 1 line addition + explanatory comments in `src/main/agents/definitions.ts`

## Why This Works

| `OPENCODE_CLI_RUN_MODE` | Plugin behavior | Result |
|---|---|---|
| Not set (current) | Sets `question: "allow"` | Agent hangs waiting for stdin |
| `"true"` (this fix) | Sets `question: "deny"` | Agent runs without blocking |

- **Safe for environments without the plugin** — the env var has no effect when oh-my-opencode is not installed
- **Doesn't conflict with existing config** — works alongside `OPENCODE_CONFIG_CONTENT` which already sets `question: "deny"`

## Testing

- [x] `npm run lint` (tsc) — passes
- [x] `npm run lint:eslint` — passes  
- [x] `npx vitest run` (envBuilder tests, 37 tests) — all pass
- [x] Verified against oh-my-opencode v3.9.0 source code
- [x] Manually tested in production environment (OpenCode v1.2.15 + oh-my-opencode + Maestro v0.14.5) with `customEnvVars` workaround — confirmed the fix resolves the hang

## Scope Note

This PR addresses the plugin interaction root cause of #338. The broader issue (no timeout/progress indication for any hung agent) may warrant a separate enhancement.

## Summary by CodeRabbit

* **New Features**
  * Added support for headless/batch operation mode signaling to plugins, enabling them to adapt behavior when running in automated environments.

### DIFF
--- a/src/main/agents/definitions.ts
+++ b/src/main/agents/definitions.ts
@@ -221,7 +221,12 @@ export const AGENT_DEFINITIONS: AgentDefinition[] = [
 		// - "question": false in tools block (original approach)
 		// The question tool waits for stdin input which hangs batch mode
 		// Users can override by setting customEnvVars in agent config
+		//
+		// OPENCODE_CLI_RUN_MODE: Signals to plugins (e.g., oh-my-opencode) that OpenCode is
+		// running in headless/batch mode. Without this, plugins may override permission settings
+		// and re-enable the question tool, causing Auto Run to hang indefinitely. See #338.
 		defaultEnvVars: {
+			OPENCODE_CLI_RUN_MODE: 'true',
 			OPENCODE_CONFIG_CONTENT:
 				'{"permission":{"*":"allow","external_directory":"allow","question":"deny"},"tools":{"question":false}}',
 		},


Chocothin:
Closing this PR after further analysis. 

`OPENCODE_CLI_RUN_MODE` is an internal variable specific to the oh-my-opencode plugin, not an official OpenCode env var. Hardcoding a third-party plugin's internal variable into Maestro's core is not the right approach.

The proper fix should be in oh-my-opencode itself — the plugin should respect `question: "deny"` from `OPENCODE_CONFIG_CONTENT` instead of overriding it with `question: "allow"` when `OPENCODE_CLI_RUN_MODE` is not set.

Users affected by this can work around it today by adding `OPENCODE_CLI_RUN_MODE: "true"` to their agent's `customEnvVars` in Maestro settings.

Will file an issue on the oh-my-opencode repo instead.

PR #506, opened 2026-03-03
fix: auto-focus input when toggling AI/Shell mode

kianhub:
## Summary
- Pressing Cmd+J to toggle between AI and Shell mode now auto-focuses the input field, so users can start typing immediately without an extra click
- Follows the existing focus pattern used by the `newTab` shortcut in the same handler

## Test plan
- [x] Press Cmd+J to switch to Shell mode — cursor should be in the input field
- [x] Press Cmd+J again to switch back to AI mode — cursor should be in the input field
- [x] Verify no regression: Cmd+T (new tab) still auto-focuses input

## Summary by CodeRabbit

* **New Features**
  * Toggle-mode shortcut now automatically focuses the main input for quicker interaction.
  * New tabs auto-focus the input so you can start typing immediately.
  * Input-focus timing standardized for more consistent, smoother UX.

* **Tests**
  * Added focus-related tests that simulate timing to validate and stabilize input-focus behavior.

### DIFF
--- a/src/__tests__/renderer/hooks/useMainKeyboardHandler.test.ts
+++ b/src/__tests__/renderer/hooks/useMainKeyboardHandler.test.ts
@@ -462,9 +462,12 @@ describe('useMainKeyboardHandler', () => {
 		});
 
 		it('should allow toggleMode shortcut (Cmd+J) when only overlays are open', () => {
+			vi.useFakeTimers();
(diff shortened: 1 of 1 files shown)


kianhub:
failing lint-test comes from files that were not touched in this PR.

pedramamini:
only thing is i'm working on replacing the shell with xterm.js ... though that'll take a while. can take this for the time being

PR #519, opened 2026-03-05
fix: reset resultEmitted on each OpenCode step_start to capture final summary (#512)

### LINKED ISSUE #512: OpenCode - Incomplete Agent Response
I am sometimes getting instances like this where it seems the agent responds with the first step of the task but doesn't provide the rest of the thinking or a summary of what it actually completed, even though these messages are in the opencode session.    In this case it happened twice, it told my the first step of what it was going to do, but never provided the summary of the completed task.

### Maestro
Image

### Opencode
Image
.....Lots More Thinking....

Image

chr1syy:
## Summary

Fixes #512 — OpenCode incomplete agent response where Maestro shows only the first step but not the completed task summary.

- **Root cause:** OpenCode emits multiple steps per run (`step_start → [text] → tool_use → step_finish(tool-calls) → repeat → step_start → text → step_finish(stop)`). The `resultEmitted` gate locked on the **first** intermediate `text` event, silently dropping all subsequent text events — including the final summary.
- **Fix:** Reset `resultEmitted` and `streamedText` on each new `step_start` for OpenCode only, so the last `text` event (the real answer) always gets through. Intermediate thinking texts also flow through, matching OpenCode's native UI behavior.
- **Scope:** OpenCode-only guard — `claude-code`, `codex`, and other agents are unaffected.

## Test plan

- [x] 3 new unit tests in `StdoutHandler.test.ts`:
  - Multi-step scenario: both intermediate thinking and final summary are emitted
  - Gate reset: `resultEmitted` resets to `false` on each `step_start` for opencode
  - Isolation: `resultEmitted` is NOT reset on `step_start` for `claude-code`
- [x] All 41 `StdoutHandler` tests pass
- Manual verification: ran `opencode run --format json "List all .ts files..."` and confirmed the output pattern matches what the fix addresses (intermediate `text` in step 2 + final summary in step 3)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed OpenCode multi-step workflows to emit the final step result correctly and reset per-step state.

* **Improvements**
  * Per-field agent configuration now persists on blur with safer async error handling and more reliable model selection commits.

* **Tests**
  * Added tests covering multi-step OpenCode result emission and per-step isolation.

* **Windows Improvements**
  * Improved stdin delivery logic for agent spawning on Windows to ensure reliable prompt delivery.

### DIFF
--- a/src/__tests__/renderer/components/Settings/tabs/EncoreTab.test.tsx
+++ b/src/__tests__/renderer/components/Settings/tabs/EncoreTab.test.tsx
@@ -1004,7 +1004,7 @@ describe('EncoreTab', () => {
 
 			expect(window.maestro.agents.setConfig).toHaveBeenCalledWith(
 				'claude-code',
-				expect.objectContaining({ model: 'test-model' })
+				expect.objectContaining({ model: 'claude-3-opus' })
 			);
 		});
 	});


chr1syy:
@coderabbitai re-review

PR #526, opened 2026-03-07
Terminal Upgrade!

pedramamini:
Massive terminal upgrade! We use xterm.js now and it opens as a tab side-by-side with our AI chat and file preview tabs. Multiple terminals can be opened. Terminals can be renamed. Content is persistent between focus changes.

## Summary by CodeRabbit

* **New Features**
  * Persistent PTY-backed terminal tabs: create, switch, close, rename, and restore across sessions
  * SSH-backed remote terminals and per-tab shell/working-directory support
  * In-terminal search with previous/next navigation and keyboard shortcuts

* **Improvements**
  * Terminal tab lifecycle and unified tab integration (AI, file, terminal)
  * Keyboard shortcuts: New Terminal Tab and Clear Terminal

* **Removed**
  * Terminal width configuration setting

* **Documentation**
  * Added Terminal Tab System docs describing usage and behavior


pedramamini:
@greptile @coderabbitai i need your help on this one. how's it look?

pedramamini:
@greptile look again. review please. this is a massive change. major upgrade!

pedramamini:
@greptile @coderabbitai All findings from the initial review have been addressed in commit 22980b74. Summary of fixes:

- **🔴 Command injection (SSH workingDirOverride)**: Switched from `JSON.stringify` (which does not suppress `$(...)` expansion in double-quoted strings) to single-quoting with `'"'"'`-escaped apostrophes in `process.ts`
- **🔴 Duplicate `vi.mock`**: Removed the shadowed duplicate `vi.hoisted`/`vi.mock` block in `UnifiedHistoryTab.test.tsx`
- **🟡 Indentation regressions**: Fixed missing leading tabs on `chatRawTextMode` in `MainPanel.tsx` and the `isVisible` effect block in `TerminalView.tsx`
- **🟡 Deprecation warn spam**: Gated `process:runCommand` deprecation log behind a one-time `runCommandDeprecationWarned` flag
- **🟡 Missing `onTabStateChange` dep**: Added `eslint-disable-line` comment on the `onExit` useEffect in `TerminalView.tsx` (callback uses `getState()` internally, so the reference is irrelevant)
- **🟡 Stale `mountedTerminalSessionsRef`**: Added a `mountedTerminalTabStateKey` Zustand subscription that tracks terminal tab states for all mounted sessions, and updated the render loop to use `useSessionStore.getState()` as the live source for non-active sessions — fixing the silent exit message drop
- **🟡 Unhandled `process.kill()` promise**: Added `.catch(() => {})` in `closeTerminalTab` to handle the expected race where PTY exits before tab removal
- **⚪ Missing trailing newline**: Fixed in `TabSwitcherModal.test.tsx`

Please re-review when ready.

pedramamini:
@greptile — Your latest inline comments appear to be against the original diff rather than the updated commit (`22980b74`). All 5 code issues you flagged have been fixed in that commit (command injection, both indentation regressions, stale session snapshot, and the unhandled `process.kill()` promise) — CodeRabbit verified and approved after checking the same commit.

The `TerminalOutput.tsx` auto-scroll button removal is **intentional** — removing the floating scroll-to-bottom indicator was the primary goal of this PR (see commit `73e81ebd`). There is no regression there.

pedramamini:
@greptile @coderabbitai Second round of fixes landed in commit `fdb5ff1c`. Summary:

- **🔴 Hard-coded `zsh` in `ProcessManager`**: Now uses `process.env.SHELL ?? '/bin/bash'` (Linux-safe), `process.env.ComSpec ?? 'powershell.exe'` on Windows
- **🔴 `useInputMode` terminal toggle with no tabs**: Added guard — if `terminalTabs.length === 0`, toggle is a no-op (prevents `inputMode='terminal'` with nothing to render)
- **🔴 `useDebouncedPersistence` skipping terminal cleanup**: Moved PTY runtime state reset (`pid/state/exitCode`) before the early-return so terminal-only sessions are also cleaned on persist
- **🔴 `useAgentListeners` shell log gate too coarse**: Removed `terminalTabs.length` check — `runCommand` exit events now always record to `shellLogs` regardless of whether terminal tabs exist
- **🟡 `MainPanel` inline handler factories**: Memoized `createTabStateChangeHandler`/`createTabPidChangeHandler` per session ID via `useMemo` — preserves `TerminalView`'s `memo()` contract
- **🟡 `TerminalView` missing `session.projectRoot` dep**: Added to `spawnPtyForTab` `useCallback` dep array
- **🟡 Test fixtures**: `useMergeSession` mock gains `terminalTabs: [], activeTerminalTabId: null`; `useSessionCrud` gains `activeTerminalTabId` null assertion; `tabHelpers` navigation fixtures gain real terminal tab entries

Note on `TerminalTabRenameModal` `onSave` type: no change needed — `renameTerminalTabHelper` already converts empty string → `null` (`name === '' ? null : name`), so callers don't need to special-case it.

Note on `useBatchedSessionUpdates` timing edge case: acknowledged but deferred — requires a more significant refactor and the window is narrow in practice.

PR #540, opened 2026-03-09
 feat: persistent web link

jSydorowicz21:
Adds a toggle in the Live overlay panel that lets users lock in their                                                   
  current web server URL. When enabled, the security token is saved to
  settings and reused on future startups instead of generating a fresh
  one each time. Handy for anyone sharing the link on a local network
  who doesn't want it to break every time Maestro restarts.
  
  Default state is off

  Toggling ON immediately persists the running server's token (no restart
  needed). Toggling OFF clears the stored token so a fresh one is
  generated next time.

## Summary by CodeRabbit

* **New Features**
  * "Persistent Web Link" toggle to persist or clear the web access token across restarts; ability to persist the current token on demand.

* **Accessibility**
  * Settings checkbox exposes aria-pressed for assistive technologies.

* **Settings**
  * Adds persistentWebLink setting (default off) with UI/state wiring and persistence support.

* **Tests**
  * Added tests covering token generation, reuse, and persistence behaviors.

### DIFF
--- a/src/__tests__/main/ipc/handlers/web.test.ts
+++ b/src/__tests__/main/ipc/handlers/web.test.ts
@@ -34,6 +34,7 @@ describe('web handlers', () => {
 	let mockWebServer: any;
 	let webServerRef: { current: any };
 	let mockCreateWebServer: any;
+	let mockSettingsStore: any;
 
 	beforeEach(() => {
 		vi.clearAllMocks();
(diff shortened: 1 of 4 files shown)


pedramamini:
This feels risky as default behavior though. With no auth, that key is what gives someone full access to your system. Can you paste some screenshots of what this looks like? Maybe we add a default off toggle for "persitent key" just like we have for custom port

jSydorowicz21:
> This feels risky as default behavior though. With no auth, that key is what gives someone full access to your system. Can you paste some screenshots of what this looks like? Maybe we add a default off toggle for "persitent key" just like we have for custom port

Heyo! I was still working on this, the default is off since it's better to play safe for those who may expose their instance to the public. It's set in src/main/stores/defaults.ts as persistentWebLink: false
We could consider adding addition warnings on what leaving the setting on entails if you'd like

{BA66EE62-B723-4BDF-97FA-54B3582835FF}

jSydorowicz21:
@pedramamini 
Pr should be good to go, squashed into 2 instead of 13. Greptile on my repo reports 4/5 with just test complaints for an edge of an edge case please let me know if you have any other questions, I'm in the discord if you prefer to reach out there @ devil920

PR #544, opened 2026-03-09
fix: clear stale agentSessionId on session_not_found loop

openasocket:
## Summary

- When a Claude session expires or is deleted, the stale `agentSessionId` was never cleared — every subsequent prompt tried `--resume` with the dead ID, received `session_not_found`, and looped endlessly
- Now clears `agentSessionId` at both tab and session level on `session_not_found` errors, so the next prompt starts a fresh session automatically

## Test plan

- [x] Unit test added: verifies `agentSessionId` is cleared at both tab and session level on `session_not_found`
- [ ] End-to-end: let a Claude session expire, send a new prompt — should recover without manual intervention

## Summary by CodeRabbit

* **Bug Fixes**
  * Session identifiers are now properly cleared when a session-not-found error occurs, ensuring correct state management and proper cleanup for subsequent interactions.

### DIFF
--- a/src/__tests__/renderer/hooks/useAgentListeners.test.ts
+++ b/src/__tests__/renderer/hooks/useAgentListeners.test.ts
@@ -691,6 +691,39 @@ describe('useAgentListeners', () => {
 			expect(agentErrorOpen).toBe(false);
 		});
 
+		it('clears agentSessionId on session_not_found so next prompt starts fresh', () => {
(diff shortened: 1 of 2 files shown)


pedramamini:
Hey! We're consolidating onto the `rc` branch — `0.16.0-RC` is being retired since `rc` already contains all its commits. Could you retarget this PR to `rc`? Thanks!

PR #548, opened 2026-03-10
fix: wizard SSH remote support — resume, persistence, and file operations

rothnic:
## Summary

Fixes wizard SSH remote session support. Documents now save on the remote host instead of locally, and wizard resume preserves SSH configuration.

## Changes

| Issue | Fix |
|-------|-----|
| Wizard resume not triggering | Added resume check in `App.tsx` when clicking Wizard button |
| SSH config not persisted | Added `sessionSshRemoteConfig` to serializable state in `WizardContext.tsx` |
| Files saved locally (not remote) | Pass `sshRemoteId` to `watchFolder()` and `saveDocuments()` → `writeDoc()` |

## Files Changed

- `src/renderer/App.tsx`
- `src/renderer/components/Wizard/WizardContext.tsx`
- `src/renderer/components/Wizard/screens/PreparingPlanScreen.tsx`
- `src/renderer/components/Wizard/services/phaseGenerator.ts`

## Testing

- [x] Resume modal appears when clicking Wizard with saved progress
- [x] Resume restores correct step with SSH config preserved
- [x] "Start Fresh" clears state correctly
- [x] Documents save on remote host (not locally)

Closes #547

### DIFF
--- a/src/renderer/components/Wizard/screens/PreparingPlanScreen.tsx
+++ b/src/renderer/components/Wizard/screens/PreparingPlanScreen.tsx
@@ -768,6 +768,9 @@ export function PreparingPlanScreen({ theme }: PreparingPlanScreenProps): JSX.El
 
 							// Save documents to disk in "Initiation" subfolder
 							setProgressMessage('Saving documents...');
+							const sshRemoteId = state.sessionSshRemoteConfig?.enabled
+								? (state.sessionSshRemoteConfig.remoteId ?? undefined)
+								: undefined;
 							const saveResult = await phaseGenerator.saveDocuments(
 								state.directoryPath,
 								genResult.documents,
@@ -775,7 +778,8 @@ export function PreparingPlanScreen({ theme }: PreparingPlanScreenProps): JSX.El
 									// Add file to the created files list as it's saved
 									addCreatedFile(file);
 								},
-								'Initiation' // Save in Initiation subfolder
+								'Initiation', // Save in Initiation subfolder
+								sshRemoteId // Pass SSH remote ID for remote sessions
 							);
 
 							if (saveResult.success) {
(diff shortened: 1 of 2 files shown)


rothnic:
Thanks for the reviews! I've addressed the feedback:

## Changes Made

### 1. Interface Stability (Greptile's feedback)

**Issue:** The interface rename from `openWizardModal` to `handleOpenWizardWithResumeCheck` created unnecessary churn and made the prop optional, which could allow `undefined` downstream.

**Fix:** Reverted to keeping `openWizardModal: () => void` as a required prop in the interface. `App.tsx` simply passes `handleOpenWizardWithResumeCheck` as the value for `openWizardModal`. This keeps the interface stable and maintains type safety.

### 2. Debug Logging (Greptile's feedback)

**Issue:** Debug `console.log` statements left in production code.

**Decision:** Keeping the logging for now because:

- This is a user-facing flow that can fail silently (settings read, state validation)
- The logs help diagnose issues in production without Sentry noise
- The volume is low (only fires on Wizard button clicks, not on a timer/loop)
- Easy to remove later once the feature is proven stable

### 3. Using `_loadResumeState` (CodeRabbit's feedback)

**Issue:** Re-reading settings directly instead of using the existing `_loadResumeState` function.

**Decision:** Keeping the direct read because:

- `_loadResumeState` from `useWizard()` is meant for restoring state INTO the wizard context
- We need to check if resume should happen BEFORE opening anything
- The logic we need (checking step !== 'agent-selection') is specific to the resume decision, not the load operation
- Keeping it inline makes the flow explicit: check → decide → act

### 4. Error Handling (CodeRabbit's feedback)

**Issue:** Silently catching errors with just `console.error` instead of reporting to Sentry.

**Decision:** Keeping the try/catch with console.error because:

- Settings read failures are gracefully handled (falls back to opening wizard normally)
- This is a non-critical user flow - if resume fails, user just gets a fresh wizard
- Per codebase guidelines, we should only bubble exceptions for unexpected/unrecoverable errors
- A settings read failure here is recoverable (fallback behavior)

The fix is now simpler, keeps interfaces stable, and addresses the core issue: the resume modal now appears when clicking the Wizard button when saved progress exists.

PR #558, opened 2026-03-12
feat: complete Maestro Cue end-to-end wiring — real execution, prompt delivery, clean output, and session lifecycle fixes

reachrazamair:
## Summary

Maestro Cue had all the UI, YAML parsing, engine logic, and IPC infrastructure
in place, but the actual agent execution path was never wired up. Triggers would
fire, the engine would dispatch runs, and then nothing happened — `onCueRun` was
a stub. This PR completes the final mile: real process spawning, correct prompt
delivery, clean output extraction, and several lifecycle bugs that would have
caused silent failures in production.

---

### 1. `onCueRun` was a stub — no agents ever ran
`index.ts` wired a no-op `onCueRun` that returned a hardcoded `completed` result
without spawning anything. Cue subscriptions appeared to succeed but did nothing.

### 2. Prompt was never delivered to the agent process
`cue-executor.ts` called `buildAgentArgs()` which builds CLI flags only — it does
not append the prompt as a positional argument. The executor bypassed
`ChildProcessSpawner` entirely, so the prompt was always missing. For OpenCode
this produced `Error: You must provide a message or a command` (exit code 1) on
every run.

**Root cause:** `ChildProcessSpawner` handles prompt appending at lines ~170–180
via a 3-way dispatch (`promptArgs` → `noPromptSeparator` → `'--', prompt`). The
Cue executor never replicated this step. The test mock for `buildAgentArgs` was
also incorrectly including `'--', 'prompt-content'`, masking the bug.

**Fix:** After `applyAgentConfigOverrides`, append the prompt using the same
3-way logic, guarded by `!sshRemoteConfig?.enabled` so SSH runs (which embed the
prompt via `wrapSpawnWithSsh`) are not double-appended.

### 3. Activity Log showed raw NDJSON instead of readable text
Agents like OpenCode run with `--format json`, producing NDJSON streams:
{"type":"step_start","part":{"type":"step-start"}}
{"type":"text","part":{"text":"Hello! Fun fact: honey never spoils..."}}
{"type":"step_finish","part":{"reason":"stop","tokens":{...}}}
The executor stored this verbatim as `CueRunResult.stdout`, which is what the
Activity Log and downstream `{{CUE_SOURCE_OUTPUT}}` template variable both read.

**Fix:** Post-process stdout through the agent's registered output parser
(`getOutputParser(toolType)`). Lines that produce `type: 'result'` events (mapped
from OpenCode's `{"type":"text","part":{"text":"..."}}`) are extracted and joined
as clean text. Falls back to raw stdout for plain-text agents or when no parser
is registered. The clean text is also what gets passed to chained downstream
agents via `agent.completed` → `{{CUE_SOURCE_OUTPUT}}`.

### 4. Session removal called `refreshSession` instead of `removeSession`
`useCueAutoDiscovery` handled session removals by calling
`refreshSession(prevId, '')` — refreshing with an empty path rather than actually
removing the session from the engine. This left ghost sessions in the engine's
session map after agents were deleted.

### 5. Encore feature toggle-on skipped `enable()`
When the `maestroCue` Encore Feature was toggled on, the hook refreshed all
sessions but never called `window.maestro.cue.enable()`, so the engine never
started. The dashboard would show sessions but no triggers would ever fire.

### 6. `removeSession` IPC channel was missing entirely
`cue:removeSession` had no preload binding and no type declaration in
`global.d.ts`, making it impossible to call from the renderer.

### 7. `handleEditYaml` in `CueModal` was a stub
The "Edit YAML" button handler contained a TODO comment with no implementation,
so clicking it did nothing.

### 8. `onCueRun` signature mismatch
The engine passed a structured object to `onCueRun` but the handler signature
expected positional parameters, causing a TypeScript error and runtime failure.

---

### `src/main/cue/cue-executor.ts`
- Import `getOutputParser` from `../parsers` and `ToolType` from shared types
- Add `extractCleanStdout(rawStdout, toolType)` helper that splits stdout by
  newlines, passes each line through `parser.parseJsonLine()`, collects all
  `type: 'result'` event text, and joins them. Falls back to raw stdout if the
  parser is absent or finds no result events.
- Use `extractCleanStdout` in the `finish()` closure before resolving
  `CueRunResult`
- Add prompt-appending block after `applyAgentConfigOverrides` for local
  execution: `promptArgs` → `noPromptSeparator` → `'--', prompt` (mirrors
  `ChildProcessSpawner`)
- SSH block comment clarified: `wrapSpawnWithSsh` embeds the prompt itself via
  `config.prompt`; pre-appending to `finalArgs` would double-append

### `src/main/cue/cue-engine.ts`
- `CueEngineDeps.onCueRun` signature changed from positional `(sessionId, prompt,
  event)` to a request object `{ runId, sessionId, prompt, subscriptionName,
  event, timeoutMs }`
- Add optional `onStopCueRun?: (runId: string) => boolean` to deps
- `start()` and `stop()` are now idempotent (`if (this.enabled) return` guards)
- DB persistence: `recordCueEvent()` on run start, `updateCueEventStatus()` on
  finish/stop
- Add `private manuallyStoppedRuns = new Set<string>()` to prevent double-logging
  when a run is stopped mid-flight
- `stopRun()` calls `this.deps.onStopCueRun?.(runId)` and adds to
  `manuallyStoppedRuns`

### `src/main/index.ts`
- Wire real `onCueRun` implementation: builds full `TemplateContext`, resolves
  session from store, calls `executeCuePrompt` with SSH config + agent config
  overrides, records history entry via `recordCueHistoryEntry`
- Add `onStopCueRun: (runId) => stopCueRun(runId)`
- Move `getAgentConfigForAgent` / `getCustomEnvVarsForAgent` to module scope
  (needed inside `whenReady`)
- Fix `projectRoot` priority: `projectRoot || cwd || fullPath || homedir()`

### `src/main/ipc/handlers/cue.ts`
- Register `cue:removeSession` IPC handler

### `src/main/preload/cue.ts`
- Expose `removeSession: (sessionId) => ipcRenderer.invoke('cue:removeSession', { sessionId })`

### `src/renderer/global.d.ts`
- Add `removeSession: (sessionId: string) => Promise<void>` to `MaestroAPI.cue`

### `src/renderer/hooks/useCueAutoDiscovery.ts`
- Session removal: `window.maestro.cue.removeSession(prevId)` instead of
  `refreshSession(prevId, '')`
- Feature toggle-on: call `enable()` first, then `Promise.all(refreshSession)`

### `src/renderer/components/CueModal.tsx`
- `handleEditYaml`: replace stub with
  `getModalActions().openCueYamlEditor(session.sessionId, session.projectRoot)`

---

## Tests

| File | Changes |
|------|---------|
| `cue-executor.test.ts` | Fix `mockBuildAgentArgs` (was incorrectly including prompt, masking the bug); add 4 prompt-appending tests; add mock for `getOutputParser`; add 4 stdout-extraction tests |
| `cue-engine.test.ts` | Update `onCueRun` mock to object params; add `onStopCueRun` mock; update all `toHaveBeenCalledWith` assertions; add `stopRun signals executor callback` test |
| `cue-ipc-handlers.test.ts` | Add missing mock methods (`removeSession`, `getSettings`, `triggerSubscription`, `getQueueStatus`, `getGraphData`); add `cue:removeSession` describe block |
| `useCueAutoDiscovery.test.ts` | Add `mockRemoveSession`; update removal test to assert `removeSession` is called; add enable/disable toggle tests |
| `CueModal.test.tsx` | Mock `modalStore`; change "handler is a stub" test to verify `openCueYamlEditor` is called with correct args |
| `UsageDashboard/*.test.tsx` | Add missing `Briefcase` and `Coffee` icon mocks |

---

## Test plan

- Create a session with a heartbeat trigger (5 min interval), attach a
  prompt, confirm the Activity Log shows a completed run with readable output
- Chain two agents via `agent.completed` + `source_session`; confirm Agent B
  receives Agent A's clean output via `{{CUE_SOURCE_OUTPUT}}`
- Toggle Maestro Cue OFF then ON in Encore Features; confirm the engine
  restarts and sessions are re-scanned
- Delete an agent that has Cue configured; confirm no ghost sessions remain
- Click "Edit YAML" on a session in the Cue dashboard; confirm the YAML
  editor opens
- Run `npm run test` — 0 failures

### DIFF
--- a/src/__tests__/renderer/components/QuickActionsModal.test.tsx
+++ b/src/__tests__/renderer/components/QuickActionsModal.test.tsx
@@ -1751,7 +1751,7 @@ describe('QuickActionsModal', () => {
 			expect(dialog).toHaveAttribute('aria-label', 'Switch Agent');
 		});
 
-		it('shows only agent jump actions in agents mode', () => {
+		it('shows only raw agent names in agents mode', () => {
 			const props = createDefaultProps({
 				initialMode: 'agents',
 				sessions: [
@@ -1761,9 +1761,10 @@ describe('QuickActionsModal', () => {
 			});
 			render(<QuickActionsModal {...props} />);
 
-			// Agent jump actions should be visible
-			expect(screen.getByText('Jump to: Agent Alpha')).toBeInTheDocument();
-			expect(screen.getByText('Jump to: Agent Beta')).toBeInTheDocument();
+			// Agent names should be shown without "Jump to:" prefix
+			expect(screen.getByText('Agent Alpha')).toBeInTheDocument();
+			expect(screen.getByText('Agent Beta')).toBeInTheDocument();
+			expect(screen.queryByText(/Jump to/)).not.toBeInTheDocument();
 
 			// Non-agent actions should NOT be present
 			expect(screen.queryByText('Create New Agent')).not.toBeInTheDocument();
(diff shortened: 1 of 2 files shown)


pedramamini:
this PR is quite early. we'll do this later.

PR #597, opened 2026-03-20
feat: add Cursor agent support (#480)

### LINKED ISSUE #480: Feature: Cursor (CLI) agent support
Cursor recently released a CLI-based agent: https://cursor.com/docs/cli/overview

It'd be cool if that could be supported by Maestro as well. I sometimes use both Claude & Cursor in my dev workflows.

jSydorowicz21:
## Summary

Closes #480.

- Add Cursor (via its `agent` CLI) as a new supported agent in Maestro
- Full integration: agent ID, definition, capabilities, metadata, context window, output parser, error patterns, session storage, tests, and documentation

## Changes

- Register `cursor` in AGENT_IDS tuple
- Add agent definition with CLI args (`-p --output-format stream-json`), resume (`--resume`), read-only (`--mode ask`), YOLO (`--force`), model selection (`--model`)
- Define 23 capability flags (resume, JSON output, session ID, image input, streaming, thinking display, etc.)
- Add metadata with Beta status
- Set default context window (200,000 tokens)
- Implement `CursorOutputParser` (410 lines) for stream-json output format
- Register 158 error patterns across 7 categories (auth, tokens, rate limit, network, permissions, crash, session)
- Implement `CursorSessionStorage` as no-op (Cursor storage format not yet documented)
- Add comprehensive tests: parser (556 lines), error patterns (177 lines), storage (94 lines), integration (44 lines)
- Add CLI research document

## Test plan

- [ ] Cursor agent appears in agent list when CLI is installed
- [ ] Cursor agent spawns correctly with `agent -p --output-format stream-json`
- [ ] Output parser handles all stream-json event types
- [ ] Error patterns correctly categorize Cursor CLI errors
- [ ] Resume works with `--resume` flag
- [ ] Model selection works with `--model` flag
- [ ] Tests pass (`npm run test`)

## Summary by CodeRabbit

* **New Features**
  * Added support for the Cursor agent throughout the app, including agent selection, supported-agent lists, and beta labeling.
  * Cursor now appears as an available provider with its own default settings and command-line integration.

* **Bug Fixes**
  * Improved detection and handling for Cursor-specific CLI output and error cases.
  * Added support for recognizing Cursor sessions and related output in the app.

### DIFF
--- a/CLAUDE-AGENTS.md
+++ b/CLAUDE-AGENTS.md
@@ -8,6 +8,7 @@ Agent support documentation for the Maestro codebase. For the main guide, see [[
 | --------------- | ------------- | ---------- | ---------------------------------------------------------------- |
 | `claude-code`   | Claude Code   | **Active** | Primary agent, `--print --verbose --output-format stream-json`   |
 | `codex`         | Codex         | **Active** | Full support, `--json`, YOLO mode default                        |
+| `cursor`        | Cursor        | **Beta**   | Cursor CLI agent, `-p --output-format stream-json --force`       |
 | `opencode`      | OpenCode      | **Active** | Multi-provider support (75+ LLMs), stub provider session storage |
 | `factory-droid` | Factory Droid | **Active** | Factory's AI coding assistant, `-o stream-json`                  |
 | `terminal`      | Terminal      | Internal   | Hidden from UI, used for shell sessions                          |
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -106,6 +106,7 @@ Maestro is an Electron desktop app for managing multiple AI coding assistants si
 | --------------- | ------------- | ---------- |
 | `claude-code`   | Claude Code   | **Active** |
 | `codex`         | OpenAI Codex  | **Active** |
+| `cursor`        | Cursor        | **Beta**   |
 | `opencode`      | OpenCode      | **Active** |
 | `factory-droid` | Factory Droid | **Active** |
 | `terminal`      | Terminal      | Internal   |
--- a/docs/cursor-cli-research.md
+++ b/docs/cursor-cli-research.md
@@ -1,0 +1,200 @@
+---
+type: research
+title: Cursor CLI Research - Agent Integration
+created: 2026-03-17
+tags:
+  - cursor
+  - cli
+  - agent-integration
+related:
+  - '[[AGENT_SUPPORT]]'
+  - '[[CLAUDE-AGENTS]]'
+---
+
+# Cursor CLI Research
+
+Research findings for integrating Cursor's CLI agent into Maestro.
+Based on https://cursor.com/docs/cli/overview, headless mode docs, and installation docs.
+
+## Binary Name & Installation
+
+- **Binary name**: `agent` (NOT `cursor` - the product is Cursor, but the CLI binary is `agent`)
+- **Installation (macOS/Linux/WSL)**: `curl https://cursor.com/install -fsS | bash`
+- **Installation (Windows PowerShell)**: `irm 'https://cursor.com/install?win32=true' | iex`
+- **Installation path**: `~/.local/bin/agent` (Linux/macOS). Windows path TBD.
+- **PATH setup**: Users must add `~/.local/bin` to PATH
+- **Updates**: `agent update` (also auto-updates)
+- **Version check**: `agent --version`
+- **Not installed on this system**: Neither `agent` nor `cursor` CLI found in PATH
+
+## Subcommands
+
+| Subcommand | Description |
+|-----------|-------------|
+| `agent` | Start interactive session |
+| `agent "prompt"` | Start with initial prompt |
+| `agent ls` | List previous conversations |
+| `agent resume` | Resume latest conversation |
+| `agent update` | Manual update |
+
+## Operational Modes
+
+| Mode | Purpose | Activation |
+|------|---------|------------|
+| Agent (default) | Full access to all tools for complex coding tasks | Default mode |
+| Plan | Design approach with clarifying questions, no file changes | `--plan`, `--mode=plan`, `/plan`, Shift+Tab |
+| Ask | Read-only exploration without making changes | `--mode=ask`, `/ask` |
+
+## Key Flags & Options
+
+### Headless/Batch Mode (Non-Interactive)
+- **`-p, --print`**: Enables non-interactive scripting/automation mode (similar to Claude Code's `--print`)
+- **`--force` / `--yolo`**: Allows agent to make direct file changes without confirmation (YOLO mode)
+- Without `--force`, proposed changes are displayed but NOT applied
+
+### Output Format
+- **`--output-format text`**: Clean, final-answer-only responses (default)
+- **`--output-format json`**: Structured analysis output
+- **`--output-format stream-json`**: Message-level progress tracking with real-time updates
+- **`--stream-partial-output`**: Incremental streaming of text deltas for live feedback
+
+### Model Selection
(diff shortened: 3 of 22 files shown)


jSydorowicz21:
@greptile re-review please

pedramamini:
@jSydorowicz21 Thank you for this contribution! Adding Cursor is a genuinely useful expansion of Maestro's agent lineup, and the depth here is appreciated: the research doc, 556 lines of parser tests, error pattern coverage, and the integration test all show real care. This is a solid foundation.

I went through the Greptile, CodeRabbit, and Codex reviews and verified the findings against the code rather than taking them at face value. A few are real and worth addressing before merge, and one I think you can safely ignore.

## First: rebase needed

The PR currently shows `MERGEABLE: CONFLICTING` against `rc`, and it's still marked as a draft. `rc` has moved a fair amount since the last commit (`c878ef2f`, March 29). Could you please rebase onto the latest `rc` and resolve the conflicts so we get a clean merge? Once that's done, please also mark the PR ready for review.

## Blocking issues (verified)

**1. `--force` is never applied on normal runs, so Cursor won't actually write files.**

`definitions.ts` puts `--force` in `yoloModeArgs`. But `buildAgentArgs()` (`src/main/utils/agent-args.ts:155`) only appends `yoloModeArgs` when the caller explicitly passes `yoloMode: true`, and the only place that does is Cue (`src/main/cue/cue-spawn-builder.ts:103`). Normal desktop prompts and Auto Run go through `buildSpawnConfigForAgent()` without it.

Your own research doc says it plainly: "Without `--force`, proposed changes are displayed but NOT applied." So implementation tasks would appear to run successfully while leaving the workspace untouched.

Compare how the other agents handle this:
- Claude Code puts `--dangerously-skip-permissions` directly in base `args` (always on)
- Codex puts `--dangerously-bypass-approvals-and-sandbox` in `batchModeArgs`, which `buildAgentArgs()` auto-applies unless `readOnlyMode` is set

`batchModeArgs` looks like the right fit for Cursor, since it's already skipped when read-only is active, which matches your comment about `--force` conflicting with `--mode ask`.

**2. The `result` event masks the assistant's actual answer.**

`cursor-output-parser.ts:143-146` sets `text` to `Completed in ${duration_ms}ms` on the result event. `StdoutHandler.ts:610` resolves the final payload as `event.text || managedProcess.streamedText`, so the duration string wins and the streamed answer is discarded. The user sees `Completed in 4321ms` instead of Cursor's response.

All three bots flagged this independently, and the code path checks out. Leaving `text` undefined on the result event is the smallest fix, since `streamedText` already has the answer accumulated.

**3. Session IDs are never captured, so follow-up turns can't resume.**

`extractSessionId()` (line 259) returns `event.sessionId`, but `transformMessage()` never sets `sessionId` on any branch, including the `system`/`init` case. That means the process manager never emits `session-id`, nothing is stored on the tab, and every follow-up prompt in the same tab starts a fresh conversation.

You've set `supportsSessionId: false` with a comment saying IDs aren't in the stream-json output, but Codex's review claims current Cursor init/user/assistant events do carry `session_id`. Worth verifying against a live `agent -p --output-format stream-json` run. If it's there, wiring it up would make `resumeArgs` actually useful and give Cursor real multi-turn behavior. If it genuinely isn't there, that's a significant caveat to note in the Beta labeling, since a coding agent that can't hold a conversation is a rough experience.

## Non-blocking, but please consider

**4. Generic `agent` binary collision has no guard.** You already flagged the risk in the comment on `binaryName`, but `checkBinaryExists('agent')` will happily accept any unrelated binary named `agent` on PATH and then launch it with Cursor-specific flags. An `agent --version` fingerprint check before marking it available would close this. Both CodeRabbit and Codex raised it.

**5. Auth error pattern gap.** None of the five `auth_expired` patterns match Cursor's common unauthenticated failure, `Authentication required. Please run 'agent login' first`. Users who haven't logged in get a generic crash instead of the recoverable auth guidance. A pattern like `/authentication required|run ['"]?agent login/i` would cover it.

**6. `supportsThinkingDisplay: true` doesn't match the parser.** The parser has no reasoning branch, only normal `assistant` text. Unless Cursor emits a separate reasoning stream you're handling elsewhere, this should be `false` so the runtime contract matches reality.

**7. No-op storage registered as available.** Registering `CursorSessionStorage` in `storage/index.ts` makes storage-availability checks report Cursor as storage-backed while `supportsSessionStorage` is `false` and every method returns empty. Consumers that check availability directly will render an empty history as "no sessions" rather than "not implemented." Either skip registration or make the availability check honor the capability flag.

**8. CLI batch mode rejects Cursor.** With `usesJsonLineOutput: false` and no special case, `spawnAgent('cursor', ...)` falls through to `Unsupported agent type for batch mode` (`src/cli/services/agent-spawner.ts:1113`). Probably acceptable for Beta, but worth calling out explicitly in the docs so playbook users aren't surprised.

## One finding you can skip

Greptile's "duplicate `--model` flag" report is a pre-existing pattern, not something you introduced. Every agent in `definitions.ts` defines both `modelArgs` and a `model` `configOption`. If that dedup interaction is a real problem it's a separate issue for all agents, so please don't chase it in this PR.

## Also worth noting

`contextWindow` defaults to `200000`, which your research doc says is undocumented and model-dependent. Not a blocker, but the initial context-usage display will be wrong until the user overrides it. Consider whether a conservative default or leaving it unset reads better.

---

Items 1 and 2 are the ones that would make Cursor look broken in real use, so those are the priority. Once the rebase is done and those are addressed, ping us and we'll take another pass. Thanks again for putting this together!

pedramamini:
@jSydorowicz21 Thanks again for the work on Cursor support! Just a friendly check-in, since this has been quiet since our review on August 11.

The branch is still at `c878ef2f`, still a draft, and still shows merge conflicts against `rc`. Could you please rebase onto the latest `rc` and resolve those conflicts so we can get a clean merge? The review items from last time still apply, and items 1-3 (`--force` not applied on normal runs, the `result` event hiding the streamed answer, and session IDs never being captured) are the ones to prioritize.

`rc` has moved since then in ways that affect this rebase, so a few pointers to save you some digging:

- **The renderer files this PR edits no longer exist as single files.** `src/renderer/components/NewInstanceModal.tsx` and `src/renderer/components/Wizard/screens/AgentSelectionScreen.tsx` are now directories (`NewInstanceModal/`, `AgentSelectionScreen/`). You shouldn't need to hand-edit a provider list in either of them anymore, see the next point.
- **All three provider pickers now read one registry.** `AGENT_PICKER_META` in `src/shared/agentMetadata.ts` is a `Record<AgentId, ...>`, so adding `cursor` to `AGENT_IDS` won't compile until it has an entry there (or an explicit `null`). A new provider also needs a `case` in `AgentLogo` (`Wizard/screens/AgentSelectionScreen/components/AgentLogo.tsx`) and a glyph in `src/renderer/constants/agentIcons.ts`. [PROVIDER-SUPPORT.md, Step 2.6](https://github.com/RunMaestro/Maestro/blob/rc/PROVIDER-SUPPORT.md#step-26-register-the-provider-in-the-pickers) has the full checklist.
- **The error pattern bank moved.** The patterns now live in `src/shared/agentErrorPatterns.ts`, and `src/main/parsers/error-patterns.ts` is just the main-process wrapper around them. That's where the Cursor patterns (including the `Authentication required. Please run 'agent login' first` case from item 5) should go.

One more thing worth flagging: #1401 also adds Cursor CLI as a beta agent, and it isn't a draft. Would you mind taking a look at it? If you're still planning to carry this one forward, it would help to coordinate there so the two efforts don't overlap.

Once it's rebased and marked ready for review, ping us and we'll take another pass. Thanks!

PR #605, opened 2026-03-21
Llm guard

Ashraf-Ali-aa:
## Summary by CodeRabbit

* **New Features**
  * LLM Guard: local scanning of AI inputs/outputs for PII, secrets, prompt injection, malicious URLs, and dangerous code; custom regex patterns, per-session overrides, inter-agent scanning, input preview, and audit export/import.
* **Security UX**
  * Indicator, session badge, scan progress, sanitization diff, security panel, toasts, and actionable recommendations.
* **Docs**
  * Comprehensive LLM Guard documentation, settings, best practices, troubleshooting, and config import/export guidance.
* **Tests**
  * Extensive new test coverage for security APIs, scanning logic, UI components, exports, and recommendations.


pedramamini:
@Ashraf-Ali-aa first off, thank you for this contribution. LLM Guard is a genuinely ambitious piece of work - 93 files, PII/secrets/injection/code/URL scanning, a PII vault, per-session policy overrides, inter-agent scanning for Group Chat, audit export, and a settings UI, all gated behind an Encore flag. The architecture is sound and the test coverage you brought along is well above what we usually see. We appreciate the effort here.

That said, I don't think this is ready to merge yet. Below is what we'd like addressed.

## 1. Please rebase onto `rc`

This PR was opened on 2026-03-21 and `rc` has moved a long way since. GitHub currently reports the branch as `CONFLICTING`, and a local test merge produces conflicts in **39 files**, including several that are core to your change:

- `src/main/process-manager/handlers/StdoutHandler.ts`, `StderrHandler.ts`, `ExitHandler.ts`
- `src/main/process-manager/types.ts`, `src/main/ipc/handlers/process.ts`, `src/main/preload/process.ts`
- `src/main/group-chat/group-chat-router.ts`
- `src/renderer/stores/settingsStore.ts`, `src/renderer/hooks/settings/useSettings.ts`, `src/renderer/types/index.ts`
- `src/renderer/components/RightPanel.tsx`, `SettingsModal.tsx`, `InputArea.tsx`, `MainPanel.tsx`, `App.tsx`
- `package.json` / `package-lock.json` / `docs/docs.json`

Could you rebase on the latest `rc` and resolve these? The process-manager handlers in particular have been reworked since March, so the guard integration points will need a careful re-application rather than a mechanical conflict resolution. That'll also refresh CI and give the bots a clean pass to re-review against.

## 2. Blocker: the audit log writes raw secrets to disk

This is the one that has to change before merge. We traced it end to end and confirmed it:

- `collectMatches()` (`src/main/security/llm-guard/index.ts:2458-2480`) populates `LlmGuardFinding.value` with the raw matched text, and `redactSecrets()` / `anonymizePii()` carry that straight through to `guardResult.findings`.
- `StdoutHandler.applyOutputGuard()` (`src/main/process-manager/handlers/StdoutHandler.ts:542-549`) passes `findings: guardResult.findings` verbatim into `logSecurityEvent()`.
- `appendEventToFile()` (`src/main/security/security-logger.ts:156-158`) does `JSON.stringify(event)` and appends it to `~/.maestro/security-events.jsonl`.

Net effect: every API key, private key, SSN, and credit card number LLM Guard detects gets written to a plaintext append-only file on disk. That inverts the purpose of the feature - turning the guard on would create a single file concentrating every secret that passed through the app, and it survives well beyond the in-memory buffer.

Worth noting you already do the right thing on the IPC path: the `SecurityEventData` mapping just above only forwards `findingTypes`, not values. The disk path needs the same treatment. Suggested direction: redact/hash `value` at the point the finding is constructed for logging (a truncated prefix plus a SHA-256 of the full value is usually enough to correlate repeat findings without storing the secret), and keep raw values in memory only if a UI surface genuinely needs them. `security-logger.ts:501` also renders `f.value.substring(0, 100)` into the exported HTML report, which has the same problem.

## 3. Duplicate security event logging on the streaming path

`applyOutputGuard()` is invoked per streaming `thinking-chunk` (`StdoutHandler.ts:336`) and again on the assembled final text (`StdoutHandler.ts:379`, `:407`, `:454`), with no dedup between them. A single secret in a 20-chunk response produces up to 21 logged events, which inflates the audit log and skews the stats and recommendations engine.

Related and worth thinking through while you're in there: scanning per-chunk means a secret straddling a chunk boundary matches in neither chunk. The final-text scan does catch it, but only after the unredacted chunk has already been emitted to the renderer. Some buffering or overlap-window at the chunk boundary would close that gap.

## 4. The bot findings worth acting on

CodeRabbit flagged roughly two dozen major items. We haven't verified each one, but these looked substantive on inspection:

- **`config-export.ts:314`** - regenerating custom pattern IDs on import breaks `mergeCustomPatterns()` dedup (`index.ts:1847-1868`) and the `f.patternId` to `pattern.id` lookup used for block decisions (`index.ts:2018-2021`). Round-tripping an exported config will silently duplicate patterns and drop block actions.
- **`config-export.ts:240`** - invalid `banTopicsPatterns` import as warnings only, then get silently skipped at scan time. The imported policy looks active while those bans are inert. Failing the import is safer than a warning for a security policy.
- **`config-export.ts:182`** - `output.detectOutputInjection` is missing from the boolean whitelist, so a non-boolean can reach the runtime config.
- **`vault.ts:16`** - `PiiVault` stores and returns caller-owned object references; copy on `add()` and on `toJSON()` so external mutation can't corrupt deanonymization.
- **ReDoS** - user-supplied `banTopicsPatterns` and custom patterns cross the IPC boundary and are compiled in the main process without complexity limits or an execution timeout. A pathological pattern hangs the main process.

## 5. Minor

- `applyOutputGuard()` is duplicated between `StdoutHandler.ts` and `ExitHandler.ts`. Per our dedup guidance in `CLAUDE.md`, please extract it to a shared helper.
- `getLlmGuardSettings()` is called inside the per-participant loop in `group-chat-router.ts`; hoist it above the loop.
- The recommendations engine reads only the in-memory buffer, which isn't rehydrated from the persisted file on startup, so recommendations reset every app restart. Might be intended, but flagging in case it isn't.

## One bot finding you can ignore

Greptile's P1 about stateful module-level regexes (`EMAIL_REGEX`, `PHONE_REGEX`, `SSN_REGEX`, `IPV4_REGEX`, `CREDIT_CARD_REGEX` retaining `lastIndex` across calls) is a **false positive**. Those five are only ever consumed through `PII_PATTERNS` into `collectMatches()`, which clones via `new RegExp(regex.source, regex.flags)` before executing. No shared `lastIndex`, no missed matches. Don't spend time on it.

---

**Summary:** rebase onto `rc`, fix the raw-secret persistence in `security-logger.ts` (blocker), dedup the streaming-path logging, and work through the config-export/vault/ReDoS items. Once that's done, ping us and we'll take another pass.

Thanks again for putting this together - we'd like to see it land.

PR #677, opened 2026-03-28
fix(quit): add safety timeout to prevent app lingering on Windows (#623)

### LINKED ISSUE #623: Windows: Maestro sometimes remains running in the background after closing and cannot be reopened until all processes are killed
On Windows, Maestro sometimes does not fully exit after closing the app.

Symptoms:
- The Maestro window disappears as expected after closing it
- But Maestro is still visible in Task Manager as a running process
- At that point, trying to open Maestro again does not work
- The only workaround is to manually end all Maestro-related processes in Task Manager
- After killing all remaining Maestro processes, Maestro can be opened normally again

This does not happen every time, but it happens often enough to be disruptive.

Expected behavior:
- Closing Maestro should fully terminate the application process
- Maestro should be able to reopen normally without requiring manual process cleanup

Actual behavior:
- Maestro sometimes remains running in the background after the window is closed
- Relaunch fails until the lingering processes are manually terminated

OS:
- Windows

chr1syy:
## Summary
- Adds a 3-second safety timeout to the quit confirmation flow so that if the renderer never responds (e.g., window mid-teardown, renderer crashed, or webContents in a transitional state), the app force-quits instead of lingering in the background with no window (issue #623)
- Clears the safety timeout when the renderer does respond (confirmed or cancelled)
- Also cleans up power save blocker and group chat moderator interval during shutdown to prevent stale OS-level state

## Test plan
- [x] Verified fix works on Windows — app no longer lingers after closing
- [x] Updated `quit-handler.test.ts` with mocks for `powerManager` and `stopSessionCleanup`
- [x] Existing quit handler tests pass with new dependencies

Fixes #623

## Summary by CodeRabbit

* **Improvements**
  * Added a 5-second quit-confirmation safety timeout to avoid hangs when the renderer doesn't respond.
  * Shutdown now runs session cleanup and clears system sleep-prevention state so the app exits more reliably and frees resources.
* **Tests**
  * Updated tests to cover the confirmation timeout and expanded shutdown cleanup behavior.
* **Documentation**
  * Normalized formatting in release notes documentation.

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -96,7 +96,7 @@ The major contributions to 0.14.x remain:
 - Leaderboard tracking now works across multiple systems and syncs level from cloud 🏆
 - Agent duplication. Pro tip: Consider a group of unused "Template" agents ✌️
 - New setting to prevent system from going to sleep while agents are active 🛏️
-- The tab menu has a new "Publish as GitHub Gist" option  📝
+- The tab menu has a new "Publish as GitHub Gist" option 📝
 - The tab menu has options to move the tab to the first or last position 🔀
 - [Maestro-Playbooks](https://github.com/pedramamini/Maestro-Playbooks) can now contain non-markdown assets 📙
 - Improved default shell detection 🐚
@@ -125,10 +125,12 @@ Thanks for the contributions: @t1mmen @aejfager @Crumbgrabber @whglaser @b3nw @d
 - TAKE TWO! Fixed Linux ARM64 build architecture contamination issues 🏗️
 
 ### v0.13.1 Changes
+
 - Fixed Linux ARM64 build architecture contamination issues 🏗️
 - Enhanced error handling for Auto Run batch processing 🚨
 
 ### v0.13.0 Changes
+
 - Added a global usage dashboard, data collection begins with this install 🎛️
 - Added a Playbook Exchange for downloading pre-defined Auto Run playbooks from [Maestro-Playbooks](https://github.com/pedramamini/Maestro-Playbooks) 📕
 - Bundled OpenSpec commands for structured change proposals 📝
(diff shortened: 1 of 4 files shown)


chr1syy:
closing #623

chr1syy:
Addressed the review feedback:

**Critical — timeout bypasses user confirmation modal:** The safety timeout is cleared by `clearConfirmationTimeout()` as soon as the renderer responds with either `quitConfirmed` or `quitCancelled`. It only fires if the renderer **never responds at all** (crash, mid-teardown, etc.). Bumped timeout from 3s to 5s to give extra margin for slow renderer startup.

**docs/releases.md comments:** Reverted `releases.md` to upstream state — the formatting issues are pre-existing and out of scope for this fix.

**confirmQuit() nitpick:** Applied — `confirmQuit()` now calls `clearConfirmationTimeout()` before setting `quitConfirmed`.

**setTimeout before send():** Applied — timeout is armed before `webContents.send()` so it's always active even if send throws.

pedramamini (commented):
Good fix for #623 — the safety timeout approach is solid. A few notes for follow-up:

1. **`clearAllReasons()` ordering** — Currently called before `killAll()` in `performCleanup()`. A process emitting output between those two calls could re-arm the sleep blocker. Should move after `killAll()`. Will fix post-merge.

2. **Timeout test coverage** — Tests verify the new cleanup hooks but don't exercise the 5s timeout path or the "clear timeout on confirmed/cancelled" paths. The timeout is the actual regression fix, so worth covering. Will add post-merge.

3. **`docs/releases.md` whitespace diff** — ~280 lines of CRLF→LF normalization. Content-identical but noisy. Will revert post-merge to keep the diff clean. For future PRs, consider excluding unrelated formatting changes or putting them in a separate commit.

Plan: merge as-is, then fix items 1-3 on top.

PR #712, opened 2026-04-02
[codex] expand autorun agent integrations

ShunsukeHayashi:
## Summary
- fix `plan-execute-verify` so the executor step resumes through the normal agent execution path instead of the background synopsis path
- add missing runtime/parser coverage for newly surfaced agents (`cursor-agent` and `zai`) and update related tests
- restore strict typing and skill-bus CLI/test wiring so lint and batch processor tests pass again

## Root cause
Several new integrations were only partially wired. The renderer auto-run executor was still using the short-lived synopsis helper, and newly exposed agents/skill-bus paths were missing the runtime or parser pieces needed by the app and CLI.

## Impact
- `plan-execute-verify` can now run real executor tasks without the 20s synopsis timeout or lost agent-error propagation
- `cursor-agent` batch/json-line parsing and `zai` runtime spawning now work end-to-end
- lint/test validation is green again for the affected integration paths

## Validation
- `npm run build:prompts`
- `npm run format:check:all`
- `npm run lint`
- `npm run lint:eslint`
- `npm run test`


ShunsukeHayashi:
Closed automatically. This PR was created accidentally by an autonomous agent (Codex) running in a local environment.

PR #714, opened 2026-04-03
feat: Persist and restore active session id (resume in last open sess…

scriptease:
Before:

Maestro would always start with session [0] from maestro-sessions.json

After:

Persist `activeSessionId` in the maestro-sessions.json and resume the last open session when Maestro launches, ignore when not found.

`{
	"sessions": [...],
	"activeSessionId": "c97dca4d-172f-471e-880c-7e694c59f9f4"
}`

## Summary by CodeRabbit

* **New Features**
  * The app now persists and restores the active session across launches, reactivating your last valid session and falling back to a sensible default when needed.

* **Refactor**
  * Active-session selection now prefers the persisted value when valid and synchronizes state with the host when it changes.

* **Tests**
  * Updated tests to cover active-session persistence, restoration, and IPC handler behavior.

### DIFF
--- a/src/__tests__/main/ipc/handlers/persistence.test.ts
+++ b/src/__tests__/main/ipc/handlers/persistence.test.ts
@@ -141,6 +141,8 @@ describe('persistence IPC handlers', () => {
 				'settings:set',
 				'settings:getAll',
 				'sessions:getAll',
+				'sessions:getActiveSessionId',
+				'sessions:setActiveSessionId',
 				'sessions:setAll',
 				'groups:getAll',
 				'groups:setAll',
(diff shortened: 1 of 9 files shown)


pedramamini:
@scriptease what do you think of the PR comments above?

pedramamini:
Reviewed the bot comments (Greptile P1s + CodeRabbit) — they're all based on a misread of the code:

- **"Side effect inside Zustand `set` callback"** — Incorrect. `setActiveSessionIdInternal` has no IPC call. `setActiveSessionId` correctly places the IPC call *after* `set()`.
- **"Unthrottled disk writes during Cmd+J/K cycling"** — Not applicable. The cycling path (`setActiveSessionIdInternal`) is local-only; only deliberate switches (`setActiveSessionId`) persist. This is exactly the "Option 1" the bot itself recommends, already implemented.
- **"Unhandled promise / IPC inside `set()`"** — Same misread. The fire-and-forget pattern is intentional and documented — failure only means the session won't be pre-selected on next launch (graceful fallback to first session).

No action items. LGTM.

scriptease:
> @scriptease what do you think of the PR comments above?

I addressed it in my latest patch I thought. sorry should have commented!

Thanks 👍

Issue #707, opened 2026-04-02
bug: Auto Run error banner stays visible after clicking Resume


jSydorowicz21:
## Bug Description

When Claude Code hits an API rate limit or error during an Auto Run and pauses, Maestro shows an error banner with Resume/Skip/Abort options. Clicking "Resume" correctly resumes the agent, but the error banner stays visible as if the button wasn't clicked.

## Steps to Reproduce

1. Start an Auto Run with multiple documents
2. Wait for (or trigger) an API rate limit / overloaded error
3. Maestro pauses and shows the error banner with Resume option
4. Click "Resume"
5. The agent resumes processing (confirmed by output continuing)
6. The error banner remains visible on screen

## Expected Behavior

The error banner should dismiss immediately after clicking Resume.

## Investigation

The resume flow dispatches `CLEAR_ERROR` to the Zustand batch store, which correctly sets `errorPaused: false` in the reducer (`batchReducer.ts:424-448`). The error banner is conditionally rendered based on `isErrorPaused`:

```tsx
// AutoRun.tsx:542-551
{isErrorPaused && batchError && (
    <AutoRunErrorBanner ... />
)}
```

Where `isErrorPaused = batchRunState?.errorPaused || false` (line 95).

The `AutoRun` component uses `React.memo` with a custom comparison (lines 727-770) that includes `errorPaused`:

```typescript
prevProps.batchRunState?.errorPaused === nextProps.batchRunState?.errorPaused
```

The reducer returns a new state object with `errorPaused: false`, which should trigger a re-render. The likely cause is either:

1. The Zustand store subscription in the parent component isn't propagating the updated `batchRunState` as a new prop reference
2. The `broadcastAutoRunState` call at `useBatchProcessor.ts:1977` reconstructs state with spread but the component receives it through a different path
3. A timing issue between `dispatch` and the component's subscription update

## Files Involved

- `src/renderer/components/AutoRun/AutoRun.tsx:542-551` (error banner render)
- `src/renderer/components/AutoRun/AutoRun.tsx:727-770` (memo comparison)
- `src/renderer/components/AutoRun/AutoRunErrorBanner.tsx` (popup component)
- `src/renderer/hooks/batch/useBatchProcessor.ts:1966-1993` (resume handler)
- `src/renderer/hooks/batch/batchReducer.ts:424-448` (CLEAR_ERROR reducer)
- `src/renderer/components/RightPanel.tsx:734-765` (secondary resume UI)

## Environment

[...]

PR #715, opened 2026-04-03
fix(autorun): resolve paused batch session for resume/skip/abort

nehaaprasad:
### summary

- Fix error banner staying after Resume by clearing the same batch session the UI shows.

### Changes

- src/renderer/hooks/batch/useBatchHandlers.ts — resolve paused session (prefer active session, then any errorPaused) for resume/skip/abort

fix : #707

## Release Notes

* **Bug Fixes**
  * Enhanced error recovery in batch processing. When pausing on error, the system now intelligently selects which batch to resume from—prioritizing the currently active one, then any paused batch, then the first active batch for more predictable operation handling.

### DIFF
--- a/src/renderer/hooks/batch/useBatchHandlers.ts
+++ b/src/renderer/hooks/batch/useBatchHandlers.ts
@@ -29,9 +29,31 @@ import { CONDUCTOR_BADGES, getBadgeForTime } from '../../constants/conductorBadg
 import { getActiveTab } from '../../utils/tabHelpers';
 import { generateId } from '../../utils/ids';
 import { useBatchProcessor } from './useBatchProcessor';
+import { useBatchStore } from '../../stores/batchStore';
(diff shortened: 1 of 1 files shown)


jSydorowicz21:
@nehaaprasad Please look into the feedback from coderabbit and greptile and address as needed

pedramamini:
Thanks for the contribution @nehaaprasad! Great catch on the error banner persistence issue — the root cause analysis of needing to check `errorPaused` in the batch store was spot on.

I've merged your changes into `fix/aut-bann-sty` on the main repo with a small refinement: removed the non-error-paused fallbacks so skip/resume/abort only target sessions that are actually paused on error (see the review comment above for rationale). Updated tests to match.

Closing this PR in favor of the updated branch — your work is preserved in the commit history. Thanks again!

PR #718, opened 2026-04-03
fix(group-chat): add participant remove button and allow agentId updates

### LINKED ISSUE #717: Bug: Provider switch breaks existing group chat: panel controls frozen, ag...
## Summary
Provider switch breaks existing group chat: panel controls frozen, agent removal unavailable

## Environment
- Maestro version: 0.16.5-RC
- Operating system: Windows (Windows 10 Home, 10.0.26200)

## Steps to Reproduce
1. Add an agent (configured with Codex) to a group chat.
2. Switch the agent's provider from Codex to Claude in the agent's settings.
3. Return to the existing group chat.
4. Observe: group chat still uses old provider; right panel controls are frozen (only Reset and Copy ID work); no remove-agent option is available.

## Expected Behavior
Switching an agent's provider should be reflected in all existing group chats, and the right panel should retain full controls (including agent removal) regardless of provider changes.

## Actual Behavior
After switching an agent from Codex to Claude, existing group chats still reference the old provider. The right panel becomes mostly non-functional — only Reset and Copy ID buttons work. The option to remove the agent from the group chat is entirely absent. Issue persists across restarts.

## Additional Context
Workaround: create a new group chat with the agent — the new group chat picks up the updated provider. The user attempted this to resolve the stale provider issue.

Suggested future workaround: adding slash commands to group chats (e.g. /remove-agent) would allow users to manage agents even when the right panel UI is broken.

Two related issues to be filed separately: (1) slash commands for agent management in group chats, (2) file references not working in group chats.

pedramamini:
## Summary

- **Add Remove button** (with confirmation) to `ParticipantCard` so users can remove stale participants from group chats
- **Wire up `onRemove` callback** through `GroupChatRightPanel` and `GroupChatParticipants` to the existing `removeParticipant` IPC handler (backend was already implemented, UI was missing)
- **Add `agentId` to `ParticipantUpdate` type** enabling programmatic updates to a participant's agent type when provider changes

### Root Cause

When a user switches an agent's provider (e.g., Codex → Claude) after adding it to a group chat, the participant's `agentId` is snapshotted at creation time in `metadata.json` and never updated. The stale reference causes the right panel to display the old provider, and since there was no UI to remove participants, users were stuck with no way to fix the group chat.

### Workaround (pre-fix)

Create a new group chat — it picks up the updated provider automatically.

Closes #717

## Test plan

- [ ] Create a group chat with a Codex participant
- [ ] Switch the agent's provider to Claude Code in agent settings
- [ ] Return to the group chat — verify the Remove button appears on participant cards
- [ ] Click Remove → verify confirmation prompt appears (Confirm/Cancel)
- [ ] Confirm removal → verify participant is removed from the group chat
- [ ] Re-add the agent — verify it now uses the updated provider
- [ ] Verify the moderator card does NOT show a Remove button (moderators cannot be removed)

## Summary by CodeRabbit

* **New Features**
  * Remove participants directly from the group chat participant list.
  * Two-step confirmation flow (Confirm/Cancel) to prevent accidental removals.
  * Visual feedback while removal is in progress ("Removing..." indicator) and disabled controls during the operation.
  * Remove action available on non-moderator participant entries in the Participants tab.

### DIFF
--- a/src/main/group-chat/group-chat-storage.ts
+++ b/src/main/group-chat/group-chat-storage.ts
@@ -477,7 +477,6 @@ export async function getParticipant(
 export type ParticipantUpdate = Partial<
 	Pick<
 		GroupChatParticipant,
-		| 'agentId'
 		| 'lastActivity'
 		| 'lastSummary'
 		| 'contextUsage'


pedramamini:
Thanks for the contribution, @pedramamini! Clean fix — the two-step confirm UX on the Remove button is a nice touch, and wiring up the already-implemented backend was the right call.

**One thing to address before merge:**

Greptile flagged (and I agree) that the `handleRemoveParticipant` handlers in both `GroupChatParticipants.tsx` and `GroupChatRightPanel.tsx` silently swallow errors:

```ts
try {
    await window.maestro.groupChat.removeParticipant(groupChatId, participantName);
} catch (error) {
    console.error(`Failed to remove participant ${participantName}:`, error);
}
```

Per our error handling guidelines in CLAUDE.md, unexpected errors should bubble up so Sentry can capture them. The user also gets zero feedback that the remove failed — the button just reappears silently. Please either:

1. Remove the try/catch and let errors propagate (Sentry picks them up automatically, and `ParticipantCard`'s `finally` block still runs), or
2. Re-throw after logging so the card can show a failure state.

Everything else looks good. The `agentId` addition to `ParticipantUpdate` and the `ParticipantCard` UI changes are solid. CI failures are pre-existing and unrelated to this PR.

Issue #116, opened 2025-12-29
ARM pty.node shipping with non ARM Linux releases...


whglaser:
Trying to run Maestro on Linux. I've tried using both the .deb package and also appimage. For both routes Maestro fails to launch and generates an error message indicating issues with pty.node. Inspecting pty.node reveals that it's been compiled for ARM in both cases...

When installing the .deb:
```
curl -LO https://github.com/pedramamini/Maestro/releases/download/v0.12.3/maestro_0.12.3_amd64.deb

mv maestro_0.12.3_amd64.deb /tmp/

sudo apt install /tmp/maestro_0.12.3_amd64.deb

maestro
innerError Error: Cannot find module '../build/Debug/pty.node'
Require stack:
- /opt/Maestro/resources/app.asar/node_modules/node-pty/lib/unixTerminal.js
- /opt/Maestro/resources/app.asar/node_modules/node-pty/lib/index.js
- /opt/Maestro/resources/app.asar/dist/main/process-manager.js
- /opt/Maestro/resources/app.asar/dist/main/index.js
-
    at Module._resolveFilename (node:internal/modules/cjs/loader:1084:15)
    at s._resolveFilename (node:electron/js2c/browser_init:2:114421)
    at Module._load (node:internal/modules/cjs/loader:929:27)
    at c._load (node:electron/js2c/node_init:2:13672)
    at Module.require (node:internal/modules/cjs/loader:1150:19)
    at require (node:internal/modules/cjs/helpers:119:18)
    at Object.<anonymous> (/opt/Maestro/resources/app.asar/node_modules/node-pty/lib/unixTerminal.js:34:15)
    at Module._compile (node:internal/modules/cjs/loader:1271:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1326:10)
    at Module.load (node:internal/modules/cjs/loader:1126:32) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/opt/Maestro/resources/app.asar/node_modules/node-pty/lib/unixTerminal.js',
    '/opt/Maestro/resources/app.asar/node_modules/node-pty/lib/index.js',
    '/opt/Maestro/resources/app.asar/dist/main/process-manager.js',
    '/opt/Maestro/resources/app.asar/dist/main/index.js',
    undefined
  ]
}
A JavaScript error occurred in the main process
Uncaught Exception:
[...]

PR #720, opened 2026-04-04
fix: build macOS x64/arm64 separately to prevent Intel architecture contamination

### LINKED ISSUE #719: Terminal fails on Intel Macs — "Failed to start terminal / Check system PTY availability"
The terminal fails to start on Intel Macs with the error: Failed to start terminal. The shell process could not be started. Check system PTY availability.

The root cause is a build packaging bug: the node-pty binaries in build/Release/ are compiled for arm64 (Apple Silicon), but this is an Intel machine. The correct x86_64 prebuilds are already included in the package under prebuilds/darwin-x64/ but aren't being used.

￼ # Confirms the wrong arch is in build/Release:
file .../node_modules/node-pty/build/Release/pty.node
→ Mach-O 64-bit bundle arm64   ← wrong on Intel

file .../node_modules/node-pty/build/Release/spawn-helper
→ Mach-O 64-bit executable arm64   ← wrong on Intel

Note: a similar "ARM64 build architecture contamination" issue was fixed for Linux previously — this appears to be the same problem but affecting the macOS Intel build.

Workaround

￼BASE=/Applications/Maestro.app/Contents/Resources/app.asar.unpacked/node_modules/node-pty

cp $BASE/prebuilds/darwin-x64/pty.node $BASE/build/Release/pty.node
cp $BASE/prebuilds/darwin-x64/spawn-helper $BASE/build/Release/spawn-helper

Force-quit and relaunch Maestro after running these commands. Terminal works correctly after applying the fix.

pedramamini:
## Summary

- Splits the macOS build into separate x64 and arm64 packaging steps, each with its own native module rebuild (`electron-rebuild --arch=<target>`) and architecture verification before packaging
- Removes explicit `arch` arrays from macOS targets in `package.json`, letting CLI flags (`--x64` / `--arm64`) control what gets built — mirroring the Linux fix from #116
- Adds pre-packaging `file` checks for `pty.node` and `better_sqlite3.node` on macOS (previously only Linux had these safeguards)

**Root cause**: The macOS CI ran on an ARM64 runner and built both architectures in a single `electron-builder` invocation. With `npmRebuild: false`, native modules were only compiled for the host arch (arm64), so the x64 package shipped arm64 binaries — breaking terminal on Intel Macs.

Closes #719

## Test plan

- [ ] Trigger a release build and verify both macOS x64 and arm64 DMGs/ZIPs are produced
- [ ] Verify the architecture verification steps pass in CI logs (x86_64 for x64, arm64 for arm64)
- [ ] Install the x64 DMG on an Intel Mac and confirm terminal starts without "Failed to start terminal" error
- [ ] Install the arm64 DMG on Apple Silicon and confirm terminal works
- [ ] Verify `file` on `pty.node` inside each packaged app shows the correct architecture

## Summary by CodeRabbit

* **Chores**
  * macOS packaging now produces separate x64 and arm64 builds with per-architecture native rebuilds and verification.
  * Windows-only adjustment for non-Linux native-module rebuilds.
  * Linux packaging unified to run architecture-specific rebuild-and-verify steps before packaging.
* **Documentation**
  * Normalized release notes formatting for consistent Markdown rendering.

### DIFF
--- a/.github/scripts/rebuild-and-verify-native.sh
+++ b/.github/scripts/rebuild-and-verify-native.sh
@@ -1,0 +1,56 @@
+#!/usr/bin/env bash
+# rebuild-and-verify-native.sh — Clean, rebuild, and verify native modules for a target architecture.
+# Used by release.yml to prevent architecture contamination across x64/arm64 builds.
+#
+# Usage: ./rebuild-and-verify-native.sh <arch>
+#   arch: x64 | arm64
+
+set -euo pipefail
+
+ARCH="${1:?Usage: $0 <x64|arm64>}"
+
+# Architecture-specific grep patterns for the `file` command output.
+# macOS: "Mach-O 64-bit bundle x86_64" / "Mach-O 64-bit bundle arm64"
+# Linux: "ELF 64-bit LSB shared object, x86-64" / "ELF 64-bit LSB shared object, ARM aarch64"
+case "$ARCH" in
+  x64)  PATTERN="x86_64\|x86-64\|AMD64" ;;
+  arm64) PATTERN="arm64\|ARM\|aarch64" ;;
+  *) echo "ERROR: unsupported arch '$ARCH' (expected x64 or arm64)" >&2; exit 1 ;;
+esac
+
+NATIVE_MODULES=(
+  "node-pty:pty.node"
+  "better-sqlite3:better_sqlite3.node"
+)
+
+# --- Clean ---
+for entry in "${NATIVE_MODULES[@]}"; do
+  mod="${entry%%:*}"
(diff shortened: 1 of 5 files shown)


pedramamini:
Thanks for the contribution, @pedramamini! 🙌

Reviewed the changes — the approach is solid. Splitting the macOS build into separate x64/arm64 steps with per-architecture rebuild + verification gates correctly addresses the root cause (arm64 runner contaminating x64 packages). The pattern mirrors the existing Linux fix from #116, which is great for consistency.

The `package.json` simplification and docs formatting cleanup are both clean.

LGTM — approving.

PR #728, opened 2026-04-05
idea: Adjust slash command fuzzy style to match the new settings search

scriptease:
Bildschirmfoto 2026-04-05 um 18 14 14

I don't know maybe it it too many colors.

## Release Notes

* **New Features**
  * Enhanced slash-command autocomplete with dynamic theme-aware styling. Matched and unmatched characters now display contextually based on selection state and the active theme's accent color. Font weights and opacity levels adjust accordingly, providing improved visual distinction during command navigation and selection.

### DIFF
--- a/src/renderer/components/InputArea.tsx
+++ b/src/renderer/components/InputArea.tsx
@@ -586,8 +586,13 @@ export const InputArea = React.memo(function InputArea(props: InputAreaProps) {
 								}}
 								onMouseEnter={() => setSelectedSlashCommandIndex(idx)}
 							>
-								<div className="font-mono text-sm leading-tight">
-									{highlightSlashCommand(cmd.command, inputValueLower.replace(/^\//, ''))}
+								<div className="font-mono text-sm">
+									{highlightSlashCommand(
+										cmd.command,
+										inputValueLower.replace(/^\//, ''),
+										theme.colors.accent,
+										idx === safeSelectedIndex
+									)}
 								</div>
 								<div className="text-[11px] opacity-70 leading-tight">{cmd.description}</div>
 							</button>
--- a/src/renderer/utils/search.ts
+++ b/src/renderer/utils/search.ts
@@ -221,9 +221,17 @@ export const filterSlashCommands = <T extends SlashCommandLike>(
 
 /**
  * Render a slash command with fuzzy-matched characters highlighted.
- * Returns a React node: plain string when no query, spans with bold/dim otherwise.
+ * Matched characters use accent color + semibold (matching Settings search style).
+ * When selected, unmatched characters are dimmed so matches stand out on the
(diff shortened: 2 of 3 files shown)


pedramamini:
Thanks for the contribution @scriptease! 🎯

This looks good — the changes are clean, focused, and consistent across both desktop and mobile. Using the accent color for matched characters and dimming unmatched ones on selection is a nice touch that brings visual consistency with the settings search.

Re: "maybe it is too many colors" — I think it's fine. The accent color is only applied to matched characters in the non-selected state, and on selection it falls back to font weight differentiation with opacity dimming. That's restrained and purposeful, not noisy.

LGTM — approving!

scriptease:
Yay 🎉 I love colors!

pedramamini:
@scriptease can you address the conflict please so i can merge this in?

pedramamini:
Thanks again for this one @scriptease, and sorry it's gone quiet for a while. The idea still stands up and we'd like to get it in.

Bad news first: it's still showing as conflicting. Good news: the conflicts are almost entirely file moves rather than anything wrong with your change. `rc` has moved on quite a bit since April, so rather than fighting a rebase, it's probably fastest to rebase onto latest `rc` and reapply the diff by hand. Here's the current lay of the land:

**1. `src/renderer/utils/search.ts` - unchanged, applies as-is.** `highlightSlashCommand` is still there and still has the old `fontWeight: 700` / `opacity: 0.8` styling, so your new signature (`accentColor`, `isSelected`) drops straight in with no changes.

**2. `src/renderer/components/InputArea.tsx` - moved.** That file was split into a directory. The slash command popover now lives at:

```
src/renderer/components/InputArea/overlays/SlashCommandPopover.tsx
```

Your edit goes on line 68, and the nice part is that component already receives `theme` and `selectedIndex` as props, so it's the same four-line change you already wrote:

```tsx
{highlightSlashCommand(
    cmd.command,
    inputValueLower.replace(/^\//, ''),
    theme.colors.accent,
    idx === selectedIndex
)}
```

(note it's `selectedIndex` here rather than `safeSelectedIndex`)

**3. `src/web/mobile/SlashCommandAutocomplete.tsx` - gone, please drop this hunk.** The `src/web/mobile/` directory was removed entirely. `src/web/` is now just PWA assets plus the service worker, and the browser build lives in `src/web-desktop/`. There's no equivalent call site to port to, so that third file can simply be dropped from the PR. That leaves you with two files instead of three.

No test updates needed as far as I can tell. Nothing in `src/__tests__/` asserts on the highlight styling, so nothing should go red on you.

Once that's pushed and it goes green we'll get it merged. Shout if you'd rather we just take it from here and reapply it ourselves, that's completely fine too and no hard feelings given how long it's been sitting.

PR #735, opened 2026-04-06
fix: reconcile Auto Run summary with cumulative session stats

pedramamini:
## Summary

- Auto Run summary now reconciles in-memory counters with persisted history entries, ensuring accurate cumulative stats (tasks, tokens, cost, duration) even when sessions span app restarts
- Applied the fix to both the renderer (`useBatchProcessor.ts`) and CLI (`batch-processor.ts`) codepaths

Closes #734

## Root Cause

In-memory cumulative counters (`totalCompletedTasks`, `totalInputTokens`, `totalOutputTokens`, `totalCost`) are initialized fresh each time `startBatchRun` is called. For long-running sessions that survive app restarts, manual stop/restart cycles, or process kills, the final "Auto Run stopped" summary only reflected stats from the most recent invocation — while all individual task history entries persisted on disk across restarts.

## Fix

Before generating the final summary, read all persisted history entries for the session and compute cumulative totals from task entries (filtering out loop/session summary entries). Use `Math.max(inMemory, historyDerived)` to ensure the summary is accurate regardless of restart history. Falls back gracefully to in-memory counters if history read fails.

## Test plan

- [ ] Run an Auto Run session, let it complete naturally — summary should match in-memory counters (no regression)
- [ ] Start an Auto Run, restart the app mid-session, resume the Auto Run, then stop — summary should reflect tasks from all invocations
- [ ] Verify the summary entry in history shows correct cumulative tasks, tokens, cost, and duration
- [ ] Run `npm run lint` and `npm test` — all pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Auto Run summaries now reconcile metrics with persisted history to report accurate task completion counts, elapsed time, and costs, especially after session restarts.

* **Documentation**
  * Updated multi-phase feature organization guidance to use flat directory structure instead of nested subdirectories.

* **Tests**
  * Added comprehensive tests for history reconciliation behavior during Auto Run completion.


chr1syy (commented):
## Consolidated review — request changes recommended

Reviewed the diff plus the bot reports. Two clear **blockers**, several strongly recommended fixes, and a base-branch concern worth a decision before merge.

---

### 🔴 Blocker 1 — Reconciliation over-counts on every repeat run (Greptile P1, confirmed)

Greptile's P1 is correct and, in my opinion, **regresses the common case worse than the original bug fixes the rare case**. `window.maestro.history.getAll(session.cwd, sessionId)` (renderer) and `readHistory(undefined, session.id)` (CLI) both return **every** AUTO entry ever written for the session, including entries from prior, fully-completed Auto Run invocations on the same agent.

Concrete failure:
- Run #1 on agent A → 30 task entries persisted
- Run #2 on agent A completes 20 tasks
- At summary time: `taskEntries.length = 50`, `totalCompletedTasks = 20` → guard fires → summary reports **50 tasks + Run #1's tokens/cost** on top of Run #2.

A user who repeats Auto Run more than once on the same agent (the common case) ends up with inflated stats — the exact opposite of what #734 wanted. The restart-recovery scenario (rare case) gets fixed at the cost of the common path. Needs a `batchStartTime` lower bound on the filter so only pre-restart entries are pulled in, then `final = totalCompletedTasks + taskEntries.length` for the additive case.

### 🔴 Blocker 2 — `docs/releases.md` was edited manually

[CLAUDE.md](https://github.com/RunMaestro/Maestro/blob/main/CLAUDE.md) is explicit:
> ## Do Not Edit: `docs/releases.md`
> `docs/releases.md` is generated/updated automatically during release pressing. **Never modify it manually**

The PR contains +254/−240 here from commit `0ea63689c` ("style: format docs/releases.md with prettier"). It's pure prettier churn (line endings + heading levels + bullet styles), but it shouldn't ship. Revert the file entirely — that also moots CodeRabbit's heading-hierarchy, bullet-style, and `FIle` typo nits, since none of those should land via this PR.

---

### 🟠 Strongly recommended

**3. `Document stalled:` entries aren't excluded (CodeRabbit, Major).** They're written as `type === 'AUTO'` in the same code path (around `useBatchProcessor.ts:1287-1295`) and will sneak past the current four-prefix filter, further inflating counts. Add `!entry.summary.startsWith('Document stalled:')`.

**4. CLI `finalTotal*` are unused in the terminal `complete` event (CodeRabbit, Major).** `batch-processor.ts` correctly stamps the persisted history entry with reconciled totals, but the generator still emits the in-memory `totalCompletedTasks` / `totalElapsedMs` / `totalCost` to CLI/JSONL consumers. After a resume, scripts that consume the stream still see the under-counted values. Wire `finalTotal*` into the emit alongside the history write.

**5. Silent `catch {}` violates Sentry policy (both bots + me).** [CLAUDE.md error-handling section](https://github.com/RunMaestro/Maestro/blob/main/CLAUDE.md#error-handling--sentry) says unexpected errors must surface to Sentry. Replace with `captureException(err, { sessionId, cwd: session.cwd, op: 'auto-run-summary-reconcile' })` (renderer) and a `logger.warn` + Sentry capture in the CLI path.

**6. No regression tests for either codepath.** The fix is two parallel filter+aggregate implementations with subtle gating; exactly the shape that warrants tests. The PR body's test-plan checkboxes are all unchecked. Recommend at least one renderer + one CLI test covering: (a) restart-recovery (history > in-memory) — the case being fixed, (b) repeat-run on same agent — guarding against the regression flagged above, (c) read-failure fallback.

**7. Filter is duplicated and brittle.** The four-prefix predicate is copy-pasted across renderer and CLI and couples to summary text owned by ~6 other files. Verified the prefixes match writers today (`useBatchRunner.ts`, `batchLoopSummary.ts`, `batchFinalSummary.ts`, `useBatchKillAction.ts`, `useAgentErrorListener.ts`, CLI `batch-processor.ts`), but any future rename silently breaks both. Extract a shared `isReconcilableTaskEntry(entry)` to `src/shared/` (or ideally introduce a `type: 'AUTO_TASK'` discriminator at write time so the filter becomes structural rather than string-based).

---

### ⚠️ Base branch — is `main` correct?

Two reasons to question targeting `main`:

1. **Behavioral risk.** This changes user-visible numeric output that feeds achievement/level progression and dashboard totals. The repo's `RC` label exists for "Getting soak time in RC branch now," and other recent Auto Run touch-ups by the same author (#1052, #1047, #1045) target `rc`. The over-counting regression flagged above is precisely the kind of issue an RC soak would catch.

2. **`rc` has already refactored the renderer touchpoint.** Commit `0ab779e4f` on `rc` decomposed `useBatchProcessor.ts` (2,281 → 328 LOC) and extracted the final-summary path into a pure function at `src/renderer/hooks/batch/internal/batchFinalSummary.ts` that receives counts as `FinalSummaryParams`. As written, this PR patches a call site that no longer exists on rc — when `rc` eventually merges down into `main`, this fix will either conflict or silently disappear. On the rc shape the reconciliation has to live in the **caller** that builds `FinalSummaryParams`, not inside `buildFinalSummary` (which is pure). Landing on `rc` first forces handling the new shape and avoids the silent-drop risk.

CLI side is unchanged on both branches (`batch-processor.ts` is 813 LOC on main, 849 on rc — only minor diffs) so that half ports cleanly either way.

---

### Summary

The intent is right and the under-counting issue is real, but as-is the PR introduces a worse regression for the common case (Greptile P1), edits a file CLAUDE.md prohibits, has no test coverage, and is structurally incompatible with the rc refactor it will eventually meet. Suggest:

- Revert `docs/releases.md`
- Add `batchStartTime` boundary + `Document stalled:` exclusion + tests
- Use the reconciled totals in the CLI `complete` event
- Replace silent catches with Sentry capture
- Confirm `rc` vs `main` base — recommend `rc`

Happy to take a pass at any of these if helpful.

pedramamini:
Addressed the review feedback. Pushed two commits.

**P1 - over-counting on repeat runs (greptile + greptile/coderabbit):** Fixed in both the renderer (\`useBatchProcessor.ts\`) and CLI (\`batch-processor.ts\`) paths, but not via the suggested \`e.timestamp < batchStartTime\` filter - that approach doesn't actually fix the reported scenario. Run #1 (30 tasks) completes before run #2 starts, so all 30 entries still pre-date \`batchStartTime\` and get added back to run #2's in-memory count → still 50.

Instead, I scope the history scan to entries written **after the most recent prior "Auto Run" summary entry**. A completed or stopped run always ends with such a summary; an app restart / process kill does **not**. So this correctly spans restarts (the actual bug in #734) while excluding earlier completed runs on the same agent/session. Added CLI regression tests for both the restart-reconciliation case and the no-cross-run-overcount case.

**Major - reconciled totals never reached the CLI \`complete\` event (coderabbit):** Extracted a \`computeReconciledTotals()\` helper, called once, and fed it into both the persisted summary entry and the terminal \`complete\` JSONL event so resumed runs report cumulative stats to CLI/JSONL consumers.

**P2 - silent catch blocks (greptile + coderabbit):** Both fall-back catches now log (renderer \`console.warn\`, CLI \`logger.warn\` with \`sessionId\` context). Added a test asserting the warn-and-fallback path.

**Out-of-scope docs churn (coderabbit pre-merge check):** Reverted the \`docs/releases.md\` line-ending normalization entirely - it's an auto-generated file and net-zero now. That also moots the \`FIle\` typo and heading/bullet nits, which were all in that churned file.

**Stale comment:** The \`maestro-system-prompt.md\` multi-phase guidance comment is against a revision no longer in the diff - that file isn't part of this PR anymore.

Local: \`tsc\`, \`eslint\`, \`prettier\`, and the batch-processor unit suites pass.

pedramamini:
Status from a batch merge pass: the cumulative-stats reconciliation is not on `main` and the underlying restart/resume gap still looks present, so it's still wanted - but `main` rewrote `batch-processor.ts` (+688/-573) and `useBatchProcessor.ts` (+140/-1810), so this needs the `computeReconciledTotals` logic re-applied to the new structure (confirm the bug still reproduces first). Not a mechanical rebase. Leaving open for a focused follow-up.

pedramamini:
Superseded by #1284 - a clean reimplementation on the latest base branch (this branch was unrebasable after the intervening refactor). Closing this PR and deleting the stale branch. Please review #1284 instead.

PR #736, opened 2026-04-06
fix: add SIGKILL escalation and deferred cleanup to process kill

pedramamini:
## Summary

Closes #733

- **SIGKILL escalation**: `ProcessManager.kill()` now schedules a SIGKILL fallback if the process doesn't exit within 5 seconds of SIGTERM. Previously, a hung process that ignored SIGTERM would survive indefinitely.
- **Deferred process map cleanup**: The process is no longer removed from the tracking map immediately on `kill()`. Instead, it stays until the exit event actually fires (handled by `ExitHandler` / `PtySpawner`), so the escalation timer can detect a still-running process and the UI accurately reflects the process state.

## Test plan

- [x] Existing process-manager tests updated and passing (37/37)
- [x] Type-check clean (`tsc --noEmit` on all configs)
- [x] ESLint clean
- [ ] Manual: start an Auto Run, stop mid-run — verify process dies within 5s
- [ ] Manual: simulate a hung process (e.g., `sleep 999`) — verify SIGKILL fires after timeout

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved process termination reliability with a forced termination fallback when graceful shutdown is unsuccessful.

* **Documentation**
  * Normalized markdown formatting in release notes.

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -83,7 +83,7 @@ The major contributions to 0.14.x remain:
 - Leaderboard tracking now works across multiple systems and syncs level from cloud 🏆
 - Agent duplication. Pro tip: Consider a group of unused "Template" agents ✌️
 - New setting to prevent system from going to sleep while agents are active 🛏️
-- The tab menu has a new "Publish as GitHub Gist" option  📝
+- The tab menu has a new "Publish as GitHub Gist" option 📝
 - The tab menu has options to move the tab to the first or last position 🔀
 - [Maestro-Playbooks](https://github.com/pedramamini/Maestro-Playbooks) can now contain non-markdown assets 📙
 - Improved default shell detection 🐚
@@ -112,10 +112,12 @@ Thanks for the contributions: @t1mmen @aejfager @Crumbgrabber @whglaser @b3nw @d
 - TAKE TWO! Fixed Linux ARM64 build architecture contamination issues 🏗️
 
 ### v0.13.1 Changes
+
 - Fixed Linux ARM64 build architecture contamination issues 🏗️
 - Enhanced error handling for Auto Run batch processing 🚨
 
 ### v0.13.0 Changes
+
 - Added a global usage dashboard, data collection begins with this install 🎛️
 - Added a Playbook Exchange for downloading pre-defined Auto Run playbooks from [Maestro-Playbooks](https://github.com/pedramamini/Maestro-Playbooks) 📕
 - Bundled OpenSpec commands for structured change proposals 📝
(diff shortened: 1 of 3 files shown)


kringle25:
This PR tackles two of the five root causes my bug identified:

1. SIGKILL escalation (your proposed fix #4) — After sending SIGTERM, a 5-second timer fires SIGKILL if the process is still tracked.

2. Deferred process map cleanup (your proposed fix #2, partially) — processes.delete(sessionId) is no longer called eagerly in kill(). The process stays in the map until the exit event actually fires, so Maestro won't report "stopped" while the process is still alive.

Issues I see

  1. Only partial fix — 3 of 5 root causes unaddressed

  Your issue clearly identified 5 architectural weaknesses. This PR only fixes #2 (premature map removal) and #4 (SIGTERM→SIGKILL escalation). Still open:

  - #1 (fire-and-forget kill chain) — The 6-hop chain with no confirmation back to the UI is unchanged.
  - #3 (no process-group killing) — Child processes/subagents can still be orphaned. The SIGKILL only targets the parent PID, not kill(-pgid). This is probably the most impactful gap — a hung agent with
  active tool calls will leave orphan children holding resources.
  - #5 (silent error swallowing) — The new escalation catch block logs at debug level, same pattern you flagged.

2. PTY process SIGKILL escalation path is questionable

For PTY processes (line ~222), the code calls process.ptyProcess.kill() (which sends SIGTERM via node-pty), then schedules globalThis.process.kill(pid, 'SIGKILL'). But ptyProcess.kill() may not expose a reliable PID — the process.pid stored in the tracking map may be the PTY master, not the child shell. If the PID is wrong or stale, the SIGKILL hits nothing (or worse, a recycled PID).

4. pty.onExit() may register duplicate handlers

In scheduleKillEscalation, pty.onExit(() => clearTimeout(escalationTimer)) adds a new listener every time kill() is called. If kill() is called multiple times on the same session (e.g., user clicks stop repeatedly), you stack up listeners. child.once('exit', ...) is fine since it auto-removes, but pty.onExit doesn't have a once semantic — check if node-pty's onExit returns a disposable or stacks.

5. No unref() on the escalation timer

The setTimeout in scheduleKillEscalation will keep the Node.js event loop alive for 5 seconds after the app otherwise wants to quit. Should call escalationTimer.unref() so it doesn't block app shutdown.

6. Windows path skipped entirely

The SIGKILL escalation is only scheduled on the non-Windows branches. The Windows killWindowsProcessTree path
(via taskkill) doesn't get the same escalation treatment. If taskkill fails silently (which you flagged in the issue), the process still survives with no fallback.

7. The releases.md changes are pure whitespace noise

The entire docs/releases.md diff is just normalizing trailing spaces and blank lines. It's ~250 lines of diff noise that makes the actual 2-file change harder to review. Should be a separate commit ordropped.

Summary
It's a solid incremental fix for the most critical symptom (process reported dead while still alive), but it's a partial fix at best. The process-group killing (#3) is arguably the biggest gap for real-world reliability — a SIGKILL on just the parent PID won't clean up orphan subprocesses from tool calls. I'd flag the unref() and onExit stacking issues as bugs to fix before merge if possible without a major refactor or collateral complexity.

Hope this helps!

chr1syy:
Closing as fixed in #846

PR #748, opened 2026-04-07
feat: show clickable session ID badge next to Auto Run Active status

scriptease:
Bildschirmfoto 2026-04-06 um 20 50 46

I have a feature proposal for the people like me that need to see something sometimes, it adds the live session id to the auto run status box, but since everything is a snapshot and not updated live it opens the session with (snapshot) but you can see the last message when something is slow or you're just curious.

## Summary by CodeRabbit

* **New Features**
  * Batch run header now shows the current active agent session as a small labeled badge (e.g., "XXXX (snapshot)").
  * Badge is clickable — opens the corresponding session in a new tab and supports providing a custom session name for easier identification.
  * Improved live tracking of which agent session is currently running during batch executions.

### DIFF
--- a/src/renderer/components/RightPanel.tsx
+++ b/src/renderer/components/RightPanel.tsx
@@ -16,6 +16,7 @@ import {
 	AlertTriangle,
 	Play,
 	XCircle,
+	ExternalLink,
 } from 'lucide-react';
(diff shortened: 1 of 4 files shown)


pedramamini:
instead of merging this in, we're going to be adding a thinking stream for auto run and will link the active session from there. don't want to add this into that area.

PR #792, opened 2026-04-11
fix: prevent SIGABRT crash when toggling Symphony off

scriptease:
Add unmounted guards and .catch() to floating promises in useSymphony hook. The debounced onUpdated callback and periodic auto-sync interval could fire after the component unmounts when Symphony is disabled, causing unhandled promise rejections on half-closed IPC channels.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability when navigating away from the application state view by preventing stale state update attempts.


pedramamini:
@copilot resolve the merge conflicts in this pull request

pedramamini:
Thanks for the fix, @scriptease — sharp catch on the post-unmount IPC calls and the cleanup is exactly right. Appreciate you taking the time to chase this down and write it up clearly. Merged!

pedramamini:
Thanks for the fix, @scriptease — appreciate the clean defensive guards and the tests covering both the post-unmount debounce path and the interval cleanup. Solid contribution.

PR #796, opened 2026-04-11
Security: Predictable UUID generation using Math.random()

tuanaiseo:
## Problem

The UUID helper uses `Math.random()` to generate v4-like IDs. `Math.random()` is not cryptographically secure and can be predicted in some environments. If these IDs are used for anything security-sensitive (session identifiers, authorization references, reset tokens, etc.), an attacker could potentially guess valid IDs.

**Severity**: `medium`
**File**: `src/shared/uuid.ts`

## Solution

Replace the implementation with a CSPRNG-backed generator, e.g. `crypto.randomUUID()` (or `crypto.getRandomValues` fallback). Keep non-crypto IDs clearly separated from any identifier used in trust or access-control decisions.

## Changes

- `src/shared/uuid.ts` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced UUID generation to use cryptographically secure random sources instead of non-cryptographic methods.
  * UUID generation now throws an error when secure randomness APIs are unavailable in the environment.

### DIFF
--- a/src/shared/uuid.ts
+++ b/src/shared/uuid.ts
@@ -1,16 +1,26 @@
 /**
  * Simple UUID v4 generator.
  *
- * Generates RFC 4122 compliant version 4 UUIDs using Math.random().
- * This is suitable for non-cryptographic purposes like session IDs
- * and history entry IDs.
+ * Generates RFC 4122 compliant version 4 UUIDs using a cryptographically
+ * secure random number source.
  *
  * @returns A UUID v4 string (e.g., "550e8400-e29b-41d4-a716-446655440000")
  */
 export function generateUUID(): string {
-	return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, (c) => {
-		const r = (Math.random() * 16) | 0;
-		const v = c === 'x' ? r : (r & 0x3) | 0x8;
-		return v.toString(16);
-	});
+	const cryptoObject = (globalThis as any).crypto;
+	if (cryptoObject?.randomUUID) {
+		return cryptoObject.randomUUID();
+	}
+
+	if (!cryptoObject?.getRandomValues) {
+		throw new Error('Secure UUID generation is not supported in this environment');
+	}
+
+	const bytes = new Uint8Array(16);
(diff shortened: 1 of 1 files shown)


pedramamini:
Thank you for the contribution, @tuanaiseo! This is a solid, well-scoped security improvement.

The implementation looks correct — `crypto.randomUUID()` as the primary path with a proper `getRandomValues` fallback, correct RFC 4122 v4 version and variant bit manipulation, and a clear error for unsupported environments (which shouldn't be hit in our Electron/Node runtime).

Happy to approve this one. Nice work! 🔒

pedramamini:
Thanks for the PR. Closing this one with context on the reasoning.

I traced every `generateUUID()` caller and none cross a trust boundary:

- `src/cli/services/batch-processor.ts` (4x) — local history entry IDs (`type: 'AUTO'`)
- `src/cli/services/agent-spawner.ts` — Claude `--session-id` for Auto Run isolation (passed *to* the agent CLI, not used as a token)
- `src/renderer/hooks/cue/usePipelineCanvasCallbacks.ts` — React node keys for the pipeline canvas

Maestro is a local desktop Electron app — there's no remote attacker to predict IDs against, and `src/main/stats/utils.ts:32` already comments *"Do not replace with generateUUID()"* for cases that need stable PKs. The PR description's threat model (session identifiers, authorization references, reset tokens) doesn't map to any actual use here.

That said, swapping to `crypto.randomUUID()` would be free hardening — but the proposed diff is 3× longer than it needs to be:

- The `(globalThis as any).crypto` cast is unnecessary — `crypto` is a typed global in both `lib.dom` and Node's globals
- The `getRandomValues` fallback is dead code — every runtime Maestro ships to (Electron main = Node ≥20, renderer = Chromium ≥120) has `crypto.randomUUID()`
- The new "throws if not supported" branch is a behavior regression on a previously-infallible helper

If you'd like to re-open with a one-line replacement (`return crypto.randomUUID();`), happy to take that. Otherwise closing as not-a-security-issue.

PR #797, opened 2026-04-11
Security: Remote script injection over insecure HTTP in renderer devtools connector

tuanaiseo:
## Problem

The renderer conditionally injects a script from `http://localhost:8097` without integrity verification. Any local process binding to that port can execute arbitrary JavaScript in the renderer context when hostname is `localhost`. This increases risk during development and can be abused on compromised developer machines.

**Severity**: `low`
**File**: `src/renderer/public/devtools-connect.js`

## Solution

Gate this logic behind an explicit development flag (e.g. `import.meta.env.DEV`), and avoid runtime remote script injection where possible. Prefer bundler/dev-only tooling hooks instead of loading arbitrary HTTP scripts at runtime.

## Changes

- `src/renderer/public/devtools-connect.js` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened devtools connection security by requiring explicit opt-in alongside localhost verification to prevent unintended activation.

### DIFF
--- a/src/renderer/public/devtools-connect.js
+++ b/src/renderer/public/devtools-connect.js
@@ -1,6 +1,6 @@
 // React DevTools: connects to standalone react-devtools app (npm install -g react-devtools)
 // Only attempts connection in dev mode (Vite serves on localhost:5173)
-if (window.location.hostname === 'localhost') {
+if (window.location.hostname === 'localhost' && window.localStorage.getItem('MAESTRO_ENABLE_DEVTOOLS_CONNECT') === 'true') {
 	var script = document.createElement('script');
 	script.src = 'http://localhost:8097';
 	script.async = false;


pedramamini:
Thanks for the report, but closing this one — the actual threat model doesn't justify the dev-ergonomics cost.

For an attacker to exploit `devtools-connect.js`, they need:

1. Local code execution on the developer's machine (to bind `127.0.0.1:8097`)
2. To beat React DevTools to that port
3. The user to be on a `localhost` hostname — which in Maestro means either the Vite dev server or the web build served from localhost. **The packaged Electron app loads from `file://`**, so this code path doesn't even run in production for desktop users
4. Payoff: JS execution in renderer context, which in dev is the same context where the dev's own first-party code already runs, and in the web build is sandboxed browser JS (no Node integration)

Any attacker capable of (1) has vastly better moves than racing a debug port.

The proposed fix also makes things slightly worse:

- It breaks the default dev experience — every developer would need to manually run `localStorage.setItem('MAESTRO_ENABLE_DEVTOOLS_CONNECT', 'true')` to get React DevTools working
- The script still ships in production bundles, just gated by a runtime flag
- It doesn't address the (already negligible) prod-web case meaningfully

Better solutions if we ever decided to act on this:

1. Exclude `devtools-connect.js` from production builds entirely (Vite conditional include)
2. Delete the file — React DevTools as a Chrome extension works fine inside Electron's renderer; the standalone connector is largely historical

Closing as won't-fix on the proposed approach.

PR #802, opened 2026-04-12
fix: detect and recover from silent session resume failure

adamcongdon:
## Summary

Fixes #773 — Chat loses context mid-session when the batch-mode `--resume` flag silently fails.

## Root Cause

Each user message spawns a new `claude` process with `--resume <agentSessionId>`. When the session ID becomes stale (server-side expiry, state loss, restart), `--resume` fails silently and the agent starts fresh. The old code **rejected** the new session ID, keeping the stale one — creating a permanent death spiral where every subsequent message also lost context.

### Detection & Recovery (`useAgentListeners.ts`)
- When agent returns a different session ID than expected, accept the new one (breaks death spiral)
- Clear stale `usageStats` and reset context gauge to 0%
- Add system log entry warning about context loss
- Fire toast notification to alert the user

### Spawn Guard (`useInputProcessing.ts`)
- Console warning when spawning batch agent without `agentSessionId` for a tab with existing logs

### Tests (`useAgentListeners.test.ts`)
- Resume failure detection (accepts new ID, clears stats, logs warning)
- No false positives on successful resume (same ID returned)
- Context gauge preservation on successful resume

## Test Results

- 22,453 tests pass
- Build succeeds
- Lint clean

## Files Changed

| File | Lines |
|------|-------|
| `src/renderer/hooks/agent/useAgentListeners.ts` | +55, -1 |
| `src/renderer/hooks/input/useInputProcessing.ts` | +13 |
| `src/__tests__/renderer/hooks/useAgentListeners.test.ts` | +98 |

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session resume failure detection with a warning toast notifying users when previous context is lost during session recovery.
  * Session usage data and context are now properly reset when a session resume fails.

* **Tests**
  * Extended test coverage for agent listener behavior across session resume success and failure scenarios.

### DIFF
--- a/src/__tests__/renderer/hooks/useAgentListeners.test.ts
+++ b/src/__tests__/renderer/hooks/useAgentListeners.test.ts
@@ -608,6 +608,14 @@ describe('useAgentListeners', () => {
 				id: 'tab-1',
 				agentSessionId: 'old-session-id',
 				awaitingSessionId: false,
+				usageStats: {
+					inputTokens: 100,
+					outputTokens: 50,
+					cacheReadInputTokens: 0,
+					cacheCreationInputTokens: 0,
+					totalCostUsd: 0.001,
+					contextWindow: 200000,
+				},
 			});
 			const session = createMockSession({
 				id: 'sess-1',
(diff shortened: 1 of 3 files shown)


pedramamini:
Thanks for the fix, @adamcongdon — solid root-cause analysis on the death-spiral and the recovery flow is exactly right. Pushed one small follow-up to address the remaining review feedback:

- Seeded `usageStats` on the resume-failure test so the `toBeUndefined()` assertion actually exercises the clearing logic (was trivially passing on `undefined`).
- Added a "no resume-failure log" assertion to the success-path test so it matches its name.

All 39 tests in the file pass. Ready to merge from my side.

pedramamini:
Merged. Thanks again @adamcongdon — really appreciate the thoughtful root-cause write-up on the death-spiral and the clean recovery flow. This is a meaningful fix for users who hit stale session IDs mid-conversation. Looking forward to your next contribution.

PR #823, opened 2026-04-13
refactor: migrate to canonical shared hooks (Phase 09)

jSydorowicz21:
## Summary

Migrates a conservative subset of hook patterns to canonical shared hooks. **This phase is intentionally narrow** - most call sites had semantic subtleties that would have caused regressions if blindly replaced.

**Net: -19 lines across 4 files**

### 09A - useEventListener migrations (4 sites)

Replaced raw `addEventListener` / `removeEventListener` pairs with the canonical `useEventListener` hook from `src/renderer/hooks/utils/useEventListener.ts`:

- `src/renderer/hooks/ui/useTourActions.ts` - `tour:action` event
- `src/renderer/components/SessionList/SessionList.tsx` - `tour:action` event
- `src/renderer/hooks/session/useHandsOnTimeTracker.ts` - `beforeunload` event
- `src/renderer/components/MarketplaceModal.tsx` - `keydown` event in PlaybookDetailView

### 09A - useFocusAfterRender migrations (0 sites)

**Intentionally zero** - the canonical hook uses `useLayoutEffect` with NO dep array and re-runs every render while its `condition` is true. Replacing the common `useEffect(() => { setTimeout(() => ref.current?.focus(), 50); }, [])` pattern would cause the focus to be re-stolen on every subsequent render, breaking the app.

The one legitimate use of the canonical hook (transition-triggered focus in `App.tsx`) is already wired up.

### 09B - (0 sites migrated)

**`useActiveSession()` semantic gap**: `selectActiveSession` falls back to `state.sessions[0]` when there's no active session; inline `sessions.find(s => s.id === activeSessionId)` returns `undefined`. Blindly migrating all ~28 call sites would surface stale-session data instead of the "no active session" branch the components rely on for modal/UI routing.

**Debounce/throttle misfits**: Every hand-rolled debounce/throttle in the codebase has a reason the canonical hook doesn't fit - `useAutoRunUndo` uses dep-scoped cleanup, `useSessionDebounce` keys timers per-session, `useTabHoverOverlay` shares timers for open/close, `useThemeStyles` uses RAF-batched per-element tracking, `GitStatusWidget` uses bidirectional clear.

### Sites intentionally skipped in 09A

- Conditional registration on `isOpen` or `capture: true` / `passive: true` options (canonical hook doesn't support listener options) - DirectorNotesModal, SettingsSearch, NewInstanceModal, EditAgentModal, PhaseReviewScreen, UsageDashboardModal, LayerStackContext, activityBus, useThemeStyles, useRemoteHandlers
- Multi-event handlers or handlers paired with initial imperative calls - useMobileLandscape, useTour, TabBar

## Test plan

- [ ] `npm run lint` passes (all 3 tsconfigs)
- [ ] `npx prettier --check .` passes
- [ ] 312 targeted hook tests pass
- [ ] Tour events still fire
- [ ] Marketplace playbook detail keyboard shortcuts work
- [ ] beforeunload handler fires (active-time tracking)

## Risk

**Very low**. 4 behavior-preserving migrations, no component changes.

## Note

The playbook target was ~490 lines. The delta reflects an honest audit of the hook surface area rather than optimistic projections - many of the ~490 lines estimate come from call sites whose semantics don't map cleanly to the canonical hooks, and forcing them would regress UX. Leaving those deliberately unmigrated until a larger hook API surface (listener options, custom fallbacks) is designed.

### DIFF
--- a/src/renderer/components/MarketplaceModal.tsx
+++ b/src/renderer/components/MarketplaceModal.tsx
@@ -28,6 +28,7 @@ import type { MarketplacePlaybook } from '../../shared/marketplace-types';
 import { useLayerStack } from '../contexts/LayerStackContext';
 import { MODAL_PRIORITIES } from '../constants/modalPriorities';
 import { useMarketplace } from '../hooks/batch/useMarketplace';
+import { useEventListener } from '../hooks/utils/useEventListener';
 import {
 	REMARK_GFM_PLUGINS,
 	generateProseStyles,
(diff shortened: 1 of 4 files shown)


jSydorowicz21:
Superseded by #834 (same branch, this PR was accidentally closed during force-push operations).

PR #830, opened 2026-04-13
refactor: consolidate constants and minor dedup (Phase 12)

jSydorowicz21:
## Summary

Removes duplicate `AUTO_RUN_FOLDER_NAME` constants and cleans up related wrapper/barrel exports across the wizard / session pipeline.

**Net: -37 lines across 17 files**

### Changes

- 3 previous local definitions of `AUTO_RUN_FOLDER_NAME` now import `PLAYBOOKS_DIR` from `src/shared/maestro-paths.ts`
- Dropped deprecated service wrappers that existed solely to forward to the canonical implementation
- Dropped orphan barrel re-exports from `Wizard/index.ts` and `Wizard/services/index.ts`

### Why this is smaller than the playbook target (~126 lines)

- `DEFAULT_CAPABILITIES` was already consolidated in [Phase 02](https://github.com/RunMaestro/Maestro/pull/811) - double-counted in the original playbook
- CSS class-name extraction was intentionally skipped - most Tailwind utility combos either appear in only 2-3 call sites (not worth a shared module) or are context-dependent (extracting them would force adding a theme/variant prop). Leaving this for a future round once a dedicated design-token system is in place.

## Test plan

- [x] `npm run lint` passes (all 3 tsconfigs)
- [x] `npx prettier --check .` passes
- [x] 262 targeted tests pass (useSessionCrud, useSessionRestoration, useWizardHandlers, existingDocsDetector, WizardIntegration)
- [x] Wizard resumes existing Auto Run docs correctly (uses `PLAYBOOKS_DIR` for the scan)
- [x] Session creation routes to correct `.maestro/playbooks/` directory

## Risk

**Very low**. Pure constant consolidation - same value, different import path.

### DIFF
--- a/src/__tests__/renderer/components/Wizard/WizardIntegration.test.tsx
+++ b/src/__tests__/renderer/components/Wizard/WizardIntegration.test.tsx
@@ -197,7 +197,6 @@ vi.mock('../../../../renderer/components/Wizard/services/phaseGenerator', () => 
 		isGenerationInProgress: vi.fn().mockReturnValue(false),
 		abort: vi.fn(),
 	},
-	AUTO_RUN_FOLDER_NAME: '.maestro/playbooks',
 }));
 
 // Mock theme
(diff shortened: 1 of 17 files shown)


pedramamini:
Thanks for the contribution, @jSydorowicz21! Clean work here — the consolidation of `AUTO_RUN_FOLDER_NAME` into a single `PLAYBOOKS_DIR` import is exactly the right call, and removing the deprecated re-export wrappers and orphan barrel exports keeps things tidy.

The test updates look correct too, including the expected-value fix in `useSessionRestoration.test.ts` where the old mock was using a stale `.maestro-autorun` value.

LGTM — approving.

PR #831, opened 2026-04-13
fix: resolve symlinks in Files panel and Auto Run folder tree

### LINKED ISSUE #829: Bug/Missing feature: Symlinked subdirectories are invisible in Files panel and Auto Run folder picker
## Summary

Symbolic links to directories are silently dropped in two places: the Files panel (Right Bar) and the Auto Run folder's document tree. Because `Dirent.isDirectory()` returns `false` for symlinks, symlinked subdirectories are never recursed into and never shown.

### 1. Files Panel (Right Bar → Files tab)

- `src/main/ipc/handlers/filesystem.ts` lines 123–132: uses `entry.isDirectory()` / `entry.isFile()` from `Dirent` — both return `false` for symlinks
- Result: symlinked subdirectories are not expandable; symlinked files are not shown at all
- SSH path (`remote-fs.ts`) correctly detects the `@` suffix from `ls -F` but then maps `isFile: false`, so symlinks are also dropped there

### 2. Auto Run Folder Document Tree

- `src/main/ipc/handlers/autorun.ts` `scanDirectory()` lines 98–107: recurses only when `entry.isDirectory()` is `true` — so symlinked subdirectories are never walked
- Result: when an Auto Run folder contains subdirectories that are symlinks, those subdirectories and all their `.md` documents are invisible
- SSH path (`scanDirectoryRemote()` line 166) explicitly checks `!entry.isSymlink`, so symlinked dirs are also excluded remotely

## Steps to Reproduce

1. Create a symlinked subdirectory inside a project:
   `ln -s /some/other/path ./my-project/linked-dir`
2. Open Maestro → Right Bar → Files tab: `linked-dir` does not appear
3. Set the project as an Auto Run folder: documents inside `linked-dir` are not discovered

## Root Cause

`Dirent.isDirectory()` and `Dirent.isFile()` do not follow symlinks — they inspect the link itself, not the target. Neither the Files panel nor Auto Run call `Dirent.isSymbolicLink()` or use `fs.lstat()` to resolve the target type.

---

## Fix Prompt

> **Files panel** (`src/main/ipc/handlers/filesystem.ts`, lines 123–132): After calling `fs.readdir()` with `withFileTypes: true`, check `entry.isSymbolicLink()`. If true, call `fs.stat(fullPath)` (which follows the symlink) to determine whether the target is a directory or a file, and set `isDirectory`/`isFile` accordingly. Add `isSymlink: boolean` to the `DirectoryEntry` interface in `src/main/preload/fs.ts` and pass it through so the renderer can show a visual indicator. Apply the same fix to the remote mapping in lines 113–119 (currently drops symlinks entirely).
>
> **Auto Run** (`src/main/ipc/handlers/autorun.ts`, `scanDirectory()` lines 79–118): After `entry.isSymbolicLink()` is detected, call `fs.stat()` on the entry to check whether the target is a directory. If so, recurse into it the same way a real directory would be handled. Apply the equivalent fix to `scanDirectoryRemote()` line 166 (remove the `!entry.isSymlink` guard and resolve the target type via a remote `stat` or `ls -ld` call).

pedramamini:
## Summary

- Symlinked subdirectories and files are now correctly resolved in the Files panel (Right Bar) and Auto Run document tree
- `Dirent.isDirectory()`/`isFile()` return `false` for symlinks — we now check `isSymbolicLink()` and follow with `fs.stat()` to determine the actual target type
- Remote (SSH) path: augmented the `ls -1AF` command in `readDirRemote` to also identify symlinks whose targets are directories, so they can be recursed into
- Removed overly-strict `!entry.isSymlink` guards in Auto Run that excluded symlinked `.md` files and prevented recursion into symlinked directories
- Fixed `fs:countItems` to also follow symlinks when counting items recursively

Closes #829

## Test plan

- [ ] Create a symlinked subdirectory inside a project (`ln -s /some/path ./project/linked-dir`)
- [ ] Open Maestro → Right Bar → Files tab: verify `linked-dir` appears and is expandable
- [ ] Place `.md` files inside the symlinked directory
- [ ] Set the project as an Auto Run folder: verify documents inside the symlinked directory are discovered
- [ ] Verify broken symlinks don't crash the file listing (should appear as files or be skipped)
- [ ] Verify existing tests pass (22,451 tests passing)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved symbolic-link handling across directory scans, listings, image discovery, counting, and backup removal — symlinked files and folders are now resolved, recursed into when appropriate, and broken symlinks are gracefully skipped; markdown and image detection now include valid symlinked files. Added recursion depth protection to avoid infinite loops from directory cycles.

* **Tests**
  * Added and updated tests for symlink resolution, broken-symlink behavior, recursion, and cycle protection.

### DIFF
--- a/src/__tests__/main/ipc/handlers/autorun.test.ts
+++ b/src/__tests__/main/ipc/handlers/autorun.test.ts
@@ -408,6 +408,103 @@ describe('autorun IPC handlers', () => {
 			expect(result.tree).toHaveLength(2);
 		});
 
+		it('should include symlinked .md files as documents', async () => {
+			vi.mocked(fs.stat).mockImplementation((p: any) => {
+				// First call: the top-level folder. Subsequent calls: symlink resolution.
+				if (p === '/test/folder') {
+					return Promise.resolve({ isDirectory: () => true, isFile: () => false } as any);
+				}
+				// Symlink target is a file
+				return Promise.resolve({ isDirectory: () => false, isFile: () => true } as any);
+			});
+
+			vi.mocked(fs.readdir).mockResolvedValue([
+				{
+					name: 'linked-doc.md',
+					isDirectory: () => false,
+					isFile: () => false,
+					isSymbolicLink: () => true,
+				},
+				{
+					name: 'real.md',
+					isDirectory: () => false,
+					isFile: () => true,
+					isSymbolicLink: () => false,
+				},
+			] as any);
+
+			const handler = handlers.get('autorun:listDocs');
+			const result = await handler!({} as any, '/test/folder');
+
+			expect(result.success).toBe(true);
+			expect(result.files).toEqual(['linked-doc', 'real']);
+		});
+
+		it('should recurse into symlinked folders containing .md files', async () => {
+			vi.mocked(fs.stat).mockImplementation((p: any) => {
+				if (p === '/test/folder') {
+					return Promise.resolve({ isDirectory: () => true, isFile: () => false } as any);
+				}
+				// Symlink target is a directory
+				return Promise.resolve({ isDirectory: () => true, isFile: () => false } as any);
+			});
+
+			// Root contains a symlinked folder; the folder contains nested.md
+			vi.mocked(fs.readdir)
+				.mockResolvedValueOnce([
+					{
+						name: 'linked-folder',
+						isDirectory: () => false,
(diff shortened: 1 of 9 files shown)


pedramamini:
@greptile re-analyze please

PR #834, opened 2026-04-13
refactor: migrate to canonical shared hooks (Phase 09)

jSydorowicz21:
## Summary

Migrates a conservative subset of hook patterns to canonical shared hooks. **This phase is intentionally narrow** - most call sites had semantic subtleties that would have caused regressions if blindly replaced.

**Net: -19 lines across 4 files**

### 09A - useEventListener migrations (4 sites)

Replaced raw `addEventListener` / `removeEventListener` pairs with the canonical `useEventListener` hook from `src/renderer/hooks/utils/useEventListener.ts`:

- `useTourActions.ts` - `tour:action` event
- `SessionList.tsx` - `tour:action` event
- `useHandsOnTimeTracker.ts` - `beforeunload` event
- `MarketplaceModal.tsx` - `keydown` event in PlaybookDetailView

### 09A - useFocusAfterRender migrations (0 sites)

**Intentionally zero** - the canonical hook uses `useLayoutEffect` with NO dep array and re-runs every render while its `condition` is true. Replacing the common `useEffect(() => { setTimeout(() => ref.current?.focus(), 50); }, [])` pattern would cause the focus to be re-stolen on every subsequent render, breaking the app.

### 09B - (0 sites migrated)

**`useActiveSession()` semantic gap**: `selectActiveSession` falls back to `state.sessions[0]` when there's no active session; inline `sessions.find` returns `undefined`. Blindly migrating would surface stale-session data instead of the "no active session" branch.

**Debounce/throttle misfits**: Every hand-rolled debounce/throttle has a reason the canonical hook doesn't fit.

Supersedes #823 (accidentally closed during rebase).

## Test plan

- [ ] `npm run lint` passes (all 3 tsconfigs)
- [ ] `npx prettier --check .` passes
- [ ] Tour events still fire
- [ ] Marketplace playbook detail keyboard shortcuts work
- [ ] beforeunload handler fires (active-time tracking)

## Risk

**Very low**. 4 behavior-preserving migrations, no component changes.

## Summary by CodeRabbit

* **Refactor**
  * Centralized event listener handling across the app (preview keyboard shortcuts, tour actions, session menu toggles, and session time tracking). Behavior is unchanged; no visible impact to users.

### DIFF
--- a/src/renderer/components/MarketplaceModal.tsx
+++ b/src/renderer/components/MarketplaceModal.tsx
@@ -28,6 +28,7 @@ import type { MarketplacePlaybook } from '../../shared/marketplace-types';
 import { useLayerStack } from '../contexts/LayerStackContext';
 import { MODAL_PRIORITIES } from '../constants/modalPriorities';
 import { useMarketplace } from '../hooks/batch/useMarketplace';
+import { useEventListener } from '../hooks/utils/useEventListener';
 import {
 	REMARK_GFM_PLUGINS,
 	generateProseStyles,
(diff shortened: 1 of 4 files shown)


pedramamini:
Thanks for the contribution, @jSydorowicz21! This is a clean, well-scoped refactor.

The 4 `useEventListener` migrations are all behavior-preserving — the hook's `handlerRef` pattern means closures stay fresh without needing deps. The detailed rationale for what was intentionally *not* migrated (listener options, conditional registration, semantic gaps in `useActiveSession`, etc.) is especially appreciated — it shows real care in avoiding regressions.

No merge conflicts, net -19 lines, LGTM. Approving.

PR #839, opened 2026-04-15
Add Settings-managed maestro-cli status and install/update

chr1syy:
## Summary
Introduces Settings-managed Maestro CLI detection and install/update flow.

Users can now verify whether `maestro-cli` is available, confirm version parity with the running Maestro app, and install/update the CLI directly from **Settings → General**.

## What Changed
- Added `MaestroCliManager` in the main process to:
  - Resolve bundled `maestro-cli.js`
  - Detect existing CLI command availability in app PATH and expanded shell PATH
  - Compare installed CLI version to app version (strict match)
  - Install/update a user-level shim in `~/.local/bin` (or Windows equivalent)
  - Update user PATH where possible and report failures clearly
- Added IPC handlers + preload API for:
  - `maestroCli.checkStatus()`
  - `maestroCli.installOrUpdate()`
- Added shared types for CLI status/install result payloads.
- Added a new **Maestro CLI** section in **General Settings**:
  - Shows PATH detection, command path, installed/expected version
  - Highlights mismatch/missing CLI state
  - Provides **Check now** and **Install / Update CLI** actions
  - Includes a single accessible live region for status/error/install announcements

## Follow-up Fixes from Review
- Replaced unsafe shell quoting in CLI shim generation and switched to app runtime execution.
- Added expanded PATH visibility (`inShellPath`) and corrected install-needed logic to focus on presence/version.
- Improved shell profile behavior:
  - Prefer shell-specific rc files (`.zshrc`, `.bashrc`)
  - Use `.profile` only as fallback
- Surface PATH update failures to the UI (`pathUpdateError`).
- Removed sync-path anti-pattern (`existsSync` in async flow).
- Made PATH export idempotency more robust.
- Derived POSIX PATH export entry from the configured install directory instead of hardcoded path.
- Added runtime-safe app-version fallback in General settings rendering to avoid test/runtime crashes when `__APP_VERSION__` is not injected.

## Validation
- `npm run lint`
- `npm run lint:eslint`
- `npm run format:check`
- `npx vitest run src/__tests__/renderer/components/Settings/tabs/GeneralTab.test.tsx src/__tests__/renderer/components/SettingsModal.test.tsx`

## Notes
- CLI target remains user-scoped (non-root) installation.
- Version compatibility requires exact app/CLI version match.


pedramamini:
Hey @chr1syy — thanks for the contribution! This is a well-structured addition and follows the existing patterns (IPC handlers, preload API, shared types, settings UI) nicely. The sequence diagram from Greptile's bot is a good illustration of how clean the architecture is.

After reviewing the diff, here are a few things I'd like to see addressed before merging:

### P0 — Unix shim shell injection via `JSON.stringify`

In `writeUnixShim` (maestro-cli-manager.ts ~line 98):

```typescript
const script = `#!/usr/bin/env bash\nnode ${JSON.stringify(bundledCliPath)} "$@"\n`;
```

`JSON.stringify` doesn't escape shell-significant characters inside bash double quotes (`$`, backticks, `!`). If the app is installed under a path containing `$` or backticks, this becomes a code-execution vector. Please switch to single-quoting with proper escaping:

```typescript
const safePath = bundledCliPath.replace(/'/g, "'\\''");
const script = `#!/usr/bin/env bash\nnode '${safePath}' "$@"\n`;
```

### P1 — `ensurePosixPathExport` unconditionally adds `.profile`

The `.profile` is always in the `rcFiles` set (line ~128), even when the user's shell is zsh or bash. The intent seems to be "add `.profile` only as a fallback when shell is unknown." As-is, a zsh user gets the PATH export appended to both `.zshrc` **and** `.profile`, which is unexpected. Consider only adding `.profile` inside the `!shellName` fallback branch.

### P1 — Partial success not surfaced to user

`ensureWindowsUserPath` returns `false` on failure (line ~187) and logs an error, but `installOrUpdate` can still return `success: true` if the binary itself runs. The user sees "CLI installed" in the UI but won't find `maestro-cli` in a new terminal because PATH wasn't updated. Per project convention, unexpected failures like a failed `SetEnvironmentVariable` should bubble up or at minimum be surfaced in the result so the UI can show a meaningful warning.

### P2 — `fs.existsSync` in async method

`checkStatus` (line ~175) uses `fs.existsSync` inside an otherwise fully-async method. This blocks the main process event loop. Minor in practice but easy to fix:

```typescript
const installShimExists = await fs.promises.access(installPath, fs.constants.F_OK).then(() => true).catch(() => false);
```

### P2 — Idempotency check uses substring matching

The check in `ensurePosixPathExport` (line ~143) uses `contents.includes(expectedEntry)` which matches substrings. If a user has `$HOME/.local/bin-custom` in their rc file, it would falsely skip the export. Checking for the marker comment `# Added by Maestro CLI installer` would be more robust.

---

Overall the structure and approach are solid. Happy to approve once the P0 and P1s are addressed!

chr1syy:
Addressed latest feedback in `5c31556`:

- Fixed `ensurePosixPathExport(installDir)` to derive PATH export from `installDir` (with `$HOME` normalization) instead of hardcoding `~/.local/bin`.
- Added a runtime-safe fallback for `__APP_VERSION__` usage in `GeneralTab` to prevent test/runtime `ReferenceError` when not injected.
- Kept the single live-region accessibility behavior for Maestro CLI status/install messaging.

Verification:
- `npm run lint`
- `npm run lint:eslint`
- `npm run format:check`
- scoped settings tests locally (`GeneralTab`, `SettingsModal`)
- GitHub Actions CI now passing (`lint-and-format`, `test`).

nehaaprasad:
### AI Code Trust — ship readiness

**Score:** 98 · **Verdict:** SAFE · **Model:** deterministic-v1

No blocking issues detected by automated checks.

**Top issues:**
- **maintainability** (medium) — File is very long (1100 lines); consider splitting. `src/main/index.ts:1`
- **maintainability** (medium) — File is very long (525 lines); consider splitting. `src/main/preload/index.ts:1`
- **maintainability** (medium) — File is very long (522 lines); consider splitting. `src/renderer/components/Settings/searchableSettings.ts:1`
- **performance** (low) — useEffect with an empty dependency array runs once — confirm dependencies are complete. `src/renderer/components/Settings/tabs/GeneralTab.tsx:201`
- **maintainability** (medium) — File is very long (1323 lines); consider splitting. `src/renderer/components/Settings/tabs/GeneralTab.tsx:1`
- **maintainability** (medium) — File is very long (3140 lines); consider splitting. `src/renderer/global.d.ts:1`
- **testing** (medium) — No test files detected alongside UI/source files. `src/main/index.ts`

PR #840, opened 2026-04-15
fix(terminal): inherit full parent env for Unix PTY to fix zsh

chr1syy:
## Summary
- Unix PTY terminal sessions used a minimal 6-variable environment (HOME, USER, SHELL, TERM, LANG, PATH) while Windows inherited the full parent env
- zsh depends on inherited variables (FPATH, XDG_*, TMPDIR, etc.) for ZLE line editor and plugin initialization — without them, input becomes invisible and navigation breaks
- Now inherits full parent environment on Unix (matching Windows), strips Electron/IDE vars, and overrides TERM/LANG/PATH with controlled values

## Test plan
- [x] Existing envBuilder tests pass (41/41)
- [x] New tests verify Unix env inheritance and Electron var stripping
- [x] Type check, lint, and prettier clean
- [x] Manual: open Maestro terminal with zsh — verify input is visible, navigation works, vi/vim `:q` renders correctly
- [x] Manual: verify bash terminal still works as before

## Summary by CodeRabbit

* **Tests**
  * Added unit tests validating terminal environment handling on Unix, including inheritance of user variables and removal of internal framework variables.

* **Bug Fixes**
  * Terminal environment now inherits parent environment on Unix/macOS while explicitly removing internal/framework-specific variables for improved isolation and consistent TERM/LANG/PATH defaults.

### DIFF
--- a/src/main/process-manager/utils/__tests__/envBuilder.test.ts
+++ b/src/main/process-manager/utils/__tests__/envBuilder.test.ts
@@ -485,14 +485,22 @@ describe('envBuilder - Global Environment Variables', () => {
 			try {
 				process.env.ELECTRON_RUN_AS_NODE = '1';
 				process.env.ELECTRON_NO_ASAR = '1';
+				process.env.ELECTRON_EXTRA_LAUNCH_ARGS = '--enable-features=something';
 				process.env.CLAUDECODE = 'true';
+				process.env.CLAUDE_CODE_ENTRYPOINT = '/path/to/entrypoint';
+				process.env.CLAUDE_AGENT_SDK_VERSION = '1.0.0';
+				process.env.CLAUDE_CODE_ENABLE_SDK_FILE_CHECKPOINTING = 'true';
 				process.env.NODE_ENV = 'test';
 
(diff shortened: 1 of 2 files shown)


pedramamini:
Thanks for the contribution, @chr1syy! This is a well-scoped and well-motivated fix.

**Review notes:**

- The spread-then-override pattern (`...process.env` → override TERM/LANG/PATH → strip Electron vars) is clean and correctly aligns Unix behavior with Windows, which already inherited the full parent env.
- Test coverage is thorough — all 8 `STRIPPED_ENV_VARS` are verified, and the platform mock cleanup in `finally` blocks is solid.
- JSDoc updates accurately reflect the new behavior.

One truly minor nit (not blocking): the `@see` comment on the `STRIPPED_ENV_VARS` constant (line ~109) still only references `buildChildProcessEnv()` — now that `buildPtyTerminalEnv()` also uses it on Unix, you could add a second `@see` there. Totally optional.

Looks good to merge! 🎉

ksylvan (approved):
LGTM!

PR #841, opened 2026-04-15
fix: `Cmd+Left/Right` tab shortcuts broken when sidebar has focus — scroll jumps instead

scriptease:
## Summary

- `Cmd+Left/Right Arrow` configured as previous/next tab shortcuts had no effect when the sidebar had focus — groups collapsed/expanded instead

## Root Cause

`handleSidebarNavigation` in `useKeyboardNavigation.ts` was catching all `ArrowLeft`/`ArrowRight` events when the sidebar had focus, regardless of modifier keys. Only `Alt+Cmd+Arrow` was previously excluded. The handler called `preventDefault()` and returned `true`, preventing the `nextTab`/`prevTab` shortcut handlers from ever being reached.

## Fix

Broadened the guard in `handleSidebarNavigation` to skip any `Cmd/Ctrl+Arrow` combination (with or without Alt). Plain `ArrowLeft`/`ArrowRight` (no modifiers) still handles group collapse/expand in the sidebar.

## Test plan

- [ ] Configure "Next Tab" → `Cmd+Right Arrow`, "Previous Tab" → `Cmd+Left Arrow`
- [ ] Click an agent in the sidebar to give it focus
- [ ] Press `Cmd+Right Arrow` — should navigate to next tab, not collapse a group
- [ ] Press `Cmd+Left Arrow` — should navigate to previous tab, not collapse a group
- [ ] Plain `ArrowLeft`/`ArrowRight` (no Cmd) still collapses/expands groups in sidebar
- [ ] `Alt+Cmd+Arrow` still toggles layout as expected

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined sidebar arrow-key keyboard navigation behavior when using Cmd/Ctrl modifier keys for more consistent shortcut handling.

### DIFF
--- a/src/renderer/hooks/keyboard/useKeyboardNavigation.ts
+++ b/src/renderer/hooks/keyboard/useKeyboardNavigation.ts
@@ -137,10 +137,12 @@ export function useKeyboardNavigation(
 				return false;
 			}
 
-			// Skip if Alt+Cmd+Arrow is pressed (layout toggle shortcut)
-			const isToggleLayoutShortcut =
-				e.altKey && (e.metaKey || e.ctrlKey) && (e.key === 'ArrowLeft' || e.key === 'ArrowRight');
-			if (isToggleLayoutShortcut) return false;
+			// Skip if Cmd/Ctrl+Arrow is pressed (with or without Alt) — could be a user-configured
+			// tab navigation shortcut (e.g. Cmd+Left/Right) or layout toggle (Alt+Cmd+Arrow).
+			// Plain ArrowLeft/Right (no modifiers) is reserved for group collapse/expand.
+			const isModifiedArrow =
+				(e.metaKey || e.ctrlKey) && (e.key === 'ArrowLeft' || e.key === 'ArrowRight');
+			if (isModifiedArrow) return false;
 
 			// Only handle arrow keys and space
 			if (!['ArrowUp', 'ArrowDown', 'ArrowLeft', 'ArrowRight', ' '].includes(e.key)) {


scriptease:
@pedramamini I tried to set CMD+left and right but the navigation view would keep stealing the arrow keys

pedramamini:
Hi @scriptease - thanks for hunting this down and fixing it! Narrowing the guard so plain arrows still drive group collapse/expand while `Cmd/Ctrl+Arrow` passes through to the tab shortcut handlers is exactly the right approach, and the comment you added makes the intent clear.

One thing I'd like addressed before merging (Greptile flagged this too, as P2): the guard only covers the horizontal arrows, so the same bug still bites the vertical ones.

```ts
const isModifiedArrow =
    (e.metaKey || e.ctrlKey) && (e.key === 'ArrowLeft' || e.key === 'ArrowRight');
if (isModifiedArrow) return false;
```

Because `ArrowUp`/`ArrowDown` aren't in that condition, any tab-nav shortcut bound to `Cmd/Ctrl+ArrowUp` or `Cmd/Ctrl+ArrowDown` is still swallowed by the sidebar's up/down navigation when it has focus. This isn't purely hypothetical: the default `nextUnreadTab` shortcut is `Alt+Meta+ArrowDown`, and it gets eaten by the sidebar today for this same reason.

Could you broaden the guard to include the vertical arrows so the fix is symmetric?

```ts
const isModifiedArrow =
    (e.metaKey || e.ctrlKey) &&
    (e.key === 'ArrowLeft' ||
        e.key === 'ArrowRight' ||
        e.key === 'ArrowUp' ||
        e.key === 'ArrowDown');
if (isModifiedArrow) return false;
```

Plain (unmodified) Up/Down still handle session navigation, so this only lets modifier+arrow combos through to the shortcut handlers, consistent with how you're already treating Left/Right.

A couple of smaller notes so you don't get sidetracked:

- No rebase needed. The branch merges cleanly against `rc`.
- CodeRabbit left two inline comments on `src/main/index.ts` (the RC-icon / empty-catch logic). That file isn't part of this PR's diff, so those don't apply to your change here and you can safely ignore them.

Once the guard covers Up/Down, I'm happy to approve. Thanks again for the contribution!

Issue #436, opened 2026-02-21
During Looping Auto-Run, UI loop counter is not accurate


ksylvan:
## What Happened

I did a looping AutoRun. In the history, I see messages like this:

<img width="785" height="432" alt="Image" src="https://github.com/user-attachments/assets/2f12e1b3-a0bc-4396-8f30-7476410cc9ae" />

At the same time, I see this  at the bottom of the panel:

<img width="805" height="151" alt="Image" src="https://github.com/user-attachments/assets/e615789e-dcfc-4acf-9f8f-26dcc084202f" />

## Expected Behavior

I'd like to see the loop counter change in real time and be accurate.

kianhub:
I think the loop count is accurate as it is the first run of all the documents, not any doc-internal loops.
The UI counter is showing index `i`, history can have index `j`.

pedramamini:
Is this on the latest RC? Because I did make some changes here.

ksylvan:
I think it was on the latest RC at that time. I should get into the habit of including the git hash in any issues I report.

pedramamini:
Not your fault. Can't wait to get this release out so we can just start incrementing the version number.

pedramamini:
@ksylvan let me know if you see this issue again please.

ksylvan:
Okay. Will do.

jeffscottward:
@pedramamini  close as resolved?

pedramamini:
This was fixed.

PR #846, opened 2026-04-16
Fix Auto Run reliability and task counting bugs

### LINKED ISSUE #229: AutoRun Playbooks picker incorrectly counts check-boxes that are in an example markdown block
For example, here:

Image

The picker shows "63%" when in reality 100% of the tasks have been completed.

### LINKED ISSUE #242: Incorrect Total Time on Auto Run during errors
## Screenshot

Image

Only did 1 Auto Run with ~30minutes before the error happened. Kept Re-trying until i restarted the app and got presented with another AutoRun-Badge for 1 hour (lvl 2 / 11).

## Suspected Issue

AutoRun was working on Regex for different types of errors, the agent might have given back info about the changed code which got interpreted as error in Maestro itself and appeared as SSH Connection refused error. With `retry`  multiple sessions got started over and over again and might have added time to the Auto Run progress through `App.tsx`:

```typescript
           // Multiply by number of concurrent sessions so each active Auto Run contributes its time
            // e.g., 2 sessions running for 1 minute = 2 minutes toward cumulative achievement time
            const deltaMs = elapsedMs * activeBatchSessionIds.length;

            // Update achievement stats with the delta
            const { newBadgeLevel } = updateAutoRunProgress(deltaMs);
```

It appears that errors during active AutoRuns just set `errorPaused: true` while the AutoRun still counts for `activeBatchSessionIds`, so time gets counted.

### LINKED ISSUE #338: Auto Run: agent hangs silently with no progress indication or timeout

### LINKED ISSUE #498: [BUG] Auto Run wizard unavailable with Factory Droid
The Auto Run wizard is unavailable for agent that use Factory Droid. I'll try to look into this later this week.

Image

### LINKED ISSUE #680: Auto Run Index lags while switching agents
Hi There! 

It seems that Maestro v0.16.4-RC has a bug where the playbook file view (via SSH) does not refresh properly when switching between agents running on different servers.

For example, if playbooks are generated, watched, or started on Agent 1, then after switching to Agent 2, the playbook view still shows the playbooks from Agent 1.

Hope this makes sense :)

### LINKED ISSUE #801: Bug: Auto-run loops forever when agent stalls without writing status file
## Problem

When an auto-run agent completes all work (commits pushed, tasks marked `[x]`) but stalls before writing its `.status` completion signal, the auto-run loop waits indefinitely. The agent process remains alive (sleeping at ~0% CPU) but produces no further output. Maestro never detects the stall and never advances or stops the run.

This has occurred twice in production use — both times the agent exhausted its context window during a large validation gate (running full test suites, fixing 100+ test failures) and hung after its final commit but before signaling completion.

## Root Cause Analysis

The auto-run engine's task completion detection relies on a simple file-polling loop:

```bash
while [ ! -f <task_id>.status ]; do sleep 5; done
```

This has no fallback for the case where the agent process stalls or crashes without writing the file:

### 1. No execution timeout

There is no upper bound on how long a single task iteration can run. An agent that hangs after 90 minutes of work is indistinguishable from one actively working.

### 2. No output-staleness detection

The agent's output stream (`.output` file) stops growing when it stalls, but nothing monitors this. In both observed incidents, the output file stopped growing 10+ minutes before the run was manually killed — a clear signal that went undetected.

### 3. No process-liveness verification

The agent process (Claude Code CLI) transitions to a sleeping state (~0% CPU, no I/O) but remains technically alive. The polling loop only checks for the `.status` file, not whether the process is still doing useful work.

### 4. Context exhaustion is a predictable failure mode

Validation gates that run full test suites and fix dozens of failures produce very large output (500+ KB). This reliably approaches or exceeds the context window, causing the agent to hang after its last tool call completes but before it can compose a final response and write the status signal.

## Observed Behavior

| Incident | IMPL Spec | Trigger | Output Size | Outcome |
|----------|-----------|---------|-------------|---------|
| 1 | IMPL-27 | Context limit during validation gate | — | Documented in INDEX.md |
| 2 | IMPL-37 | 137 test failures fixed in validation gate | 541 KB | Agent PID alive at 0.1% CPU, no `.status` written, polling loop running indefinitely |

In both cases, all work was committed and pushed successfully — the only failure was the completion signal.

## Proposed Fix

1. **Output staleness timeout** — Monitor the `.output` file's `mtime`. If no new output is written for N minutes (e.g., 5-10), treat the task as stalled. This catches the exact failure mode observed: agent completes work but hangs composing its response.

2. **Total execution timeout** — Set a configurable maximum wall-clock time per task iteration. If exceeded, kill the agent and mark the task for retry or manual review.

3. **Process activity check** — Periodically sample the agent process's CPU usage. A process sleeping at ~0% CPU for several minutes after previously being active is a strong stall signal.

4. **Graceful recovery** — On stall detection, check whether the agent's last commit matches the task's expected outcome (all checkboxes marked). If so, write the `.status` file on behalf of the agent and advance. If not, mark the task as failed/stalled for user review.

Related: #733 covers the inverse problem (user-initiated stop doesn't kill the process). This issue is about Maestro detecting the hang autonomously without user intervention.

## Environment

- macOS Darwin 25.4.0
- Maestro desktop app (Electron)
- Agent: Claude Code CLI (Claude Opus, 1M context)
- Auto-run mode: single-document loop

### LINKED ISSUE #829: Bug/Missing feature: Symlinked subdirectories are invisible in Files panel and Auto Run folder picker

chr1syy:
Closes #680, closes #733, closes #801, closes #242, closes #498, closes #829, closes #754, closes #338, closes #229

## Summary
This PR addresses a batch of Auto Run reliability and correctness bugs across process lifecycle handling, file watching, symlink discovery, and markdown task counting.

## Issues Addressed
- #680 - Auto Run index lag while switching agents
- #733 - Stop/kill reliability for hung auto-runs
- #801 - Auto-run loop hangs when agent stalls without status file
- #242 - Incorrect total time accounting during paused-error states
- #498 - Auto Run wizard unavailable with Factory Droid
- #829 - Symlinked subdirectories invisible in Files panel and Auto Run picker
- #754 - Right panel task progress not updating reliably during batch runs
- #338 - Agent hangs silently without timeout fallback
- #229 - Task counting incorrectly includes fenced markdown examples

### Process lifecycle and stall handling
- Improved stop behavior by killing all active batch-related process IDs for the session.
- Added watchdog safeguards for batch task execution:
  - inactivity timeout (no output)
  - hard wall-clock timeout
- Propagated watchdog failure context so task failures are surfaced with meaningful error messages.

### Auto Run state and counting correctness
- Excluded paused-error sessions from active-session time accounting.
- Reworked markdown task counting to ignore fenced code blocks.
- Normalized CRLF/CR line endings in task counting logic to prevent Windows/mixed-line-ending undercounts.

### Right panel/document refresh behavior
- Added stale-load guards during session switching to avoid cross-session document bleed-through.
- Added polling-based refresh for selected-document content during active runs.
- Updated watcher debounce to per-file timers to avoid dropped rapid change events.

### Symlink support
- Added symlink target resolution in filesystem directory listings (local + remote).
- Added symlink-aware recursion and cycle guards for Auto Run tree scanning.

### Agent capability fix
- Enabled Auto Run wizard support for Factory Droid.

## Validation
Targeted Vitest suites executed:
- `src/__tests__/renderer/hooks/useBatchProcessor.test.ts`
- `src/__tests__/renderer/hooks/useAutoRunHandlers.test.ts`
- `src/__tests__/renderer/hooks/useAutoRunDocumentLoader.test.ts`
- `src/__tests__/main/ipc/handlers/autorun.test.ts`
- `src/__tests__/main/ipc/handlers/filesystem.test.ts`
- `src/__tests__/renderer/stores/batchStore.test.ts`

Result: **477 passed, 0 failed**.

## Follow-up QA
- #436 (loop counter accuracy) is expected to be improved by related refresh/counting fixes, but should be explicitly validated via manual looping Auto Run repro before closure.

## Summary by CodeRabbit

* **New Features**
  * Wizard output enabled for the factory-droid agent
  * Directory listings now explicitly mark symlinks

* **Bug Fixes**
  * Markdown task counting skips fenced code blocks
  * Symlink targets resolved to avoid incorrect file/directory types
  * Better error propagation and structured error kinds for agent spawns
  * Active session selection excludes error-paused sessions

* **Improvements**
  * Per-file debounce for file-watching and safer watcher shutdown
  * More robust remote polling and progress polling; removed extra debug logs
  * Inactivity and timeout watchdogs for Auto Run batch sessions

* **Tests**
  * Updated tests to match new agent exit signatures and capability flag changes


chr1syy:
@greptile re-review please

pedramamini:
@chr1syy — thanks for the thorough batch, code review came back clean. Three of the four ready-to-merge PRs just went in (#875, #877, #857); this one can't merge yet because it's conflicting with `rc`. Could you rebase onto the latest `rc` and force-push?

Code itself is in good shape — I'm happy to merge once it's rebased and CI re-greens. Two small **optional** follow-ups (not blockers, don't hold up the rebase for them):

1. **`src/renderer/hooks/agent/useAgentExecution.ts:472`** — `BATCH_HARD_TIMEOUT_MS` is hardcoded at 60 minutes, but the inactivity timer is user-configurable up to 600 min (10 hrs). A user who raises the inactivity floor for a long-running agent will still get killed at 60 min by the hard timer. Either bump this to match the inactivity max or expose it as a second setting. Fine as a follow-up issue.

2. **`src/renderer/hooks/batch/useBatchProcessor.ts:488`** — `updateBatchStateAndBroadcast` now delegates to `_scheduleDebouncedUpdate(sessionId, updater, immediate)`. Worth confirming the `immediate` flag actually propagates through the underlying debounced scheduler (the prior `_immediate` underscore suggests it may have been ignored). If it's silently dropped, `flushDebouncedUpdate` calls in completion paths become the only thing forcing a sync flush — doable but worth a doc comment pinning the contract.

Everything else (watchdog correctness, symlink cycle guards via `realpath` dedup set, markdown-task counting with fenced-code skip + CRLF normalization, exit-signature widening being additive rather than breaking, Factory Droid wizard capability) checks out. 477 tests green on the right paths.

Ping me when rebased.

PR #856, opened 2026-04-18
fix(cue): production-harden pipeline editor — persistence, identity, manual triggers, animation fidelity

reachrazamair:
## Summary

Closes a batch of field-reported bugs in the Cue pipeline editor that collectively broke the "what I see = what's in YAML = what the engine does" invariant. All fixes are backward-compatible with existing `.maestro/cue.yaml` files, each has dedicated regression tests verified to fail when the fix is stashed, and the full suite (**25,602 tests**) passes.

Four commits:

1. Core rendering + identity + manual-trigger fixes (the main push)
2. Regression tests for per-instance prompt-file cleanup on agent delete
3. Per-edge animation fidelity (only the edge feeding the running agent animates)
4. Per-trigger animation fidelity (only the trigger whose sub is running spins)

---

### Rendering ↔ YAML contract
- **Multi-trigger prompt leakage**: when an agent had multiple incoming triggers, the second/third trigger's textarea rendered the first trigger's prompt. Removed the `inputPrompt` fallback at the selection layer and replaced it with event-type barebones templates (`DEFAULT_EVENT_PROMPTS` now covers every `CueEventType`).
- **Pipeline color drift**: colors changed across Dashboard tab switches, modal reopens, and app restarts. New optional `pipeline_color` field on every subscription makes YAML authoritative.
- **Pipeline grouping broke on subscription rename**: editing a subscription's `name` used to split a pipeline (grouping was done by stripping `-chain-N` off the sub name). New `pipeline_name` field is now the authoritative grouping key; the suffix convention is kept only as a legacy fallback.
- **Duplicate palette**: `src/shared/cue-pipeline-types.ts` and `src/renderer/components/CuePipelineEditor/pipelineColors.ts` had different 12-color lists. Consolidated to the canonical shared palette.
- **"Two agents swapped" on reload**: chain subscriptions referenced upstream agents by `sessionName` only, so renames/recreations caused chain edges to bind to the wrong node. New dual-written `source_session_ids` (stable UUIDs) is ID-preferred on load with name fallback.
- **Unresolved agent references silently swapped**: when `agent_id` pointed to a deleted session, the loader fell through to a heuristic that could pick the wrong agent. Unresolved references now surface as a visible red error node (new `ErrorNode` component), and save is blocked until resolved.
- **"Vanishing pipeline" on fan-out with differing prompts**: the validator required `prompt` or `prompt_file`, but our per-agent file externalization emitted neither — the lenient partition dropped the sub and the entire pipeline disappeared from the UI. Validator now accepts `fan_out_prompt_files` or `fan_out_prompts` as satisfying the prompt requirement.
- **Fan-out per-agent prompt asymmetry**: when fan-out targets had differing prompts, one prompt went to a `.md` file and the rest were inline-arrayed. New `fan_out_prompt_files` externalizes each per-agent prompt to its own file; the normalizer resolves them into `fan_out_prompts` at load time so the runtime dispatch path keeps reading one authoritative field.
- **Deterministic subscription ordering**: `subscriptionsToPipelines` now sorts by (initial-trigger-first, chain index, name) so re-serialized YAML can't produce an isomorphically-different graph.

### Manual trigger
- **Chain triggers unreachable from the Play button**: the trigger node's Play button fired `pipelineName`, which only matched the first sub. Chain triggers (GitHub PR/Issue polls, scheduled, etc.) were unreachable. Each trigger node now carries its own `subscriptionName` (stamped by `yamlToPipeline`) and the Play button fires that exact sub.

### Disk hygiene
- **`cue:deleteYaml` left orphan prompts**: clicking "Remove Cue configuration" deleted `cue.yaml` but left every `.maestro/prompts/*.md` behind. Now prunes with an empty keep-set and removes the prompts directory when empty.
- **GitHub `cue_github_seen` rows accumulated**: `clearGitHubSeenForSubscription` existed and was tested but never called. Wired into `refreshSession` (diff old vs new sub IDs) and `removeSessionInternal` (clear all).
- **Layout-JSON dedup**: `savePipelineLayout` now dedupes pipelines by `id` as a defensive backstop.
- **Removed `# Pipeline: …` header comment**: pipeline identity lives in the authoritative `pipeline_name` / `pipeline_color` fields on each subscription.

### Animation fidelity
- **All edges animated when any run was active**: pipeline-wide `isRunning` flag was set on every edge. New rule: an edge animates iff its target is an agent whose `sessionName` is in the pipeline's active-agents set.
  - Linear chain: exactly the edge feeding the running agent animates; animation shifts one edge at a time as the run advances.
  - Fan-out: every edge whose target is running animates concurrently; queued targets stay static.
  - Fan-in: all incoming edges to the running target animate; upstream edges stay static unless that source is itself running.
  - Cross-pipeline: runs in pipeline A don't animate pipeline B's edges even when both contain agents with the same `sessionName`.
  - Non-agent targets (cli_output, error nodes) never animate.
- **All trigger icons spun when any sub ran**: in a multi-trigger pipeline, a single sub firing flagged all three trigger nodes as running. New rule: a trigger animates iff its own `subscriptionName` is in the pipeline's running-subs set. Concurrent subs → multiple triggers spin; single sub → single trigger; legacy unsaved triggers fall back to the pipeline-wide flag (Play button hidden anyway).

---

## Schema additions (all optional, backward-compatible)

On `CueSubscription`:

| Field | Purpose |
|---|---|
| `pipeline_name` | Authoritative grouping key. Editing a sub's `name` no longer splits a pipeline. |
| `pipeline_color` | Hex color, round-tripped. Stable across Dashboard tab switches and restarts. |
| `source_session_ids` | Stable UUID reference for chain sources. Survives upstream rename. |
| `fan_out_prompt_files` | Per-agent fan-out prompt file paths. Externalizes differing prompts symmetrically. |

Legacy YAML without these fields loads unchanged; the next save stamps them in.

---

## Tests

- 16 new test files / describe blocks covering every surface.
- Verified by **stash-and-run** for each fix — the regression test fails when the fix is reverted, passes after restoration.
- Files added:
  - `cue-config-normalizer-fanout.test.ts`
  - `cue-fanout-engine-load.test.ts`
  - `cue-fanout-vanishing-pipeline.test.ts`
  - `cue-agent-delete-prune.test.ts`
  - `pipelineColorPalette.test.ts`
  - `pipelineColor.roundTrip.test.ts`
  - `pipelineRoundTrip.integration.test.ts`
  - `promptDefaults.test.ts`
  - `yamlToPipeline.chainIds.test.ts`
  - `yamlToPipeline.errorNodes.test.ts`
  - `yamlToPipeline.pipelineName.test.ts`
  - `usePipelineSelection.multiTrigger.test.ts`
- Existing tests updated where the old behavior was the bug (validator error messages, layout-JSON precedence, palette indices, `isRunning` default).

---

## Migration notes

- **No data migration required.** Legacy YAML loads via the existing fallback paths; first save after upgrade stamps the new fields.
- **Colors may shift once on first load** for YAMLs written before `pipeline_color` landed — palette fallback produces the same colors the user was seeing pre-upgrade, then the next save locks them in.
- **`dev:main` is a one-shot compile.** Restart `npm run dev` (or run `npm run build:main`) after pulling so `dist/main/*.js` picks up the validator / engine changes. Renderer-only changes hot-reload via Vite.

---

## Verification

- `npm run test` — **25,602 pass** / 107 skipped / 0 fail
- `npm run lint` — TypeScript clean
- Manual smoke tests in the pipeline editor:
  - Create a multi-trigger pipeline, assign distinct prompts per trigger, save → reload → prompts round-trip.
  - Three-agent fan-out with differing prompts → three `.md` files emitted, each per-agent; pipeline still renders on reopen.
  - Rename an upstream agent → chain edge resolves via ID, no error node.
  - Delete an upstream agent → red error node appears, save blocked until resolved.
  - Click Play on a GitHub trigger in a multi-trigger pipeline → that specific sub fires (previously unreachable).
  - Start a linear chain run → animation walks edge-by-edge through the chain as agents complete.
  - Start runs on two triggers concurrently → both triggers spin, third stays still.

---

## Commits

- `1b1afd56` fix(cue): resolve rendering contradictions, prompt leakage, agent identity, and manual-trigger bugs
- `de733644` test(cue): regression guards for per-instance prompt-file cleanup on agent delete
- `435983ef` fix(cue): animate only the edge feeding the currently-running agent
- `1cfb57e2` fix(cue): animate only the trigger node whose subscription is actually running

## Summary by CodeRabbit

* **New Features**
  * Visual error nodes for unresolved agents; persisted pipeline names/colors; per-subscription “Play” target; expanded default prompt API; external per-agent prompt files for fan-out.

* **Bug Fixes**
  * Fixes for fan-out persistence, prompt-file pruning and empty-dir cleanup, stable chain-source resolution, more accurate node/edge running animations, pipeline layout deduplication, and save blocked for unresolved-error pipelines.

* **Tests**
  * Large suite of regression/integration tests for fan-out, prompts, round-trips, error nodes, multi-trigger isolation, and layout/selection guards.


pedramamini:
Thanks for the thorough work on this, @reachrazamair — the regression coverage and the YAML-contract framing are really solid, and the animation-fidelity rule is a nice cleanup.

Before I label this `approved`, could you take a look at the automated review findings? A few of the "Major" ones look load-bearing for the invariants this PR is establishing:

- **Duplicate agent names in fan-out collide on prompt file paths** (`src/renderer/components/CuePipelineEditor/utils/pipelineToYaml.ts:226`) — both Greptile and CodeRabbit flagged this. `cuePromptFilePath(agentName, pipeline.name)` will overwrite earlier prompt content when two fan-out targets share a `sessionName`. Worth a stable disambiguator.
- **Stale `source_session_ids` falling through to name fallback** (`utils/yamlToPipeline.ts:188`) — if the stable ID is present but unresolved, silently falling back to `source_session` by name can reintroduce the exact swap bug the ID field was added to prevent. Probably should surface as unresolved instead.
- **Fan-in sort ordering** (`utils/yamlToPipeline.ts:378`) — `getChainIndex()` returns `0` for `*-fanin`, so the new deterministic sort may place fan-in targets before their chain links and resolve incorrectly.
- **Active-run mapping by stripped names** (`hooks/cue/usePipelineState.ts:303`) — these maps still use `subscriptionName.replace(...) === pipeline.name`, which conflicts with `pipeline_name` being the new authoritative grouping key (per-trigger animation fidelity could misattribute once a user renames a subscription).
- **Redundant shadowed `pipelineKey`** at `utils/yamlToPipeline.ts:844` — minor, but worth cleaning up while you're in there.

Also the `test` CI job is still running — let's make sure that's green before we land.

Once those are addressed (or you've pushed back on any you disagree with), happy to re-review and get this labeled.

PR #862, opened 2026-04-18
fix(cli): disable background tasks for Claude Code batch spawns

### LINKED ISSUE #861: maestro-cli send (batch mode): background tasks silently fail — should auto-set CLAUDE_CODE_DISABLE_BACKGROUND_TASKS=1
## Summary

When `maestro-cli send` is used in **batch/default mode** (no `--live` flag), the spawned Claude Code agent runs as a headless, short-lived process. If the agent uses `run_in_background` on any tool call, the task starts but the batch session exits before the background task completes — and the result is permanently lost. There is no persistent session to receive the async notification.

## The Issue

When using Maestro Discord, once in a while I cannot get the final result from the agent because it tells me that it spawned a background task, but it never returns. For a CLI call, it should not use background tasks and ask you to come later.

Image

## Reproduction

1. Set up any programmatic caller of `maestro-cli send` (a Discord bot, a CI script, etc.)
2. Trigger an agent that runs a long operation (e.g. a `repo sync` + code review)
3. The agent responds with something like *"Review running in background (ID xyz). I'll be notified when it finishes."*
4. The caller receives that message and the session ends
5. The background task completes inside the already-exited Claude Code process — the result is never delivered

## Root cause

`CLAUDE_CODE_DISABLE_BACKGROUND_TASKS` (introduced in Claude Code v2.1.4) exists precisely to prevent this. But batch callers of `maestro-cli send` have no way to know they need to set it, and the CLI doesn't set it automatically.

Background tasks in a batch session are always a footgun — there is no mechanism for the result to surface after the session exits. The `run_in_background` feature only makes sense in a persistent interactive session.

## Proposed fix

In the `spawnAgent()` / `spawnJsonLineAgent()` code path inside `maestro-cli send` (the non-`--live` batch path in `send.ts`), automatically inject:

```
CLAUDE_CODE_DISABLE_BACKGROUND_TASKS=1
```

See https://claudelog.com/faqs/what-is-disable-background-tasks-in-claude-code/

into the spawned Claude Code process environment. This ensures any agent running in batch mode executes all tool calls sequentially and returns a complete result before the session exits.

The `--live` path (which routes through the desktop app's renderer with a persistent session should be unaffected.

## Impact

Any programmatic or third-party caller of  — Discord bots, Slack integrations, CI pipelines, scripts — is silently affected by this today. The agent appears to respond, but long-running tasks never deliver their output.

EOF
)

scriptease:
Adds a new `batchModeEnvVars` field to agent definitions that is read only by the CLI spawners (spawnClaudeAgent, spawnJsonLineAgent), not by the desktop UI path or --live mode. Populates Claude Code with CLAUDE_CODE_DISABLE_BACKGROUND_TASKS=1 so run_in_background tool calls do not silently fail when the short-lived `maestro-cli send` batch session exits before the async task completes.

Shell-env-wins precedence matches the existing defaultEnvVars pattern, so users can still opt out by setting the var in their shell.

The original issues it that maestro-cli requests can cause background tasks to spawn that then never report back and the result is lost.

IMG_5030 png

The setting is documented here https://claudelog.com/faqs/what-is-disable-background-tasks-in-claude-code/

But Claude Opus 4.7 couldn't figure it out and had to be carried to the finish line.

Bildschirmfoto 2026-04-18 um 17 10 56

Fixes #861

## Summary by CodeRabbit

* **Chores**
  * Enhanced batch mode behavior by disabling background tasks for agent operations during CLI spawning, preventing interference with batch processes. Environment variable configuration respects pre-existing shell settings and ensures consistent batch operation execution.

### DIFF
--- a/src/__tests__/cli/services/agent-spawner.test.ts
+++ b/src/__tests__/cli/services/agent-spawner.test.ts
@@ -1400,6 +1400,55 @@ Some text with [x] in it that's not a checkbox
 				);
 			}
 		});
+
+		it('should set CLAUDE_CODE_DISABLE_BACKGROUND_TASKS=1 for claude-code batch spawns (#861)', async () => {
+			const resultPromise = spawnAgent('claude-code', '/project', 'prompt');
+			await new Promise((resolve) => setTimeout(resolve, 0));
+
+			const [, , options] = mockSpawn.mock.calls[0];
+			expect(options.env.CLAUDE_CODE_DISABLE_BACKGROUND_TASKS).toBe('1');
+
+			mockStdout.emit('data', Buffer.from('{"type":"result","result":"Done"}\n'));
+			mockChild.emit('close', 0);
+			await resultPromise;
+		});
+
+		it('should set CLAUDE_CODE_DISABLE_BACKGROUND_TASKS=1 even in read-only mode', async () => {
+			const resultPromise = spawnAgent('claude-code', '/project', 'prompt', undefined, {
+				readOnlyMode: true,
+			});
+			await new Promise((resolve) => setTimeout(resolve, 0));
+
+			const [, , options] = mockSpawn.mock.calls[0];
+			expect(options.env.CLAUDE_CODE_DISABLE_BACKGROUND_TASKS).toBe('1');
+
+			mockStdout.emit('data', Buffer.from('{"type":"result","result":"Done"}\n'));
+			mockChild.emit('close', 0);
+			await resultPromise;
+		});
+
+		it('should let a pre-set CLAUDE_CODE_DISABLE_BACKGROUND_TASKS from shell env win', async () => {
+			const originalValue = process.env.CLAUDE_CODE_DISABLE_BACKGROUND_TASKS;
+			process.env.CLAUDE_CODE_DISABLE_BACKGROUND_TASKS = '0';
+
+			try {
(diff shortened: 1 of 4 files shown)


pedramamini:
Thanks for the clean fix, @scriptease! 🙏

Reviewed the diff and I'm happy with this:

- **Scope is tight**: new `batchModeEnvVars` field is applied exclusively in the CLI spawners (`spawnClaudeAgent` / `spawnJsonLineAgent`), so the desktop UI and `--live` paths are untouched. Exactly right for a fix that targets short-lived `maestro-cli send` sessions.
- **Precedence is consistent**: the `!env[k]` shell-env-wins check matches the existing `defaultEnvVars` pattern a few lines down, so users can still opt out via their shell.
- **Good test coverage**: normal spawn, read-only mode, and the shell-override case are all exercised — plus a definition-level assertion. Nice.
- **Docs**: the inline comment on `batchModeEnvVars` explicitly calls out that it's CLI-only and points at #861, which is exactly the kind of breadcrumb future us will appreciate.

The CodeRabbit nitpick about extracting a helper / using a strict presence check is reasonable but not worth blocking on — staying consistent with the existing `defaultEnvVars` loop is the right call for now, and we can refactor both together if/when a third case lands.

Approving. 🟢

PR #873, opened 2026-04-19
feat(cue): resolve shared-workspace conflicts via owner_agent_id

### LINKED ISSUE #867: Multiple agents sharing a project root run every Cue subscription N times — no way to designate an owner
## Summary

When two or more Maestro agents are registered against the **same project directory**, they each independently load the project's `.maestro/cue.yaml` and each spin up their own copy of every subscription. The result: every trigger fires N times (once per agent sharing the workspace), every scheduled time runs N agents in parallel, every `task.pending` poll fans out N ways. There is currently no way to declare which agent in a shared workspace should "own" the Cue config.

## Concrete repro from my setup

I have an Obsidian vault registered as two agents (different model/profile, same workspace):

```
Obsidian          claude-code  $HOME/.../V1   id 160dff0f-...
Obsidian(Sonnet)  claude-code  $HOME/.../V1   id fe7c6b37-...
```

`$HOME/.../V1/.maestro/cue.yaml` contains a `time.scheduled` subscription at 18:00 and a `task.pending` subscription on `research.md`. As soon as the engine registers both agents, both fire every subscription — the 18:00 summary runs twice, the research-task pipeline kicks off two parallel investigations clobbering each other's outputs in the same `Research/<date>-<slug>/` folder.

This is a legitimate (and increasingly common) workflow: keep one workspace, switch between Opus and Sonnet agents against it depending on the task. Today, doing so multiplies the cost and chaos of every Cue trigger.

## Why it's a scheduler concern, not a YAML-author concern

A subscription already supports `agent_id` for *fan-out targeting*, but that doesn't help here — there's no way to say "only the agent with id X should ever evaluate this whole config". A YAML author who wants single-fire semantics today has to:

- duplicate the config per agent and gate each one with `agent_id` filters, OR
- pick one agent in the UI and remember never to use the other one for unattended work

Both are footguns. The engine knows which agents share a project root; it should be the one to dedupe.

## Proposed fix

Add an optional `owner_agent_id` (or similarly named) field to the top-level `settings:` block of `cue.yaml`:

```yaml
settings:
  owner_agent_id: 160dff0f-db39-4245-9756-39662cd75499  # Obsidian — Opus
  timeout_minutes: 30
  …

subscriptions:
  - name: "research queue"
    event: task.pending
    watch: "research.md"
    prompt_file: .maestro/prompts/cue-research.md
```

Behavior:

- If `owner_agent_id` is set, only the agent whose id matches loads/runs the subscriptions in this file. All other agents whose `projectRoot` equals this project see the config, log a one-line `[CUE] "Obsidian(Sonnet)" skipping cue.yaml — owner_agent_id targets "Obsidian"`, and register no triggers.
- If `owner_agent_id` is **absent**, fall back to a sensible default — see "Default policy" below.
- Per-subscription `agent_id` continues to work for fan-out within the owner.

## Default policy when `owner_agent_id` is unset

This is the trickier half. Two reasonable options:

1. **First-registered wins** — the first agent to call `initSession` for a given `projectRoot` "claims" the config; subsequent agents on the same root skip with a warning. Deterministic on a sorted session list, easy to reason about, but order-dependent if the user reshuffles agents.
2. **Warn loud, run anyway** — preserve today's behavior, but on detection of multiple agents sharing a project root with `cue.yaml`, surface a prominent banner in the Cue modal: *"2 agents share `/path/to/V1` — every trigger will fire twice. Set `settings.owner_agent_id` to disambiguate."*

I'd argue for (1) plus the banner from (2) — it's safer and the banner explains how to make it deterministic.

## Suggested fix bundle

1. Add `owner_agent_id?: string` to the `settings` schema.
2. In `cue-session-runtime-service.ts`'s `initSession`, after `loadCueConfigDetailed` returns ok, check if `config.settings.owner_agent_id` is present and != `session.id` — if so, log + return early.
3. If `owner_agent_id` is unset, group sessions by `projectRoot`, pick the first by stable order, and treat it as implicit owner. Log the implicit choice so users can see what claimed the config.
4. Surface a banner in the Cue modal when multiple sessions share a project root, regardless of whether `owner_agent_id` is set.
5. Docs: add a "Sharing a workspace across agents" section to `maestro-cue-configuration.md`.

## Related

- #865 — `task.pending` first-scan seeding
- #866 — dormant agents not visible in dashboard

(Found together while wiring up a research pipeline against an Obsidian vault that I happened to have registered as two agents.)

scriptease:
## Summary

Fixes #867 — when two or more agents share a `projectRoot`, every *unowned* Cue subscription (one without an explicit `agent_id`) currently fires once per agent. In the issue reporter's setup (an Obsidian vault registered as both an Opus and a Sonnet agent) the 18:00 summary ran twice, the `task.pending` research pipeline kicked off two parallel investigations, and their outputs clobbered each other in the same `Research/<date>-<slug>/` folder.

This PR resolves the conflict at the engine level rather than asking YAML authors to duplicate configs or pick one agent and remember never to unattended-run the other. It introduces a single optional field, `settings.owner_agent_id`, and a deterministic fallback — and surfaces the resulting state in the Cue dashboard so users can see which agent took the config and why.

## Why this is necessary

- The engine already knew which agents shared a project root; it just didn't dedupe. Every shared workspace silently multiplied cost and caused output collisions.
- `agent_id` on a subscription is for *fan-out targeting* within the owner, not for "which agent owns the config" — so it couldn't be used as a workaround.
- Users should not have to duplicate YAML or pick one agent out-of-band. Making this a property of the config keeps the rule next to the rules it governs.

## Resolution matrix

The ownership check is a pure utility (`computeOwnershipWarning`) — its return value is the single source of truth: a non-empty string means "non-owner, show the red ⚠️ in the dashboard and use this text as the tooltip"; `undefined` means "owner, no indicator".

| `owner_agent_id` | Agents in the projectRoot | Outcome for this session | Dashboard |
|---|---|---|---|
| Set, matches this session's id or name | — | Owner — fires unowned subs | No indicator |
| Set, matches *another* session in the root | This session is a non-match sibling | Skips unowned subs | Red ⚠️ — tooltip points to the owner |
| Set, matches **nobody** in the root | Every agent there is a non-match | Skips unowned subs | Red ⚠️ on every row — tooltip says the value matches no agent |
| Unset | Only one agent in the root | Owner — fires unowned subs | No indicator |
| Unset | >1 agent in the root, this is first in the session list | Owner — fires unowned subs | No indicator |
| Unset | >1 agent in the root, this is not first | Skips unowned subs | Red ⚠️ — tooltip names the winner and points to `owner_agent_id` as the override |
| Any value | Config was loaded from an ancestor dir | Ancestor fallback already filters to targeted subs — gate is a no-op | No indicator |

Subscriptions that carry an explicit `agent_id` continue to fan out regardless — the gate applies only to unowned subs.

## Files touched

- `src/shared/cue/contracts.ts` — `CueSettings.owner_agent_id`, `CueSessionStatus.ownershipWarning`
- `src/main/cue/config/cue-config-validator.ts` — validates the field
- `src/main/cue/cue-session-state.ts` — extracts the pure `computeOwnershipWarning` utility + stores warning on `SessionState`
- `src/main/cue/cue-session-runtime-service.ts` — reduces the ownership gate in `initSession` to a call to the utility
- `src/renderer/components/CueModal/SessionsTable.tsx` — red `AlertTriangle` + tooltip next to the session name
- `src/prompts/_maestro-cue.md`, `src/renderer/hooks/cue/useCueAiChat.ts` — teach the YAML-editor chat agent about the new field
- `docs/maestro-cue-configuration.md` — new "Sharing a workspace" section

## Test plan

- [x] `npm run lint` — clean
- [x] `npx vitest run src/__tests__/main/cue/` — 984 tests pass (10 new cases in `cue-session-state.test.ts` covering every row of the matrix above, including ancestor bypass and whitespace-only `owner_agent_id`)
- [x] Full `validate:push` suite via pre-push hook — 25,687 tests pass
- [ ] Manual: register the same vault as two agents; verify non-winner row in the Cue dashboard shows the red ⚠️ with the expected tooltip; verify setting `owner_agent_id` to the other agent flips which row fires

## Summary by CodeRabbit

* **New Features**
  * settings.owner_agent_id to pin which agent will run unowned subscriptions in shared project directories.

* **Documentation**
  * Added guidance on ownership resolution, accepted owner identifiers, behavior when owner is unset/ambiguous/missing, and recommended usage.

* **Improvements**
  * Dashboard shows warning icons/tooltips when a session will not run unowned subscriptions.
  * Non‑owners no longer run unowned subscriptions; explicit per‑subscription agent ids still run as before.
  * Completion/dispatch now targets the resolved owner for shared unowned subscriptions.
  * Validation rejects empty/whitespace owner_agent_id values.

* **Tests**
  * Added coverage for shared-workspace ownership scenarios and completion dispatch behavior.


pedramamini:
Thanks for tackling this @scriptease — the matrix in the PR description is a great framing, and consolidating the ownership gate behind a single pure utility is the right call. A few concerns surfaced by the automated reviews (coderabbit / greptile) that I think are worth addressing before merge:

**1. Implicit ownership can be won by a non-Cue agent** (`src/main/cue/cue-session-state.ts:149`)
`allSessions` is sourced from `deps.getSessions()`, which includes every application session — not just those that have loaded a `cue.yaml`. If a non-Cue agent happens to sort first for the `projectRoot`, `firstForRoot` picks it and the actually-configured Cue session gets a warning and has its unowned subs suppressed. The candidate list for the implicit fallback should be scoped to Cue-configured sessions (or at minimum to sessions that could own the config).

**2. Explicit owner lookup can resolve to multiple sessions** (`src/main/cue/cue-session-state.ts:143`)
`ownerExists` matches on id **or** name, and `isOwner` is then computed the same way against the current session. That means:
- If two agents share a display name, both can satisfy `isOwner` and both fire unowned subs.
- If `owner_agent_id` matches another agent's id but this session's name happens to equal that id string, this session wrongly claims ownership.

Recommend resolving a single owner candidate first (prefer id match, fall back to unique name match; treat ambiguous name matches as \"unresolvable\" with a warning), then comparing the current session against that resolved candidate's id.

**3. Ownership warning not reachable by keyboard / touch / screen readers** (`src/renderer/components/CueModal/SessionsTable.tsx:78-87`)
The actionable text is only in `title`, and `aria-label` is the generic \"Cue ownership conflict\". Since this is a red error indicator, users who most need the detail can't get at it. Either fold the warning text into the accessible name (`aria-label={s.ownershipWarning}` with `role=\"img\"`) or render a real tooltip component.

**4. Minor: prompt wording** (`src/renderer/hooks/cue/useCueAiChat.ts:31`, `src/prompts/_maestro-cue.md:86`)
The chat prompt describes `owner_agent_id` as \"required\" when >1 agent shares a projectRoot, but the runtime has a deterministic first-wins fallback. \"Recommended\" / \"used to override the default winner\" is more accurate and keeps the AI editor from over-prescribing.

No merge conflicts — base is clean against `rc`. Once the two ownership-resolution bugs are sorted I'm happy to take another look.

scriptease:
@greptile plz review

pedramamini:
Merged — thank you for this @scriptease! 🙏

The matrix-driven framing made the review trivial to reason about, and consolidating the gate behind a single pure utility (`computeOwnershipWarning`) is exactly the kind of factoring that keeps this code reviewable as more shared-workspace edge cases inevitably surface. Solid work all around.

I tacked on one follow-up commit covering the two outside-diff items CodeRabbit flagged on later passes — extending the ownership gate to `hasTimeBasedSubscriptions` / `countActiveSubscriptions` / refresh activeCount, and to the `agent.completed` dispatch path in `cue-completion-service`. Without them a non-owner could still keep the machine awake or fire duplicate completion chains. Two regression tests added in `cue-completion-chains.test.ts`.

Looking forward to your next one.

PR #875, opened 2026-04-20
fix(terminal): unblock OpenCode TUI and fix zsh PATH for claude/codex

chr1syy:
## Summary

Two related fixes that together make the Maestro Terminal tab actually usable for running CLI agents (`opencode`, `claude`, `codex`):

1. **`fix(build): actually skip esbuild re-minification of xterm.js core`** — stops the production `vendor-xterm` chunk from double-minifying `@xterm/xterm@6.0.0`'s pre-minified core. The prior `skip-xterm-minify` plugin was a no-op because its `renderChunk` hook had no `enforce` order, so Vite's `enforce: 'post'` esbuild-transpile minifier ran afterward and mangled the `requestMode` DECRQM handler. At runtime this threw `ReferenceError: <letter> is not defined` the first time a TUI sent `CSI ? N $ p`, poisoning the parser and freezing the tab (dead output + dead keystrokes). Replaced with a capture/restore pair: `renderChunk` caches the pre-minify `vendor-xterm` code, `generateBundle` writes it back after every `renderChunk` hook.

2. **`fix(terminal): include user install locations in PTY PATH`** — PTY terminal tabs overrode `PATH` with a narrow `buildUnixBasePath()` (version-manager bins + `/opt/homebrew:/usr/local:/usr/bin`). `opencode` worked because it's reachable via nvm; `claude` (symlink → `~/.local/share/claude/versions/…`) and `codex` (ELF in `~/.local/bin`) were unreachable. Bash hid this because Debian's `bash -l -i` sources `.bashrc`; zsh only sources `.zprofile`/`.zshrc` when present, so zsh users saw `command not found`. Switched to `buildExpandedPath()` — same canonical dev PATH that child-process agents already use — which backfills `~/.local/bin`, `~/.claude/local`, `~/.opencode/bin`, npm-global, Homebrew, etc. Shell rc files still run and their customizations still win.

## Repro (before)

In a Maestro Terminal tab on Linux:
- `opencode` → TUI renders for an instant, cursor hides, then hard freeze — Ctrl+C ignored, no echo. DevTools shows `ReferenceError: i is not defined at xd.requestMode`.
- On zsh: `claude` / `codex` → `command not found` despite being installed.

`opencode run "hi"` (batch mode, skips xterm.js) worked; VS Code's integrated terminal (also xterm.js) also worked, confirming the break was in our Vite bundle rather than xterm.js upstream.

## Test plan

- [x] Run targeted unit tests in scope: `envBuilder.test.ts` (44/44 pass, includes new `should include common user install locations in PATH on Unix`), `process-manager.test.ts` (43/43 pass), `LocalCommandRunner.test.ts` (1/1 pass).
- [x] `prettier --check` + `eslint` clean on the three touched files.
- [x] Built the renderer (`npx vite build`) and diff'd the emitted `vendor-xterm-*.js` `requestMode` against upstream `node_modules/@xterm/xterm/lib/xterm.mjs` — byte-equivalent.
- [x] Manual reproduction fixed on Linux: `opencode`, `claude`, `codex` all launch and stay interactive in both bash and zsh terminal tabs.
- [ ] Reviewer: confirm on macOS — the same DECRQM crash path was suspected for the `vim` hang tracked in a sibling worktree; this change should resolve it.

## Risk notes

- The PATH change is additive; it only widens what's reachable from a minimal-config shell. Users' own rc-file `export PATH=...` still runs and still takes precedence.
- The Vite plugin replaces the chunk content in `generateBundle`; Rollup's content-hashed filename is computed from the pre-restore code, so the hash may not reflect the final bytes. Acceptable because a rebuild always reproduces the same restored content — cache-busting stays sound within a given build.

## Summary by CodeRabbit

* **Refactor**
  * Unix terminal environment now sources PATH from the shared expanded-path utility, improving consistency for spawned shells.

* **Chores**
  * Adjusted build tooling to preserve the terminal vendor chunk during packaging for more reliable runtime behavior.

* **Tests**
  * Added Unix-specific test coverage to verify terminal PATH includes user-local bin locations.

### DIFF
--- a/src/__tests__/main/process-manager/runners/LocalCommandRunner.test.ts
+++ b/src/__tests__/main/process-manager/runners/LocalCommandRunner.test.ts
@@ -1,12 +1,12 @@
 import { EventEmitter } from 'events';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
-const { mockPtySpawn, mockResolveShellPath, mockBuildInteractiveShellArgs, mockBuildUnixBasePath } =
+const { mockPtySpawn, mockResolveShellPath, mockBuildInteractiveShellArgs, mockBuildExpandedPath } =
 	vi.hoisted(() => ({
 		mockPtySpawn: vi.fn(),
 		mockResolveShellPath: vi.fn(),
 		mockBuildInteractiveShellArgs: vi.fn(),
-		mockBuildUnixBasePath: vi.fn(),
+		mockBuildExpandedPath: vi.fn(),
 	}));
 
 vi.mock('node-pty', () => ({
@@ -28,9 +28,13 @@ vi.mock('../../../../main/process-manager/utils/pathResolver', () => ({
 	buildWrappedCommand: vi.fn(),
 }));
 
-vi.mock('../../../../main/process-manager/utils/envBuilder', () => ({
-	buildUnixBasePath: mockBuildUnixBasePath,
-}));
+vi.mock('../../../../shared/pathUtils', async (importOriginal) => {
+	const actual = await importOriginal<typeof import('../../../../shared/pathUtils')>();
+	return {
+		...actual,
+		buildExpandedPath: mockBuildExpandedPath,
+	};
+});
 
 vi.mock('../../../../shared/platformDetection', () => ({
 	isWindows: vi.fn(() => false),
@@ -43,7 +47,7 @@ describe('LocalCommandRunner', () => {
 		vi.clearAllMocks();
 		mockResolveShellPath.mockReturnValue('/bin/zsh');
 		mockBuildInteractiveShellArgs.mockReturnValue(['-l', '-i', '-c', 'ls']);
-		mockBuildUnixBasePath.mockReturnValue('/usr/bin:/bin');
+		mockBuildExpandedPath.mockReturnValue('/usr/bin:/bin');
 	});
 
 	it('resolves and emits stderr when PTY spawn throws', async () => {
--- a/src/main/process-manager/runners/LocalCommandRunner.ts
+++ b/src/main/process-manager/runners/LocalCommandRunner.ts
@@ -7,7 +7,7 @@ import { EventEmitter } from 'events';
 import * as pty from 'node-pty';
 import { logger } from '../../utils/logger';
 import type { CommandResult } from '../types';
-import { buildUnixBasePath } from '../utils/envBuilder';
+import { buildExpandedPath } from '../../../shared/pathUtils';
 import {
 	resolveShellPath,
 	buildInteractiveShellArgs,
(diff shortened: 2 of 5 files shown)


chr1syy:
Addressed the P2 and investigated the failing CI check.

**P2 — `LocalCommandRunner` asymmetry** → Fixed in `264ade681`. `LocalCommandRunner` now uses `buildExpandedPath()` to match `buildPtyTerminalEnv()`, so `runCommand` invocations see the same reachable PATH as PTY tabs. The stripped env (HOME/USER/SHELL/TERM/LANG/PATH only) is preserved — only the PATH value widens. Mock updated accordingly in the LocalCommandRunner test (1/1 passing, and the full `envBuilder.test.ts` 44/44 still green).

**CI `test` job failure** → Unrelated to this PR. The failure is a `ReferenceError: window is not defined` unhandled rejection raised during teardown of `src/__tests__/renderer/components/UsageDashboard/AutoRunStats.test.tsx`:

```
❯ getCurrentEventPriority node_modules/react-dom/cjs/react-dom.development.js:10993:22
❯ requestUpdateLane node_modules/react-dom/cjs/react-dom.development.js:25495:19
❯ dispatchSetState node_modules/react-dom/cjs/react-dom.development.js:16648:14
❯ src/renderer/components/UsageDashboard/AutoRunStats.tsx:138:4  (setLoading(false) in finally)
```

Classic "async setState fires after happy-dom window is torn down" pattern. The touched files in this PR (`vite.config.mts`, `envBuilder.ts`, `envBuilder.test.ts`, now `LocalCommandRunner.ts` + its test) don't import React, don't touch the UsageDashboard, and don't alter test environment setup. The report itself shows `710 passed | 1 skipped (711)` — every test passed; the run only fails because of the single unhandled rejection that escaped into the process. Recent `rc` CI runs are mixed (one of the last ten failed too), consistent with the test being pre-existing flake rather than something introduced here. The push of the P2 fix will re-trigger CI; if it still fails in the same spot, we can open a separate issue to stabilize that test.

pedramamini:
Thanks for the thorough investigation and fix, @chr1syy — really appreciate the contribution.

The diagnosis on both fronts is great:
- The `renderChunk` vs `enforce: 'post'` ordering on the xterm.js minify-skip plugin is a subtle one, and the capture-in-`renderChunk` / restore-in-`generateBundle` pattern is a clean way around it. Good call-out on the hash-vs-content trade-off in risk notes; with `modulePreload: false` that's harmless and deterministic per build.
- The zsh `.zshrc`-not-sourced PATH issue is exactly the kind of cross-shell footgun that's easy to miss on bash. Switching to `buildExpandedPath()` keeps user rc-file customizations winning while widening the floor for minimal-config shells. The follow-up fix to `LocalCommandRunner` to remove the asymmetry is the right move.

CI is green, mergeable is clean, Greptile P2 addressed, CodeRabbit has no actionable comments. Shipping. The CoderRabbit nitpick about asserting prepend order in the PATH test is a reasonable hardening but not a blocker — feel free to tack it on if you like, otherwise we can revisit later.

parsakhaz:
fwiw i think there may be a simpler version of the xterm half of this now.

i just re-hit the same opencode `requestMode` crash in Pane and tested the Vite output across targets/minifiers. the capture/restore chunk approach makes sense, but the lower-touch fix that worked for us was just:

```ts
build: {
  target: 'es2021'
}
```

that keeps esbuild minification on and avoids the broken downlevel output where xtermjs 6's `r ||= {}` path turns into an undeclared `void 0 || (r = {})` assignment. my ranking after repro was:

1. `build.target: 'es2021'`, best for Electron/Vite apps
2. terser, works but changes minifier and adds a dep
3. preserve/restore xterm chunk, works but has more build-pipeline risk
4. disable minify
5. DECRQM parser interception

we added a post-build guard too so the emitted `TerminalPanel` chunk fails the build if the broken shape comes back: https://github.com/dcouple/Pane/pull/176

PR #876, opened 2026-04-21
feat(cue): security hardening + backend test coverage

reachrazamair:
## Summary

Lands two Cue refactoring phases in a single PR:

- **Security Hardening:** four defense-in-depth guards (glob/path traversal, env-var sanitization, prompt-file containment, DB file permissions).
- **Backend Test Coverage:** race-condition, fan-in edge-case, and end-to-end integration tests, plus reusable `InMemoryCueDb` infrastructure so engine-level integration tests run deterministically in CI without relying on the native `better-sqlite3` binary.

No production behavior changes outside the four security guards. Every change is covered by new tests; no existing tests were modified other than adding 3 new output-prompt scenarios to `cue-run-manager.test.ts`.

## Security Hardening

| # | Guard | Files changed |
|---|---|---|
| 11A | Reject `watch` patterns with `..`, leading `/`, `\`, or Windows drive letters. Runtime guard in file-watcher drops events whose resolved path escapes `projectRoot`. | `cue-config-validator.ts`, `cue-file-watcher.ts` |
| 11B | New single-purpose `cue-env-sanitizer.ts`. Spec-exact blocklist (`PATH`, `HOME`, `USER`, `SHELL`, `LD_PRELOAD`, `DYLD_INSERT_LIBRARIES`, `NODE_OPTIONS`) + POSIX identifier regex. Dropped vars surface as `warn` via the existing `onLog` hook. | `cue-env-sanitizer.ts` (new), `cue-spawn-builder.ts` |
| 11C | `readPromptFile()` now refuses prompt files whose resolved path is outside the project root (mirrors the write-side guard in `cue-config-repository.ts`). | `cue-config-normalizer.ts` |
| 11D | `fs.chmodSync(dbPath, 0o600)` after DB open, wrapped in try/catch (non-fatal on Windows NTFS / network mounts). | `cue-db.ts` |

### SRP / SOC

Each guard lands in the module that owns the concern: glob-shape validation stays in the validator, runtime path containment in the watcher, env sanitization in a dedicated module called once from spawn-builder, prompt-read containment in the normalizer, DB permissions in the DB lifecycle module. No cross-cutting abstractions introduced.

### Tests (42)

New `src/__tests__/main/cue/cue-security.test.ts`:

- Glob validator rejection (11 tests): `../*.ts`, bare `..`, mid-path `..`, Windows backslash traversal, `/etc/passwd`, `\Windows\...`, `C:\...`, `D:/...`, plus acceptance of normal globs and `task.pending` watch patterns.
- File-watcher runtime guard (3 tests): drops escapes with `warn`, accepts in-root events, survives missing `onLog`.
- Env sanitizer (23 tests): all 7 blocklisted names, 6 invalid-regex names, case-sensitive policy (lowercase `path` allowed), mixed batches, order preservation, `getBlockedEnvVarNames` canonical list.
- Prompt-file containment (4 tests): legitimate resolution, relative traversal refused, absolute-outside refused, absolute-inside allowed.
- DB chmod (2 tests): `0o600` applied on POSIX (`skipIf(!isPosix)`); ENOENT path logs warn and initialization still completes.

## Backend Edge-Case Tests

| File | Tests |
|---|---|
| `cue-race-conditions.test.ts` (new) | 9 — `reset()` and `stopRun()` mid-flight, late `onCueRun` resolution discarded, queue drop-oldest under saturation, FIFO drain ordering, new `execute` during drain joins tail, documented `stopAll`-drain race |
| `cue-fan-in-edge-cases.test.ts` (new) | 10 — never-completes timeout in `continue` and `break` modes, per-sub overrides, mid-wait cleanup, duplicate completion idempotency, timer-not-extended invariant, post-timeout / `expireTracker` safety |
| `cue-run-manager.test.ts` (+3) | Output-prompt failure preserves main-task stdout, output-prompt rejection exits cleanly, `stopRun` during output-prompt phase kills both parent and child processRunId |

## Backend Integration Tests

New shared helper: `src/__tests__/main/cue/cue-integration-test-helpers.ts` (~390 lines)

- `createInMemoryCueDb()` — high-fidelity mirror of the cue-db module contract. `better-sqlite3` is a native module built against Electron's ABI and fails to load under vitest's Node runtime; the mirror preserves the SQL semantics that matter for the rest of the engine (insert ordering, `INSERT OR IGNORE` uniqueness, prune-by-age cutoff, heartbeat single-row upsert, safe-wrapper no-throw).
- `buildCueDbModuleMock(getDb)` — module-shape delegate for `vi.mock('.../cue-db', ...)`.
- `canLoadBetterSqlite3()` — probe-instantiates a `:memory:` DB so `describe.skipIf` reflects the binary's real availability (mere `require()` gives false positives on ABI mismatch).

| File | Tests |
|---|---|
| `cue-db-integration.test.ts` (new) | 25 + 1 skipped real-SQLite smoke — lifecycle, event journal ordering/LIMIT/filter/INSERT OR REPLACE, safe wrappers, 100-event concurrent writes, heartbeat upsert, prune math, GitHub seen-set uniqueness + scoped cleanup + pruning, restart simulation |
| `cue-engine-integration.test.ts` (new) | 9 end-to-end — real `CueEngine` wired with real session registry / fan-in tracker / run manager / heartbeat / dispatch / completion / cleanup services; only `onCueRun` + `loadCueConfig` + trigger-source constructors mocked. Heartbeat → DB round-trip, agent-completion chain propagation, hot-reload subscription replacement, `removeSession` teardown, restart preserves completed rows, multi-session isolation |

## Test plan

- `npx tsc --noEmit --pretty` — 0 errors
- `npm run lint` (three tsconfig projects) — 0 errors
- `npm run lint:eslint` — 0 errors, 0 warnings
- Cue backend suite: 56 files / 1,124 passing / 1 skipped / 0 failing
- Full suite: 722 files / **25,884 passing / 108 skipped / 0 failing** (~116s)

## Files

**New (7)**
- `src/main/cue/cue-env-sanitizer.ts`
- `src/__tests__/main/cue/cue-security.test.ts`
- `src/__tests__/main/cue/cue-race-conditions.test.ts`
- `src/__tests__/main/cue/cue-fan-in-edge-cases.test.ts`
- `src/__tests__/main/cue/cue-integration-test-helpers.ts`
- `src/__tests__/main/cue/cue-db-integration.test.ts`
- `src/__tests__/main/cue/cue-engine-integration.test.ts`

**Modified (7)**
- `src/main/cue/config/cue-config-validator.ts`
- `src/main/cue/cue-file-watcher.ts`
- `src/main/cue/cue-spawn-builder.ts` (+ latent type narrowing fix surfaced by stricter `tsconfig.lint.json`)
- `src/main/cue/config/cue-config-normalizer.ts`
- `src/main/cue/cue-db.ts`
- `src/__tests__/main/cue/cue-run-manager.test.ts` (+3 output-prompt scenarios)
- `src/renderer/components/CuePipelineEditor/utils/pipelineToYaml.ts` (removed unused `eslint-disable no-console` directive that ESLint flagged as dead)

## Accepted limitations (documented in-code)

- **Symlink traversal:** `path.resolve` does not follow symlinks; a link inside `projectRoot` pointing outside slips past the 11A / 11C runtime guards. Project-trust boundary, not a Cue concern.
- **Windows NTFS chmod:** 0o600 is partially honored on NTFS; `chmodSync` is wrapped in try/catch to avoid failing init on mounts without POSIX permissions.
- **better-sqlite3 ABI:** the real-SQLite smoke test in `cue-db-integration.test.ts` uses `describe.skipIf(!canLoadBetterSqlite3())`. CI with mismatched ABI runs only the semantic in-memory tests; local runs with a matching binary catch mirror-vs-native drift.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevent path traversal for globs and prompt-file reads; file watcher now drops events outside project root and logs warnings.
  * DB file permission tightening is best-effort and logs a warning on failure.
  * Stopping all runs now clears the run queue early to avoid queuing during shutdown.

* **New Features**
  * Sanitizes custom environment variables for spawned processes and reports dropped names.

* **Tests**
  * Added extensive unit and integration tests covering DB, engine, fan-in, race, security, lifecycle, and helpers.

* **Style**
  * Removed an unnecessary ESLint suppression before a console warning.

### DIFF
--- a/src/__tests__/main/cue/cue-db-integration.test.ts
+++ b/src/__tests__/main/cue/cue-db-integration.test.ts
@@ -478,8 +478,13 @@ describe.skipIf(!canLoadBetterSqlite3())('Phase 15B — real SQLite smoke test',
 			os.tmpdir(),
 			`maestro-cue-smoke-${Date.now()}-${Math.random().toString(36).slice(2)}.db`
 		);
+		// Capture the cue-db module lazily so the finally block can close the
+		// SQLite handle even if an assertion above throws. Leaving the handle
+		// open before `fs.unlinkSync` would fail on Windows (file locked) and
+		// leak the connection on POSIX.
+		let cueDb: typeof import('../../../main/cue/cue-db') | null = null;
 		try {
-			const cueDb = await import('../../../main/cue/cue-db');
+			cueDb = await import('../../../main/cue/cue-db');
 			cueDb.initCueDb(undefined, dbPath);
 			cueDb.recordCueEvent({
 				id: 'smoke-1',
@@ -492,8 +497,14 @@ describe.skipIf(!canLoadBetterSqlite3())('Phase 15B — real SQLite smoke test',
 			const events = cueDb.getRecentCueEvents(0);
 			expect(events).toHaveLength(1);
 			expect(events[0].id).toBe('smoke-1');
-			cueDb.closeCueDb();
 		} finally {
+			if (cueDb) {
+				try {
+					cueDb.closeCueDb();
+				} catch {
+					/* best effort — double-close is safe, other errors are non-fatal here */
+				}
+			}
 			try {
 				fs.unlinkSync(dbPath);
 			} catch {
--- a/src/__tests__/main/cue/cue-db-integration.test.ts
+++ b/src/__tests__/main/cue/cue-db-integration.test.ts
@@ -1,0 +1,504 @@
+/**
+ * Phase 15B — Cue database contract / integration tests.
+ *
+ * Exercises the contract cue-engine depends on (ordering, UNIQUE, prune-by-age,
+ * heartbeat upsert, safe-wrapper no-throw) through the in-memory mirror
+ * defined in `cue-integration-test-helpers.ts`. We cannot use real
+ * `better-sqlite3` under vitest because the native binary is built for
+ * Electron's ABI and fails to load in plain Node; the in-memory mirror
+ * preserves the SQL semantics that actually matter for the rest of the
+ * engine (ordering, UNIQUE constraints, prune cutoff).
+ *
+ * A `describe.skipIf(!canLoadBetterSqlite3())` block at the bottom runs one
+ * real-SQLite smoke round-trip when the binary is available locally — this
+ * catches drift between the mirror and the native module without breaking CI
+ * on hosts that can't load it.
+ */
+
+import { describe, it, expect, beforeEach } from 'vitest';
+import * as path from 'path';
+import * as os from 'os';
+import * as fs from 'fs';
+import {
+	createInMemoryCueDb,
+	canLoadBetterSqlite3,
+	type InMemoryCueDb,
+} from './cue-integration-test-helpers';
+
+describe('Phase 15B — cue-db in-memory contract', () => {
+	let db: InMemoryCueDb;
+
+	beforeEach(() => {
+		db = createInMemoryCueDb();
+		db.initCueDb();
+	});
+
+	// ─── Lifecycle ────────────────────────────────────────────────────────
+
+	describe('lifecycle', () => {
+		it('is ready after init and not ready after close', () => {
+			expect(db.isCueDbReady()).toBe(true);
+			db.closeCueDb();
+			expect(db.isCueDbReady()).toBe(false);
+		});
+
+		it('initCueDb is idempotent — second call is a no-op', () => {
+			// Calling init twice must not throw nor reset the current state.
+			db.recordCueEvent({
+				id: 'e1',
+				type: 'time.heartbeat',
+				triggerName: 't',
+				sessionId: 'session-1',
+				subscriptionName: 'sub',
+				status: 'running',
+			});
+			db.initCueDb(); // second call
+			expect(db.getRecentCueEvents(0).length).toBe(1);
+		});
+
+		it('getRecentCueEvents throws when DB is not initialized', () => {
+			db.closeCueDb();
+			expect(() => db.getRecentCueEvents(0)).toThrow(/not initialized/);
+		});
+	});
+
+	// ─── Event journal ────────────────────────────────────────────────────
+
+	describe('event journal', () => {
+		it('records and retrieves a single event', () => {
+			db.recordCueEvent({
+				id: 'evt-1',
+				type: 'time.heartbeat',
+				triggerName: 'hb',
+				sessionId: 'session-1',
+				subscriptionName: 'sub-1',
+				status: 'running',
+				payload: '{"x":1}',
+			});
+			const events = db.getRecentCueEvents(0);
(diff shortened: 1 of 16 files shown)


pedramamini:
Thanks for the thorough work here @reachrazamair — the dedicated `cue-env-sanitizer`, the in-memory `better-sqlite3` mirror, and the race/fan-in edge-case coverage are all excellent additions. A few items flagged by the automated reviewers look worth addressing before this comes out of draft, especially since the headline of the PR is security hardening:

**Security gaps in the env-var blocklist** (`src/main/cue/cue-env-sanitizer.ts`)
The current set blocks `LD_PRELOAD` / `DYLD_INSERT_LIBRARIES` / `NODE_OPTIONS` but omits the direct analogues for other runtimes the user's agents can spawn:
- `LD_LIBRARY_PATH` (Linux dynamic-library resolution — sibling of the `LD_PRELOAD` entry already present)
- `PYTHONPATH`, `RUBYLIB`, `PERL5LIB`, `JAVA_TOOL_OPTIONS` (per-language code-injection vectors through `customEnvVars` in YAML)

Given the threat model that motivated 11B, these feel in-scope. Worth adding to both `BLOCKED_ENV_VARS` and the blocklist test table.

**Case-sensitive `startsWith` on case-insensitive filesystems** (`src/main/cue/config/cue-config-normalizer.ts:43`)
`path.resolve` preserves input casing. On default APFS (macOS) and NTFS (Windows) a legitimate prompt path whose casing differs from `projectRoot` will be silently rejected. Since macOS is the primary target, this can bite real users. Consider `path.relative(projectRoot, resolved)` + checking the result doesn't start with `..`, or normalizing both sides with `fs.realpathSync` / case-folded compare.

**Windows drive-relative path bypass** (`src/main/cue/config/cue-config-validator.ts:29`)
The current regex `/^[A-Za-z]:[\\/]/` blocks `C:\\...` but lets `C:secret\\*.txt` through (drive-relative, not drive-absolute). Loosening to `/^[A-Za-z]:/` closes the gap; worth a regression test in `cue-security.test.ts`.

**`stopAll()` race test codifies the bug** (`src/__tests__/main/cue/cue-race-conditions.test.ts:406`)
The test currently locks in behavior where `stopAll()` returns with a fresh active run after draining `sub-2`. For an engine-shutdown API that's surprising; consider fixing the production path so the queue is cleared before stopping, then asserting no active runs remain, rather than freezing the current semantics in a test.

**Minor test hygiene**
- `cue-db-integration.test.ts:502` — move `closeCueDb()` into a `finally` so a failing assertion doesn't leak a handle and break `unlinkSync`.
- `cue-integration-test-helpers.ts:26` — header still references `buildWiredCueEngine(overrides)`, which doesn't exist in this helper anymore.

Happy to re-review once the security-blocklist additions and the case-folding fix land. The rest is polish.

PR #880, opened 2026-04-22
feat(cue): reliability, observability, performance

reachrazamair:
## Summary

Lands all three remaining Cue refactoring phases in a single cohesive PR:

- **Reliability:** queue persistence across restarts, user-visible queue overflow toasts, GitHub rate-limit exponential backoff, and a fan-in health API.
- **Observability:** Sentry expansion into heartbeat + GitHub paths, a new in-process metrics module, and a typed `CueLogPayload` discriminated union replacing ad-hoc log objects.
- **Performance:** visibility-aware polling in `useCue`, extracted/memoized PipelineCanvas sub-regions, and a ref-stabilized `isValidConnection` callback.

All existing public APIs (`CueEngine`, `window.maestro.cue.*`, `cueService`) are preserved. Zero behavioral regressions.

### Reliability

| Change | Files |
|---|---|
| **12A Queue persistence** — new `cue_event_queue` table + `cue-queue-persistence.ts` façade. Queued events survive app crash / quit. Restore drops stale, malformed, and session-missing rows (each logged as a typed `queueDropped` payload + recorded as `timeout` in `cue_events` for staleness). | `cue-db.ts`, `cue-queue-persistence.ts` (new), `cue-run-manager.ts`, `cue-engine.ts` |
| **12B Queue overflow warning** — `onQueueOverflow` dep on run manager → engine emits `queueOverflow` payload via `cue:activityUpdate` → `useCue` toasts the user per drop. Previously silent except in logs. | `cue-run-manager.ts`, `cue-engine.ts`, `useCue.ts` |
| **12C GitHub rate-limit backoff** — new `isGitHubRateLimitError` helper matches primary/secondary/HTTP 403/429. Replaces `setInterval` with self-rescheduling `setTimeout` chain so exponential backoff (cap 1h) actually takes effect. Rate-limit errors skip Sentry; non-rate-limit errors are now reported. | `cue-github-poller.ts` |
| **12D Fan-in health** — `checkHealth()` returns trackers stalled >50% of timeout (clamps clock-backward to 0). Exposed via `CueEngine.getFanInHealth()` + `cue:getFanInHealth` IPC. | `cue-fan-in-tracker.ts`, `cue-engine.ts`, IPC + preload + service |

### Sentry + observability

| Change | Files |
|---|---|
| **13A Sentry expansion** — heartbeat gains a consecutive-failure counter (emits once at threshold=3, resets on success/stop). GitHub poller reports `resolveGh`/`resolveRepo`/`doPoll`/`seedMarker` failures to Sentry with distinct operation tags. Rate-limit errors intentionally skip Sentry. | `cue-heartbeat.ts`, `cue-github-poller.ts` |
| **13B Metrics module** — `cue-metrics.ts` with `CueMetricsCollector` (snapshot / increment / reset). Engine wraps all subsystem `onLog` calls via `meteredOnLog` and translates typed payloads into counter bumps — subsystems stay decoupled from the metrics module. Exposed via `CueEngine.getMetrics()` + `cue:getMetrics` IPC. | `cue-metrics.ts` (new), `cue-engine.ts`, IPC + preload + service |
| **13C Structured log types** — `CueLogPayload` discriminated union in `src/shared/cue-log-types.ts` covering 15 event types. Every emit site uses `satisfies CueLogPayload` so schema drift fails at compile time. Renderer `useCue` narrows on `payload.type` — 12B's overflow toast is the first consumer. | `cue-log-types.ts` (new), all emit sites |

### Performance

| Change | Files |
|---|---|
| **14A Visibility-aware polling** — `useCue` accepts optional `{ pollIntervalMs }`; default 10s. Interval skips when `document.visibilityState !== 'visible'`. Activity listener unchanged (push-based, always warm). | `useCue.ts` |
| **14B Memoized sub-regions** — extracted `PipelineLegend` + `PipelineEmptyState` as `React.memo` components. Wrapped `NodeConfigPanel`, `EdgeConfigPanel`, `CueSettingsPanel` in `React.memo`. Eliminates re-renders on every node-drag tick. | `PipelineLegend.tsx` (new), `PipelineEmptyState.tsx` (new), `PipelineCanvas.tsx`, 3 panels |
| **14C Stable `isValidConnection`** — ref pattern (`nodesRef`/`edgesRef`) replaces volatile deps. Callback identity now stable across drag ticks; ReactFlow no longer re-registers validation bookkeeping per frame. | `usePipelineCanvasCallbacks.ts` |

## Issues fixed along the way

1. Heartbeat DB failures previously swallowed forever — now surfaced to Sentry at threshold.
2. GitHub poller hammered the API through rate limits with no retry backoff — now exponential.
3. Queue overflows were invisible to users — now a per-drop toast.
4. Queue rows were lost on crash / quit — now persist + restore with FIFO, staleness, and integrity guards.
5. GitHub seed-marker failures on first-poll error were silent — now Sentry-reported.
6. `isValidConnection` identity churned on every drag tick — ref pattern keeps it stable.
7. Pipeline legend + empty-state overlay re-rendered on every drag tick — extracted + memoized.
8. `NodeConfigPanel` / `EdgeConfigPanel` / `CueSettingsPanel` weren't memoized — now wrapped.
9. `useCue` polled every 10s even when the tab/window was hidden — now visibility-gated.

## Test stats

- **Before:** 722 files / 25,884 passing / 108 skipped / 0 failing.
- **After:** 737 files / **26,100 passing** / 108 skipped / 0 failing.
- **Net: +216 tests / +15 files / 0 regressions.**

New test files:
- `cue-metrics.test.ts` (11)
- `cue-heartbeat.test.ts` (8)
- `cue-queue-persistence.test.ts` (11)
- `PipelineLegend.test.tsx` (5)
- `PipelineEmptyState.test.tsx` (4)

Extended test files:
- `cue-github-poller.test.ts` (+11)
- `cue-fan-in-tracker.test.ts` (+7)
- `cue-run-manager.test.ts` (+9: 6 persistence + 3 overflow)
- `cue-engine-integration.test.ts` (+3 metrics)
- `useCue.test.ts` (+5: 3 visibility + 2 overflow toast)
- `usePipelineCanvasCallbacks.test.ts` (+3 callback stability)
- `services/cue.test.ts` (+5: metrics + fan-in health)

Plus InMemoryCueDb mirror and 11 backend mock factories updated to include the new `cue_event_queue` methods.

## Verification

- `npx tsc --noEmit --pretty` — 0 errors.
- `npm run lint:eslint` — 0 errors.
- `npm run test` — 737/26,100 green.

## Test plan

- **Baseline smoke:** toggle Cue OFF → ON; create `time.heartbeat`, `file.changed`, `task.pending` subscriptions; verify each fires.
- **Fan-out:** 1 trigger → 2 agents dispatches both.
- **Fan-in:** 2 agents → 1 aggregator; fires when both complete.
- **Manual trigger:** "Run Now" fires and appears in activity log.
- **Stop run / stop all:** individual + bulk stops halt cleanly.
- **YAML hot-reload:** external edits apply without duplicate triggers.
- **Command node:** both `shell` and `cli send` modes work.
- **queue persistence:** `max_concurrent: 1`, `queue_size: 5`. Trigger 1 active + 2 queued, force-quit app, relaunch — verify queued runs dispatch automatically.
- **stale drop:** `timeout_minutes: 1`, queue an event, quit, wait 2+ min, relaunch — dropped event appears as `timeout` in activity log.
- **overflow toast:** `max_concurrent: 1`, `queue_size: 2`. Fire 4 events rapidly — 2 distinct warning toasts appear.
- **rate limit backoff:** grep logs for `rate-limited by GitHub — backing off to Nm` during extended poller run.
- **fan-in health:** stall a fan-in past 50% of `fan_in_timeout_minutes: 10` — `window.maestro.cue.getFanInHealth()` returns the stalled entry.
- **Sentry:** make `cue.db` read-only, force a write failure — Sentry captures the event with the correct operation tag.
- **metrics API:** DevTools: `await window.maestro.cue.getMetrics()` returns a snapshot with all counters; `runsStarted`/`runsCompleted` increment as subscriptions fire.
- **log types:** DevTools: `window.maestro.cue.onActivityUpdate(p => console.log(p))` — every payload carries a `type` discriminator.
- **visibility polling:** DevTools Network filter `cue`; minimize window; verify no polls while hidden; refocus resumes polling.
- **memoization:** React DevTools Profiler during a node drag — legend / empty state / config panels no longer re-render.
- **connection drag:** drag a valid connection (trigger→agent) highlights; trigger→trigger rejects; duplicate target rejects.
- **Full-cycle smoke:** create pipeline with 3 nodes, save, restart app — pipeline preserved. Toggle Cue off/on 3× — no duplicate fires.

## Summary by CodeRabbit

* **New Features**
  * Persistent queue storage + restore, engine metrics collector, public metrics & fan‑in health APIs, IPC/read APIs for metrics and health, and GitHub rate‑limit backoff with logging.
* **Bug Fixes**
  * Deterministic heartbeat failure handling with controlled reporting; safer DB/queue error handling during persistence/restore; queue overflow/drop semantics stabilized.
* **Refactor**
  * Extracted/memoized pipeline editor UI components; hook optimizations to reduce callback churn.
* **Tests**
  * Extensive new and updated tests covering queue persistence, metrics, heartbeat, polling, run manager, and UI.


pedramamini:
@reachrazamair thanks for the massive, well-structured PR — the phase-by-phase breakdown and the typed `CueLogPayload` discriminated union in particular are really nice. Before we land this, a few reviewer-raised concerns look worth addressing:

**P1 — Unreachable metrics counters** (Greptile, `src/main/cue/cue-metrics.ts`)
`eventsQueued`, `queuePersisted`, `githubPollErrors`, and `heartbeatFailures` are declared on `CueMetrics` but have no corresponding `CueLogPayload` variant / `recordMetricFromPayload` case, so they will always snapshot as `0`. Either wire the increment paths (for heartbeat failures this means emitting a typed payload instead of only calling `captureException` directly) or drop the keys — silent zeros will be worse than absence.

**Major — GitHub repo auto-detect bypasses backoff** (CodeRabbit, `cue-github-poller.ts:147`)
The catch around `gh repo view` returns `null`, so `doPoll()` exits before the rate-limit branch ever runs. Subscriptions without an explicit `repo` will keep hammering at the baseline interval when auto-detection is rate-limited. Re-throw `isGitHubRateLimitError(err)` from that catch.

**Major — Fail-open DB failures silently swallowed** (CodeRabbit, `cue-queue-persistence.ts`)
`clearSession`, `clearAll`, and the `getQueuedEvents()` failure path continue on error but only log locally. Per CLAUDE.md's Sentry guidance, please `captureException` with context before the fail-open (keep the fail-open behavior, just make it visible).

**P2 / minor, please also address:**
- **Unsafe cast + duplicate toasts** in `useCue.ts:155-156`: `payload as unknown as {...}` skips runtime validation, and the "intentionally unique" toast title is actually identical across drops from the same session. Narrow via `payload.type === 'queueOverflow'` and add a `queuedAt`/counter suffix so per-drop toasts don't collapse.
- **`queueDropped` with empty `sessionId`** (`cue-queue-persistence.ts:223`): aggregated stale/malformed/session-missing drops all set `sessionId: ''` but the union types it as required `string`. Make it optional or split the variant so consumers don't treat `''` as a real id.
- **Dead `clearSession()`** on `CueQueuePersistence` — never invoked; `clearQueue()` loops `remove()` instead. Either wire it up or drop it.
- **Heartbeat catch is too broad** (`cue-heartbeat.ts:49`): all `updateHeartbeat()` errors are treated as transient and hidden until threshold. Gate only known recoverable DB-lock style errors and re-throw / capture unexpected ones immediately.
- **Restore-time drops aren't recorded for missing-session / malformed rows** (only stale calls `safeRecordCueEvent`). Worth normalizing across all three discard branches so the event history matches the `queueDropped` counter.
- **Restored `queuedAt` is reset to `Date.now()`** in `cue-engine.ts:411` — across repeated crash/restart cycles an event could evade the staleness check indefinitely. Either preserve the original `queuedAt` through restore or document the intent.
- **Fake timers never restored** in `cue-fan-in-tracker.test.ts` and `cue-metrics.test.ts` — add `afterEach(() => vi.useRealTimers())` so the last fake clock doesn't leak into later suites in the same worker.
- **Typo**: `entrie(s)` in the restored-queue log message (`cue-queue-persistence.ts:234`) — user-facing.

No merge conflicts, and lint/test/Greptile/CodeRabbit checks are green — once the above are addressed I think this is ready. Let me know if you'd like to punt any of these to follow-ups.

PR #883, opened 2026-04-22
fix(ssh): broaden remote agent detection PATH + use POSIX command -v

### LINKED ISSUE #878: SSH remote agent detection fails when remote PATH is restricted
## Description

When configuring an SSH remote for agent execution, Maestro fails to detect installed agents on the remote host if the agent binary is not in the default `PATH` of the SSH non-login shell. This is common because:

1. SSH command execution (`ssh host 'which claude'`) uses a **non-login, non-interactive shell** which loads a minimal PATH (typically `/usr/bin:/bin`)
2. Agent binaries are typically installed in user-local directories like `~/.local/bin`, `~/.claude/local/`, or `~/go/bin/` which are only added to PATH by login shell profiles (`.bashrc`, `.profile`, `.zshrc`)
3. The current `detectAgentsRemote()` implementation relies solely on `which <binary>` — when this fails, the agent is marked as "Not Found" even though it's installed and usable

## Steps to Reproduce

1. Configure an SSH remote in Maestro (Settings → SSH Remotes)
2. Install an agent (e.g., Claude Code) on the remote host at `~/.local/bin/claude`
3. Ensure `~/.local/bin` is only added to PATH in `.bashrc` or `.profile` (standard installation)
4. Open "Create New Agent" modal and select the SSH remote
5. Observe that the agent shows **"Not Found"** despite being installed

## Expected Behavior

Maestro should detect the agent as available on the remote host, regardless of whether the SSH non-login shell's PATH includes the installation directory.

## Actual Behavior

Agent shows "Not Found" in the agent picker. The logs show:
```
[INFO] [AgentDetector] Agent "Claude Code" not found on remote host
```

The `which claude` command returns exit code 1 because the non-login SSH shell doesn't have `~/.local/bin` in PATH.

## Root Cause

`detectAgentsRemote()` in `src/main/ipc/handlers/agents.ts` uses only `which <binaryName>` to locate agents. Non-login SSH shells have a restricted PATH that excludes common user-local installation directories.

chr1syy:
## Summary

Fixes #878. SSH remote agent detection failed for users whose agent binary lived in an install directory that wasn't in `BASE_SSH_PATH_DIRS` (e.g. Claude Code's local-install layout under `~/.claude/local`, or any Go-based CLI under `~/go/bin`).

Two changes:

1. **Extend `BASE_SSH_PATH_DIRS`** (`src/main/utils/ssh-command-builder.ts`) with the install locations the issue cites plus a few other common ones found in the wild:
   - `~/.claude/local` — Claude Code local-install / auto-update bundle
   - `~/go/bin` — Go default `GOBIN` (Factory Droid and other Go-based CLIs)
   - `~/.bun/bin` — Bun + bunx-installed CLIs
   - `~/.deno/bin` — Deno-installed CLIs
   - `~/.nix-profile/bin` — Nix user profile binaries
   - `/snap/bin` — Linux snap-installed binaries

   `~/.local/bin` (the path the reporter reproduced with) was already covered.

2. **Swap `which` → POSIX `command -v`** in both `detectAgentsRemote()` and the `agents:get` SSH path (`src/main/ipc/handlers/agents.ts`). `command -v` is a shell builtin (no PATH lookup needed, no dependency on `/usr/bin/which` being installed), and behaves consistently across bash/dash/zsh. The probe already runs inside `/bin/bash --norc --noprofile -c '...'` via `buildSshCommand()`, so the builtin resolves correctly.

The reporter's claim that `which` is invoked against a minimal SSH non-login PATH is no longer accurate as of the current `rc` — `buildSshCommand()` wraps every remote command in a bash invocation that prepends a curated PATH. But the curated list was missing the install locations they hit, so the symptom is real.

## Test plan

- [x] `vitest run src/__tests__/main/utils/ssh-command-builder.test.ts` — extended PATH-coverage test asserts every new entry appears in both the bash `-c` wrapper and the stdin script.
- [x] `vitest run src/__tests__/main/ipc/handlers/agents.test.ts` — new `SSH remote detection (issue #878)` block exercises `agents:detect` with an `sshRemoteId` and verifies (a) every probe uses `command -v` rather than `which`, (b) the resolved path is parsed and surfaced, (c) non-zero exit marks the agent unavailable.
- [x] `prettier --check` and `eslint` clean on all touched files.
- [ ] Manual SSH detection against a remote with `claude` installed under `~/.claude/local/claude` (recommended for the reporter to confirm).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed SSH remote agent detection to properly search additional common installation paths where agent binaries may be located (e.g., package manager directories).
  * Improved remote agent availability resolution over SSH connections with more robust binary location detection methods.

* **Tests**
  * Added comprehensive test suite validating SSH remote agent detection and installation path resolution.

### DIFF
--- a/src/__tests__/main/ipc/handlers/agents.test.ts
+++ b/src/__tests__/main/ipc/handlers/agents.test.ts
@@ -290,6 +290,88 @@ describe('agents IPC handlers', () => {
 			expect(result[0].configOptions[0].argBuilder).toBeUndefined();
 			expect(result[0].configOptions[0].key).toBe('test');
 		});
+
+		describe('SSH remote detection (issue #878)', () => {
+			let mockSettingsStore: {
+				get: ReturnType<typeof vi.fn>;
+				set: ReturnType<typeof vi.fn>;
+			};
+
+			beforeEach(() => {
+				mockSettingsStore = {
+					get: vi.fn().mockReturnValue([
+						{
+							id: 'remote-1',
+							host: 'dev.example.com',
+							username: 'dev',
+							port: 22,
+							enabled: true,
+						},
+					]),
+					set: vi.fn(),
+				};
+
+				handlers.clear();
+				registerAgentsHandlers({
+					...deps,
+					settingsStore: mockSettingsStore as any,
+				});
+
+				vi.mocked(buildSshCommand).mockResolvedValue({
+					command: 'ssh',
+					args: ['-o', 'BatchMode=yes', 'dev@dev.example.com', 'mock'],
+				});
+			});
+
+			it("uses POSIX 'command -v' (not 'which') to probe each agent binary", async () => {
+				vi.mocked(execFileNoThrow).mockResolvedValue({
+					exitCode: 0,
+					stdout: '/home/dev/.local/bin/claude\n',
+					stderr: '',
+				});
+
+				const handler = handlers.get('agents:detect');
+				await handler!({} as any, 'remote-1');
+
+				// Every probe should invoke 'command -v <binary>', never 'which'
+				const calls = vi.mocked(buildSshCommand).mock.calls;
+				expect(calls.length).toBeGreaterThan(0);
+				for (const [, options] of calls) {
+					expect(options.command).toBe('command');
+					expect(options.args[0]).toBe('-v');
+				}
+			});
+
+			it('marks the agent available and records the resolved remote path', async () => {
+				vi.mocked(execFileNoThrow).mockResolvedValue({
+					exitCode: 0,
+					stdout: '/home/dev/.claude/local/claude\n',
+					stderr: '',
+				});
+
+				const handler = handlers.get('agents:detect');
+				const result = await handler!({} as any, 'remote-1');
+
+				const claude = result.find((a: any) => a.id === 'claude-code');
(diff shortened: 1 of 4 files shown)


pedramamini:
Thanks for the clean, well-scoped fix @chr1syy! 🙏

Really appreciate the care here:
- Extending `BASE_SSH_PATH_DIRS` with the specific locations that triggered #878 (`~/.claude/local`, `~/go/bin`) plus the other conventional install dirs (`~/.bun/bin`, `~/.deno/bin`, `~/.nix-profile/bin`, `/snap/bin`) is exactly the right shape — no speculative entries, each one has a concrete ecosystem behind it.
- Swapping `which` → `command -v` is a nice correctness bonus. Making the probe independent of `/usr/bin/which` being installed removes a failure mode we'd have hit eventually on minimal remote hosts.
- The inline comments explaining *why* each PATH entry exists will save future-us a lot of archaeology.
- Test coverage is solid — asserting every `buildSshCommand` probe uses `command -v` (not `which`) is the right regression guard, and the PATH coverage test locks in the new entries in both the bash wrapper and the stdin-script paths.

Approving and labeling.

PR #885, opened 2026-04-23
fix(ssh): readDirRemote fails on dirs without symlinks (for-loop exit regression)

chr1syy:
## Summary

- **Regression introduced by #??? (`ea95a418c`, Apr 13)** — *fix: resolve symlinks in Files panel and Auto Run folder tree*. `readDirRemote` appends a `for` loop that probes each entry for symlink-to-dir status; when the target dir has no such entries the loop's last `[ -L "\$f" ]` test returns non-zero, becoming the overall SSH exit code. Because every stderr is silenced by `2>/dev/null`, the caller only saw Node's generic ``Command failed: <cmd>`` even though `ls` output was valid. Surfaced in production as `fs:readDir` errors opening the Files panel on remote containers that host only regular directories (e.g. `/home/` with a single user dir).
- **Fix**: append `|| true` to the symlink scan so a successful `ls` can't be masked by the for-loop tail; additionally accept a non-zero exit as long as `__SYMDIR__` is present in stdout (defense-in-depth — existing tests mock `execSsh` with `exitCode: 0`, so this regression was invisible to CI).
- **Tests**: regression case asserting `success: true` when stdout is valid but `exitCode === 1`, plus a shape assertion that the emitted remote command ends with `done 2>/dev/null || true`.

## Reproduction

```sh
bash -c 'ls -1AF --color=never /home/ 2>/dev/null || echo "__LS_ERROR__"; echo "__SYMDIR__"; for f in /home/* /home/.[!.]* /home/..?*; do [ -L "\$f" ] && [ -d "\$f" ] && basename "\$f"; done 2>/dev/null; echo "EXIT=\$?"'
# EXIT=1   (without the patch)
# EXIT=0   (with `|| true`)
```

## Test plan

- [x] `npx vitest run src/__tests__/main/utils/remote-fs.test.ts` (55 passed)
- [x] `npx vitest run src/__tests__/main/ipc/handlers/filesystem.test.ts` (43 passed)
- [x] `npx vitest run src/__tests__/main/ipc/handlers/autorun.test.ts` (84 passed)
- [x] `npx prettier --check` on touched files
- [x] `npm run lint:eslint` on touched files (no errors)
- [ ] Manual: open Files panel against an SSH remote whose target dir has no symlink-to-dir entries; confirm listing renders instead of surfacing ``Error invoking remote method 'fs:readDir': Error: Command failed: …``

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a regression in remote directory reads involving symlinked targets so valid listings no longer fail when the remote scan produces a tolerated non-zero exit status; operations still fail appropriately when real errors are detected.

* **Tests**
  * Added tests covering symlink-related remote directory reads, ensuring parsing succeeds for tolerated exit scenarios and failures are preserved for genuine errors.

### DIFF
--- a/src/__tests__/main/utils/remote-fs.test.ts
+++ b/src/__tests__/main/utils/remote-fs.test.ts
@@ -204,6 +204,34 @@ describe('remote-fs', () => {
 			const remoteCommand = call[call.length - 1];
 			expect(remoteCommand).toMatch(/done 2>\/dev\/null \|\| true$/);
 		});
+
+		// The marker-based tolerance is intentionally narrow — anything that
+		// doesn't match the known for-loop-tail shape must still surface as
+		// a failure so we don't silently swallow SSH/transport errors.
+		it('propagates failure when marker is present but stderr is non-empty', async () => {
+			const deps = createMockDeps({
+				stdout: 'src/\n__SYMDIR__\n',
+				stderr: 'ssh: connect to host X port 22: Connection refused',
+				exitCode: 1,
+			});
+
+			const result = await readDirRemote('/home/user/project', baseConfig, deps);
+
+			expect(result.success).toBe(false);
+			expect(result.error).toContain('Connection refused');
+		});
+
+		it('propagates failure when marker is present but exit code is not 1', async () => {
+			const deps = createMockDeps({
+				stdout: 'src/\n__SYMDIR__\n',
+				stderr: '',
+				exitCode: 255, // typical ssh transport failure
+			});
+
+			const result = await readDirRemote('/home/user/project', baseConfig, deps);
+
+			expect(result.success).toBe(false);
+		});
 	});
 
 	describe('readFileRemote', () => {
--- a/src/__tests__/main/utils/remote-fs.test.ts
+++ b/src/__tests__/main/utils/remote-fs.test.ts
@@ -168,6 +168,42 @@ describe('remote-fs', () => {
 			// Path should be properly escaped in the command
 			expect(remoteCommand).toContain("'/path/with spaces/and'\\''quotes'");
 		});
+
+		// Regression: when the target directory contains no symlink-to-dir
+		// entries, the trailing `for` loop's last `[ -L "$f" ]` test returns
+		// non-zero, which becomes the overall exit code of the remote command.
+		// Before the `|| true` fix, readDirRemote treated a successful `ls` as
+		// a failure (stderr is swallowed by `2>/dev/null`, so the only signal
+		// was Node's "Command failed" boilerplate). See PR description.
+		it('still succeeds when for-loop tail exits non-zero but ls output is valid', async () => {
+			const deps = createMockDeps({
+				stdout: 'src/\npackage.json\n__SYMDIR__\n',
+				stderr: '',
+				exitCode: 1,
+			});
+
+			const result = await readDirRemote('/home/user/project', baseConfig, deps);
+
+			expect(result.success).toBe(true);
+			expect(result.data).toEqual([
+				{ name: 'src', isDirectory: true, isSymlink: false },
+				{ name: 'package.json', isDirectory: false, isSymlink: false },
+			]);
+		});
+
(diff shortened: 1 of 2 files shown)


pedramamini:
@chr1syy Thanks for the great writeup and the targeted fix here — the regression analysis is super clear and the test additions are exactly what was missing from CI.

One thing worth considering before we merge, echoing CodeRabbit's review comment: the `reachedSymlinkMarker` defense-in-depth check at `src/main/utils/remote-fs.ts:253` treats *any* non-zero exit as success when `__SYMDIR__` is in stdout. With `|| true` already in place, the for-loop tail can no longer produce a non-zero exit on its own — so any non-zero exit reaching that branch likely indicates a real transport/SSH failure that occurred *after* the marker was emitted, which we'd then silently swallow.

Could you tighten that condition to only tolerate the specific known shape, e.g.:

```ts
const toleratedSymlinkScanExit =
    result.exitCode === 1 &&
    reachedSymlinkMarker &&
    !hasLsErrorMarker &&
    result.stderr.trim() === '';
```

This keeps the defense-in-depth intent while staying aligned with the project's "don't silently swallow errors" guideline. Once that's in (or you've decided against it with reasoning), happy to approve.

chr1syy:
@pedramamini good point, agreed — pushed `0c2e144e2` narrowing the tolerance to the specific known for-loop-tail shape:

```ts
const toleratedSymlinkScanExit =
    result.exitCode === 1 &&
    reachedSymlinkMarker &&
    !hasLsErrorMarker &&
    result.stderr.trim() === '';

if (result.exitCode !== 0 && !hasLsErrorMarker && !toleratedSymlinkScanExit) { … }
```

Anything that doesn't match that signature (different exit code, any stderr, transport reset mid-pipeline) still propagates. Added two regression tests asserting real transport failures keep surfacing:

- `propagates failure when marker is present but stderr is non-empty` (exit 1 + `Connection refused`)
- `propagates failure when marker is present but exit code is not 1` (exit 255)

Scoped tests still green (57 in remote-fs, 43 in filesystem handler). Also resolves CodeRabbit's inline comment at the same line.

pedramamini:
Thanks @chr1syy — nice diagnosis on the last-iteration `[ -L ]` exit bleed-through. That's a legitimate path to the failure, and your repro is clean.

We pushed a fix for this earlier today that I believe supersedes this PR. There are actually **two** independent routes to the "Command failed" symptom, and `|| true` only catches one of them:

1. **bash** (your case): unmatched globs are passed literally, but the loop's final `[ -L "$f" ]` test returns non-zero for any non-symlink entry, and `$?` leaks as the pipeline exit. `|| true` fixes this.
2. **zsh** (default shell on modern macOS remotes): unmatched globs trigger `NOMATCH` and abort the shell *before* the `||` chain can run. Verified locally:
   ```
   $ zsh -c 'for f in /tmp/nolinks/* /tmp/nolinks/.[!.]* /tmp/nolinks/..?*; do …; done 2>/dev/null || true; echo "EXIT=$?"'
   (no output — shell aborted before `||` or the trailing echo)
   ```
   So `|| true` alone leaves Mac-to-Mac SSH still broken, which is how this got reported to us.

The fix we landed replaces the glob-based `for` loop with `find <path> -mindepth 1 -maxdepth 1 -type l | while read -r f; do [ -d "$f" ] && basename "$f"; done`. No shell globbing, so neither NOMATCH nor last-iter-exit can occur. Portable across GNU and BSD `find`.

One small correction: the glob patterns you're patching were introduced in `33dd12ba2` (6 days ago, "extract shared symlink resolver"), not `ea95a418c` (Apr 13). The earlier `ea95a418c` used `<path>/.*`, which incidentally sidestepped both failure modes because `.` and `..` always matched.

Closing as duplicate with thanks — your write-up made it obvious that the bash path existed as a separate issue from the one I'd originally diagnosed, and your test-case style (valid stdout + non-zero exit) is a pattern worth keeping in mind.

Issue #317, opened 2026-02-06
Centralized Process Spawning


pedramamini:
## Problem

8 distinct code paths for spawning agent processes with inconsistent feature support. Only 3/8 use `wrapSpawnWithSsh()`. Only 6/8 use `buildAgentArgs()`. Only 5/8 use `applyAgentConfigOverrides()`. The main `process.ts` handler has **200 lines of inline SSH logic** duplicating the wrapper utility.

## Current State

| Code Path | File | buildAgentArgs | applyConfig | SSH Wrapper | SSH Support |
|-----------|------|:-:|:-:|:-:|:-:|
| Main IPC handler | `ipc/handlers/process.ts` | ✅ | ✅ | ❌ inline | ✅ |
| Group chat moderator | `group-chat-router.ts` | ✅ | ✅ | ✅ | ✅ |
| Group chat participants | `group-chat-router.ts` | ✅ | ✅ | ✅ | ✅ |
| Group chat addition | `group-chat-agent.ts` | ✅ | ✅ | ✅ | ✅ |
| Context groomer | `context-groomer.ts` | ✅ | ❌ **missing** | ❌ | delegated |
| Tab naming | `tabNaming.ts` | ✅ | partial | ❌ inline | ✅ |
| Wizard | `conversationManager.ts` | ❌ **hardcoded** | ❌ | ❌ | delegated |
| CLI batch | `agent-spawner.ts` | ❌ **separate** | ❌ | ❌ | ❌ |

## Critical Issues

1. **SSH logic duplicated in 3 places**: `process.ts` (~200 lines inline), `tabNaming.ts` (~90 lines inline), and `wrapSpawnWithSsh()` utility (correct abstraction, underused)
2. **Context groomer missing agent config**: Doesn't call `applyAgentConfigOverrides()`, so grooming ignores user's agent model/args/env configuration
3. **Wizard has hardcoded per-agent args**: Lines 632-689 of `conversationManager.ts` manually build args for each agent type instead of using `buildAgentArgs()`
4. **CLI batch is completely separate**: Uses `child_process.spawn()` directly, no ProcessManager, no SSH, no agent config
5. **Inconsistent SSH prompt handling**: process.ts always uses stdin, wrapper uses 4000-char threshold, tabNaming checks agent capabilities

## Proposed Solution

Create single entry point:

```typescript
// src/main/utils/agent-spawner.ts
export async function spawnAgentProcess(config: {
  agentId: string;
  cwd: string;
[...]

PR #887, opened 2026-04-23
fix(group-chat): pass global shell env vars to moderator/participant spawns

### LINKED ISSUE #886: Group chat moderator ignores global shell env vars (ANTHROPIC_API_KEY not merged)
### Summary

The Group Chat moderator (and synthesis) spawn paths in `group-chat-router.ts` call `processManager.spawn()` without passing `shellEnvVars`, so environment variables set in **Settings → General → Shell Configuration → Global Environment Variables** are not merged into the moderator's child process environment. Individual (1-on-1) chats route through `ipc/handlers/process.ts` and pass `shellEnvVars` correctly, so this only breaks Group Chats.

For users who rely on `ANTHROPIC_API_KEY` (or `ANTHROPIC_AUTH_TOKEN`, `ANTHROPIC_BASE_URL`) configured globally, the moderator launches the Claude CLI with no credentials and immediately returns a synthetic auth-failed stub:

```json
{
  "model": "<synthetic>",
  "stop_reason": "stop_sequence",
  "error": "authentication_failed"
}
```

### Reproduction

1. Do **not** run `claude login` (no `~/.claude/.credentials.json`).
2. In Maestro, set `ANTHROPIC_API_KEY=<valid key>` in **Settings → General → Shell Configuration → Global Environment Variables**.
3. Start a 1-on-1 Claude Code chat — works fine.
4. Start a Group Chat with Claude Code as the moderator — moderator fails immediately with the synthetic auth error above.

### Root cause

Two env-merging paths exist; one is missing a parameter.

**Correct path — individual chats** (`src/main/ipc/handlers/process.ts` ~line 514):
```ts
shellEnvVars: globalShellEnvVars,
```
Read from `settingsStore.get('shellEnvVars', {})` (~line 230).

**Broken path — group chat moderator** (`src/main/group-chat/group-chat-router.ts` ~lines 806–822 and 1716–1726):
```ts
processManager.spawn({
  // ... customEnvVars: spawnEnvVars, ...
  // shellEnvVars NEVER set
});
```

Downstream, `buildChildProcessEnv(customEnvVars, isResuming, globalShellEnvVars)` (`src/main/process-manager/utils/envBuilder.ts` ~lines 199–238) expects the third argument. When it's undefined, global shell vars are skipped. `envBuilder.ts` also strips `CLAUDECODE`/`CLAUDE_CODE_*`/`ELECTRON_*` (~lines 111–125), so the CLI launches with nothing to authenticate.

This is the same class of inconsistency catalogued in #317 ("Centralized Process Spawning"); the group-chat spawn sites appear to have been missed.

### Suggested fix

Read the global shell env vars in `group-chat-router.ts` and pass them to both `processManager.spawn()` call sites:

\`\`\`ts
const globalShellEnvVars = settingsStore.get('shellEnvVars', {});

processManager.spawn({
  // ...existing fields...
  customEnvVars: spawnEnvVars,
  shellEnvVars: globalShellEnvVars,
});
\`\`\`

### Workaround

Move the env var from the global Shell Configuration to the **Moderator's per-agent Custom Environment Variables** in the Group Chat's Edit dialog. `customEnvVars` does flow through (`router.ts:725,815`).

### Environment

- Maestro: latest as of filing
- macOS 15 (Darwin 25.4.0)
- Claude Code CLI with `ANTHROPIC_API_KEY` auth (no OAuth `claude login` session)

josecsotomorales:
Fixes #886.

## Summary

Group Chat spawns in `group-chat-router.ts` were calling `processManager.spawn()` without `shellEnvVars`, so environment variables set in **Settings → Shell Configuration → Global Environment Variables** were not merged into the child process environment. Individual (1-on-1) chats go through `ipc/handlers/process.ts` which already passes `shellEnvVars` correctly, so this bug only affected Group Chats.

In practice, users relying on a globally-configured `ANTHROPIC_API_KEY` (or `ANTHROPIC_AUTH_TOKEN` / `ANTHROPIC_BASE_URL`) saw the moderator immediately fail with a synthetic auth-failed stub from the Claude CLI:

```json
{ "model": "<synthetic>", "error": "authentication_failed" }
```

because `envBuilder.ts` strips `CLAUDECODE` / `CLAUDE_CODE_*` / `ELECTRON_*` and the merged `shellEnvVars` layer was missing, so the CLI launched with no credentials.

## Changes

- `src/main/group-chat/group-chat-router.ts`
  - Add `getGlobalShellEnvVars()` helper that reads from `getSettingsStore()` (mirrors `ipc/handlers/process.ts:230`).
  - Pass `shellEnvVars: getGlobalShellEnvVars()` at all **four** `processManager.spawn()` sites: moderator (initial routing), participant, moderator synthesis, and participant recovery.
- `src/main/group-chat/group-chat-moderator.ts`
  - Add optional `shellEnvVars?: Record<string, string>` to the `IProcessManager.spawn` config type so the router's calls typecheck. Matches the existing `ProcessConfig.shellEnvVars` field.
- `src/__tests__/main/group-chat/group-chat-router.test.ts`
  - Mock `stores/getters` so the router tests don't blow up on uninitialized stores (the 17 spawn-path tests previously passed because the router never called `getSettingsStore()`; now they need it stubbed).

## Test plan

- [x] `npm run lint` — clean
- [x] `npx vitest run src/__tests__/main/group-chat/group-chat-router.test.ts` — 61/61 pass
- [x] `npx vitest run` (full suite) — 22521/22522 pass; the single failure (`stats/integration.test.ts > electron-rebuild verification for better-sqlite3`) is a native-module env check unrelated to this change and fails on `main` in environments where `electron-rebuild`'s gyp build can't complete.
- [ ] Manual: set `ANTHROPIC_API_KEY` in Settings → Shell Configuration, start a Group Chat with Claude Code as moderator, confirm moderator responds without `authentication_failed`.

## Notes

- This is the class of bug catalogued in #317 ("Centralized Process Spawning") — the group-chat spawn sites were apparently missed in that pass. A follow-up centralizing all eight spawn sites behind a single helper would prevent recurrence, but I've kept this PR scoped to the minimal behavioural fix.
- Precedence preserved: `customEnvVars` (per-agent / per-moderator) still wins over `shellEnvVars` (global) per `envBuilder.ts:199-238`.

## Summary by CodeRabbit

* **Tests**
  * Added targeted tests to verify propagation and merging of global shell environment variables into group-chat spawn flows, with per-test isolation to prevent state leakage.

* **Refactor**
  * Consistently apply globally configured shell environment variables across moderator, synthesis, and participant subprocesses; per-session custom variables take precedence on conflict.

### DIFF
--- a/src/__tests__/main/group-chat/group-chat-router.test.ts
+++ b/src/__tests__/main/group-chat/group-chat-router.test.ts
@@ -47,10 +47,14 @@ vi.mock('../../../main/utils/ssh-spawn-wrapper', () => ({
 }));
 
 // Mock stores/getters: router reads global shellEnvVars via getSettingsStore().
-// Return an empty settings store so spawns receive an empty shellEnvVars map.
+// The mock is driven by `mockedShellEnvVars` so individual tests can assert the
+// value actually flows through to processManager.spawn() and (for SSH) into
+// wrapSpawnWithSsh's customEnvVars input.
+let mockedShellEnvVars: Record<string, string> = {};
 vi.mock('../../../main/stores/getters', () => ({
 	getSettingsStore: () => ({
-		get: (_key: string, defaultValue: unknown) => defaultValue,
+		get: (key: string, defaultValue: unknown) =>
+			key === 'shellEnvVars' ? mockedShellEnvVars : defaultValue,
 	}),
 }));
 
@@ -101,6 +105,9 @@ describe('group-chat-router', () => {
 
 		// Set the mock userData path to our test directory
 		mockUserDataPath = testDir;
+
+		// Reset the global-settings shell env mock between tests
+		mockedShellEnvVars = {};
 
 		// Create a fresh mock for each test
 		mockProcessManager = {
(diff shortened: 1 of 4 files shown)


pedramamini:
Thanks for the fix @josecsotomorales — nice diagnosis and the local-spawn path is clearly correct. Before we merge I'd like to get the SSH case sorted as well.

CodeRabbit flagged this and after walking through the code I think it's a real gap: in all four `processManager.spawn()` sites you're passing `shellEnvVars: getGlobalShellEnvVars()`, but for SSH sessions the spawn has already been transformed by `wrapSpawnWithSsh()` at that point. That helper embeds only `customEnvVars` into the remote SSH stdin script (`src/main/utils/ssh-spawn-wrapper.ts:171`) and then returns `customEnvVars: undefined` (line 189). So `shellEnvVars` only lands on the *local* `ssh` client process — `ANTHROPIC_API_KEY` never makes it to the remote agent, and users with SSH-configured group chats hit the same `authentication_failed` you're fixing.

The individual-chat path handles this at `src/main/ipc/handlers/process.ts:437`:
```ts
const mergedSshEnvVars = { ...globalShellEnvVars, ...(effectiveCustomEnvVars || {}) };
```
…and feeds that merged object into the SSH command builder. Mirroring that here — merging `getGlobalShellEnvVars()` into the `customEnvVars` you pass to `wrapSpawnWithSsh()` at all four call sites, with the per-agent `effectiveCustomEnvVars` taking precedence — would close the same gap for SSH group chats.

Also worth tightening the test: the current `getSettingsStore` mock always returns the default, so a regression that stops threading `shellEnvVars` through would still pass. Returning a sentinel map and asserting it reaches `mockProcessManager.spawn` for at least one moderator and one participant path would catch that.

Lint/tests all look clean and the scoping is good — just need the SSH path before this goes in. Thanks again!

josecsotomorales:
Thanks for the review @pedramamini — great catch on the SSH path. Pushed `daf43cf8` addressing all three review items:

**1. SSH gap closed.** Added `mergeGlobalShellWithCustomEnv()` helper and applied it at all four `wrapSpawnWithSsh()` call sites (moderator / participant / synthesis / recovery). Per-agent `effectiveCustomEnvVars` takes precedence over globals, mirroring `ipc/handlers/process.ts:437`.

**2. `GenericProcessManager` widened** (CodeRabbit's suggestion). The interface in `ipc/handlers/groupChat.ts` was missing `shellEnvVars` plus `promptArgs` / `shell` / `runInShell` / `sendPromptViaStdin` / `sendPromptViaStdinRaw` — already used by the router. Aligned with `IProcessManager` / `ProcessConfig`.

**3. Test sentinel** (Greptile's nit). The mock now reads from a mutable `mockedShellEnvVars` (reset in `beforeEach`) rather than always returning the default. Three new assertions:

- moderator local spawn receives the sentinel as `shellEnvVars`
- participant local spawn receives the sentinel as `shellEnvVars`
- SSH-participant path's `wrapSpawnWithSsh` input has the sentinel merged into `customEnvVars`, with per-session vars winning for the same key

Lint clean, 64/64 router tests pass (3 new), full suite 22524/22525 (same unrelated `better-sqlite3` electron-rebuild env check fails locally regardless of branch). Ready for another look when you have a moment.

josecsotomorales:
@greptileai

Issue #706, opened 2026-04-02
bug: custom Claude Code skill descriptions show generic fallback on Linux/WSL (case-sensitive filesystem)


jSydorowicz21:
## Bug Description

Custom Claude Code skills always display "A Claude Code Skill" or "Agent command" as their description in Maestro, instead of the actual description from the skill's frontmatter.

## Root Cause

In `src/main/ipc/handlers/claude.ts` line 1672, the skill scanner looks for `skill.md` (lowercase):

```typescript
const skillPath = path.join(dir, entry.name, 'skill.md');
```

But Claude Code skills use uppercase: `SKILL.md` (e.g., `~/.claude/skills/Research/SKILL.md`).

On case-sensitive filesystems (Linux, WSL), `skill.md` != `SKILL.md`, so the lookup silently fails. The skill name is discovered via the init message's `slash_commands` array, but without a matching description from the file, the UI falls back to generic text.

## Full Chain

1. Claude Code reports skill names via init message (`slash_commands: ["Research", ...]`)
2. `agents:discoverSlashCommands` in `agents.ts:1067` returns just `{ name }` without descriptions
3. Renderer's `getSlashCommandDescription()` doesn't find custom skills in its hardcoded builtin list
4. Falls back to generic "Agent command" description

## Files Affected

- `src/main/ipc/handlers/claude.ts:1672` - Case-sensitive `skill.md` lookup
- `src/main/ipc/handlers/agents.ts:1067` - Missing description in skill discovery
- `src/renderer/hooks/agent/useAgentListeners.ts:988` - Uses fallback instead of actual descriptions

## Proposed Fix

1. Try `SKILL.md` first, fall back to `skill.md` (or do a case-insensitive directory listing)
2. When reporting skills from `slash_commands`, fetch descriptions from disk and return `{ name, description }`
3. Use actual descriptions in the renderer instead of `getSlashCommandDescription()` fallback

## Environment

- Platform: Linux/WSL (case-sensitive filesystem)
- macOS may also be affected depending on filesystem configuration
- Windows NTFS is case-insensitive so this works there by accident

chr1syy:
Closing as fixed in #890 
If you encounter anything feel free to open a new issue.

Issue #298, opened 2026-02-04
Issues running python command using MCP Desktop Commander - Windows


cesalo:
### **Description**

When using an MCP server (Desktop Commander) through Maestro, the start_process tool consistently fails with MCP error -32000: Connection closed, while all other tools from the same MCP server work correctly.

### **Environment**

Maestro version: (current)
Desktop Commander version: 0.2.34
OS: Windows 11
Agent type: claude-code (v2.1.31)
Node.js: 22.14.0

### **Steps to Reproduce**

Configure Desktop Commander as an MCP server in Maestro
Use any file-based DC tool (e.g., read_file, list_sessions, get_config) — works fine
Use start_process (e.g., start_process("python --version", timeout_ms: 10000)) — fails

### **Expected Behavior**

start_process should spawn the process and return output, as it does when running Desktop Commander from Claude Code directly in PowerShell.

### **Actual Behavior**

MCP error -32000: Connection closed

### **Key Observations**

Other DC tools work: read_file, list_sessions, get_config, list_directory all succeed through Maestro in the same session
Works outside Maestro: The exact same start_process call works when running Claude Code directly in PowerShell — Desktop Commander spawns the process and returns output normally
DC config confirms PowerShell: defaultShell: "powershell.exe" is set, and Python is available in PowerShell
Pattern: The tools that work are quick request/response (file reads, config queries). The tool that fails (start_process) involves spawning a child process and streaming output over a longer duration

### **Hypothesis**

Maestro's MCP transport layer may be prematurely closing or timing out the connection for longer-running MCP tool calls like start_process, which involve process spawning and output streaming, versus instant file I/O operations that return immediately.

pedramamini:
I think the issue here a core architectural decision made with Maestro, which is to use the batch interface of the various supported providers. What this means is that we call Claude Code (for example) and it will run until the completion of that request and then return a response. So any MCP tool usage that is temporary it'll wait for to complete. But if there's going to be a start process that's going to live for a very long time or permanently then that will only work on if you're using the actual interactive mode where the process is long-living.

Because the way it works now is we send a request we spawn the process we get the response process is gone. But we do have a session ID that we use to resume that very same conversation. So I think this might just be a functional limitation and if you've got MCP servers or other commands you want to run they're going to be long-running they're going to have to be done through the terminal which I know the terminal kind of sucks today. It's better than what Claude Code provides but I'm going to be bringing in full terminal support via xterm.js (#166 

Not 100% sure on this.

[...]

Issue #187, opened 2026-01-13
windows cloudflared detection (bug)


maceip:
I'll submit a PR for this but in general there's a a bunch of path issues I think floating around 

Codex can't be found (it's in path and when I add the path c:\\ or c:\ or /mnt/c it still can't find it 

& The whole wsl/win thing is hard for everyone rn

pedramamini:
We've got a dedicated channel to #windows-support on the Discord, there are two user-developers who are working on some PRs to address this. I've been flying blind on Windows. It really is a pain between WSL/native, we will find, document, and implement into the UX the happy path.

dmaynor:
Closing as duplicate of #159 — report has been merged into that issue.

chr1syy:
Closing as a duplicate of #159, which tracks the broader Windows WSL/native path handling work. The discussion was already merged there on 2026-02-14.

Issue #170, opened 2026-01-10
"No module named 'distutils'" when running npm install


jlengrand:
Hey there, 

Was just trying to run Maestro on my MBP (Sequoia 15.6.1) and got the following error 

```
  File "/Users/julienlengrand-lambert/Developer/maestro/node_modules/node-gyp/gyp/gyp_main.py", line 42, in <module>
    import gyp  # noqa: E402
    ^^^^^^^^^^
  File "/Users/julienlengrand-lambert/Developer/maestro/node_modules/node-gyp/gyp/pylib/gyp/__init__.py", line 9, in <module>
    import gyp.input
  File "/Users/julienlengrand-lambert/Developer/maestro/node_modules/node-gyp/gyp/pylib/gyp/input.py", line 19, in <module>
    from distutils.version import StrictVersion
ModuleNotFoundError: No module named 'distutils'
✖ Rebuild Failed
```

The internet seemed to direct me towards reinstalling the xCode dev tools, but turns out the issue was something else. 

I run Python 3.14, node-gyp seems to depend on distutils which was removed between 3.12 and 3.14. 

Solution is to install setuptools (`pip3 install setuptools`) and run `npm install` again. 

Maybe this helps someone. And maybe the install script can propose the solution in case the error appears again?

jlengrand:
I'm letting this open in case a maintainer wants to have a look at it, but feel free to close it :)

pedramamini:
Interesting, thanks for sharing, will see what change can be made to improve robustness here.

charles-dyfis-net:
The flake.nix added in #149 installs setuptools for exactly this reason. Granted, that doesn't help folks who don't choose to use direnv+nix for dependency management.

dmaynor:
Closing as duplicate of #191 — your workaround has been merged into that issue.

dmaynor:
## Merged from #191 (DevX: npm install fails with latest python)

Reported by @ksylvan — same `ModuleNotFoundError: No module named 'distutils'` error on macOS with Python 3.14. node-gyp's `gyp` module depends on `distutils` which was removed in Python 3.12+. Fails during `electron-rebuild` of `better-sqlite3`.

**Additional workaround from #191:** Use a Python 3.11 virtual environment:
```
uv venv -p 3.11
source .venv/bin/activate
npm install
```

(This issue's workaround of `pip3 install setuptools` also works.)

Closing #191 as a duplicate of this issue.

pedramamini:
[...]

PR #890, opened 2026-04-23
fix: open-issue cleanup (readDirRemote tolerance, Claude skills on WSL, Python 3.12 guard)

chr1syy:
## Summary

Three small, independent fixes picked off the open-issue backlog.

### 1. Claude Code skill descriptions on case-sensitive filesystems — #706
Claude Code writes `SKILL.md` (canonical uppercase). The existing `scanSkillsDir` only looked for `skill.md`, so on Linux/WSL the lookup silently failed and the UI fell back to the generic "Agent command" label. Fixed by:

- `scanSkillsDir` (`src/main/ipc/handlers/claude.ts`) now tries `SKILL.md` first and falls back to `skill.md` for legacy layouts.
- `agents:discoverSlashCommands` (`src/main/ipc/handlers/agents.ts`) now enriches Claude's reported `slash_commands` list with the description read from each matching skill's frontmatter, via a new `readClaudeSkillDescriptions()` helper.
- The IPC return type gets an optional `description` field (`src/main/preload/agents.ts`, `src/renderer/global.d.ts`).
- `useWizardHandlers` (`src/renderer/hooks/wizard/useWizardHandlers.ts`) prefers the disk-derived description when present, falling back to the built-in table otherwise. The `/skills` empty-state message now documents the canonical `SKILL.md` name.

### 2. Python 3.12 install-time guard — #170
node-gyp (pulled in transitively by `better-sqlite3` and `node-pty`) still imports `distutils`, which was removed from Python's stdlib in 3.12 — so `npm install` dies during the `electron-rebuild` postinstall step with `ModuleNotFoundError: No module named 'distutils'`. Added a non-blocking `preinstall` check (`scripts/check-python.mjs`, wired in `package.json`) that warns early when the resolved Python is ≥3.12 without `setuptools`, pointing to the two known workarounds. Silenceable with `MAESTRO_SKIP_PYTHON_CHECK=1`.

### 3. SSH `readDirRemote` exit-code handling (review follow-ups from closed PR #885)
Two commits responding to review feedback on #885:

- **`fix(ssh): readDirRemote succeeds when symlink-scan for-loop exits non-zero`** — the remote command ends with a `for` loop probing for symlinks-to-dirs; when no such symlink exists the final `[ -L "$f" ]` returns non-zero and becomes the SSH command's exit code. With stderr silenced, the caller only saw Node's generic "Command failed" boilerplate despite valid `ls` output. Fixed on both sides: `|| true` on the symlink scan, plus a defense-in-depth tolerance in `readDirRemote` that accepts a non-zero exit when the `__SYMDIR__` marker is present.
- **`fix(ssh): narrow readDirRemote marker-based success override`** — the initial tolerance was too wide (any non-zero exit with `__SYMDIR__` in stdout was accepted). Narrowed to the specific known for-loop-tail shape: exit 1, marker present, no `__LS_ERROR__`, empty stderr — so a real SSH/transport failure still propagates.

## Also closes
- **#187** (Windows cloudflared detection) — closed as duplicate of #159.
- **#298** (MCP Desktop Commander on Windows) — closed as a by-design batch-mode limitation.

## Test plan
- [x] `npx prettier --check` on all touched files — clean
- [x] `npx eslint` on modified source files — clean
- [x] `npx vitest run` scoped to `claude.test.ts`, `agents.test.ts`, `useWizardHandlers.test.ts` — **215/215 passing**
- [x] Added regression tests:
  - `claude.test.ts`: skill discovery via `SKILL.md` (primary) and `skill.md` (legacy fallback)
  - `agents.test.ts`: `discoverSlashCommands` enriches skill commands with frontmatter description
  - `useWizardHandlers.test.ts`: IPC-returned description survives through to the session's `agentCommands`
  - `readDirRemote` regression tests (in prior SSH commits) covering non-zero-exit-with-valid-stdout, non-empty-stderr, and non-1-exit-code paths
- [x] Manually verified `scripts/check-python.mjs` fires the warning on Python 3.12.3 without `setuptools` and exits 0; also verified `MAESTRO_SKIP_PYTHON_CHECK=1` silences it

## Regression notes
- On case-insensitive filesystems (Windows NTFS, default macOS APFS) both `SKILL.md` and `skill.md` resolve to the same file, so behaviour is unchanged there.
- `useWizardHandlers` uses `cmd.description ?? getSlashCommandDescription(...)` — when `description` is absent the old lookup-table behaviour is preserved, so non-skill slash commands are unaffected.
- The preinstall script only warns; it never blocks install, so users who explicitly choose to run an unsupported Python aren't forced off it.

## Summary by CodeRabbit

* **New Features**
  * Slash commands can show descriptions extracted from skill metadata.
  * Added an advisory pre-install Python compatibility check (skippable via env var).

* **Improvements**
  * Skill discovery now prefers canonical uppercase SKILL.md with fallback and better case-sensitive handling.
  * Better messaging guiding creation/location of skill metadata.

* **Bug Fixes**
  * More robust error handling during skill scans (non-missing IO errors surface appropriately).

* **Tests**
  * Added unit tests covering skill discovery, description enrichment, and persistence.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
 		"package:linux": "node scripts/set-version.mjs npm run build && node scripts/set-version.mjs electron-builder --linux",
 		"start": "electron .",
 		"clean": "rm -rf dist release node_modules/.vite",
+		"preinstall": "node scripts/check-python.mjs",
 		"prepare": "node scripts/setup-git-hooks.mjs",
 		"postinstall": "electron-rebuild -f -w node-pty,better-sqlite3",
 		"lint": "tsc -p tsconfig.lint.json && tsc -p tsconfig.main.json --noEmit && tsc -p tsconfig.cli.json --noEmit",
--- a/scripts/check-python.mjs
+++ b/scripts/check-python.mjs
@@ -1,0 +1,74 @@
+// Warn early if the Python that node-gyp will pick up is 3.12+ without
+// `setuptools` installed. node-gyp (pulled in transitively by better-sqlite3
+// and node-pty) still imports `distutils`, which was removed from the
+// standard library in Python 3.12 — so `npm install` dies with
+// `ModuleNotFoundError: No module named 'distutils'` during the postinstall
+// electron-rebuild step. See https://github.com/RunMaestro/Maestro/issues/170.
+
+import { spawnSync } from 'node:child_process';
+
+const skip =
+	process.env.MAESTRO_SKIP_PYTHON_CHECK === '1' || process.env.MAESTRO_SKIP_PYTHON_CHECK === 'true';
+if (skip) process.exit(0);
+
+const candidate = process.env.PYTHON || process.env.npm_config_python || 'python3';
+
+function resolvePython() {
+	const bins = candidate === 'python' ? ['python'] : [candidate, 'python'];
+	for (const bin of bins) {
+		const probe = spawnSync(bin, ['--version'], { stdio: 'pipe', encoding: 'utf8' });
+		if (probe.error || probe.status !== 0) continue;
+		const raw = (probe.stdout || probe.stderr || '').trim();
+		const match = raw.match(/Python (\d+)\.(\d+)/);
+		if (!match) continue;
+		return { bin, major: Number(match[1]), minor: Number(match[2]), raw };
+	}
+	return null;
+}
+
+const python = resolvePython();
+if (!python) {
+	// No Python found — let node-gyp raise its own error if it actually needs one.
+	process.exit(0);
+}
+
+// distutils was removed in Python 3.12; node-gyp's vendored gyp still imports it.
+const needsDistutilsWorkaround = python.major > 3 || (python.major === 3 && python.minor >= 12);
+if (!needsDistutilsWorkaround) process.exit(0);
+
+const hasSetuptools = spawnSync(python.bin, ['-c', 'import setuptools'], {
+	stdio: 'pipe',
+	encoding: 'utf8',
+});
+// Treat "setuptools present" AND spawn failures (status === null from a
+// transient EAGAIN, sandbox, or binary-gone-away) as non-problems: the
+// advisory is best-effort and we'd rather stay silent than misattribute
+// a spawn error to a missing module.
+if (hasSetuptools.error || hasSetuptools.status === null || hasSetuptools.status === 0) {
+	process.exit(0);
+}
+
+const reset = '\x1b[0m';
+const yellow = '\x1b[33m';
+const bold = '\x1b[1m';
+// eslint-disable-next-line no-console
+console.warn(
+	[
+		'',
+		`${yellow}${bold}[maestro] Python toolchain warning${reset}`,
+		`  Detected ${python.raw} at '${python.bin}', but \`setuptools\` is not installed.`,
+		`  node-gyp (via better-sqlite3 / node-pty) still imports \`distutils\`, which was`,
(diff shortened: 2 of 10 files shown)


pedramamini:
@chr1syy thanks a lot for picking off three items from the open-issue backlog in one go — the `SKILL.md` case-sensitivity fix, the Python 3.12 preinstall guard, and the `readDirRemote` exit-code tolerance are all solid and come with good regression tests.

Two things before we can merge:

1. **Merge conflicts with `rc`** — GitHub is showing this branch as conflicting. Could you kindly rebase onto the latest `rc` and resolve the conflicts so we get a clean merge?

2. **Minor follow-up from the CodeRabbit review** worth considering while you're rebasing (not blocking, but nice to tighten):
   - In `readClaudeSkillDescriptions` (`src/main/ipc/handlers/agents.ts` ~L104), the `description:` regex runs against the entire file content rather than just the YAML frontmatter block. `claude.ts`'s `parseSkillFile` already does the bounded version — worth mirroring that here so a stray `description:` line in the skill body can't win over the real frontmatter value.
   - The `readdir`/`readFile` catches in the same helper silently swallow all errors. Narrowing those to `ENOENT`/`ENOTDIR` and re-throwing the rest would let unexpected failures (`EACCES`, `EIO`, etc.) surface to Sentry, matching the project's error-handling guideline.

Ping me once the rebase is done and I'll take another pass.

chr1syy:
@pedramamini rebased onto `rc` and resolved all outstanding review comments. Now a single commit on top of `upstream/rc` (your `957312618` `find`-based fix supersedes the two `readDirRemote` commits I was carrying, so I dropped them during the rebase).

What's addressed:

- **Rebase conflicts** — gone; single clean commit now.
- **CodeRabbit (first pass) on `agents.ts`**:
  - Bounded the frontmatter regex via `extractFrontmatter()` so a body line can't masquerade as `description:`.
  - Added `isMissingEntryError()` and narrowed `readdir`/`readFile` catches to `ENOENT`/`ENOTDIR` — other errors propagate to Sentry.
- **CodeRabbit (second pass) on `claude.ts`**:
  - Same error narrowing applied to `parseSkillFile` and `scanSkillsDir` so `EACCES`/`EIO` on the skills dir no longer silently drops skills.
- **CodeRabbit inline on `agents.ts` enrichment site**:
  - Wrapped the `readClaudeSkillDescriptions(cwd)` call at the `discoverSlashCommands` call site in its own try/catch. Enrichment is best-effort — a permission error captures to Sentry but no longer tears down the user's slash-command list.
- **CodeRabbit nitpicks on `check-python.mjs`**:
  - Candidate list dedupes when `PYTHON=python`.
  - `spawnSync` failures (`status === null`) are no longer misattributed to missing `setuptools` — advisory stays silent in that case.

New regression tests (219 scoped tests passing):

- `claude.test.ts`: `getSkills` propagates `EACCES` from `scanSkillsDir` and `EIO` from `parseSkillFile` instead of silently returning an empty list.
- `agents.test.ts`: skill enrichment failure still returns slash commands as names-only (updated from the earlier "propagates" expectation to match the new best-effort semantics).

Prettier + eslint clean on all touched files.

pedramamini:
Thanks @chr1syy for the thorough work on these three fixes — the case-sensitive `SKILL.md` discovery, the Python 3.12 preinstall guard, and the SSH `readDirRemote` follow-ups all look great. Appreciate the regression tests and how cleanly each concern was split. Also nice job iterating through the bot review feedback (frontmatter bounding, fs error propagation, enrichment resilience). Merging.

Issue #871, opened 2026-04-19
Cue dashboard Active Runs: checkbox-shaped affordance acts as delete with no confirmation


scriptease:
## Summary

In the Cue modal's **Active Runs** section, each row has what looks like a checkbox at the leading edge. Clicking it does **not** select the row — it immediately and irreversibly **deletes** (cancels/removes) the run, with no confirmation prompt. This is a UX trap: the affordance promises selection, the action is destruction.

<img width="1213" height="334" alt="Image" src="https://github.com/user-attachments/assets/76285267-1bb0-4f97-b708-c431030d2e7e" />
See red checkbox in front of the active run.

## Repro

1. Trigger any subscription so it appears under Active Runs (or wait for one to fire).
2. In the Active Runs section, click the small box-shaped indicator on the left of the row.
3. Observed: the run is removed instantly, no confirm dialog, no undo.
4. Expected: either selection (with deletion via a separate button), or an unambiguous delete affordance with confirmation.

## Why this is wrong

Two failure modes overlap here:

- **Wrong icon for the action.** A checkbox is a *selection* control. A trash icon is a *delete* control. Using a checkbox shape for an immediate destructive action breaks every UI convention the rest of the app already follows.
- **No confirmation on a destructive irreversible action.** Active runs may represent in-flight agent work that's been running for minutes (or longer for `time.heartbeat` jobs). One stray click and that work is gone with no way to recover.

## Suggested fixes (pick one)

1. **Trash icon, parity with the Sessions Table.** The top Sessions With Cue table already uses a trash icon for its delete action. Use the same icon here. No checkbox-looking element, no confusion. Ideally still confirm before destroy (single-click destroy is fine for trivial actions, but cancelling an in-flight run isn't trivial).

2. **True checkbox + bulk delete pattern.** If the goal is multi-select bulk operations, keep the checkbox but make it actually select (not delete). Add a **Delete Selected** button that appears once one or more rows are checked — that button is the destructive control, and it carries an implicit confirmation step (the user has to deliberately click both "select" and "delete").

Either fix is acceptable; today's hybrid (checkbox affordance + immediate delete) is the worst of both worlds.

## Related dashboard issues

- #865 — `task.pending` first-scan seeding never fires
- #866 — dormant agents not visible in dashboard
- #867 — multiple agents sharing a project root run every subscription N times
- #868 — Run Now only fires the first subscription
- #869 — View in Pipeline always opens the first pipeline
[...]

Issue #870, opened 2026-04-19
Cue dashboard: show pipeline count when an agent has more than one pipeline


scriptease:
## Summary

In the Cue modal's **Sessions With Cue** table, the Pipelines column shows colored dots for each pipeline on an agent but never surfaces a numeric count. When an agent has more than one pipeline this is hard to read at a glance — and when several agents are listed it becomes very easy to miss that an agent has additional pipelines beyond the first one or two visible dots. Adding an explicit count (e.g. `●●● 3` or `3 pipelines`) when `pipelineCount > 1` would make the surface honest about how much is configured.

## Repro / observed behavior

<img width="316" height="217" alt="Image" src="https://github.com/user-attachments/assets/eede09dc-92a5-4a2a-b451-aeb5d82f401e" />

> Shows that the agent does in fact have multiple pipelines configured, visible only after clicking into the dropdown.

<img width="1240" height="287" alt="Image" src="https://github.com/user-attachments/assets/4ac75ba6-822f-456c-9e07-ffa060def63b" />

> Several agents are present but the column doesn't communicate which ones own multiple pipelines and the red and orange are missing. Without a count, the user assumes one pipeline per agent and is confused where their pipeline is.

## Expected

The Pipelines column should show a numeric count alongside the dots whenever an agent has `> 1` pipeline. Examples of acceptable presentations:

- `●●● ×3`
- `3 ●●●`
- `● ● ● (3)`
- A small badge `3` overlaid on the dot cluster

Single-pipeline rows can keep the bare dot — adding `×1` everywhere would be noise.

## Why this matters

- **Discoverability** — users need to know multiple pipelines exist before they think to click into them.
- **Pairs with #869** — the broken "View in Pipeline" routing already makes the multi-pipeline case painful; a missing count makes the same case invisible.
- **Pairs with #868** — Run Now fires only the first subscription, so users need to know how many subscriptions exist to even understand what they're missing.

## Suggested fix

In the Sessions table cell renderer for the Pipelines column, append a count whenever the pipeline list has more than one entry. The count is already known at render time (it's the source of the dots).

## Related

- #865 — `task.pending` first-scan seeding
- #866 — dormant agents not visible in dashboard
- #867 — multiple agents sharing a project root run every subscription N times
- #868 — Run Now only fires the first subscription
- #869 — View in Pipeline always opens the first pipeline

jSydorowicz21:
This was added in #893. If you still see this issue please reopen the issue.

Issue #869, opened 2026-04-19
Cue dashboard 'View in Pipeline' always opens the first pipeline regardless of which row was clicked


scriptease:
## Summary

In the Cue modal's Sessions Table, the **View in Pipeline** button always opens the *first* pipeline in the Pipeline Editor, regardless of which row was clicked. The sibling **Edit YAML** button on the same rows works correctly (opens the right agent's YAML). This is a routing/wiring issue — the source pipeline ID isn't being threaded through to the Pipeline Editor's filter on navigation.

## Repro

1. Have at least two agents (or one agent with multiple pipelines via the `# Pipeline:` comment + `name-chain-N` convention) so the Pipeline Editor has more than one pipeline.
2. Open the Cue modal (`Option+Q`) → Sessions With Cue.
3. Click **View in Pipeline** on the *second* row.
4. Observed: Pipeline Editor opens, focus is on the *first* pipeline.
5. Expected: Pipeline Editor opens filtered/focused to the pipeline owned by the row that was clicked — i.e., parity with the docs:

   > "View in Pipeline — Jump to the Pipeline Editor filtered to that agent"

## Why it's likely a threading issue (not a deeper bug)

- The sibling **Edit YAML** button on the same row works fine — it correctly opens the YAML for the right agent. So the row context (agent / session / pipeline ID) is available where the click happens.
- The Pipeline Editor itself can be filtered by pipeline name (clicking a pipeline in the top bar selects it), so the destination view supports targeted navigation.
- The bug appears to be that the View in Pipeline click handler navigates to the Pipeline Editor without forwarding the source pipeline/agent ID, so the editor falls back to its default selection (the first pipeline).

## Suggested fix

Thread the row's pipeline ID (or agent ID + pipeline name) through the navigation call. Pseudocode:

```ts
// today
onViewInPipeline(row) => navigate('/cue/pipelines')

// proposed
onViewInPipeline(row) => navigate('/cue/pipelines', { selectPipeline: row.pipelineId })
```

Then have the Pipeline Editor read `selectPipeline` on mount and pre-apply the same filter that clicking a pipeline in the top bar applies.

## Workaround today

After clicking View in Pipeline, manually click the right pipeline name in the editor's top bar.

## Related

- #865 — `task.pending` first-scan seeding
- #866 — dormant agents not visible in dashboard
- #867 — multiple agents sharing a project root run every subscription N times
- #868 — Run Now only fires the first subscription on a multi-subscription agent

[...]

Issue #865, opened 2026-04-19
task.pending: first-scan seeding never fires, even when tasks already exist


scriptease:
## Summary

When a `task.pending` subscription is added (or hot-reloaded) against a file that **already** contains unchecked tasks, no event ever fires for those existing tasks. The scanner silently seeds the file's hash on its first scan and only fires on **subsequent** content changes. This is invisible from the docs and from the Activity Log (which shows no entries at all), making it look like the subscription is broken.

## Repro

1. Create `.maestro/cue.yaml` with a fresh `task.pending` subscription:
   ```yaml
   subscriptions:
     - name: "research queue"
       event: task.pending
       watch: "research.md"
       poll_minutes: 1
       prompt_file: .maestro/prompts/cue-research.md
   ```
2. Create `research.md` with one or more `- [ ]` items already present.
3. Wait several poll intervals.
4. Open the Cue modal → Activity Log is empty. `cue_events` table in `cue.db` has no rows for this subscription.

## Expected

Either (a) the first scan fires for existing pending tasks, or (b) the docs make the seed-then-fire behavior explicit so users know they need to modify the file (or hit Run Now) to bootstrap the subscription.

## Actual

`src/main/cue/cue-task-scanner.ts` lines 132–135:

```ts
// On first scan, seed the hash but don't fire events
if (prevHash === undefined) {
    continue;
}
```

The first scan always silently seeds. Combined with the fact that adding a new subscription to a file that already has tasks is a *very* natural first-use scenario, this is a confusing dead-on-arrival experience.

## Docs gap

The Cue events reference at https://docs.runmaestro.ai/maestro-cue-events under `task.pending` says only:

> Only fires when the task list changes (new tasks appear or existing ones are modified)

That phrasing is ambiguous — a reader reasonably interprets "tasks appear" to include "tasks present when the engine first sees the file". The seeding step is not mentioned anywhere.

GitHub poller docs explicitly call out the equivalent behavior:

> On first run, seeds the "seen" list with existing PRs — only **new** PRs trigger events

The same disclosure is missing for `task.pending`.

## Suggested fixes (any one of these would help)

1. **Doc fix** — add a note to `task.pending` matching the wording used for `github.pull_request`: *"On first scan, seeds the file hash — only subsequent changes trigger events. Use Run Now (or `maestro-cli cue trigger`) to bootstrap a subscription against a file that already contains tasks."*
[...]

PR #893, opened 2026-04-24
fix(cue): five dashboard polish fixes (#828, #865, #869, #870, #871)

reachrazamair:
## Summary

Five small bugs in the Cue dashboard, all on the `cue-polish` branch.

**#828 — YAML editor shows "Config must be a non-null object" on first open**
`yaml.load()` returns `null` for a comments-only file (which is what the default template is). `validateCueConfigDocument` was treating any falsy value as invalid. Fixed by short-circuiting on `null`/`undefined` and returning a clean `{ valid: true, errors: [] }` — so comments-only and blank files are treated as an empty config rather than an error. Also covers the case where a user manually clears their file down to just comments.

**#865 — task.pending first-scan seeding is silent**
On startup, `cue-task-scanner` seeds file hashes without firing events (correct), but emitted no log entry, making it look like the scanner hadn't run at all. Added an `info` log per seeded file, consistent with the pattern `cue-github-poller` already uses. No behavior change.

**#869 — "View in Pipeline" always opens the first pipeline**
`handleViewInPipeline` was ignoring its `session` argument and just switching the tab. Fixed by resolving the session's pipeline via `getPipelineColorForAgent` + `dashboardPipelines`, storing it as `pendingPipelineId` in `CueModal` state, and passing it as a new `initialPipelineId` prop to `CuePipelineEditor`. The editor calls `selectPipeline` once the pipeline list has loaded.

**#870 — No indication when a session belongs to multiple pipelines**
The Sessions table rendered one dot per pipeline but gave no count. Added a `×N` badge after the dots when `colors.length > 1`.

**#871 — Active Runs stop button deletes without confirmation**
The per-run stop button (previously a `Square` icon) called `onStopRun` immediately on click with no warning. Wrapped it in `getModalActions().showConfirmation(...)` and replaced the icon with `Trash2` to match other destructive actions in the UI. The "Stop All" button is unchanged.

## Files changed

| File | Change |
|---|---|
| `src/main/cue/config/cue-config-validator.ts` | Treat `null`/`undefined` as valid empty config |
| `src/main/cue/cue-task-scanner.ts` | Log seeded files on first scan |
| `src/renderer/components/CueModal/CueModal.tsx` | Resolve pipeline from session in `handleViewInPipeline`; thread `initialPipelineId` |
| `src/renderer/components/CuePipelineEditor/CuePipelineEditor.tsx` | Add `initialPipelineId` prop; `useEffect` calls `selectPipeline` once list loads |
| `src/renderer/components/CueModal/SessionsTable.tsx` | `×N` badge when `colors.length > 1` |
| `src/renderer/components/CueModal/ActiveRunsList.tsx` | Stop button → confirmation dialog; `Square` → `Trash2` |
| `src/__tests__/main/cue/cue-yaml-loader.test.ts` | Update null test to assert `valid: true`; add separate test for non-null non-object |
| `src/__tests__/renderer/components/CueModal.test.tsx` | Update stop-run test to confirm dialog before asserting `stopRun` called |

## Test Plan

- `npm run test` — 26,283 passed, 0 failed
- `npx tsc --noEmit` — no errors
- ESLint — no warnings or errors

## Summary by CodeRabbit

* **New Features**
  * "View in pipeline" now preselects the matching pipeline in the editor.

* **UI Changes**
  * Per-run stop now requires confirmation; stop icon updated to Trash2.
  * Pipeline color display shows a dim ×N count when multiple pipelines present.
  * Queue-overflow toast includes millisecond timestamps.
  * Drawer open delay and cleanup improved to avoid stale state.

* **Bug Fixes**
  * Initial GitHub poll failures no longer stop rescheduling.
  * Heartbeat error reporting deduplicates repeated signatures.
  * Null/empty Cue configs are treated as valid.
  * Startup/refresh semantics refined so startup events fire correctly after stops and for late-discovered sessions.

### DIFF
--- a/src/__tests__/main/cue/cue-ipc-handlers.test.ts
+++ b/src/__tests__/main/cue/cue-ipc-handlers.test.ts
@@ -226,10 +226,10 @@ describe('Cue IPC Handlers', () => {
 	});
 
 	describe('cue:enable', () => {
-		it('should call engine.start()', async () => {
+		it('should call engine.start() with system-boot reason', async () => {
 			const handler = registerAndGetHandler('cue:enable');
 			await handler(null);
-			expect(mockEngine.start).toHaveBeenCalledOnce();
+			expect(mockEngine.start).toHaveBeenCalledWith('system-boot');
 		});
 	});
 
--- a/src/__tests__/main/cue/cue-session-lifecycle.test.ts
+++ b/src/__tests__/main/cue/cue-session-lifecycle.test.ts
@@ -588,10 +588,11 @@ describe('CueEngine session lifecycle', () => {
 			engine.stop();
 		});
 
-		it('a second system-boot in the same process lifecycle still dedups', () => {
-			// Edge case: simulates someone (e.g. a test) calling start('system-boot')
-			// twice without a real Electron restart. The startup keys persist across
-			// stop/start so the second boot is a no-op for app.startup.
+		it('a second system-boot start after stop re-fires app.startup (dedup keys cleared on stop)', () => {
+			// stop() resets the startup dedup keys so that re-enabling Cue (which
+			// calls start('system-boot')) fires startup subscriptions again. This
+			// matches the expected UX: toggling Cue off then on is treated as a
+			// new Cue "boot" from the user's perspective.
 			mockLoadCueConfig.mockReturnValue(makeStartupConfig());
 			const deps = createMockDeps();
 			const engine = new CueEngine(deps);
@@ -599,9 +600,9 @@ describe('CueEngine session lifecycle', () => {
 			engine.start('system-boot');
 			expect(deps.onCueRun).toHaveBeenCalledTimes(1);
 
-			engine.stop();
-			engine.start('system-boot');
-			expect(deps.onCueRun).toHaveBeenCalledTimes(1);
+			engine.stop(); // clears startup dedup keys
+			engine.start('system-boot'); // should fire again
+			expect(deps.onCueRun).toHaveBeenCalledTimes(2);
 
 			engine.stop();
 		});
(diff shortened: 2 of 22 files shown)


pedramamini:
Thanks for the contribution, @reachrazamair — nice batch of polish fixes, and the write-up is clear.

Both Greptile and CodeRabbit independently flagged two P1-ish issues around the new "View in Pipeline" plumbing that I agree are worth addressing before this merges:

1. **`CuePipelineEditor.tsx:~126` — `useEffect` re-fires on pipeline list mutations.** The dep array `[initialPipelineId, pipelines.length]` means any pipeline create/delete will re-run the effect and snap the user back to the originally-requested pipeline, overriding whatever they've selected since. Should be gated with a one-shot `useRef` so it only applies on initial load.

2. **`CueModal.tsx:~141` — `pendingPipelineId` is never cleared after consumption.** Because `CuePipelineEditor` unmounts when switching away from the pipeline tab, remounting re-applies the stale `initialPipelineId` even if the user didn't click "View in Pipeline" again. Related: re-clicking "View in Pipeline" for the same session is a no-op on `useState` (same value), so that path also doesn't re-trigger. A request-token shape (`{ id, nonce }`) or clearing after consumption handles both.

Also worth a look (lower priority):
- **`useCue.ts:148`** — Greptile noted the raw ISO timestamp is being appended to the user-facing toast title. Using `getMilliseconds()` or similar keeps titles distinct without leaking the wire value.
- **`cue-heartbeat.ts:~112`** — CodeRabbit flagged that the per-streak dedup only tracks the most recent signature, so alternating A/B/A/B errors would re-fire `onFailure` repeatedly and break the "exactly once per failure run" contract in `cue-engine.ts`. A `Set<string>` of reported signatures, or decoupling Sentry dedup from `onFailure`, would preserve that.

Happy to re-review once those are in. Thanks again!

PR #894, opened 2026-04-24
fix(cli): write CLI discovery file when web server is restarted (+ tests/refactor)

### LINKED ISSUE #859: maestro-cli open-file fails: CLI server discovery file (cli-server.json) not written when turning the network interface off and on
## Version
- Maestro: 0.16.9-RC (Electron 28.3.3)
- OS: macOS
- CLI: `/usr/local/bin/maestro-cli`

## Summary

`maestro-cli open-file <path>` (and all commands requiring a desktop connection: `refresh-files`, `refresh-auto-run`, `auto-run`, `send -t`) always exits with:

```
Error: Maestro desktop app is not running
```

even when Maestro is running.

Image

## Root Cause

The CLI detects the desktop app by reading a discovery file at `~/Library/Application Support/maestro/cli-server.json` with schema `{ port, token, pid, startedAt }`, then connects via WebSocket to `ws://localhost:{port}/{token}/ws`.

The app is confirmed running (HTTP health endpoint responds), but `cli-server.json` is never written to disk and the WebSocket server is not started — confirmed by 404 responses on all WebSocket paths.

## Steps to Reproduce

1. Launch Maestro 0.16.9-RC
2. Run `maestro-cli status` or `maestro-cli open-file /any/valid/path`

**Expected:** command succeeds, file opens as a preview tab
**Actual:** `Error: Maestro desktop app is not running`

## What Still Works

`maestro-cli list`, `maestro-cli send`, `maestro-cli settings` — these read local JSON files and do not require the WebSocket server.

## Workaround

None. The discovery path is hardcoded; there is no flag to override it.

## Bug

I've traced the bug. Here's what's going on:

**Root cause:** `live:startServer` in `src/main/ipc/handlers/web.ts:269-294` never calls `writeCliServerInfo()`. Only `ensureCliServer()` (fired once at app launch) writes the discovery file.

**The sequence that bricks CLI discovery:**
1. App boots → `ensureCliServer()` writes `cli-server.json` with port+token.
2. User stops the web server (via UI / "disable all live") → `live:stopServer` runs `deleteCliServerInfo()` + `setWebServer(null)`.
3. User restarts it → `live:startServer` creates a brand-new `WebServer` (new token, new random port) → **no discovery file is written**.
4. CLI `readCliServerInfo()` returns null → "Maestro desktop app is not running."

Your running app matches this exactly: port 63866 is bound (new port, new WebServer), but `cli-server.json` is absent.

**Fix options**

- **A (minimal):** mirror `ensureCliServer` inside `live:startServer` — after the server starts, call `writeCliServerInfo({port, token: webServer.getSecurityToken(), pid, startedAt})`. Two lines.
- **B (more robust):** refactor so `live:startServer` delegates to `ensureCliServer` (same create/start-or-reuse path). Avoids future drift.

Option A is the surgical fix and matches the spirit of CLAUDE.md. Want me to apply it?

chr1syy:
## Summary

Polished continuation of @scriptease's #860. Ships their fix for #859 and adds the follow-ups called out in review:

- **Extract `refreshCliDiscoveryFile(port, token)` helper.** Collapses four near-identical \`writeCliServerInfo({port, token, pid: process.pid, startedAt: Date.now()})\` blocks (two in \`ensureCliServer\`, two in \`live:startServer\`) to one call site. Prevents future drift on pid/startedAt semantics or new \`CliServerInfo\` fields. (This was CodeRabbit's nitpick on #860.)
- **Mock \`cli-server-discovery\` in the web-handlers tests.** Before this change, the existing \`live:startServer\` tests silently wrote \`~/.config/maestro/cli-server.json\` on the developer's machine every time they ran.
- **Fix the \`WebServer.start()\` mock return.** Added \`token: 'mock-security-token'\` so the mock matches the real \`{port, token, url}\` shape — the PR's new code destructures \`token\` and was silently receiving \`undefined\`.
- **Cover the new behavior.** Four new regression tests assert \`writeCliServerInfo\` is called with the expected args in each \`live:startServer\` branch (fresh-create, restart-existing, already-running) and *not* called when \`start()\` throws. Added \`deleteCliServerInfo\` assertions to \`live:stopServer\` and \`live:disableAll\` so the full CLI-discovery lifecycle is pinned.

\`scriptease\`'s original commit is preserved unchanged as the first commit on this branch; the polish is a separate commit.

Fixes #859.
Supersedes #860.

## Test plan

- [x] \`vitest run src/__tests__/main/ipc/handlers/web.test.ts\` passes (35/35, up from 31).
- [x] \`prettier --check\` clean on both changed files.
- [x] No new type errors in scoped files.
- [x] Launch Maestro → verify \`~/Library/Application Support/maestro/cli-server.json\` exists.
- [x] \`maestro-cli status\` / \`maestro-cli open-file <path>\` succeed.
- [x] Stop the web server from the UI → verify \`cli-server.json\` is deleted and \`maestro-cli\` reports "not running".
- [x] Start the web server again from the UI → verify \`cli-server.json\` is rewritten with the new port/token and \`maestro-cli open-file\` succeeds without restarting the app.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Server discovery writes centralized and treated as non-fatal so startups proceed even if discovery write fails.
  * When persistent web links are disabled, servers are rotated to mint a fresh token; when enabled, servers are reused without rotation.
  * Cleanup reliably removes discovery info on stop/disable; rotation failures preserve the existing server.

* **Tests**
  * Expanded coverage for rotation, reuse, discovery refresh on create/restart, no-refresh-on-start-error, non-fatal discovery-write errors, and cleanup/delete behavior.


pedramamini:
Thanks for the thorough polish on this, @chr1syy — the helper extraction, mock hygiene fix, and the four new regression tests are exactly what #860 needed. Great work, and nice catch on the `token: undefined` silently flowing through the existing tests.

A couple of small items before we merge:

1. **Discovery write failures can now mask a successful server start** (CodeRabbit's outside-diff comment, legit). In `src/main/ipc/handlers/web.ts`, `refreshCliDiscoveryFile` runs synchronously inside the outer `try/catch` in `live:startServer`. If `fs.writeFileSync`/`fs.renameSync` throws (disk full, permissions, unwritable config dir), `webServer.start()` has already succeeded but we'd fall into the outer catch at line 297 and return `{ success: false, error }` — so the UI shows "failed to start" while the server is genuinely running. `ensureCliServer` (lines 96–99) already treats this as non-fatal; `live:startServer` should mirror that. A local `try/catch` around each `refreshCliDiscoveryFile` call that logs and continues would fix it.

2. **Minor test gap** (Greptile P2): the zero-session branch of `live:disableAll` still calls `webServer.stop()` + `deleteCliServerInfo()` in the implementation, but the test doesn't assert the latter — worth a one-line addition to match the two-session test and pin the full lifecycle.

The fix itself and the rest of the tests look great — once those two are addressed this is ready to go.

chr1syy:
Addressed both review items in `cf66457e4`:

**1. Non-fatal discovery-write in `live:startServer`** — wrapped both `refreshCliDiscoveryFile(...)` calls in local `try/catch` blocks that log and continue. A `fs.writeFileSync` failure no longer masks a successfully-started server; the handler returns `{ success: true, url }` as long as `webServer.start()` (or the already-running check) succeeds. Mirrors `ensureCliServer`'s behavior.

**2. Zero-session `live:disableAll` lifecycle** — added `expect(mockWebServer.stop).toHaveBeenCalled()` and `expect(deleteCliServerInfo).toHaveBeenCalledTimes(1)` assertions so the full cleanup path is pinned.

Also added two regression tests covering the new non-fatal behavior (fresh-start write failure + already-running write failure).

37/37 tests pass · prettier clean · no new type errors.

PR #896, opened 2026-04-24
feat(web): mobile composer polish + port Max Output Lines setting

chr1syy:
## Summary
- **Composer polish (mobile)**: `CommandInputBar` now reports its measured height via a `ResizeObserver`; the chat view reserves matching bottom padding so the last message stays visible when the bar expands with multi-line drafts.
- **Stacked composer buttons**: re-mounted voice / slash / thinking action buttons on the bottom row of the stacked phone composer, alongside the send button, so the send button no longer floats with an awkward gap.
- **Port "Max Output Lines per Response"**: added the desktop setting to the web UI with the 15 / 25 / 50 / 100 / All toggle group. Default on web is **All**. Infinity serializes as `null` on the wire; desktop rehydrates. `MessageHistory` now applies line-based truncation only when a finite cap is configured.

Note: also includes the previously-landed-locally `feat(cli): honor session config overrides + SSH remote when spawning agents` commit, which was already on the branch head before this work.

## Test plan
- [x] CI: type-check, ESLint, Prettier clean on touched files
- [x] Unit tests: `src/__tests__/web/mobile/MessageHistory.test.tsx` and `src/__tests__/web/mobile/CommandInputBar.test.tsx` updated + passing
- [x] Desktop: verify existing `Max Output Lines` setting on Display tab still works (null ↔ Infinity round-trip)
- [x] Web: on a phone viewport, type a multi-line draft and verify the composer grows without hiding the last chat line, and that action buttons reflow to the left of the send button when stacked
- [x] Web: open Settings → Behavior → verify `Max Output Lines per Response` renders with `All` selected by default; changing it updates message truncation in-place and syncs to desktop

## Summary by CodeRabbit

* **New Features**
  * Added "Max Output Lines per Response" setting to control message truncation in the mobile UI.
  * Implemented dynamic input bar height adjustment for improved mobile layout.
  * Enhanced settings synchronization between desktop and web clients.

* **Bug Fixes**
  * Fixed environment variable precedence handling in agent spawning to follow documented layering order.


pedramamini:
Thanks for the contribution, @chr1syy! 🎉

Reviewed the changes and this looks great:

- **Composer height observer**: `ResizeObserver` in `CommandInputBar.tsx` → `inputBarHeight` state in `App.tsx` is a clean way to reserve scroll padding; the initial `getBoundingClientRect()` call avoids a one-frame layout pop before the observer fires.
- **Stacked-mode action buttons**: re-mounting voice/slash/thinking on the bottom row fixes the awkward floating send button. Updated test in `CommandInputBar.test.tsx` correctly inverts the old "not in document" assertion.
- **Max Output Lines port**: the `Infinity ↔ null` wire encoding is threaded consistently through `WebSettings`, `web-server-factory`, `CallbackRegistry` default, `messageHandlers` allowlist, `useSettings.setMaxOutputLines`, and the mobile `SettingsPanel`. `MessageHistory` gating line-based truncation behind `Number.isFinite(maxOutputLines)` (with the char threshold still firing) matches desktop semantics, and the new "All (Infinity)" test case locks that down.
- **CLI spawner overrides + SSH**: solid parity work — `applyAgentConfigOverrides` reuse, dynamic `import()` of `ssh-spawn-wrapper` to keep the SSH chain out of the local hot path, and failing loudly via `sshUnresolvedFailure()` instead of silently falling back locally is the right call. Test coverage in `agent-spawner.test.ts` (+426 lines) across local overrides, SSH resolution success/failure, and env precedence is thorough.

No merge conflicts (base: `rc`), and the PR description already flags that the CLI commit rode along from the prior branch state, which matches what I see in `git log`.

CI (lint-and-format, test, CodeRabbit, Greptile) is still pending — please keep an eye on those and address anything that surfaces, but from a code-review standpoint I'm happy. Approving.

chr1syy:
@greptileai re-review please

PR #909, opened 2026-04-27
fix(cli): register copilot-cli parser + honor promptArgs in send

chr1syy:
## Summary

- `maestro-cli send` against any `copilot-cli` agent threw `Error: No parser available for agent type: copilot-cli` because `src/cli/services/agent-spawner.ts` had its own duplicated parser switch that was missing the entry. Replaced it with `createOutputParser()` from `parser-factory.ts` so the registry is single-source (copilot-cli is already wired there).
- Same spawner ignored each agent definition's `promptArgs` builder, so even after registration Copilot would have spawned with the prompt as a positional arg after `--` instead of via `-p`. Mirrored `ChildProcessSpawner`'s precedence (`promptArgs` → `noPromptSeparator` → `--`); Copilot now spawns as `copilot --allow-all --output-format json -p "<prompt>"`.
- Added a regression test that asserts the argv shape and round-trips a `session.start` + final `assistant.message` event into a successful result with the captured session id.

## Repro

```
$ maestro-cli send <copilot-agent> "hi"
Error: No parser available for agent type: copilot-cli
    at createParser (.../dist/cli/maestro-cli.js:56686:13)
    at spawnJsonLineAgent (...)
    at spawnAgent (...)
    at _Command.send (...)
```

## Test plan

- [x] `npm run lint` — all three tsc projects clean
- [x] `prettier --check` clean on changed files
- [x] `eslint` clean on `src/cli/services/agent-spawner.ts`
- [x] `vitest run src/__tests__/cli/services/agent-spawner.test.ts` — 85/85 passing (84 pre-existing + 1 new copilot regression)
- [x] Manual: `maestro-cli send <copilot-agent> "hi"` returns a response instead of throwing

## Summary by CodeRabbit

* **Tests**
  * Added test coverage validating agent spawn behavior for the copilot-cli flow, including flag handling and JSON-line output parsing to ensure successful session responses.

* **Refactor**
  * Improved agent spawner to centralize output parser creation and streamline prompt-argument handling for more reliable local and wrapped executions.

### DIFF
--- a/src/__tests__/cli/services/agent-spawner.test.ts
+++ b/src/__tests__/cli/services/agent-spawner.test.ts
@@ -1441,6 +1441,47 @@ Some text with [x] in it that's not a checkbox
 			await resultPromise;
 		});
 
+		it('should spawn copilot-cli with -p prompt arg and parse its result event', async () => {
+			const resultPromise = spawnAgent('copilot-cli', '/project', 'Hello copilot');
+			await new Promise((resolve) => setTimeout(resolve, 0));
+
+			const [cmd, args] = mockSpawn.mock.calls[0];
+			expect(cmd).toBeTruthy();
+
+			// Copilot-CLI batch mode: copilot --allow-all --output-format json -p "prompt"
+			expect(args).toContain('--allow-all');
+			expect(args).toContain('--output-format');
+			expect(args).toContain('json');
+			expect(args).toContain('-p');
+			expect(args).toContain('Hello copilot');
+			// promptArgs path replaces the '--' separator
+			expect(args).not.toContain('--');
+
+			// Emit a session.start init event
+			mockStdout.emit(
+				'data',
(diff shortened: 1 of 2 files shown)


pedramamini:
Thanks @chr1syy — appreciate the contribution! 🙏

Nice fix. Replacing the duplicated parser switch with the shared `createOutputParser()` factory removes a real footgun (every new agent had to be added in two places), and threading `def.promptArgs` through both the local and SSH-wrapped paths is the right move so Copilot/Gemini-style flag-based prompts work in batch mode. Bonus points for moving the parser resolution above `spawn()` to address Greptile's orphaned-process P1.

CodeRabbit's nitpick about tightening the `-p` argv assertion to index-based (à la the `--model` test) is reasonable but not blocking — happy to merge as-is.

Approved.

PR #923, opened 2026-04-28
feat(cli): add session list/show for external history polling (PR2)

chr1syy:
## Summary

PR2 of the Maestro CLI plan — splits read-only conversation-state inspection out of `dispatch`'s write path so external pollers (Maestro-Discord, Cue follow-ups) can pick up where Maestro left off without owning a persistent channel. Pair `dispatch --new-tab` (writes, returns tabId) with `session show <tabId>` (reads, applies `--since`/`--tail` filters) for a stateless poll loop.

- `maestro-cli session list [--json]` — flattens every open AI tab across every Maestro agent into addressable entries `{tabId, agentId, agentName, toolType, name, agentSessionId, state, createdAt, starred}`.
- `maestro-cli session show <tabId> [--since <ts>] [--tail <n>]` — returns the tab's conversation history. `--since` accepts ISO-8601 or an epoch number (auto-detected as ms vs sec by magnitude, so a `Date.now()/1000` cursor and a `Date.now()` cursor both work without a unit flag).
- New `list_desktop_sessions` / `get_session_history` WebSocket message types backed by stateless callbacks that read straight from the persisted session store — no renderer round-trip and no broadcast side-effects, so polling cadence is bounded only by IPC.
- `LogEntry.source` mapped to a coarse `role` (`user` | `assistant` | `system` | `tool` | `thinking` | `error` | `unknown`) for conversational consumers; raw `source` preserved alongside for callers that need to discriminate beyond role.
- Errors land in the typed response (`MISSING_TAB_ID` / `TAB_NOT_FOUND` / `NOT_CONFIGURED` / `MAESTRO_NOT_RUNNING` / `INVALID_OPTION`) rather than as a generic `error` so the CLI's request/response pairing stays deterministic.

## Test plan

- [x] `npx vitest run src/__tests__/cli/commands/session.test.ts` — 16 new tests cover JSON output shape, `--since` ISO/epoch parsing, `--tail` validation, transport-error mapping, and `TAB_NOT_FOUND` surfacing
- [x] `npx vitest run src/__tests__/main/web-server/handlers/messageHandlers.test.ts` — 6 new handler tests cover `list_desktop_sessions` payload, `MISSING_TAB_ID` / `TAB_NOT_FOUND` paths, `--tail` negative-coercion guard
- [x] `npx vitest run src/__tests__/main/web-server/web-server-factory.test.ts` — mock `WebServer` updated with new setters; all 37 tests pass
- [x] All 270 PR2-relevant tests pass (CLI + web-server scope); prettier & ESLint clean on touched files
- [ ] Manual: `maestro-cli dispatch <agent> "test" --new-tab`, capture `tabId`, then `maestro-cli session show <tabId>` returns the prompt + assistant reply
- [ ] Manual: `maestro-cli session show <tabId> --since "$(date -u +%FT%TZ)"` + a follow-up dispatch returns only the new turn
- [ ] Manual: `maestro-cli session list` lists all open tabs across agents

## Summary by CodeRabbit

* **New Features**
  * Added `session list` to enumerate open AI tabs (compact or JSON) and `session show <tab-id>` to view conversation history with `--since`, `--tail`, and `--json` options.

* **Behavior Improvements**
  * Improved `--since` parsing (ISO or epoch seconds/ms), strict `--tail` validation (accepts 0), clearer transcript/list formatting, and deterministic error codes for missing/unmatched tabs and connection failures.

* **Tests**
  * Expanded unit and integration tests for CLI commands, websocket handlers, callback wiring, input validation, and error mapping.

* **Documentation**
  * CLI docs updated with usage, JSON envelope, filters, and failure semantics.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -150,6 +150,113 @@ JSON output includes full session metadata:
 
 Currently supported for `claude-code` agents.
 
+### Session Inspection
+
+Inspect open AI tabs across the running Maestro desktop app and read their conversation history. Pair `dispatch --new-tab` (writes, returns a `tabId`) with `session show <tabId>` (reads, supports `--since` and `--tail`) to build a stateless poll loop without owning a persistent connection — used by Maestro-Discord and Cue follow-ups.
+
+Both verbs talk to the running desktop over the same WebSocket as `dispatch`. There is no on-disk fallback: if the app is not running, the CLI exits with code `MAESTRO_NOT_RUNNING`.
+
+#### List Open Tabs
+
+Flatten every open AI tab across every Maestro agent into addressable entries:
+
+```bash
+# Default: compact text (one tab per line)
+maestro-cli session list
+
+# JSON for scripting
+maestro-cli session list --json
+```
+
+Default text columns: `state` (`busy` / `idle`), star (`★` if starred), `tabId`, agent name + id, tab name, `createdAt` (relative). One tab per line so the output pipes cleanly into `grep`, `awk`, etc.
+
+JSON envelope:
+
+```json
+{
+	"success": true,
+	"sessions": [
+		{
+			"tabId": "tab-1",
+			"sessionId": "tab-1",
+			"agentId": "a1b2c3d4-...",
+			"agentName": "Backend",
+			"toolType": "claude-code",
+			"name": "Refactor parser",
+			"agentSessionId": "claude-uuid-1",
+			"state": "idle",
+			"createdAt": 1714268000000,
+			"starred": false
+		}
+	]
+}
+```
+
+To extract just `tabId`s with `jq`: `maestro-cli session list --json | jq '.sessions[].tabId'`.
+
+#### Show Conversation History
+
+Print a tab's conversation log, with optional cursor (`--since`) and cap (`--tail`) filters applied desktop-side so the wire payload stays small even on long conversations.
+
+```bash
+# Default: formatted transcript (header + per-message blocks)
+maestro-cli session show <tab-id>
+
+# JSON for scripting
+maestro-cli session show <tab-id> --json
+
+# Only messages newer than an ISO-8601 timestamp
+maestro-cli session show <tab-id> --since "2026-04-28T10:00:00Z"
+
+# `--since` also accepts a bare epoch number (auto-detects ms vs sec by magnitude,
+# so both `Date.now()` and `Date.now() / 1000` cursors work without a unit flag)
+maestro-cli session show <tab-id> --since 1714268000
+
+# Cap at the last N messages (applied after `--since`)
+maestro-cli session show <tab-id> --tail 20
+
+# Combine cursor + cap for poll loops
+maestro-cli session show <tab-id> --since "$LAST_TS" --tail 50
+```
+
+| Flag                  | Description                                                                                |
+| --------------------- | ------------------------------------------------------------------------------------------ |
+| `--since <timestamp>` | Only return messages strictly after this timestamp (ISO-8601, or epoch ms/sec auto-scaled) |
+| `--tail <n>`          | Cap output to the last N messages (non-negative integer; applied after `--since`)          |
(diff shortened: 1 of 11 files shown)


pedramamini:
@chr1syy Thanks for this — the design (write/poll split, typed error codes, stateless callbacks straight from the persisted store) is exactly the right shape for external pollers, and the test coverage on the CLI + handler layers is solid. A few things I'd like addressed before we approve:

**P1 — `--tail 0` returns the full transcript instead of an empty list** (`src/main/web-server/web-server-factory.ts:1204-1206`). `Array.prototype.slice(-0)` is identical to `slice(0)` because `-0 === 0`, so the guard `options.tail >= 0` lets `0` through and the caller receives every message. Suggest:
```ts
if (options?.tail !== undefined && options.tail >= 0) {
    logs = logs.slice(Math.max(logs.length - options.tail, 0));
}
```
Worth adding a `--tail 0` case to the handler tests too — the existing suite covers negative coercion but not this one.

**P2 — `--json` declared on `sessionShow` but silently ignored** (`src/cli/commands/session.ts:561-565` + the `--json` registration in `src/cli/index.ts`). `sessionShow` always emits JSON, so the flag is a no-op. Either drop it from `SessionShowOptions` and the CLI registration, or add the human-readable branch to mirror `sessionList`. Given the consumers here are pollers, I'd lean toward dropping it and keeping JSON-only — but please don't leave a registered flag that does nothing.

**P3 — `Number.parseInt(options.tail, 10)` accepts partial input** (`src/cli/commands/session.ts:717`). `parseInt('5abc', 10) === 5` and `parseInt('1.9', 10) === 1`, so `--tail 5abc` silently works as 5 and `--tail 1.9` silently truncates. A `/^\d+$/` precheck (or `Number.isInteger(Number(options.tail))`) before `parseInt` keeps the validation honest.

No merge conflicts, just these correctness fixes. Once those land I'm happy to approve.

chr1syy:
## Manual test notes — 0.16.17-RC build (2026-04-30)

CLI happy path + error cases ran green against a live build with 326 sessions. Filing four small **spec ↔ implementation deltas** to address — feature works, but the surface differs from how it was documented in the test plan / would likely be documented externally:

| Area | Test plan / expected | Actual implementation |
|------|----------------------|------------------------|
| `session list --json` shape | bare array, `jq '.[].id'` | object-wrapped: `{success, sessions: [...]}` — use `jq '.sessions[].tabId'` |
| `session show` history field | `history` | `messages` |
| `session show` limit flag | `--limit N` | `--tail N` (also supports `--since <ISO-8601 \| epoch ms/sec>`) |
| `session show --json` flag | documented as a flag | flag does not exist; output is always JSON |

### Verified cases
- **CLI-01** `session list` — 326 sessions, identical md5 across two consecutive runs
- **CLI-02** `session list --json` — parses cleanly via `jq`
- **CLI-03 / CLI-04** `session show <tabId>` — returns `{success, tabId, sessionId, agentId, agentSessionId, messages[]}`; messages have `{id, role, source, content, timestamp}`
- **CLI-05** `session show <tabId> --tail 3` — returns 3 most-recent messages, oldest-first
- **CLI-07** `session show bogus-id-12345` — exits 1 with `{success: false, error: "Tab not found: …", code: "TAB_NOT_FOUND"}`, no stack trace leaked
- **CLI-08** `session show` (no arg) — exits 1 with `error: missing required argument 'tab-id'`

### Skipped
- **CLI-06** Maestro-not-running path — skipped to avoid disrupting the live session running this validation.

### Suggested follow-ups
1. Either align CLI output with the documented `{ id, history }` shape, **or** update the plan / forthcoming docs to match the as-built shape (`tabId`, `messages`, `--tail`).
2. If `session show` is intended to also support a non-JSON pretty mode, add it; otherwise drop the `--json` reference from external documentation since output is always JSON.
3. Consider a default text-mode column for `session list` that includes `state` and `createdAt` (currently JSON-only), since the plan implied those would be visible in the table view.

chr1syy:
@greptileai re-review please

PR #926, opened 2026-04-29
feat(usage-dashboard): Cue stats tab

pedramamini:
## Summary

Adds a **Cue** tab to the Usage Dashboard modal. The tab fuses Cue lineage with per-agent token data and surfaces totals, per-pipeline / per-agent / per-subscription rollups, a time series, and a chain-tree view. Implementation spans schema, ingestion, aggregation, IPC, and renderer.

- **Schema (Phase 01):** three additive columns on `cue_events` — `pipeline_id`, `chain_root_id`, `parent_event_id` — populated by an idempotent `PRAGMA table_info` migration. Lineage is plumbed through `cue-engine` → `cue-run-manager` → `cue-dispatch-service` → `cue-completion-service` → `cue-fan-in-tracker` → `cue-queue-persistence`. Writes are gated on `encoreFeatures.usageStats === true` (lineage stays NULL when off).
- **Token attribution (Phase 02):** new `src/main/cue/stats/cue-token-accessor.ts` reads per-agent storage (claude-code, codex, opencode, factory-droid, copilot-cli) behind a 30s cache. Audit notes in `src/main/cue/stats/AGENT-TOKEN-AUDIT.md`.
- **Aggregation (Phase 03):** `getCueStatsAggregation` in `src/main/cue/stats/cue-stats-query.ts`, shared types in `src/shared/cue-stats-types.ts`, IPC handler at `src/main/ipc/handlers/cue-stats.ts`, preload bridge at `src/main/preload/cueStats.ts`. Handler throws `'CueStatsDisabled'` when either Encore flag is off so the renderer can render the friendly disabled note.
- **Renderer (Phase 04):** `src/renderer/components/UsageDashboard/CueStats.tsx` — summary cards, time-series chart, group tables, agent token chart, chain list, coverage warnings, skeleton, disabled, error, and empty states. Tab wired into `UsageDashboardModal.tsx`.

## Review fixes (post-playbook)

A self-review surfaced two production bugs that the playbook tests didn't catch — both fixed in this PR before pushing:

1. `'CueStatsDisabled'` equality check failed against Electron's wrapped IPC error (`Error invoking remote method '...': Error: CueStatsDisabled`). The renderer rendered the generic error/retry note instead of the disabled state. Switched to substring detection and normalized the error sentinel; test now exercises both bare and wrapped forms.
2. Cue tab visibility was gated on `encoreFeatures.maestroCue` only, but the IPC handler requires both `maestroCue` AND `usageStats`. Fixed the renderer to match handler gating; integration test now covers all three off-flag permutations.

## Known follow-ups

- `getRecentCueEvents` has no LIMIT on `timeRange === 'all'`. Acceptable for now; worth a follow-up to cap and surface a "results truncated" coverage warning if `cue_events` grows large.
- Token-accessor cache has TTL but no explicit eviction; entries linger until next read for that session ID. Low-priority memory drift, fine for current usage.
- `flattenChain` walks parent pointers without a `visited` set. UUIDs make a real cycle near-impossible, but defense-in-depth would be cheap.
- Manual smoke test in Electron with both flags toggling — deferred (see Phase 05 doc); needs a human to drive the GUI.

## Test plan

- [ ] Both flags on: Cue tab visible, populated state renders summary cards / time-series / pipeline table / agent chart / subscription table / chain list
- [ ] `maestroCue=false` OR `usageStats=false` OR both off: Cue tab hidden, base four tabs still work
- [ ] Toggle a flag off while Cue tab is active: falls back to Overview without crash
- [ ] `cue_events` migration runs cleanly on an existing DB (no errors, columns added)
- [ ] Lineage NULL on rows written when `usageStats` was off — aggregation groups them as `Unattributed` instead of breaking
- [ ] No console errors in DevTools during any of the above

## Summary by CodeRabbit

* **New Features**
  * Cue analytics dashboard and gated “Cue” tab with time-series, pipeline/agent/chain views, coverage warnings, export, and new charts (radial, year-in-pixels).
  * Local Cue telemetry outbox with batched submission, backoff, and fire-and-forget flush on autorun/quit.
  * Worktree-aware usage metrics, Agent Overview cards, Worktree analytics, and Achievement share button (copy/save image).
  * New header settings to show Session ID / Cost pills and worktree display options.

* **Behavior Changes**
  * Run lineage and pipeline attribution persisted when usage stats are enabled; recorded as null when disabled.

* **Tests**
  * Extensive new and updated test suites for stats, lineage, telemetry, token attribution, persistence, and UI.

* **Documentation**
  * New Cue telemetry and agent token audit docs.


pedramamini:
@coderabbitai review

@greptileai review

Merged latest `rc` and addressed the prior round of feedback in commit 8fd2a78d (+ a follow-up sparkline anchor fix in c8441f5b). Highlights:

**Major fixes**
- `preload/cueStats.ts`: normalize Electron's wrapped IPC error so renderers receive the bare `CueStatsDisabled` sentinel — no more substring-match dance at every callsite.
- `ipc/handlers/cue-stats.ts`: run the disabled-flag gate outside `withIpcErrorLogging` so `CueStatsDisabled` doesn't pollute Sentry/log output for the expected recovery path.
- `cue-telemetry.ts`: guard `countTelemetryEvents()` in a try/catch so DB failures cannot bubble out of `recordTriggerFired` / `recordRunCompleted` (telemetry must stay non-throwing); fix the over-payload overflow path to drop the OLDEST half of the batch (`getTelemetryBatch` returns oldest-first, so the previous logic was discarding the freshest data).
- `quit-handler.ts` + `stats.ts`: route telemetry-flush rejections through Sentry (`captureException`) instead of empty catches / unhandled promises.
- `history-manager.ts` + `marketplace.ts`: capture unexpected watcher errors in Sentry; keep ENOENT/EPERM/UNKNOWN warn-only.
- `AchievementShareButton`: send copy/download failures to Sentry instead of swallowing.

**Chart fixes (worktree-aware aggregations)**
- `AgentComparisonChart` + `AgentEfficiencyChart`: backfill historical totals from `data.byAgent` so deleted/closed agents stay on the chart once worktree splitting kicks in.
- `AgentOverviewCards`: only fall back to the provider total when this is the sole visible session for that provider; render `0` otherwise to stop duplicating the same number across multiple cards. Regression test added.
- `chartUtils.findSessionByStatId`: delimiter-aware longest-prefix match so `sess-1` no longer shadows `sess-10`. Regression tests added.
- `SummaryCards`: densify the last-7-days sparkline window (anchored on the latest date in `byDay`) so absent days zero-fill instead of compressing.
- `AutoRunStats`: tooltip now says "tasks attempted" (matches the `count` field) and guards divide-by-zero.
- `DurationTrendsChart`: tooltip width/whitespace now align with the clamp assumptions.

**Minor / quick wins**
- `SessionItem` rename input commits through `onBlur` on Enter to prevent IPC double-fire.
- `ActivityHeatmap` heading uses `.card-enter` class so `prefers-reduced-motion` can disable it.
- `YearInPixelsStrip`: import `MouseEvent`/`KeyboardEvent`/`FocusEvent` types directly (modern JSX transform); cells now `tabIndex={0}` with focus/key handlers so keyboard users can read the per-day tooltip.
- `UsageDashboardModal`: empty-state gate now uses the filtered (non-terminal) sessions list.
- `CueStats` Retry button: explicit `type="button"`.
- `index.css`: lowercase `currentcolor`.
- `settingsStore.getSettingsActions()`: expose the new pill setters so non-React callers can toggle them.

**Known follow-ups deferred (not blocking)**
- `cue-stats-query.getCueStatsAggregation` still has no LIMIT on `timeRange === 'all'` — heavy-lift fix tracked for a follow-up.
- The two CodeRabbit comments about test selectors (`cue-stats-chain*`, `cue-stats-agent-table`) were obsoleted by the diagnostics-refocus commit 5fc2b89 — the agent panel is still a chart with `data-testid="cue-stats-agent-chart"` and the chains list was intentionally removed (see new guard test at `CueStats.test.tsx:482-494`).

Lint clean (`npm run lint` + `npm run lint:eslint`). Vitest suite passes (768 files, 1 pre-existing unrelated AutoRun act() warning).

pedramamini:
@greptile re-review

PR #931, opened 2026-04-30
fix(worktrees): keep nested worktrees from disappearing on restart

chr1syy:
## Summary

- `scanWorktreeDirectory` only read immediate children of the configured `basePath`, so worktrees from slash-named branches (e.g. `fix/worktree-removal` → `<basePath>/fix/worktree-removal`) never appeared in `gitSubdirs`. On every app restart, `scanWorktreeConfigs` then bulk-removed those nested children because their `cwd` was absent from the scan.
- Recurse one level into non-git subdirs (configurable `MAX_DEPTH = 1`) so the common `<basePath>/<group>/<branch>` layout is covered. Errors while reading a nested group dir are swallowed locally and do **not** flag `scanFailed` — that flag is reserved for the top-level read so the renderer's removal-skip fallback continues to work.
- Bump the chokidar watcher from `depth: 0` to `depth: 1` so live `addDir` / `unlinkDir` events fire for nested worktrees too. The existing `is-inside-work-tree` + `show-toplevel` checks in the addDir handler correctly reject the intermediate group dir (e.g. `fix/`).

## Repro

1. Create a worktree from a slash-named branch: `git worktree add <basePath>/fix/foo fix/foo`.
2. Open the agent in Maestro so the child session is registered.
3. Restart the app. The session for `fix/foo` disappears even though the worktree is still on disk.

After this PR, the session survives the restart (and external `git worktree add`/`remove` for nested branches now triggers the live listener).

## Test plan

- [x] `vitest run src/__tests__/main/ipc/handlers/git.test.ts` (156 tests, all green) — includes the existing scan/watcher coverage plus three new cases:
  - nested worktrees from slash-named branches are discovered alongside flat ones
  - recursion stops at one level (`<basePath>/a/b/c` is **not** scanned)
  - read errors on a nested group dir don't fail the whole scan or set `scanFailed`
- [x] `vitest run src/__tests__/renderer/hooks/useWorktreeHandlers.test.ts` (60 tests)
- [x] `vitest run src/__tests__/main/preload/git.test.ts src/__tests__/renderer/components/WorktreeRunSection.test.tsx src/__tests__/renderer/components/AppWorktreeModals.test.tsx`
- [x] Updated the existing watcher test to expect `depth: 1` (was the only spec asserting on the old value).

## Summary by CodeRabbit

* **New Features**
  * Discover nested worktrees in subdirectories, including branch names that contain slashes.

* **Bug Fixes**
  * Scan tolerates partial directory read errors and still returns available worktrees; top-level remote read failures mark scans as failed.
  * Enforced one-level recursion to avoid deeper nested discovery.
  * File-watcher now monitors one directory level deeper to detect nested worktree changes.

* **Tests**
  * Expanded coverage for nested discovery, recursion limits, SSH/read failures, and watcher behavior.

### DIFF
--- a/src/__tests__/main/ipc/handlers/git.test.ts
+++ b/src/__tests__/main/ipc/handlers/git.test.ts
@@ -4178,6 +4178,181 @@ branch refs/heads/bugfix-123
 			expect(result.gitSubdirs[0].branch).toBe('feature-branch');
 			expect(result.scanFailed).toBeFalsy();
 		});
+
+		it('should discover nested worktrees from slash-named branches', async () => {
+			// Regression: branches like "fix/worktree-removal" produce a nested
+			// path <basePath>/fix/worktree-removal. Without one-level recursion,
+			// the scan misses these and the renderer wrongly removes the session.
+			vi.mocked(mockFs.readdir).mockImplementation(async (dir: any) => {
+				if (String(dir) === '/parent') {
+					return [
+						// Flat worktree (existing happy path)
+						{ name: 'flat-branch', isDirectory: () => true },
+						// Group directory containing nested worktrees
+						{ name: 'fix', isDirectory: () => true },
+					] as any;
+				}
+				if (String(dir) === path.join('/parent', 'fix')) {
+					return [
+						{ name: 'worktree-removal', isDirectory: () => true },
+						{ name: 'files-restart', isDirectory: () => true },
+					] as any;
+				}
+				return [] as any;
+			});
+
+			vi.mocked(execFile.execFileNoThrow).mockImplementation(async (cmd, args, cwd) => {
+				const cwdStr = String(cwd);
+
+				// Group directory itself is NOT a git repo
+				if (cwdStr === path.join('/parent', 'fix')) {
+					if (args?.includes('--is-inside-work-tree')) {
+						return { stdout: '', stderr: 'fatal: not a git repository', exitCode: 128 };
+					}
+				}
+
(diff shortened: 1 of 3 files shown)


pedramamini:
Thanks for the thoughtful fix, @chr1syy! Reviewed the PR end-to-end:

- Repro case is real and the one-level recursion is the right minimal fix for slash-named branches.
- The refactor of `scanWorktreeDirectory` into `readSubdirs` / `inspectSubdir` / `scanLevel` is clean, and the SSH top-level failure path correctly throws (with `code='ENOENT'`) so the outer catch still sets `scanFailed: true` — pinned by the new `should set scanFailed when SSH readDirRemote fails at the top level` test.
- Nested-read errors are appropriately swallowed, and the comment explaining why we forward `unlinkDir` for the intermediate group dir (renderer filters by registered child cwds) is exactly the right call-out.
- Re-ran `vitest run src/__tests__/main/ipc/handlers/git.test.ts` (157 ✅) and `src/__tests__/renderer/hooks/useWorktreeHandlers.test.ts` (61 ✅) locally — all green.

Greptile's P1 about SSH `scanFailed` not being set is a misread of the new control flow (the throw propagates up through `scanLevel`'s top-level `await readSubdirs(dir)` since only the recursive call is wrapped). CodeRabbit's nitpick about a redundant `--show-toplevel` is already addressed at line 1213 (`repoRoot = toplevel`).

Looks good — approving and merging.

PR #921, opened 2026-04-28
feat(web): mobile parity additions for tokens, cache, and worktree branches

HumpfTech:
## Summary

- include reasoning token counts in the mobile session token display
- show cache read/write token counts in the mobile token tooltip
- add worktree branch badges to mobile session pills
- bound the web WebSocket deduplication cache with LRU-style eviction
- include a Prettier-only cleanup commit for existing release-note format drift

## Validation

Reported from pre-push validation on the final branch:

- `npm run lint`
- `npm run lint:eslint`
- `npm run format:check:all`
- `npm run test` — 22,528 passed / 22,635 total, 107 skipped

## Notes

Playwright web coverage and web e2e scripts are intentionally not included in this upstream branch; this PR contains only the production web/mobile changes.

## Summary by CodeRabbit

* **New Features**
  * Mobile: execution-queue panel with move/remove controls, worktree branch badges on session pills, and expanded token/cycle display
  * Message history: tool execution entries with dedicated styling and status indicators

* **Real-time & Sync**
  * Web/preload/websocket broadcasts for git status, execution-queue, and tool execution; remote commands now enqueue rather than drop; session state includes token/timing fields

* **Performance**
  * Bounded, recency-aware message deduplication

* **Documentation**
  * Standardized release notes formatting

* **Tests**
  * Updated tests and mocks to cover queue/remote behaviors and command flow

PR #895, opened 2026-04-24
feat(web): bring AutoRun on WebUI/mobile to parity with desktop

chr1syy:
## Summary

- Closes the AutoRun parity gap between the desktop and the WebUI/mobile interface.
- Adds **playbook CRUD, per-doc task reset, error-pause recovery (resume / skip / abort), subfolder tree view, lock state, in-document search, and undo/redo** to the mobile Auto Run surface.
- All routes reuse the existing \`window.maestro.autorun.*\` / \`window.maestro.playbooks.*\` IPC, so SSH remote sessions work transparently without special-casing.

### Server / WebSocket (\`src/main/web-server/\`)
- 9 new message types with input validation:
  - \`list_playbooks\`, \`create_playbook\`, \`update_playbook\`, \`delete_playbook\`
  - \`reset_auto_run_doc_tasks\`
  - \`resume_auto_run_error\`, \`skip_auto_run_document\`, \`abort_auto_run_error\`
- \`AutoRunState\` extended with \`errorPaused\` / \`errorMessage\` / \`errorType\` / \`errorRecoverable\` / \`errorDocumentIndex\` / \`errorTaskDescription\`.
- \`broadcastAutoRunState\` now forwards the error fields so mobile can show recovery UI without a second roundtrip.
- Callbacks threaded through \`CallbackRegistry\`, \`WebServer\`, and \`web-server-factory.ts\` (new \`remoteRequest\` helper for request-response IPC).

### Renderer
- \`useBatchHandlers\` re-exports raw \`resumeAfterError\` / \`skipCurrentDocument\` / \`abortBatchOnError\` (sessionId-targeted) so web-triggered recovery can target a specific session.
- \`useAppRemoteEventListeners\` handles the new remote events.
- \`getAutoRunDocs\` now populates a \`folder\` field for tree rendering.

### Web / mobile
- \`useAutoRun\` hook gains playbook CRUD + reset/error-recovery methods.
- \`AutoRunPanel\` groups documents by subfolder, shows a lock badge while running, and exposes per-doc **Reset** buttons.
- \`AutoRunSetupSheet\` adds a collapsible **Playbooks** panel with load / save / update / delete and modification detection (switches the Save button between Create / Update).
- \`AutoRunIndicator\` switches to a column layout that, when \`errorPaused\`, renders the error context plus **Resume / Skip / Abort** buttons (Resume hidden when \`errorRecoverable === false\`).
[...]

PR #933, opened 2026-04-30
fix(web): honor 'All' for Max Output Lines per Response on mobile/web

chr1syy:
## Summary

- The web/mobile `MessageHistory` had an unconditional 500-char truncation that fired regardless of the user's "Max Output Lines per Response" choice. Selecting **All** still collapsed any AI response longer than ~500 chars with a "tap to expand" hint.
- The dead `LINE_TRUNCATE_THRESHOLD = 8` fallback was also unreachable behind a `!hasLineCap` early-return.
- Drop the char-based threshold and the dead fallback so the web UI matches desktop `TerminalOutput.tsx:350` exactly: **All (Infinity / unset) disables truncation entirely**; a finite cap drives line-based collapse only.

## Why now

Originally introduced in `725491d` ("port Max Output Lines setting"). Commit message stated the intent — "MessageHistory now applies line-based truncation only when a finite cap is configured" — but the legacy 500-char path was left behind, so "All" never actually meant All.

Verified this is **not** caused by the soaked PRs (#895, #921, #923) — none of them touch the truncation logic. PR #921 only appends a `ToolCard` helper to `MessageHistory.tsx`.

## Code change

`src/web/mobile/MessageHistory.tsx` — `shouldTruncate` / `getTruncatedText` short-circuit on `!hasLineCap` and otherwise gate on line count alone. `CHAR_TRUNCATE_THRESHOLD` and the unreachable `LINE_TRUNCATE_THRESHOLD` fallback are removed.

## Test plan

- [x] Replace char-based truncation assertions with line-based equivalents.
- [x] Add regression: long single-line content + `maxOutputLines={Infinity}` → no expand hint.
- [x] Add regression: 15-line content + `maxOutputLines={Infinity}` → no expand hint.
- [x] Add regression: 600-char single-line + finite cap → no expand hint (only line counts gate truncation).
- [x] Add regression: omitted `maxOutputLines` defaults to "All" — no truncation.
- [x] `npx vitest run src/__tests__/web/mobile/MessageHistory.test.tsx` — 81/81 pass.
- [x] `npx tsc --noEmit -p tsconfig.json` and `-p tsconfig.lint.json` — clean.
- [x] `npx prettier --check` and `npx eslint` on the touched files — clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed message truncation to use line-count instead of character-based truncation
  * Fixed "All" option to properly disable truncation regardless of message length

* **Tests**
  * Refactored test suite to validate line-count truncation behavior

### DIFF
--- a/src/__tests__/web/mobile/MessageHistory.test.tsx
+++ b/src/__tests__/web/mobile/MessageHistory.test.tsx
@@ -240,47 +240,77 @@ describe('MessageHistory', () => {
 	});
 
 	describe('Message Truncation', () => {
-		const longText = 'A'.repeat(600); // > CHAR_TRUNCATE_THRESHOLD (500)
-		const manyLines = 'Line\n'.repeat(15); // > LINE_TRUNCATE_THRESHOLD (8)
+		// 15 newline-separated lines; > any finite cap we exercise below.
+		const manyLines = Array.from({ length: 15 }, (_, i) => `Line ${i + 1}`).join('\n');
+		// 600 characters on a single line — used to assert that char length alone
+		// does NOT trigger truncation (matches desktop TerminalOutput behavior).
+		const longSingleLine = 'A'.repeat(600);
 
-		it('truncates text longer than 500 characters', () => {
-			const logs: LogEntry[] = [createLogEntry({ text: longText })];
-			render(<MessageHistory logs={logs} inputMode="ai" />);
-			expect(screen.getByText('▶ expand')).toBeInTheDocument();
-		});
-
-		it('truncates text with more than 8 lines when a finite line cap is configured', () => {
+		it('truncates by lines when line count exceeds the configured cap', () => {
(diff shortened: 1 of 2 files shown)


pedramamini:
Thanks for the fix, @chr1syy! 🙏

Reviewed and this is a clean, well-targeted change:
- Removing the 500-char truncation correctly restores "All" semantics on web/mobile and matches `TerminalOutput.tsx` desktop behavior.
- Dropping the unreachable `LINE_TRUNCATE_THRESHOLD` fallback (gated behind the earlier `!hasLineCap` return) is good dead-code hygiene.
- Test coverage is thorough — Infinity, omitted prop, finite cap, and the char-length-alone regression all have explicit assertions.
- CI, CodeRabbit, and Greptile are all green; no merge conflicts.

Approving. ✅

PR #934, opened 2026-04-30
fix(renderer): prevent renderer-unloading nav from inline-code in <a>

chr1syy:
## Summary

When the AI emits a markdown link of the form `` [`file.md`](file.md) ``, the rendered `<a>` contains a clickable `<code>`. The inline-code click handler at `src/renderer/utils/inlineCodeCopy.ts` called `e.stopPropagation()` but not `e.preventDefault()`, so:

1. The `<code>` handler ran (copy + "Copied to Clipboard" flash, propagation stopped).
2. The outer `<a>`'s React `onClick` (which would have called `preventDefault()`) never fired.
3. The browser's default link navigation **still** ran, resolving the relative href against the current page (`file:///opt/Maestro/resources/app.asar/dist/renderer/index.html`).
4. Electron's `will-navigate` allowed any `file://` URL inside the renderer directory through, so the renderer navigated to a non-existent in-bundle file and unloaded the app — appearing as an "Electron restart" on Linux right after the "Copied" flash.

## Changes

- **`src/renderer/utils/inlineCodeCopy.ts`** — `buildInlineCodeHandlers` now calls `e.preventDefault()` alongside `e.stopPropagation()`. No-op for standalone `<code>`, decisive for `<code>` nested inside `<a>`.
- **`src/main/app-lifecycle/window-manager.ts`** — defense-in-depth: tightened `will-navigate` so production only allows navigation to the renderer entry HTML itself (`pathToFileURL(rendererPath)`), not any file inside the renderer directory. A future stray relative `<a>` can no longer unload the app.
- **Tests** — new `src/__tests__/renderer/utils/inlineCodeCopy.test.tsx` with the regression case (clicking nested `<code>` inside `<a>` must mark the click `defaultPrevented`); updated `src/__tests__/main/app-lifecycle/window-manager.test.ts` with a regression case asserting `file:///path/to/TEST-PLAN-0.16.15-RC.md` is now blocked.

## Test plan

- [x] `npx vitest run src/__tests__/renderer/utils/inlineCodeCopy.test.tsx` — 9 pass
- [x] `npx vitest run src/__tests__/main/app-lifecycle/window-manager.test.ts` — 30 pass (new regression case included)
- [x] `npx vitest run src/__tests__/renderer/components/MarkdownRenderer.test.tsx src/__tests__/renderer/utils/markdownConfig.test.ts` — 219 pass (consumers unaffected)
- [x] `npx prettier --check` on all changed files — clean
- [x] `npx eslint` on all changed source files — 0 errors / 0 warnings
- [ ] Manual repro on Linux: click an AI-emitted `` [`file.md`](file.md) `` link where the file isn't in the project tree — should no longer reload the app

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed inline code copy when code is nested inside links—clicking now copies without triggering navigation.
  * Hardened renderer navigation: only the exact renderer entry URL is allowed; non-entry file:// navigation and mismatched/modified entry URLs are blocked, including stricter dev-server-origin checks.

* **Tests**
  * Added comprehensive tests for inline code copy behavior (click and keyboard) and navigation regressions, including handler registration and repeated-call behavior.

### DIFF
--- a/src/__tests__/main/app-lifecycle/window-manager.test.ts
+++ b/src/__tests__/main/app-lifecycle/window-manager.test.ts
@@ -624,7 +624,7 @@ describe('app-lifecycle/window-manager', () => {
 			expect(mockEvent.preventDefault).toHaveBeenCalled();
 		});
 
-		it('should allow file:// navigation within renderer directory in production', async () => {
+		it('should allow file:// navigation to the renderer entry HTML in production', async () => {
 			const { createWindowManager } = await import('../../../main/app-lifecycle/window-manager');
 
 			const windowManager = createWindowManager({
@@ -646,7 +646,7 @@ describe('app-lifecycle/window-manager', () => {
 			);
 			const navigateHandler = willNavigateCall![1];
 
-			// Should allow file:// navigation within the renderer's directory (/path/to/)
+			// Should allow file:// navigation to the renderer entry HTML itself.
 			const mockEvent = { preventDefault: vi.fn() };
 			navigateHandler(mockEvent, 'file:///path/to/index.html');
 			expect(mockEvent.preventDefault).not.toHaveBeenCalled();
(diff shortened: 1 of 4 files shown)


pedramamini:
Thanks for the fix, @chr1syy! Nicely diagnosed — the `stopPropagation` without `preventDefault` interaction with Electron's `will-navigate` allowing any file URL inside the renderer directory was a sneaky one. The two-layer fix (precise click handling + tightened exact-URL match in `will-navigate`) is exactly the right shape, and the regression tests for both layers make this safe to land. Approving.

PR #935, opened 2026-04-30
fix(updater): bypass busy-agent quit gate so Windows installer isn't orphaned

chr1syy:
## Summary

- On Windows, `autoUpdater.quitAndInstall(false, true)` spawns the NSIS installer bound to our PID **before** `app.quit()` fires. The existing `before-quit` handler intercepts the quit to ask the renderer about busy agents (`event.preventDefault()`), which leaves the spawned installer waiting for a parent exit — only the 5s safety timeout rescues it, and unreliably.
- The user already opted in via `UpdateCheckModal`, so the busy-agent confirmation is redundant on the install path. We now mark the quit pre-confirmed before calling `quitAndInstall`, using the existing `QuitHandler.confirmQuit()` escape hatch.

## Implementation

- `src/main/auto-updater.ts` — `initAutoUpdater` accepts `{ onBeforeQuitAndInstall? }`. The hook fires inside the `updates:install` handler before `quitAndInstall(false, true)`. Failures in the hook are logged and don't block the install. Added `__setAutoUpdaterForTesting` (test-only) so vitest can inject a mock past the dynamic `require('electron-updater')` that bypasses the module mocker.
- `src/main/app-lifecycle/window-manager.ts` — new optional `getConfirmQuit` dep on `WindowManagerDependencies`; passed into `initAutoUpdater` as `onBeforeQuitAndInstall`. Lazy because the quit handler is constructed after the window.
- `src/main/index.ts` — hoist a `let quitHandler: QuitHandler | null = null` forward declaration above `createWindowManager`, wire `getConfirmQuit: () => quitHandler?.confirmQuit`. The closure resolves at install-click time, by which point the handler exists.

## Test plan

- [x] `npx vitest run src/__tests__/main/auto-updater.test.ts src/__tests__/main/app-lifecycle/window-manager.test.ts src/__tests__/main/app-lifecycle/quit-handler.test.ts` — 55/55 pass
- [x] Type check (`npm run lint`) clean for changed files (pre-existing failures elsewhere are unrelated)
- [x] Prettier and ESLint clean
- [ ] Manual Windows verification: trigger an update install, confirm the app exits without the 5s safety-timeout fallback firing and the installer takes over cleanly
- [ ] Manual macOS / Linux smoke: confirm install path still works (no functional change expected on those platforms)

## Summary by CodeRabbit

* **New Features**
  * Auto-updater now invokes the app's quit-confirmation hook (if present) before installing updates.

* **Bug Fixes**
  * Installation proceeds safely even if the quit-confirmation hook throws or is absent; errors are reported without blocking install.

* **Tests**
  * Added tests covering auto-updater installation behavior, quit-confirmation integration, and error-reporting edge cases.

### DIFF
--- a/src/__tests__/main/app-lifecycle/window-manager.test.ts
+++ b/src/__tests__/main/app-lifecycle/window-manager.test.ts
@@ -442,6 +442,66 @@ describe('app-lifecycle/window-manager', () => {
 			expect(mockInitAutoUpdater).toHaveBeenCalled();
 		});
 
+		it('passes onBeforeQuitAndInstall to auto-updater that invokes confirmQuit', async () => {
+			const { createWindowManager } = await import('../../../main/app-lifecycle/window-manager');
+			const confirmQuit = vi.fn();
+			const getConfirmQuit = vi.fn(() => confirmQuit);
+
+			const windowManager = createWindowManager({
+				windowStateStore: mockWindowStateStore as unknown as Parameters<
+					typeof createWindowManager
+				>[0]['windowStateStore'],
+				isDevelopment: false,
+				preloadPath: '/path/to/preload.js',
+				rendererPath: '/path/to/index.html',
+				devServerUrl: 'http://localhost:5173',
+				useNativeTitleBar: false,
+				autoHideMenuBar: false,
+				getConfirmQuit,
+			});
+
+			windowManager.createWindow();
+
+			// initAutoUpdater(window, options)
+			expect(mockInitAutoUpdater).toHaveBeenCalledWith(
+				expect.anything(),
+				expect.objectContaining({ onBeforeQuitAndInstall: expect.any(Function) })
+			);
+
+			// Pull the option and confirm it routes to confirmQuit
+			const options = mockInitAutoUpdater.mock.calls[0][1] as {
+				onBeforeQuitAndInstall: () => void;
+			};
+			options.onBeforeQuitAndInstall();
+
+			expect(getConfirmQuit).toHaveBeenCalled();
+			expect(confirmQuit).toHaveBeenCalledTimes(1);
+		});
+
+		it('onBeforeQuitAndInstall is a no-op when quit handler is not yet wired', async () => {
+			const { createWindowManager } = await import('../../../main/app-lifecycle/window-manager');
+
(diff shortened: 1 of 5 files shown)


chr1syy:
## Review feedback addressed (06cd5ee2)

**CodeRabbit — Sentry reporting on hook failures** ✅
`onBeforeQuitAndInstall` failures now flow to \`captureException\` with \`{ module: 'AutoUpdater', hook: 'onBeforeQuitAndInstall', operation: 'updates:install' }\` context. The install still proceeds — telemetry no longer drops on the floor. Added a regression test plus a non-\`Error\` throw case (string thrown → wrapped before reporting).

**Greptile P2 — Test-only export leakage** ✅
\`__setAutoUpdaterForTesting\` body is now hard-gated behind \`process.env.NODE_ENV !== 'production'\`. The symbol still has to exist for the test import (TS doesn't support conditional exports), but a stray call in a production bundle is a no-op and cannot subvert the real updater singleton.

**Greptile P2 — \`onBeforeQuitAndInstall\` overwrite shadowed by IPC-guard** ↩️ no change
This is flagged as "no regression from this PR" and the analysis isn't quite right: the install handler reads the *module-level* \`onBeforeQuitAndInstall\` \`let\` (not a closure capture from setupIpcHandlers). Updates to that variable on a re-init *would* be picked up. \`initAutoUpdater\` is also only called once today (\`window-manager.ts\` runs it inside \`createWindow\`, which the rest of the app calls a single time). Fixing the singleton pattern more broadly is out of scope for this PR.

Tests: 56/56 pass across \`auto-updater\`, \`window-manager\`, and \`quit-handler\` suites.

pedramamini:

Approach is clean — using `QuitHandler.confirmQuit()` to pre-set `quitConfirmed` is the right escape hatch here, and the lazy `getConfirmQuit` getter neatly sidesteps the construction-order issue between `windowManager` and `quitHandler`. The follow-up commit addressing CodeRabbit (Sentry capture on hook failure with contextual tags + non-Error throw coverage) and Greptile (gating `__setAutoUpdaterForTesting` behind `NODE_ENV !== 'production'`) tightened things up nicely. Test coverage on all three branches (success, missing hook, hook throws) is solid.

Approved. ✅

PR #936, opened 2026-05-01
perf: hot-path wins (persistence diff + Cue idle + stats batch)

reachrazamair:
## Summary

Three coherent perf bundles + one Cue UX polish, all on the streaming hot path:

- **session persistence + categorization:** stop cloning the entire sessions tree on every change, replace per-session `JSON.stringify(cliActivity)` with a shallow primitive compare, and collapse the four chained `useMemo`s in `useSessionCategories` into one pass.
- **background work hygiene:** Cue scanners (file watcher, task scanner, GitHub poller) pause when the app is hidden via a new `cue:setActive` IPC; stats query-event writes batch to SQLite in 50-event / 500ms windows wrapped in a single transaction.
- **Cue UX polish:** restore the drag-preview line when connecting two pipeline nodes — works in both pan and select canvas modes now.

12 commits. 26,946+ tests pass. `tsc` clean. `eslint` clean.

### `sessions:setMany` (the load-bearing change)

- **Before:** every debounced flush mapped `prepareSessionForPersistence` over **all** sessions and shipped the full tree across IPC. CLAUDE-PERFORMANCE.md observed >500 MB short-lived heap churn from this pattern.
- **After:** first flush after load uses `setAll` to seed the main process and capture a baseline. Subsequent flushes diff `sessions` against the baseline by reference equality per element (Zustand's immutable update pattern guarantees mutated sessions get a fresh reference) and ship only the changed subset + tombstone ids via the new `setMany` IPC handler.
- **IPC payload:** O(N) → O(dirty), typically 1–2 of 30+ sessions during streaming.
- **Why hook-level diff and not store-level dirty tracking:** 200+ existing `setSessions(prev => prev.map(...))` call sites use the functional-updater form. Wrapping every mutator was invasive and risky; the diff captures the same information in one place with no caller changes.

Commits: `315eeb402` (characterization), `03a9503be` (handler), `5aa3eb60b` (renderer switch).

### cliActivity shallow compare

- **Before:** `JSON.stringify(prev.cliActivity) !== JSON.stringify(curr.cliActivity)` per session per flush. With 30+ sessions × 200ms cadence during streaming, thousands of stringify ops per minute.
- **After:** 4-step primitive comparison. The producer (`useCliActivityMonitoring`) only ever sets the field to `undefined` or `{ playbookId, playbookName, startedAt }`, so a field-by-field check is equivalent at every real call site.
- **Per-session diff cost:** O(stringify ~100B × 2) → O(1).

Commit: `01a3063df`.

### `useSessionCategories` memo consolidation

- **Before:** four `useMemo`s daisy-chained, each invalidated by the previous. Every session mutation triggered four cascading recomputes.
- **After:** one `useMemo` keyed on `[sessions, sortedSessions]`, computes all three Maps in one pass. `getWorktreeChildren` stays as a `useCallback` — same identity contract, same output.
- **Render cost:** session-list traversal cost drops roughly proportional to the number of memos eliminated.

Commit: `f5eedbf1d`.

### Cue scanners visibility-aware pause

- **Before:** `cue-task-scanner`, `cue-file-watcher`, and `cue-github-poller` ran unconditionally regardless of whether the Maestro window was visible. Task scanner did per-minute directory walks + markdown reads. GitHub poller hit `gh` CLI every 5 min. Result: laptop fans spinning while users weren't looking.
- **After:** module-level `cueActive` flag (`cue-active-state.ts`) gates per-tick work in each scanner. New `useCueVisibilityWiring` hook in `App.tsx` forwards `document.visibilityState` via `cue:setActive` IPC. When inactive: task scanner skips the walk, GitHub poller skips the gh fetch, file watcher drops dispatch (chokidar subscription stays alive — OS file-watch is nearly free). Pause-don't-stop preserves clean resume on the next tick.
- **Default-active** is preserved when `isActive` is omitted, so CLI / tests / headless agents are unaffected.

Commits: `e8d58a591` (engine + scanners), `7f2c10295` (renderer wiring).

### Stats DB write batching

- **Before:** every `stats:record-query` IPC fired a single `stmt.run()` against SQLite. Many fsyncs per second on the streaming hot path.
- **After:** events buffered for 500ms or 50 events (whichever first). Flush wraps all buffered inserts in one `db.transaction(() => { ... })` — N WAL+fsync collapses into one. Same cadence as the existing logger / thinking-chunk batching landed in `e475c0bad`.
- Generated id returned synchronously so callers (which all `.catch()` and discard the id) don't wait for the disk write.
- `before-quit` hook flushes any in-flight buffer; events lost on flush failure (best-effort telemetry vs. infinite-retry risk).
- The process-listener path (auto-run batch mode, has its own retry logic) deliberately untouched.

Commit: `64f5c936f`.

## Cue pipeline UX — drag-preview line restored

The connection-line that follows the cursor while dragging from one node handle to another was effectively invisible. ReactFlow's default styling is `stroke: #b1b1b7` at 1px — washed out against our dark theme. Users saw the resulting edge appear on mouse-up with no visual cue during the drag itself. Compounded by `nodesConnectable` being gated on canvas pan/select mode, which broke the preview entirely in pan mode.

Fixes layered three ways:

1. **Custom `connectionLineComponent`** — `PipelineConnectionLine` paints its own SVG path: dashed bezier in `CUE_COLOR` at 2px with a small dot at the cursor tip. Bypasses any styling/specificity issues with the default render path.
2. **`connectionLineStyle` + `connectionLineType` props** as belt-and-suspenders for any path that falls back to the default ConnectionLine.
3. **CSS rule** on `.react-flow__connection-path` for the same belt-and-suspenders reason in case a future change reverts the prop.
4. **Decouple `nodesConnectable` from canvas mode.** Hand mode still disables node-body dragging (so left-drag pans through), but handles always work — matches n8n / Zapier / Figma's connector tool.

Commits: `ae6c28f1c`, `ba2e42a18`, `8f46c752b`, `8e195ddc1`.

## Risk profile

| Risk | Mitigation |
| --- | --- |
| Reference-diff misses a session mutation done in-place | Same constraint React.memo and Zustand selectors require — already enforced project-wide. Verified by 10 new tests covering update / add / remove / no-op / rapid-mutation / unmount paths. |
| `setAll` semantics change | First flush still calls `setAll` (preserves bootstrap path); 75 existing test assertions on `setAll` continue to pass. |
| `setMany` mishandles add/remove edge cases | 17 dedicated tests cover empty, single, mixed, conflict (remove wins), unseen-id, no-broadcast, error paths. |
| cliActivity comparison drifts from JSON.stringify behavior | 8 lock-in tests pin every (prev, curr) transition that must broadcast or stay silent — pass on both old and new code. |
| Memo consolidation changes output shape | 5 lock-in tests assert deep-equal Map content + reference identity across renders for a 10-session fixture — pass on both old and new code. |
| Cue scanners miss events during long-hidden windows | Pause-don't-stop preserves OS-level file watching. On resume, downstream re-scan paths catch up. The 24h GitHub prune timer is unaffected. |
| Stats buffer loses events on crash | Best-effort telemetry. The `before-quit` hook flushes on graceful exit. Hard crashes lose at most ~50 events / 500ms. |

## Test plan

- `npm run test` — **26,946 passed, 0 failed** (108 pre-existing skipped)
- `npm run lint` — clean across `tsconfig.lint.json`, `tsconfig.main.json`, `tsconfig.cli.json`
- `npm run lint:eslint` — clean
- +48 new tests across persistence (8 cliActivity diff, 17 setMany, 10 dirty-only flush, 5 categorization lock-in), Cue (5 active-state, 12 scanner gates, 6 visibility hook), and stats (15 buffer cases)

## Summary by CodeRabbit

* **New Features**
  * Cue now respects renderer visibility: background scanning pauses when the window is hidden and resumes when visible.
  * Incremental session persistence added to update/remove sessions more efficiently.

* **Improvements**
  * Query events are buffered and flushed in batches for more resilient analytics and shutdown handling.
  * Improved pipeline connection preview visuals and connection interaction affordances.

* **Tests**
  * Expanded coverage around visibility gating, incremental persistence, buffered query events, and related wiring.

### DIFF
--- a/src/__tests__/main/cue/cue-active-state.test.ts
+++ b/src/__tests__/main/cue/cue-active-state.test.ts
@@ -1,0 +1,48 @@
+/**
+ * Tests for the Cue active-state module.
+ *
+ * The visibility-aware pause that PR-B 1.4 wires into every scanner.
+ */
+
+import { describe, it, expect, beforeEach } from 'vitest';
+import {
+	isCueActive,
+	setCueActive,
+	resetCueActiveForTests,
+} from '../../../main/cue/cue-active-state';
+
+describe('cue-active-state', () => {
+	beforeEach(() => {
+		resetCueActiveForTests();
+	});
+
+	it('defaults to active', () => {
+		expect(isCueActive()).toBe(true);
+	});
+
+	it('flips to inactive when setCueActive(false) is called', () => {
+		setCueActive(false);
+		expect(isCueActive()).toBe(false);
+	});
+
+	it('flips back to active when setCueActive(true) is called', () => {
+		setCueActive(false);
+		setCueActive(true);
+		expect(isCueActive()).toBe(true);
+	});
+
+	it('is idempotent — repeated calls preserve the same state', () => {
+		setCueActive(false);
+		setCueActive(false);
+		expect(isCueActive()).toBe(false);
+		setCueActive(true);
+		setCueActive(true);
+		expect(isCueActive()).toBe(true);
+	});
+
+	it('resetCueActiveForTests restores the default', () => {
+		setCueActive(false);
+		resetCueActiveForTests();
+		expect(isCueActive()).toBe(true);
+	});
+});
--- a/src/__tests__/main/cue/cue-file-watcher.test.ts
+++ b/src/__tests__/main/cue/cue-file-watcher.test.ts
@@ -240,4 +240,84 @@ describe('cue-file-watcher', () => {
 
 		consoleSpy.mockRestore();
 	});
+
+	// PR-B 1.4: visibility-aware pause
+	describe('isActive gate', () => {
+		it('drops the debounced event when isActive returns false', () => {
+			const onEvent = vi.fn();
+			let active = false;
+
+			createCueFileWatcher({
+				watchGlob: '**/*.ts',
+				projectRoot: '/test',
+				debounceMs: 5000,
+				onEvent,
+				triggerName: 'test',
+				isActive: () => active,
+			});
+
+			const changeHandler = mockOn.mock.calls.find((call) => call[0] === 'change')?.[1];
+			changeHandler('src/index.ts');
+			vi.advanceTimersByTime(5000);
+
+			expect(onEvent).not.toHaveBeenCalled();
+		});
+
+		it('dispatches the next event after isActive flips to true', () => {
+			const onEvent = vi.fn();
+			let active = false;
+
+			createCueFileWatcher({
+				watchGlob: '**/*.ts',
+				projectRoot: '/test',
+				debounceMs: 5000,
+				onEvent,
+				triggerName: 'test',
+				isActive: () => active,
+			});
+
+			const changeHandler = mockOn.mock.calls.find((call) => call[0] === 'change')?.[1];
+			changeHandler('src/inactive.ts');
+			vi.advanceTimersByTime(5000);
+			expect(onEvent).not.toHaveBeenCalled();
+
+			active = true;
+			changeHandler('src/active.ts');
+			vi.advanceTimersByTime(5000);
+			expect(onEvent).toHaveBeenCalledTimes(1);
+		});
+
+		it('still subscribes to chokidar even when isActive is false (no setup teardown)', () => {
+			createCueFileWatcher({
+				watchGlob: '**/*.ts',
+				projectRoot: '/test',
+				debounceMs: 5000,
+				onEvent: vi.fn(),
+				triggerName: 'test',
+				isActive: () => false,
+			});
+
+			const registeredEvents = mockOn.mock.calls.map((call) => call[0]);
+			expect(registeredEvents).toContain('change');
+			expect(registeredEvents).toContain('add');
+			expect(registeredEvents).toContain('unlink');
+		});
+
+		it('defaults to always-active when isActive is omitted', () => {
+			const onEvent = vi.fn();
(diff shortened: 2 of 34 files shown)


pedramamini:
@reachrazamair thanks so much for the thorough work here — the perf bundling is exactly the kind of hot-path surgery that pays off, and the test coverage on the persistence diff path is excellent. Verified locally that the major automated-review feedback is all addressed in `e5d0d1a` (baseline-on-success for `setAll`/`setMany`, payload validation on `cue:setActive`, Sentry captures wired in the right places, dev-local path scrubbed). I also chased down CodeRabbit's "critical" `is_remote` flag — it's a false positive, migration v2 in `src/main/stats/migrations.ts` adds the column via `ALTER TABLE` so the INSERT is fine.

One minor item still worth tightening before this merges:

**Guard the `before-quit` registration in `registerStatsHandlers`** (src/main/ipc/handlers/stats.ts ~line 85). Right now `app.on('before-quit', ...)` is registered every time the function runs. Production only calls it once, but integration test environments and any future dynamic re-registration would accumulate listeners (and call `flushQueryEventsSync` multiple times on quit — each subsequent call is a no-op against an empty buffer, but it's a footgun waiting). A module-level `let beforeQuitRegistered = false` guard, or `app.once`, makes it idempotent in one line.

Once that's in (or you decide it's not worth the line — happy to defer if you'd rather), and CI greens, I'm happy to approve. 🙏

PR #940, opened 2026-05-02
 perf: cold-path I/O + error-pattern bank 

reachrazamair:
## Summary

- `history-manager.ts` fully migrated from sync `fs.*Sync` to `fs/promises`. Per-session reads/writes no longer block the main process; tight loops over many sessions parallelize via `Promise.all`.
- `history-bucket-cache.ts` async with in-flight read de-duplication so concurrent activity-graph renders share one disk read instead of racing.
- `error-patterns.ts` priority-ordered iteration (most-likely error types first) plus a length guard that skips the regex bank for token-noise chunks under 7 chars.

3 modules touched, ~50 call sites updated across 5 files, 8 new tests, 0 regressions.

### history-manager async (`src/main/history-manager.ts`)

Every public method went sync → async:

- `initialize`, `migrateFromLegacy`, `needsMigration`, `hasMigrated`
- `getEntries`, `addEntry`, `deleteEntry`, `updateEntry`, `clearSession`
- `listSessionsWithHistory`, `getHistoryFilePath`
- `getAllEntries`, `getAllEntriesPaginated`, `getEntriesByProjectPath`, `getEntriesByProjectPathPaginated`, `getEntriesPaginated`
- `updateSessionNameByClaudeSessionId`, `clearAll`, `clearByProjectPath`, `startWatching`

Cold-start: `initialize` no longer blocks the splash window behind `existsSync` + `readFileSync`. Steady-state: `addEntry` (called per agent turn) no longer blocks other IPC behind a sync read+write. Aggregation methods now `Promise.all` per-session reads — N sequential reads collapse into N parallel reads on the unified history view.

Migration ordering preserved: per-session writes parallelize via `Promise.all`, but the migration-marker write is sequential (must follow all session writes so a crash mid-migration can't leave a marker with partial data).

Call sites updated:
- `main/index.ts` (Cue execution path, `addEntry` fire-and-forget)
- `ipc/handlers/history.ts` (~20 sites — `await` added)
- `ipc/handlers/director-notes.ts` (~10 sites + `countAgentsAndSessions` made async, parallelized)
- `web-server/web-server-factory.ts` + `routes/apiRoutes.ts` + `types.ts` (callback signature widened to `T | Promise<T>`)
- `ipc/handlers/agentSessions.ts` (1 site)

### history-bucket-cache async (`src/main/utils/history-bucket-cache.ts`)

`get` / `set` / `invalidate` / `clear` async. Constructor no longer eagerly creates the cache directory — `ensureDir` is lazy, called from `set()`. `get()` tolerates a missing directory by returning null on `ENOENT`.

New `inflightReads` map de-duplicates concurrent disk reads for the same cache key — without it, the activity graph and unified history view could both miss the in-memory cache in the same tick and both issue parallel disk reads against the same file.

Callers in `history.ts` and `director-notes.ts` fire-and-forget the writes (`void cache.set(...)`) — the in-memory cache layer is updated synchronously, so the renderer doesn't need to wait for the disk write to settle.

### error-patterns priority + length guard (`src/main/parsers/error-patterns.ts`)

Reordered the error-type iteration to put `rate_limited` and `network_error` first (estimated ~70% of production matches based on observed agent-failure modes). First-match early-return contract preserved.

Added an opt-out length guard: `matchErrorPattern` skips the regex bank entirely for chunks shorter than `ERROR_PATTERN_DEFAULT_MIN_CHUNK_LENGTH` (7 chars — covers all real error strings we observe in production, including "timeout"). Streaming token noise (single-char chunks, short punctuation) dominates volume but never matches a pattern.

Callers can opt out via `{ minLength: 0 }` in the new `MatchErrorPatternOptions` (used by unit tests that probe individual short error strings).

## SSH history sync — verified untouched

The `.maestro/history/history-<hostname>.jsonl` cross-host shared-history feature lives in a separate module (`src/main/shared-history-manager.ts`) that this PR does not modify. Verified end-to-end:

- Two distinct managers, only one was touched (per-session local in `history-manager.ts`; per-host shared in `shared-history-manager.ts`).
- Integration in `history.ts` IPC handler still merges local + shared correctly: `await historyManager.getEntries(...)` + `readRemoteEntriesSsh` (already async) / `readRemoteEntriesLocal` (sync, untouched), then dedupe-by-id + sort.
- `history:add` still triggers all three writes: per-session local (`await historyManager.addEntry`), remote SSH mirror (`writeEntryRemote(...).catch(...)`, fire-and-forget), and local-project mirror (`writeEntryLocal`, conditional on `shareHistoryToProjectDir`).
- Bucket cache bypass intact: `getGraphData` checks `!hasShared && !hasLocalShared` before using the cache; shared-history overlay paths skip the cache entirely.
- Hostname filtering preserved: `LOCAL_HOSTNAME` constant unchanged, `readRemoteEntriesLocal` and `readRemoteEntriesSsh` still filter out the local host's own JSONL file.
- All 10 `shared-history-manager.test.ts` tests pass unchanged.

### Automated (already green)

- `npm run test`: **26,959 passed / 0 failed** (+8 new vs. baseline)
  - 2 new bucket-cache tests (concurrent-read de-dup, missing-dir tolerance)
  - 6 new error-patterns tests (length guard semantics, opt-out option, priority ordering, empty-string edge)
- `npm run lint` (`tsc`): clean across `tsconfig.lint.json`, `tsconfig.main.json`, `tsconfig.cli.json`
- `npm run lint:eslint`: clean
- All 210 history-related tests across 6 test files pass:
  - `history-manager.test.ts` (74) — bridged via `fs/promises` mock that delegates to existing sync mocks; minimal test rewrite
  - `shared-history-manager.test.ts` (10) — untouched
  - `history-bucket-cache.test.ts` (21) — `await` added throughout
  - `history.test.ts` (41) IPC handler — passes
  - `director-notes.test.ts` (31) — passes
  - `history.test.ts` shared (33) — passes

## Risk profile

| Risk | Mitigation |
| --- | --- |
| Async migration step ordering (step N reads step N-1's write) | Per-session writes parallelized via `Promise.all`; migration-marker write sequenced after — load-bearing comment + integration test covers the round-trip |
| Concurrent IPC handlers race on bucket cache disk reads | New `inflightReads` map de-duplicates same-key reads; covered by the new "concurrent reads" test |
| Length guard skips a real error string < 7 chars | Threshold chosen to cover every real error message observed (shortest is "timeout" at 7); shorter chunks are token noise that never match a pattern. Callers can opt out via `minLength: 0` |
| SSH shared-history sync regression | Separate untouched module, all 10 tests pass. Verified all 4 integration touchpoints (read/write × local/remote) preserve original semantics |
| Test surface change too invasive | `history-manager.test.ts` uses an `fs/promises`-to-`fs`-sync bridge mock so the existing 74 assertions on `mockReadFileSync` / `mockWriteFileSync` / `mockMkdirSync` continue to fire without rewrite |

## Detailed change & impact

| Change | Mechanism | User-visible impact |
| --- | --- | --- |
| `history-manager` async I/O | `fs.readFileSync` / `writeFileSync` / `existsSync` / `mkdirSync` / `unlinkSync` / `readdirSync` → `fs/promises` equivalents | Cold start blocks slightly less; per-turn writes don't block other IPC; aggregation queries (Director's Notes, web sync) parallelize reads across sessions |
| Bucket cache async + de-duped | Sync read → async; in-flight Map prevents same-key disk read races | Activity graph "instant feeling" instead of 100-200ms hitch on cold cache; concurrent renders don't double-read disk |
| Error-pattern priority + length guard | Most-likely types first, skip bank for chunks under 7 chars | Marginal CPU win during streaming (~0.5-2% per CLAUDE-PERFORMANCE.md estimate); error detection unchanged for all real-world cases |

## Summary by CodeRabbit

* **Improvements**
  * Improved error detection: minimum-length guard reduces false positives and preserves first-match priority.
  * History persistence is now async and non-blocking; writes are fire-and-forget where appropriate for snappier UI flows.
  * History cache is more robust: concurrent reads are de-duplicated and missing/stale cache data is tolerated.
  * Web/API callbacks now support async history providers.

* **Tests**
  * Expanded async-focused tests for history, watcher, cache, and error-pattern behavior.


pedramamini:
@reachrazamair thanks for the contribution! Reviewed the diff end-to-end.

**Approving** — the work is solid:

- **`history-manager` async migration**: clean conversion with thoughtful preservation of ordering invariants. The load-bearing comment around the migration-marker write (sequential after `Promise.all` of session writes, so a mid-migration crash can't strand a marker over partial data) is exactly right. Distinguishing `ENOENT` from real read failures throughout (`getEntries`, `addEntry`, `deleteEntry`, `updateEntry`, `clearSession`, `listSessionsWithHistory`) is also the right call.
- **Fire-and-forget on the Cue path** (`void historyManager.addEntry(...)` in `index.ts`) is appropriate — the caller doesn't need durability before returning, and an unhandled rejection would still surface to Sentry via the inner try/catch.
- **`history-bucket-cache` in-flight read de-dup** via the `inflightReads` map closes a real race on cold cache where the activity graph and unified history view can both miss the in-memory cache in the same tick. Lazy `ensureDir` from `set()` (instead of in the constructor) is a nice cleanup.
- **`error-patterns` priority + length guard** — preserves first-match early-return semantics, the `MatchErrorPatternOptions.minLength` opt-out keeps unit tests clean, and the threshold (7 chars to cover "timeout") is well-reasoned.
- **Web-server callback signature widening** to `T | Promise<T>` is the minimum-surface-area way to thread async through without forcing every callsite to refactor.

Test coverage looks appropriate (concurrent-read de-dup, missing-dir tolerance, length-guard semantics + opt-out). SSH shared-history module verified untouched.

Adding the `approved` label. CI checks are still in progress at the time of this review — if anything goes red, please follow up.

PR #944, opened 2026-05-02
feat(web): add unread-tab bell filter to mobile TabBar

chr1syy:
## Summary

- Ports the agent-level unread bell (already in `LeftPanel`) down to the session/tab level. When you have an agent with many tabs, toggling the bell collapses the tab bar to just the active tab plus any tab with `hasUnread` or `state === 'busy'`, making it easy to find the session that's waiting on you.
- Bell is disabled when no tab has activity, shows an accent-colored dot when off-but-unread tabs exist, fills with accent when active, and auto-disables once activity settles — same UX language as the agent bell.
- Filtered render preserves each tab's original array index so the long-press popover's Move Left / Move Right reorder math stays correct.

## Test plan

- [x] `vitest run src/__tests__/web/mobile/TabBar.test.tsx` (80 tests, including 10 new bell tests covering enabled/disabled state, filter behavior for busy / unread / active tabs, aria-label and aria-pressed transitions, auto-disable on settle, and reorder index correctness while filtered)
- [x] ESLint clean on `src/web/mobile/TabBar.tsx`
- [x] Prettier clean on both changed files
- [ ] Manual: open the mobile web UI on an agent with several tabs, mark one busy/unread, toggle the bell, confirm only that tab + the active tab remain visible

## Summary by CodeRabbit

* **New Features**
  * Bell filter for the mobile tab bar to toggle viewing all tabs or only unread/busy tabs; active tab always remains visible. Button shows an unread indicator when applicable, stays clickable with no unread, and preserves the filter across rerenders. Accessibility state updated for screen readers.

* **Tests**
  * Added tests for bell rendering, accessibility state, toggle behavior, persistence after activity clears, visibility rules, and reorder behavior while filtered.

### DIFF
--- a/src/__tests__/web/mobile/TabBar.test.tsx
+++ b/src/__tests__/web/mobile/TabBar.test.tsx
@@ -1330,4 +1330,231 @@ describe('TabBar', () => {
 			expect(module.default).toBe(module.TabBar);
 		});
 	});
+
+	describe('Bell filter', () => {
+		it('renders the bell button', () => {
+			const tabs = [createTab({ id: 'tab-1', name: 'First' })];
+			render(
+				<TabBar
+					tabs={tabs}
+					activeTabId="tab-1"
+					onSelectTab={mockOnSelectTab}
+					onNewTab={mockOnNewTab}
+					onCloseTab={mockOnCloseTab}
+				/>
+			);
+			expect(screen.getByLabelText('Filter unread tabs')).toBeInTheDocument();
+		});
+
+		it('disables the bell button when no tabs have unread activity', () => {
+			const tabs = [
+				createTab({ id: 'tab-1', name: 'First', hasUnread: false }),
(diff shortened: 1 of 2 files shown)


pedramamini:

Reviewed and this looks great:
- Cleanly mirrors the agent-level bell pattern from `LeftPanel`, so the UX language stays consistent.
- The `tabs.indexOf(tab)` trick to preserve the original index for Move Left/Right reorder math is a nice touch.
- Auto-disabling when `hasUnreadTabs` settles is the right call — keeps the user from getting stranded with a single-tab bar.
- Disabled-when-empty + accent dot indicator + `aria-pressed` give the bell button proper a11y semantics.
- Test coverage is thorough (enabled/disabled, filter behavior across busy/unread/active, auto-disable, reorder index correctness while filtered).

Approving. ✅

PR #945, opened 2026-05-02
fix(worktrees): filter discovered worktrees by parent repo identity

chr1syy:
## Summary

`scanWorktreeConfigs` and the chokidar `onWorktreeDiscovered` listener attached every worktree found under `worktreeConfig.basePath` to the parent agent that "owned" that basePath, **without verifying the worktree actually belonged to the parent's git repository**. With two parents whose basePaths overlap — or one basePath containing worktrees from multiple repos — whichever parent's loop iterated first grabbed everything.

This is the bug behind "all my worktrees got re-attached under the wrong agent" after the worktree-wipe fixed by #931 corrupted the persisted state in older builds.

## What changed

- **`useWorktreeHandlers.ts`** — resolve the parent's main-repo root once per scan via `worktreeInfo` (parent of `--git-common-dir`, so worktree parents resolve to the same canonical root as the main repo) and skip subdirs whose `repoRoot` doesn't match. When either side can't be resolved, fall back to the legacy "trust the basePath" behavior so existing setups keep working.
- Also detach already-attached children whose `cwd` is a worktree of a different repo. Without this, the wrong-agent children would never be re-attached to the correct parent — the existing-session dedup in `scanWorktreeConfigs` would block it.
- Symmetric guard in the chokidar `onWorktreeDiscovered` listener: fetch the parent and discovered-worktree `repoRoot`s in parallel and skip mismatches.

No IPC contract changes — the renderer uses the existing `git.worktreeInfo` IPC for both directions.

## Repro

1. Configure a parent agent at `repos/repo-a` with `worktreeConfig.basePath = /shared/worktrees`.
2. Drop a worktree of an unrelated repo into `/shared/worktrees/feat-other` (e.g. an existing leftover from an older project).
3. Restart the app or trigger a visibility-change rescan.

**Before:** `feat-other` is attached to repo-a's parent agent.
**After:** `feat-other` is filtered out; only worktrees that share repo-a's main repo root are attached.

## Test plan

- [x] `vitest run src/__tests__/renderer/hooks/useWorktreeHandlers.test.ts` — 67 pass (60 existing + 7 new):
  - filters scanned subdirs whose `repoRoot` doesn't match the parent
  - detaches a child whose `cwd` is a worktree of a different repo (self-heals the wrong-agent attachment)
  - two parents sharing a basePath each receive only their own repo's worktrees
  - falls back to legacy behavior when the parent's `repoRoot` can't be resolved
  - falls back to legacy behavior when subdir's `repoRoot` is null
  - chokidar `onWorktreeDiscovered` rejects a worktree from a different repo
  - chokidar `onWorktreeDiscovered` accepts a worktree from the matching repo
- [x] `vitest run src/__tests__/main/ipc/handlers/git.test.ts` — 153 pass (no regressions, no main-side changes)
- [x] `tsc --noEmit -p tsconfig.lint.json` clean
- [x] `tsc --noEmit -p tsconfig.main.json` clean
- [x] `prettier --check` clean on touched files

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented misattachment of worktrees by enforcing repository identity during scans and watcher events
  * Ensured stale or reassigned child sessions are detached before reattachment
  * Fixed ordering/deduplication so attachments aren’t blocked by pending removals

* **User-facing**
  * Notifications now distinguish "Worktree Re-assigned" from "Worktree Removed"

* **Tests**
  * Expanded tests for repo-identity filtering, discovery/attachment, and error reporting to monitoring

### DIFF
--- a/src/__tests__/renderer/hooks/useWorktreeHandlers.test.ts
+++ b/src/__tests__/renderer/hooks/useWorktreeHandlers.test.ts
@@ -470,6 +470,60 @@ describe('handleSaveWorktreeConfig', () => {
 
 		expect(mockGit.scanWorktreeDirectory).not.toHaveBeenCalled();
 	});
+
+	it('filters out scanned subdirs whose repoRoot does not match the parent repo', async () => {
+		// Same repo-identity guard as scanWorktreeConfigs, applied at the moment
+		// the user saves the worktree config. Without this, pointing the agent
+		// at a basePath that contains worktrees from another repo would attach
+		// them on the spot, before any later rescan could clean up.
+		useSessionStore.setState({
+			sessions: [
+				{
+					...mockParentSession,
+					cwd: '/repos/repo-a',
+					worktreeConfig: undefined,
+				},
+			],
+			activeSessionId: 'parent-1',
+		} as any);
+
+		mockGit.worktreeInfo.mockResolvedValueOnce({
+			success: true,
+			exists: true,
+			isWorktree: false,
+			repoRoot: '/repos/repo-a',
+		});
+
+		mockGit.scanWorktreeDirectory.mockResolvedValue({
+			gitSubdirs: [
+				{
+					path: '/shared/worktrees/feat-mine',
+					branch: 'feat-mine',
+					name: 'feat-mine',
+					repoRoot: '/repos/repo-a',
+				},
+				{
+					path: '/shared/worktrees/feat-other',
+					branch: 'feat-other',
+					name: 'feat-other',
+					repoRoot: '/repos/repo-b',
+				},
+			],
+		});
+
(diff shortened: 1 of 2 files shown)


pedramamini:
Thanks for the contribution, @chr1syy! Nice surgical fix — the symmetric guard across both `scanWorktreeConfigs` and the chokidar `onWorktreeDiscovered` path is the right call, and the fallback-to-legacy behavior when either repoRoot can't be resolved means existing setups stay safe. The self-healing detach in the removals loop is a particularly nice touch for users who got bitten by the old wipe bug. Test coverage is solid (filter, detach, two-parent shared basePath, both fallback cases, plus chokidar accept/reject). Approving.

PR #946, opened 2026-05-03
feat(web): add Run-in-Worktree toggle to mobile AutoRun launch

chr1syy:
## Summary

- Adds the **Dispatch to a separate worktree** toggle (with base-branch picker, sanitized new-branch input, and optional "create PR on completion") to the mobile Auto Run launch sheet — closing the AR-PARITY-02 gap with desktop's `WorktreeRunSection`.
- Two new SSH-aware WebSocket messages, `get_git_branches` and `list_worktrees`, proxy git directly from the main process using each session's `cwd` + `sessionSshRemoteConfig`.
- The mobile launch path threads the composed `worktree` payload through `configure_auto_run`, where the desktop's existing remote-handler already creates the worktree, checks out the branch, and (optionally) opens a PR.

## What changed

**Backend (`src/main/web-server/`)**
- `handlers/messageHandlers.ts` — new `get_git_branches` and `list_worktrees` handlers with `sessionId` validation, "not configured" fallback, and `requestId` echo.
- `types.ts` — `SessionData` / `SessionBroadcastData` gain `isGitRepo` + `worktreeBasePath`; new `GitBranchesResult`, `WorktreeEntry`, `ListWorktreesResult`, and the corresponding callback type aliases.
- `managers/CallbackRegistry.ts`, `WebServer.ts` — registry slots and setter exposure for the new callbacks.
- `web-server-factory.ts` — emits `isGitRepo` + `worktreeBasePath` on the sessions list and wires the two new callbacks via `execGit` (SSH-aware) on the session's cwd.

**Frontend (`src/web/`)**
- `hooks/useAutoRun.ts` — new `LaunchWorktreeConfig`, `WorktreeSummary`, `loadGitBranches`, `listWorktrees`; `launchAutoRun` spreads `worktree` only when `enabled === true`.
- `hooks/useWebSocket.ts` — `SessionData` gains `isGitRepo` + `worktreeBasePath`.
- `mobile/AutoRunWorktreeSection.tsx` (new) — toggle, base-branch picker, sanitized branch-name input with path preview, PR checkbox; gated on `isGitRepo` + `worktreeBasePath`.
- `mobile/AutoRunSetupSheet.tsx`, `mobile/App.tsx` — section rendered above the Prompt, loaders bound to the active session.

**Tests**
- `__tests__/main/web-server/handlers/messageHandlers.test.ts` — six new cases (validation, payload shape, callback-not-configured) for the two WS handlers.
- `__tests__/web/hooks/useAutoRun.test.ts` (new) — `launchAutoRun` worktree forwarding/strip + branch/worktree request shape and error fallbacks.
- `__tests__/web/mobile/AutoRunWorktreeSection.test.tsx` (new) — render gating, enable→config emission, PR toggle, branch-cleared→null.
- Updated `__tests__/web/mobile/App.test.tsx` mock to expose the new hook methods.

## Test plan

- [x] Mobile (web): open Auto Run setup on a git-repo session with worktree base path configured; toggle "Dispatch to a separate worktree", confirm branches load and the path preview matches `<basePath>/<branchName>`.
- [x] Launch with the toggle on → desktop creates the worktree, checks out the new branch, runs the Auto Run inside it.
- [x] Toggle "Automatically create PR when complete" on → PR is opened against the chosen base branch on completion.
- [x] Section is hidden for non-git sessions and disabled (with helper text) when no `worktreeBasePath` is configured.
- [x] SSH remote session: branches load via the session's SSH remote (`sessionSshRemoteConfig`).
- [x] `npx vitest run` on the scoped suite (158 pass / 0 fail).
- [x] Lint, Prettier, and TypeScript all clean on touched files.

## Summary by CodeRabbit

* **New Features**
  * Mobile Auto Run: dispatch runs into isolated Git worktrees with base-path, branch selector, path preview, and optional auto-PR creation; launches respect worktree enablement and validation.
  * Sessions & lazy loaders: session metadata now surfaced for Git/worktree status and on-demand branch/worktree loading to improve setup responsiveness.

* **Tests**
  * Added tests covering Web→Desktop message paths, branch/worktree loaders, and mobile AutoRun UI flows, validations, race conditions, and error handling.


chr1syy:
## Blocker: AR-PARITY-02 gap not actually closed — mobile/web launch skips child-session creation that desktop does

Found while soaking #946 into `0.16.0-RC-build` for a 0.16.17-RC test build.

The PR description says this closes the AR-PARITY-02 gap with desktop's `WorktreeRunSection`. The on-disk worktree creation, branch checkout, and \"agent runs in the worktree\" parts work correctly. But the **UI affordance** that makes the desktop flow actually usable — a new child agent session in the Left Bar that the user can switch to — is missing on the mobile/web path. From a mobile user's perspective, they tap \"Dispatch to a separate worktree\", AutoRun runs, and **nothing visible happens in their Left Bar**. There's no agent for them to navigate to or continue interacting with after the AutoRun.

### Evidence

**Desktop path** — `src/renderer/hooks/batch/useAutoRunHandlers.ts:427-457`:

```ts
} else if (
    config.worktreeTarget?.mode === 'create-new' ||
    config.worktreeTarget?.mode === 'existing-closed'
) {
    let parentForSpawn = activeSession;
    if (activeSession.parentSessionId) {
        const parent = selectSessionById(activeSession.parentSessionId)(useSessionStore.getState());
        if (parent) parentForSpawn = parent;
    }
    const newSessionId = await spawnWorktreeAgentAndDispatch(parentForSpawn, config);
    if (!newSessionId) return;
    targetSessionId = newSessionId;
}
// ...
startBatchRun(targetSessionId, config, activeSession.autoRunFolderPath);
```

→ `spawnWorktreeAgentAndDispatch` creates a **new child session** with `parentSessionId` linking it to the original. That child shows up in the Left Bar; `startBatchRun` is dispatched to *its* ID.

**Mobile/web path** — `src/renderer/hooks/remote/useAppRemoteEventListeners.ts:306-341`:

```ts
const worktree: BatchRunConfig['worktree'] | undefined =
    config.worktree && config.worktree.enabled
        ? { enabled: true, path: ..., branchName: ..., createPROnCompletion: ..., prTargetBranch: ... }
        : undefined;

const batchConfig: BatchRunConfig = {
    documents,
    prompt: config.prompt || DEFAULT_BATCH_PROMPT,
    loopEnabled: config.loopEnabled || false,
    maxLoops: config.maxLoops,
    ...(worktree ? { worktree } : {}),
};
// ...
startBatchRun(sessionId, batchConfig, folderPath).catch(...)
```

→ `startBatchRun` is called with the **original** `sessionId`. No `spawnWorktreeAgentAndDispatch`. The worktree is created on disk and the spawned agent's cwd is set to it (which is why my probe playbook reported `cwd: /home/chris/code/worktrees/main-test`, `is worktree: yes`, `git dir` ≠ `git common dir`), but no child agent appears in the UI.

### Repro

1. Mobile/web AutoRun setup on a git session with `worktreeBasePath` configured.
2. Toggle \"Dispatch to a separate worktree\" on, pick a base branch, type a fresh branch name, leave PR-on-completion off.
3. Pick the diagnostic playbook (one task that prints `pwd` / `git rev-parse --git-dir` / `git rev-parse --git-common-dir`).
4. Launch.

**Observed:**
- Worktree directory exists on disk after launch ✅
- Agent runs inside it (cwd matches `<basePath>/<branch>`, `is worktree: yes`) ✅
- **Left Bar still shows only the original agent.** No child session, no new entry, no toast. ❌
- The AutoRun completes; user has no UI surface to continue interacting with the agent that ran in the worktree.

### Suggested fix

Mirror desktop's path in the remote handler: before calling `startBatchRun`, when `config.worktree?.enabled === true`, spawn a child session via `spawnWorktreeAgentAndDispatch` (or its session-creating equivalent) and use the returned `newSessionId` as the dispatch target. Propagate that ID back to the mobile client in the `configure_auto_run_result` so the mobile client can navigate to the new session.

The minimum diff is roughly:

```ts
// In useAppRemoteEventListeners.ts, before startBatchRun:
let dispatchSessionId = sessionId;
if (worktree?.enabled) {
    const newSessionId = await spawnWorktreeAgentAndDispatch(parentForSpawn, batchConfig);
    if (!newSessionId) {
        sendRemoteConfigureAutoRunResponse(responseChannel, { success: false, error: 'Worktree spawn failed' });
        return;
    }
    dispatchSessionId = newSessionId;
}
sendRemoteConfigureAutoRunResponse(responseChannel, { success: true, sessionId: dispatchSessionId });
startBatchRun(dispatchSessionId, batchConfig, folderPath).catch(...);
```

…plus the corresponding mobile-client-side handling of the returned `sessionId` so the user can switch to the new agent. Tests in `__tests__/renderer/hooks/batch/useAutoRunHandlers.worktree.test.ts` already cover the desktop path; mirror equivalents for the remote handler would be useful.

### Other findings from the same soak (non-blocking, FYI)

- ✅ Section gating: hidden for non-git sessions; disabled (with helper) when no `worktreeBasePath`.
- ✅ Branch list loads from the session's git over `get_git_branches`; SSH path not yet exercised here.
- ✅ Path preview sanitization: `feat: foo bar!` → `foo-bar!` (colon + `feat:` prefix dropped, space→hyphen).
- ⚠️ Sanitization preserves `!` — git allows it, but it's a shell-special character (history expansion) that may surprise users when they `cd` to the worktree path interactively. Strictness call.

chr1syy:
## Afternote — child session IS being spawned, but attached to the wrong parent

Correction to my prior comment: a child session **does** get added to the Left Bar after a mobile/web worktree launch — it just attaches to **the agent immediately above the launching agent** in the Left Bar, not to the launching agent itself. I missed this on the first read because the new entry blended in among my many existing worktree sessions; chr1syy spotted it on a closer look at the Left Bar after the run.

That makes this two distinct issues, not one:

1. **Wrong-parent attachment.** The newly created child session links to a sibling/predecessor agent rather than the agent that actually launched the AutoRun. Likely a parent-resolution bug — possibly an off-by-one in whatever logic walks/derives the parent on the remote launch path, or a stale `activeSession` snapshot at the moment the spawn fires. Suspect path: the `parentForSpawn` resolution that prefers `activeSession.parentSessionId` may be reading from the wrong session object on the remote path.

2. **The earlier diagnosis still partially stands.** I had pattern-matched the symptom \"no UI affordance\" to \"no spawn at all\" based on the snippet in `useAppRemoteEventListeners.ts:306-341`, which doesn't call `spawnWorktreeAgentAndDispatch`. There must be another path inside `startBatchRun` (or downstream of it) that does spawn for the remote launch — that's what's producing the wrong-parent child. The earlier fix-suggestion (\"call `spawnWorktreeAgentAndDispatch` before `startBatchRun`\") was therefore incomplete; the right fix is probably to **align the remote launch path's spawn call to use the same `parentForSpawn` resolution as `useAutoRunHandlers.ts:427-457`**, threading the launching session through correctly.

### Repro for the wrong-parent attachment

1. Have multiple agents in the Left Bar.
2. Launch a worktree-mode AutoRun via mobile/web from a specific agent (let's call it agent X).
3. After launch, observe the Left Bar: the new child session attaches to **agent X-1** (or a sibling) instead of agent X.

Keeping the `blocking` label — both issues land on the same launch path and should be fixed together.

chr1syy:
@greptileai re-review please

chr1syy:
## Wrong-parent fix verified — unblocking

Re-tested 2026-05-07 against the `0.16.0-RC-build` soak build (PR #946 with `9ac39ddd` + `c1d06a3e`).

### What's fixed

- ✅ **Child session is now correctly parented to the launching agent.** The wrong-parent attachment regression from yesterday's report is gone. Verified by launching `WORKTREE-TEST-01.md` against a specific agent X (git repo + `worktreeBasePath` configured) — diagnostic playbook reported `cwd: ~/code/worktrees/rc-test`, `is worktree: yes`, `branch: rc-test`, and `git worktree list` includes the new entry.
- ✅ Worktree creation on disk, branch checkout, agent's cwd inside the worktree — all intact.

Removing the `blocking` label.

### Non-blocking follow-ups (mobile/web parity gaps)

Two UI affordance gaps surfaced during the same soak — these don't block this PR but should be tracked as follow-up work:

1. **Launching agent's state indicator doesn't change after launch** — stays green (idle) instead of flipping to busy/connecting. Mobile users get no in-UI confirmation the launch took.
2. **Spawned child session not visible in mobile's Left Bar** like on desktop — desktop renders the parent/child relationship, mobile doesn't.

Filed as a gist with repro and suggested fix directions: https://gist.github.com/chr1syy/67630166ff5217a97d7b1fad91201f3b

Suggested fix directions in the gist: Center Flash + jump-session click action for the launch ack (Gap 1), and mirroring desktop's `parentSessionId` rendering on the mobile/web session list (Gap 2). The Center Flash + clickAction pattern is already established in the codebase, so Gap 1 is a small lift.

Other items still untested in this pass (deferred from yesterday's plan, not part of this PR's blocker history):

- PR-on-completion toggle → real PR opens against the chosen base branch.
- Branch-cleared edge: clearing the input → expect `worktree: null` payload, no worktree.
- SSH-remote session: branches load via SSH, worktree creation on the remote host.

These can be exercised in a follow-up pass once the gist's items are addressed.

PR #947, opened 2026-05-03
feat(web): expose Playbook Exchange to mobile/web

chr1syy:
## Summary

- Mobile users can now browse, preview, and import community playbooks from inside the AutoRun setup sheet (Playbook Exchange parity with desktop's `MarketplaceModal`).
- Marketplace cache / fetch / import logic is extracted into a shared service so the IPC handler and the web-server use a single code path (no renderer round-trip).
- SSH-remote sessions on mobile import to the correct host — the server resolves `autoRunFolderPath` and SSH config from the session itself, so mobile clients can't override them.

## What changed

**Server**
- `src/main/services/marketplace-service.ts` (new) — pure helpers for cache/manifest/document/import.
- `src/main/ipc/handlers/marketplace.ts` — slimmed to thin wrappers around the service. Existing IPC behavior preserved (45 IPC tests still pass).
- `src/main/web-server/handlers/messageHandlers.ts` — four new WebSocket message types: `marketplace_get_manifest` (with `refresh: true`), `marketplace_get_document`, `marketplace_get_readme`, `marketplace_import_playbook`. Path-traversal + required-field validation at the boundary.
- `src/main/web-server/web-server-factory.ts` — wires callbacks; `importMarketplacePlaybook` resolves `autoRunFolderPath` and SSH config from the session (mobile clients only send `sessionId`, `playbookId`, `targetFolderName`).
- `WebServer.ts`, `CallbackRegistry.ts`, `types.ts` — extended for the four new callbacks.

**Mobile UI**
- `src/web/mobile/MarketplaceSheet.tsx` (new) — 90vh bottom sheet, list → detail → import flow with category chips, search, README + document preview via `MobileMarkdownRenderer`, and target-folder input.
- `src/web/mobile/AutoRunSetupSheet.tsx` — optional `onOpenMarketplace` prop and a "Browse Playbook Exchange" entry point at the top of the scrollable area.
- `src/web/mobile/App.tsx` — adds `showMarketplaceSheet` state, marketplace handlers, wires the entry point, and auto-refreshes AutoRun docs after a successful import.

## Test plan

- [x] `npm run lint` (TypeScript, all configs)
- [x] `npm run lint:eslint`
- [x] `npx prettier --check` for all changed files
- [x] `vitest run src/__tests__/main/ipc/handlers/marketplace.test.ts` (45 tests pass — IPC handler refactor preserves behavior)
- [x] `vitest run src/__tests__/main/web-server/handlers/messageHandlers.test.ts` (113 tests pass — includes 7 new tests for marketplace WS handlers + traversal rejection + unconfigured-callback paths)
- [x] `vitest run src/__tests__/web/mobile/App.test.tsx` (95 tests pass — adds mock for new `MarketplaceSheet`)
- [ ] Manual smoke on mobile: open AutoRun setup → tap Playbook Exchange → browse, search, preview README, preview document, import to a folder name → verify the imported playbook appears in the AutoRun document list (AR-PARITY-01)
- [ ] Manual smoke with SSH remote session: import lands on the remote host (parity with desktop)

## Notes

- No browse-folder picker on mobile (no native folder picker available); users type the folder name. Same constraint the desktop modal already has when remote.
- Marketplace cache TTL, file watcher hot-reload, and local-manifest behavior are unchanged.

## Summary by CodeRabbit

* **New Features**
  * Playbook Exchange (Marketplace) on mobile: browse/search by category, preview README/docs, edit import folder name, import playbooks with progress and haptic feedback; integrates with Auto Run.

* **Bug Fixes / Safety**
  * Stronger input validation and traversal protections; clearer errors when SSH remote is missing/disabled. Imports tolerate per-document failures and record only successfully imported docs/assets.

* **Tests**
  * Expanded tests covering marketplace messaging, IPC, import edge cases, and mobile flows.


chr1syy:
Merged `upstream/rc` and addressed all review feedback from greptile and coderabbitai. Pushed `f36de1fd`.

### Conflict resolution
The only conflict was `src/main/ipc/handlers/marketplace.ts` — kept this PR's thin-wrapper refactor (logic now in `src/main/services/marketplace-service.ts`) and ported the upstream watcher error handler (Windows UNKNOWN / EPERM / ENOENT classification) into `createLocalManifestWatcher`.

### Review feedback addressed

**Security (P1)**
- WS handlers reject `playbookPath` values that point at the local filesystem (absolute or `~`-prefixed). Local paths remain reachable via desktop IPC where the renderer is trusted; mobile/web clients cannot coerce arbitrary file reads via the marketplace fetch helpers. (greptile)
- WS `marketplace_import_playbook` rejects `targetFolderName` containing `..`, `/`, `\\`, `~`, or absolute paths. New `assertSafeTargetFolderName` in the service layer is the source of truth so the IPC path gets the same guard. (greptile + coderabbit)
- IPC `marketplace:importPlaybook` fails loudly when `sshRemoteId` is provided but cannot be resolved, rather than silently downgrading to a local import. Matches CLAUDE.md's SSH-spawn pattern. (coderabbit)

**Correctness (Major)**
- Persisted `newPlaybook.documents` is filtered against `importedDocs`, so partial imports no longer save references to missing files. (coderabbit)
- Reading the per-session playbooks file no longer silently resets to `[]` on non-ENOENT errors — corrupt JSON / EACCES now throws rather than overwriting existing data. (coderabbit)
- `fetchReadme` re-throws unexpected I/O and network failures (only ENOENT/404 map to "no README"), so faults surface in Sentry. (coderabbit)
- WS document/readme handlers return a structured failure when the callback resolves to `null` (unconfigured wiring), instead of treating it as a successful empty response. (coderabbit)

**Code quality (P2)**
- Replaced dynamic `require('os')` with `import os from 'os'` in marketplace-service. (greptile)

**UI polish (Minor)**
- AutoRunSetupSheet entry button: focus ring rendered via `onFocus`/`onBlur` so keyboard users still see focus after `outline: none`. (coderabbit)
- MarketplaceSheet preview: `??` instead of `||` so empty README/doc renders rather than the fallback placeholder. (coderabbit)
- MarketplaceSheet: monotonic `previewRequestIdRef` discards stale README/document responses when the user picks another playbook/file mid-flight; error responses surface in the import-error banner instead of being swallowed. (coderabbit)

### Test updates
- Replaced two SSH "fall back to local fs" tests with "fail loudly" assertions matching the new behavior.
- Added the missing `mockRejectedValueOnce({ code: 'ENOENT' })` for the per-session playbooks file in 7 tests that previously relied on the bare-catch silently swallowing `JSON.parse(undefined)` errors.
- Fixed the `os` mock in the tilde-path test to expose `default` for the new static import.
- Asserts that partial-import playbooks persist only successfully-imported docs.

### Verification
- `npm run lint` clean (TypeScript, all configs).
- `npm run lint:eslint` clean.
- `prettier --check` clean.
- Scoped suites pass: marketplace IPC (45), message handlers (113), mobile App (95).

chr1syy:
Merged \`upstream/rc\` again, fixed CI failures, and addressed the new coderabbit comment. Pushed \`1fb98426\`.

### Conflicts resolved (6 files)

Upstream's PR2 of the CLI surface refactor (\`list_desktop_sessions\` / \`get_session_history\`), AutoRunSetupSheet playbook-CRUD redesign, and FolderPickerSheet all landed on \`rc\` after the previous merge. Combined both sides:

- \`WebServer.ts\`, \`types.ts\`, \`CallbackRegistry.ts\` — kept marketplace callback registrations alongside upstream's \`listDesktopSessions\` / \`getSessionHistory\`.
- \`handlers/messageHandlers.ts\` — kept four marketplace WS handlers alongside the new \`list_desktop_sessions\` / \`get_session_history\` handlers.
- \`web/mobile/App.tsx\` — kept both \`MarketplaceSheet\` wiring and the new \`FolderPickerSheet\` + \`handleAutoRunFolderConfirm\` flow; threaded \`sendRequest\` / \`send\` / \`currentDocument\` / \`onOpenMarketplace\` into the redesigned \`AutoRunSetupSheet\`.
- \`web/mobile/AutoRunSetupSheet.tsx\` — kept the upstream playbook-CRUD refactor and re-anchored the Playbook Exchange entry button above the Playbooks section so import remains discoverable.
- \`__tests__/main/web-server/handlers/messageHandlers.test.ts\` — kept marketplace mocks/tests alongside upstream's CLI-surface mocks/tests.

### CI fix

- \`web-server-factory.test.ts\` MockWebServer was missing the four marketplace setter stubs, producing 36 \`TypeError: ... is not a function\` failures. Added them to the mock (combined with upstream's two new CLI setters).

### New coderabbit comment (Major) — typed \`marketplace_*_result\` failures

Marketplace WS validation branches were emitting generic \`type: 'error'\` frames rather than the request-scoped \`marketplace_*_result\` types, which would make typed clients hang waiting for the success type.

Introduced a small \`sendMarketplaceFailure(client, type, error, message, extra?)\` helper and routed every validation-failure path through it (missing fields, untrusted local paths, traversal in filename, separators in \`targetFolderName\`, callback unconfigured). Tests updated to assert typed-result failures, plus two new cases:

- \`marketplace_get_document\` with an absolute \`playbookPath\` returns \`marketplace_get_document_result { success: false, error: 'Local filesystem paths are not allowed…' }\`.
- \`marketplace_import_playbook\` with separators in \`targetFolderName\` returns \`marketplace_import_playbook_result { success: false, error: '… separators…' }\`.

### Verification

- \`npm run lint\` clean (TypeScript, all configs).
- \`npm run lint:eslint\` clean.
- \`prettier --check\` clean.
- Scoped suites pass: marketplace IPC (45), message handlers (123, +7 new tests), web-server-factory (95), mobile App (95). 319 total.

chr1syy:
## Two UX findings from soaking #947 into 0.16.0-RC-build

### Finding 1 — Import folder prefill contradicts the server validator (functional bug)

The mobile import flow prefills the **Import to folder** input with the marketplace manifest's full identifier (e.g. `development/readme-accuracy`). When the user taps **Import Playbook**, the server-side handler in `src/main/web-server/handlers/messageHandlers.ts` rejects it with:

> `targetFolderName must be a single folder name without separators`

So **the import does not work out of the box** — every user clicking Import on a category-namespaced playbook hits this and has to manually edit the input down to a single segment. Manually editing it down to e.g. `readme-accuracy` succeeds and the playbook appears in the AutoRun list, so the underlying import path is healthy; only the default value is wrong.

**Positive side-effect:** the path-traversal validator that catches `..` / `/` is working correctly. Step 5 of the test plan (\"craft a malicious WS message with `targetFolderName: '../foo'`\") is implicitly verified by this very interaction — the same validator that's rejecting the slash would reject `..`.

**Suggested fix (smallest surface):** client-side sanitize the prefill to a single segment before populating the input. Two reasonable options:

```ts
// Last-segment-only (simplest)
const defaultFolder = manifestId.split('/').pop() ?? manifestId;

// Or replace separators
const defaultFolder = manifestId.replace(/[/\\]/g, '-');
```

Pair with a small \"will install at: `<autoRunFolderPath>/<resolvedName>`\" preview line under the input so the user sees the resolved path before tapping Import. This also makes the validator essentially unreachable from the UI in normal use, leaving it as a defensive backstop only.

The alternative — relaxing the server validator to accept nested paths bounded to `autoRunFolderPath` — is bigger, riskier, and goes against the simplicity of the current single-segment guarantee. I'd skip that.

### Finding 2 — Bottom-nav `Run` and `Playbooks` open the same screen

In the mobile bottom-nav / left drawer, tapping **Run** and tapping **Playbooks** both open the AutoRun setup sheet. The two affordances are functionally identical right now.

The naming implies a separation:

- **Run** → AutoRun setup (pick a doc, configure, launch). What both currently do.
- **Playbooks** → the **Playbook Exchange** browser, opened directly without going through the AutoRun setup sheet first.

Right now you have to: tap Playbooks → AutoRun setup opens → tap **Browse Playbook Exchange** → 90vh sheet slides up. That's three taps to get to the Exchange when the user already said \"Playbooks.\"

**Suggested fix:** route the Playbooks bottom-nav entry directly into the Marketplace sheet (the same one currently mounted under AutoRun setup → Browse). Keep the Browse entry inside AutoRun setup so users in mid-launch flow can still discover it.

### Other findings from the same soak

- ✅ Marketplace sheet renders, manifest loads, playbooks + categories visible.
- ✅ Search and category filtering work.
- ✅ Detail view (README + document) renders via `MobileMarkdownRenderer`.
- ✅ Once the folder name is single-segment, import succeeds and refreshes the AutoRun document list (per AR-PARITY-01).
- ✅ Path-traversal validator on the server is functional (caught by Finding 1 incidentally).

Both findings are UX, not security/correctness — leaving these unlabeled. Upgrade to `blocking` if Finding 1 is considered a ship-stopper for the 0.16.17-RC release.

chr1syy:
Upgrading **Finding 1** (import folder prefill contradicts the validator) to **blocking** — confirmed with chr1syy. "Import doesn't work without manually editing the prefill" is broken-out-of-the-box behavior, not polish. Finding 2 (Run/Playbooks bottom-nav redundancy) remains a non-blocking UX redesign suggestion.

chr1syy:
@greptileai re-review please

chr1syy:
Addressed greptile's P1 inline finding on `MarketplaceSheet.tsx:92`. Pushed `60f0ce67`.

### Fix — split preview vs import error state

`importError` was dual-purpose: `handleSelectPlaybook` and `handleSelectDocument` set it on README/document fetch failures, but it rendered in the import footer next to the **Import Playbook** button. A user who tapped a playbook, saw the preview fail, and read the error in the footer would reasonably believe an import had been attempted and failed.

Introduced a dedicated `previewError` state:

- `handleSelectPlaybook` and `handleSelectDocument` now write to `previewError` only.
- `handleImport` continues to write to `importError` only.
- `previewError` renders as an inline alert inside the preview pane (above where the markdown would render), with `role="alert"` and the same theme-error coloring used elsewhere.
- Reset paths (Escape key handler, back button, new-playbook selection, new-document selection) clear both flags so a stale preview error never lingers across selections.

### Verification

- `npm run lint` clean (TypeScript, all configs).
- `npm run lint:eslint` clean.
- `prettier --check` clean.
- Scoped suite passes: mobile App (95).

Other open items from the latest review pass:
- coderabbitai's re-review request was acknowledged but no actionable comments have arrived yet.
- The previous round's nitpicks (`'Marketplace not configured'` wording, `handleAutoRunCloseSetup` resetting `showMarketplaceSheet`) are non-blocking — happy to fold them in if either bot upgrades them.

chr1syy:
## Folder prefill fix verified — unblocking

Re-tested 2026-05-07 against the soak build (`e4585946 fix(web): single-segment slug for mobile marketplace import folder` + `60f0ce67 fix(web): split preview vs import error state in marketplace sheet`).

### What's fixed

- ✅ **Import folder prefill is now single-segment.** A category-namespaced playbook (id like `category/name`) prefills as `name` only. Tapping **Import Playbook** without editing the input now succeeds; the imported playbook appears in the AutoRun document list afterward. The validator-vs-prefill contradiction from yesterday is resolved.

Removing the `blocking` label.

### Non-blocking follow-up (mobile/web parity gap)

Desktop has an **"Add Docs" / Docs Overview** affordance for discovering and adding AutoRun documents. The mobile/web equivalent is missing — users on mobile have to navigate directly to the Playbook Exchange or to already-imported docs, with no equivalent overview surface. The Playbook Exchange (this PR) is one entry point; the Docs Overview is the broader contextualizing surface around it.

Added as Gap 3 to the running mobile/web parity follow-ups gist (alongside PR #946's launch-affordance gaps): https://gist.github.com/chr1syy/67630166ff5217a97d7b1fad91201f3b

Suggested direction in the gist: mirror the desktop component on the webui, reuse the WS-side IPC via the same shared-service pattern this PR established (`marketplace-service.ts`), and have the Playbook Exchange become one item inside the broader overview rather than the only entry point.

### Other items still untested in this pass (deferred from yesterday's plan)

- Step 6 — desktop `MarketplaceModal` regression sanity (proves the IPC handler refactor didn't break the desktop path).
- Step 7 — SSH-remote import lands on the remote AutoRun folder.
- Finding 2 from yesterday — `Run` and `Playbooks` bottom-nav still open the same screen. Was filed as non-blocking and not addressed in this round; decide whether to ship as-is or push a follow-up.

chr1syy:
Merged \`upstream/rc\` again. Pushed \`7bea26d3\`.

### Conflict resolution

Single conflicted file: \`src/main/ipc/handlers/marketplace.ts\`. Upstream introduced marketplace compatibility gating (\`isCompatible\` against \`minMaestroVersion\`) inside the inline IPC import body, while this PR had refactored that body into a thin wrapper around \`src/main/services/marketplace-service.ts\`.

Resolved by:
- Keeping the thin-wrapper IPC handler (no behavior loss — service still implements the full import flow, including the partial-import doc filtering and ENOENT-only swallowing from the previous review round).
- **Moving the compatibility check into the service layer** (\`importMarketplacePlaybook\`), right after the \`marketplacePlaybook\` lookup. Both the IPC handler and the WebSocket \`marketplace_import_playbook\` callback now share the same defense-in-depth gate against incompatible \`minMaestroVersion\` — so a future CLI / deep-link / programmatic / mobile WS bypass attempt all hit the same throw-with-clear-message path.

The IPC handler imports pulled in by upstream (\`MarketplaceFetchError\`, \`MarketplaceManifest\`, etc.) were dropped; the service file already owns those.

### Verification

- \`npm run lint\` clean (TypeScript, all configs).
- \`npm run lint:eslint\` clean.
- \`prettier --check\` clean on touched files.
- Scoped suites pass: marketplace IPC, message handlers, web-server-factory, mobile App, marketplace-compatibility — 344 tests total.

PR is back to \`MERGEABLE\` / \`CLEAN\`.

chr1syy:
Addressed all seven CodeRabbit findings in `ba1017b5` (pushed to `feat/web-playbook-exchange`, so it appears on both this PR and #947). Summary:

**Service layer (`marketplace-service.ts`)**
- Import path now falls back to a stale cache when `fetchManifest()` fails, mirroring the browse path so a visible playbook can't disappear at import time.
- Refuse to persist a playbook when every document failed to write (only when the playbook actually had documents — asset-only imports still work). Throws `MarketplaceImportError` before the playbook file is touched.

**WebSocket handlers (`messageHandlers.ts`)**
- `isUntrustedLocalPath` now also rejects `.` / `..` segments and any embedded backslash, on top of the existing absolute / `~` / Windows-drive-prefix checks.
- All four marketplace catch blocks (`get_manifest`, `get_document`, `get_readme`, `import_playbook`) now route through a new `reportMarketplaceHandlerError` helper that calls `captureException` before sending the typed failure. Mirrors the existing `reportHandlerError` pattern while preserving the typed-result protocol mobile clients wait on.

**Web server factory (`web-server-factory.ts`)**
- Marketplace-import catch now reports the original exception to Sentry with `{ operation, sessionId, playbookId, targetFolderName }` context before returning `{ success: false }`.

**Mobile sheet (`MarketplaceSheet.tsx`)**
- Manifest / README / document / import catch blocks `console.error` the underlying exception so the original cause is visible in browser devtools (web bundle has no Sentry — same pattern used in `AutoRunDocumentViewer`).
- Returning to the README view from a doc preview now clears `previewError` and re-fetches the README via a new `loadReadmeFor` helper. Previously the sheet could stay stuck on a stale "No README available" / fetch error after the user tapped a doc and went back.

**Tests** (`marketplace.test.ts`, `messageHandlers.test.ts`)
- New: empty-import guard fires and no playbook is persisted.
- New: stale-cache fallback during import.
- New: parameterized `isUntrustedLocalPath` traversal/backslash rejection via the typed result.

All 205 tests in the scoped suites pass; lint + prettier clean.

pedramamini:
Reviewed and ready to land in spirit, but two things need to happen before merge:

**1. Rebase on current `rc`.** This branch is now in conflict with `rc` after #946 / #974 / #976 / others landed — the web-server scaffolding (`web-server-factory.ts`, `messageHandlers.ts`, `CallbackRegistry.ts`, `WebServer.ts`, `AutoRunSetupSheet.tsx`, etc.) overlaps with what #946 added. Please rebase onto `origin/rc` and resolve.

**2. Fix the SSH silent-downgrade in `src/main/web-server/web-server-factory.ts`.** The desktop IPC (`src/main/ipc/handlers/marketplace.ts:137-139`) explicitly throws `SSH remote not found or disabled` when a session opted into SSH but the remote can't be resolved — exactly the loud-fail pattern called out in CLAUDE.md ("when the user enabled SSH but the configured remote can't be resolved, fail loudly").

The new WS callback path (`resolveSessionSshConfig` → marketplace import) returns `undefined` in three failure modes (missing remoteId, remote not in store, `enabled: false`) and then proceeds to import locally. A user with an SSH-enabled session whose remote is misconfigured will silently land the playbook on the desktop's local filesystem instead of the remote host. Tests for this exact regression were added on the desktop side (`should fail loudly when SSH remote ID does not resolve`) but the equivalent guard is absent on the WS path.

Also: `resolveSessionSshConfig` reads `session.sshRemoteId || session.sessionSshRemoteConfig?.remoteId` but never inspects `sessionSshRemoteConfig.enabled`. `AgentSshRemoteConfig` (`src/shared/types.ts:626`) treats `enabled` as the source of truth — a session with `enabled: false` and a populated `remoteId` would be wrongly treated as remote.

**Fix:** distinguish "no SSH configured" from "SSH configured but unresolvable" in the WS callback. Return `{ success: false, error: 'SSH remote not found or disabled' }` for the latter. Gate the resolution on `sessionSshRemoteConfig?.enabled === true` before pulling `remoteId`. Add a WS-side test mirroring the desktop unresolvable-remote regression test.

Once those two are in I'll merge it.

pedramamini:
@chr1syy ping — see comment above for the rebase + SSH fix needed before this can land. Same applies to #973 which is stacked on top.

chr1syy:
@pedramamini both asks addressed in \`92576d97\`.

### 1. Rebase / merge state

Branch is now in sync with \`upstream/rc\` (round 4 merge in \`8eb5b558\`). \`gh pr view\` reports \`mergeable: MERGEABLE\`, \`mergeStateStatus: CLEAN\`. No outstanding conflicts.

### 2. SSH silent-downgrade in \`web-server-factory.ts\`

Reworked \`resolveSessionSshConfig\` to mirror the desktop IPC handler's loud-fail behavior:

- **Source of truth:** \`sessionSshRemoteConfig.enabled === true\` is now the gate. A session with \`enabled: false\` and a populated \`remoteId\` is treated as no-SSH (passes through to local import). The legacy top-level \`sshRemoteId\` field still implies enabled when present, matching how the rest of the codebase reads it (\`sessionHelpers.ts:384\`).
- **Three unresolvable conditions now throw \`'SSH remote not found or disabled'\`:**
  1. \`enabled: true\` but \`remoteId\` is null/missing.
  2. \`remoteId\` doesn't appear in \`sshRemotes\`.
  3. Matching \`sshRemotes\` entry has \`enabled: false\`.
- **Pre-resolution in the import callback:** SSH is resolved before the import try-block. Throws are caught and returned as \`{ success: false, error }\` without routing through \`captureException\` — user misconfiguration isn't an exceptional bug.

### Tests

Added \`importMarketplacePlaybookCallback behavior\` block (4 specs) in \`src/__tests__/main/web-server/web-server-factory.test.ts\`:
- \`enabled: true\` + remoteId not in store → loud fail.
- \`enabled: true\` + matching entry has \`enabled: false\` → loud fail.
- \`enabled: true\` + \`remoteId: null\` → loud fail.
- \`enabled: false\` + populated remoteId → \`importMarketplacePlaybook\` called with \`sshConfig: undefined\` (correctly treated as local).

Mocked \`marketplace-service.importMarketplacePlaybook\` and added \`captureException\` / \`electron.app\`+\`BrowserWindow\` to the existing electron mock so the import callback is testable end-to-end.

### Verification

- \`npm run lint\` clean (TypeScript, all configs).
- \`npm run lint:eslint\` clean.
- \`prettier --check\` clean.
- Scoped suites pass: web-server-factory (48 + 4 new), message handlers, marketplace IPC, mobile App — 347 tests total.

PR #948, opened 2026-05-03
feat(coworking): expose terminal scrollback to all four agents via opt-in MCP installer

chr1syy:
## Summary

Adds a **Coworking** Encore Feature: agents can read terminal scrollback **and inspect/drive in-app browser tabs** on demand via a per-agent MCP server. Opt-in via Settings -> Encore Features -> Coworking -> Install for [agent].

- One-shot installer writes a `maestro-coworking` MCP entry to each agent's **user-level** config (`~/.claude.json`, `~/.codex/config.toml`, `~/.config/opencode/opencode.json`, `~/.factory/mcp.json`). Install once, persists. Node is resolved to an absolute path at install time (falls back to `node` on PATH) so GUI-launched agents under nvm/fnm/volta still find a runtime.
- Each terminal tab gets a click-to-copy `term:N` pill so users can address them by name.
- **Terminal tools:** `list_terminals()`, `read_terminal(id, lines?)`.
- **Browser read tools:** `list_browsers()`, `get_browser_url(id)`, `read_browser(id, format?, maxChars?)` where `format` is `text | innerText | html`.
- **Browser interaction tools (opt-in per agent):** `browser_navigate(id, url)`, `browser_back/forward/reload/stop(id)`, `browser_click(id, selector)`, `browser_type(id, selector, text)`, `browser_eval(id, code)`, `browser_screenshot(id)`.
- Off by default. When the Encore flag is off the Settings panel and pills are hidden. Browser interaction is **additionally** gated behind a per-agent "Interaction" toggle (default off), enforced in the main-process bridge so a hand-configured MCP client cannot bypass the UI.

## Architecture

- **Bundled MCP server** - hand-rolled stdio server (no SDK runtime dep) written as `<userData>/coworking-mcp-server.js` on every `app.ready`. Stable absolute path survives Maestro upgrades. Advertises all tools; the agent's MCP client owns the subprocess lifecycle.
- **IPC bridge** - Unix-domain-socket on POSIX, named pipe on Windows, at a stable per-userData path. Newline-delimited JSON-RPC. Per-connection sessionId binding (env + ppid fallback) so an agent only ever sees its own session.
- **Registry** - main-process mirror of every session's terminal AND browser tabs, keyed by sessionId (no "active session" concept, for privacy). Browser tabs get a stable, monotonic, never-reused `browser:N` id assigned by the registry (no renderer-side schema change needed).
- **Browser op resolver** - read text/HTML and all interaction need the LIVE webview, so they round-trip main -> renderer over a unique, sender-validated `coworking:requestBrowserOp` channel (8s timeout, under the MCP server's 10s per-RPC cap). `list_browsers`/`get_browser_url` are metadata-only and never touch a webview.

### Browser resolution decision (activation vs background mount)

rc already keeps the active agent's browser tabs mounted (visible or hidden) via `useBrowserTabMounting` (the `browserTabKeepAlive` setting). We chose **(A') prefer an already-mounted webview handle; activate the target tab only as a fallback**:

- The responder first looks up the target tab in the per-tab `BrowserTabView` handle map (lifted from `MainPanelContent` to `MainPanel`). If the tab is already mounted (even hidden via keep-alive), we use it directly with **no focus change**.
- If it isn't mounted, we activate it (which mounts it), run the op, then restore the previously active browser tab.
- The live webview only exists for the **active agent**. For a non-focused agent, an opt-in, LRU-capped **background webview host** mounts the tab off-screen so read/interaction still work without stealing focus; with the host disabled the responder returns a clear `ok:false` (metadata tools always work cross-session). See **Hardening follow-ups**.

Hidden multi-webview mounting for non-focused sessions (option B) is now available as an **opt-in, capped** background host (default off) - see **Hardening follow-ups**. The focused-agent already-mounted fast path remains the default and the only path when the host is off.

## Files (high level)

- `src/shared/coworkingBrowser.ts` - single source of truth for the browser op/result/input/entry contract (shared across main, preload, renderer).
- `src/main/coworking/*` - registry (browser records + `browser:N` ids + interaction permission + agentType), tools (`listBrowsers`/`getBrowserUrl`/`readBrowser`/`browserInteract` + resolver), bridge dispatch + `validateInteractionOp`, server-script (browser tools in `TOOLS` + `tools/call`), audit log.
- `src/main/ipc/handlers/coworking.ts`, `src/main/preload/coworking.ts` - browser sync + sender-validated browser-op round-trip.
- `src/renderer/...` - `useCoworkingRegistrySync` mirrors browser tabs + per-agent interaction flag + hides flagged tabs; `useCoworkingBrowserResponder` resolves the live (or background) webview and runs ops; `BrowserTabView` gains `extract`/`getMeta`/`navigate`/`reload`/`stop`/`executeJavaScript`/`capturePage` + a hide-from-agent toggle; `CoworkingSetup` gets per-agent Interaction + confirm-policy controls and a global background-browsing toggle.
- Settings: `coworkingBrowserInteraction: string[]`, `coworkingBrowserInteractionConfirm: Record<string, policy>`, `coworkingBackgroundBrowsers` + `...Limit` across settingsStore/defaults/settingsMetadata/useSettings.

## Hardening follow-ups (this update)

Layered on top of the read + interaction tools above:

- **Per-call approval for state-changing ops.** Before any `browser_navigate/back/forward/reload/stop/click/type/eval/screenshot`, the renderer can require explicit user approval via a cancel-safe confirm dialog. Policy is per-agent (`off | dangerous | all`, default `dangerous` = confirm `navigate`/`eval`), surfaced in the Coworking Setup panel. Interaction-op timeouts were raised (server bridge 330s / main resolver 300s) so a human has time to respond; reads stay at 8s.
- **Per-tab "hide from agent".** A persisted `hiddenFromAgent` flag on `BrowserTab` with an eye toggle in the browser toolbar. Hidden tabs are filtered out of `list_browsers` / `read_browser` / interaction at the registry-sync boundary, so agents can neither see nor address them.
- **Browser-tool audit log.** Every dispatched browser op is recorded (timestamp, sessionId, agentType, tool, op kind, redacted arg summary, outcome `ok | error | denied`) to the system log + a JSONL under `userData`. URLs/selectors are truncated and page content / eval code is never logged (lengths only).
- **Cross-session live access (opt-in).** A capped, LRU-evicted **background webview host** (`coworkingBackgroundBrowsers`, default off; `coworkingBackgroundBrowsersLimit` default 2, clamped 1-10) mounts a non-focused agent's tab off-screen - its own `partition` preserved, `isActive=false`, `aria-hidden`, pointer-events off, no focus steal - so read + interaction work without switching the user's active agent. Disabling the setting releases every hidden webview. `capturePage` on a hidden webview returns a clear `ok:false` rather than a blank image.
- **Cleanup.** Removed the unused `@modelcontextprotocol/sdk` dependency (see Notes).

New files: `coworkingApprovalStore.ts`, `CoworkingApprovalHost.tsx`, `CoworkingBackgroundBrowsers.tsx`, `coworkingBackgroundBrowserStore.ts`, `coworking-audit.ts`.

## Notes

- `@modelcontextprotocol/sdk` was declared but unused (the server is hand-rolled); it has been **removed** from `package.json` and the lockfile regenerated (`npm install --package-lock-only`).

## Test plan

- [x] Typecheck (all three tsconfigs) + ESLint on changed files green.
- [x] Vitest: **126 coworking tests** pass (registry incl. `browser:N` ids / scoping / interaction permission + hidden-tab filtering + agentType; tools + read truncation; bridge dispatch + interaction gate + op validation + audit; renderer responder op->handle mapping incl. the cross-session background path; per-call approval store + policy; background-host LRU / cap / clear).
- [x] End-to-end MCP smoke: drove the real generated server over stdio against a mock bridge - `initialize` / `tools/list` (all 12 tools) / `tools/call` for `list_browsers`, `read_browser`, plus an interaction op that **proceeds when approved** and **surfaces a decline as an MCP error**.
- [ ] Manual in-app (requires a desktop build): toggle Coworking on, Install for an agent, open >=2 browser tabs, ask the agent to `list_browsers` / `read_browser` / (with Interaction on) `browser_navigate` + `browser_click`, confirm the approval dialog, and confirm the page changes.

## Deferred (out of scope for this PR)

- Subscriptions / push notifications, cross-tab references.
- SSH-mode coworking (agent runs remote, terminals/browsers are local).
- Inline directives fallback for any future non-MCP agent.

## Summary by CodeRabbit

* **New Features**
  * Added a Coworking settings panel to install/uninstall and check MCP setup for supported agents.
  * Introduced session-scoped coworking support for terminals and browsers, including optional browser interaction, background browsing, and confirmation prompts.
  * Added coworking “term:N” pills on terminal tabs with one-click copy.
  * Added browser tool audit logging and “hide/show from agents” controls.
* **Bug Fixes**
  * Strengthened the coworking bridge handshake and RPC validation to prevent cross-session data leakage.
  * Improved stable coworking ID assignment and terminal ordering, including legacy session compatibility and safer buffer scoping/truncation.


pedramamini:
Hey @chr1syy — thanks so much for putting this together! The Coworking design (one-shot user-level MCP install + `term:N` pills + scoped read-only tools) is a really nice shape, and I appreciate the deferred-scope breakdown.

Heads up: the branch currently has merge conflicts against `rc` (`mergeStateStatus: DIRTY`). Could you kindly rebase onto the latest `rc` and resolve the conflicts so we can get a clean merge? Once that's green I'll do a full review pass.

Thanks again!

chr1syy:
## Blocker: `list_terminals` / `read_terminal` scope is bound to the UI's focused AI tab, not the caller agent

Found while soaking #948 into `0.16.0-RC-build` for a 0.16.17-RC test build. Install round-trip and end-to-end tool execution work cleanly; this finding is about the **scoping model** of the MCP tools.

### Repro

1. Coworking flag on, install for **claude-code** (session A in this repro).
2. Open a terminal in session A → pill `term:3` (after some open/close churn).
3. Open a terminal in **session B** (a different agent in the Left Bar).
4. From within claude-code in session A, run a tool with a delay so the user can switch UI focus during the call window:
   ```bash
   sleep 8 && date
   ```
5. During the sleep, click session B's AI tab in the Left Bar so it becomes the focused tab.
6. After the sleep, call `mcp__maestro-coworking__list_terminals` from claude-code (still session A — its agent process didn't move).

**Observed:** `list_terminals` returns `{id: "term:1", title: "Terminal 1"}` — that's session **B**'s terminal, not session A's `term:3`.
**Expected:** session A's terminal(s) only — claude-code in session A should never see session B's terminals regardless of UI focus.

### Why this is a privacy issue

The PR description says \"Scoped to the **active AI tab's session only**.\" In practice that means scoped to **whichever AI tab the user is currently focused on**, not to the calling agent. With Maestro's whole point being long-running agents that work in the background, this is exploitable:

- Agent A keeps running while the user switches focus to session B (very common workflow).
- Agent A calls `list_terminals` / `read_terminal` during that window.
- Agent A silently reads scrollback from session B — different agent, possibly different project, possibly secrets in the buffer.
- No UI signal that this is happening; nothing in the buffer indicates it was read by a foreign agent.

### Architectural cause (from PR description)

> Renderer ↔ main — renderer pushes active-session terminal records via \`coworking:syncSessionTerminals\`; main asks renderer for buffers via webContents.send → responseChannel pattern.

The registry stores a **singleton \"current\" snapshot** that gets overwritten on every focus change. Any agent reading the registry reads whatever was last pushed. There's no per-session keying.

### Suggested fix

Scope must be **caller-bound**, not focus-bound:

1. **Bind MCP server to its session at install time.** Each agent's `maestro-coworking` MCP entry should carry its session ID (env var or arg), so the bridge can identify the caller.
2. **Per-session registry**, keyed by session ID. `coworking:syncSessionTerminals` should push for one session at a time without clobbering others (e.g., `Map<sessionId, TerminalRecord[]>`).
3. **`list_terminals` / `read_terminal` lookups** use the caller's session ID from the bridge handshake to read only that session's slice of the registry.
4. Optional follow-up: a UI signal (toast / pill flash) when an agent reads a terminal, so the user is aware of the read regardless of focus state.

### Other findings from the same soak (non-blocking, FYI)

- ✅ Default-off behavior correct.
- ✅ Install diff for claude-code is minimal-additive — only the `mcpServers.maestro-coworking` block added, plus the syntactically-required trailing comma. Zero collateral edits.
- ✅ MCP server starts and tools function end-to-end (`read_terminal` returned accurate scrollback).
- ✅ Stable IDs — closing `term:1` did not recycle the ID for newly opened terminals.
- ⚠️ `list_terminals` returns `cwd: \"\"` (empty string) — schema description promises a cwd value.
- ⚠️ UX papercut: tab display title renumbers positionally (\"Terminal 1\") while the pill exposes the stable monotonic handle (`term:3`); they can diverge and confuse users when an agent references a `term:N` that doesn't match the visible tab number.

chr1syy:
@greptileai re-review please

chr1syy:
## Regression — `list_terminals` / `read_terminal` now return nothing for the caller's own session

Re-tested PR #948 against today's build (`d035932f fix(coworking): bind MCP tools to caller session, not UI focus`). The privacy bug from yesterday's report is plugged — but the fix over-corrected: an agent can no longer read **its own session's** terminals.

### Repro

1. Settings → Encore Features → Coworking on, install for **claude-code**. (`~/.claude.json` install diff is still clean-additive — that part is fine.)
2. Open one or more terminal tabs in claude-code's session — verify they exist visually with `term:N` pills.
3. From within claude-code, call `mcp__maestro-coworking__list_terminals`.

**Observed:**

```json
{ "terminals": [] }
```

`read_terminal({ id: "term:N" })` for any visible `N` returns:

> `MCP error -32000: coworking tools: terminal 'term:N' not found in your session (it may have been closed)`

**Expected:** the caller's own visible terminals listed.

### Tried to rule out a sync race

- Re-probed `list_terminals` multiple times — consistently empty.
- Asked the user to refocus a terminal tab (would trigger any focus-change push) — still empty.
- Asked the user to close all terminals and open a single fresh one (would trigger the add path) — still empty.
- The error string \"not found in **your session**\" confirms the bridge is correctly identifying the caller's session — it just has nothing in the registry under that key.

### Diagnosis

The fix changed the *read* side (caller-bound lookup) but appears to have left the *write* side incompletely wired. Two candidate causes:

1. **Session-id mismatch between install-time capture and live agent.** The MCP server's `~/.claude.json` entry has no per-session identifier (just the shared socket). If the bridge identifies callers by something captured at install time but the running claude-code agent has a different session ID (e.g., after an app restart between install and use), the lookup will always miss. Possible if the claude-code subprocess inherits a session id that no longer matches what the renderer pushes under.
2. **Per-session push not firing for the caller's session.** `coworking:syncSessionTerminals` may only fire under specific lifecycle triggers (focus change, etc.) and/or push under a key that doesn't match what the read side looks up.

A quick way to disambiguate would be a log line in the bridge for both sides — \"received sync push for session=X with N terminals\" and \"list_terminals lookup for session=Y\" — to see whether X and Y match.

### Severity

**Blocker.** The feature is now non-functional for its primary use case. Yesterday's behavior at least let the agent read its own session's terminals (the bug was the focus-driven cross-session leak); today's behavior reads nothing. The new tool description (\"You will never see terminals belonging to other Maestro sessions, regardless of which tab the user has focused\") accurately describes what's happening, but the inverse — \"you WILL see your own session's terminals\" — isn't holding.

### Side note from soak testing — `term:N` ID reuse

User feedback during this re-test: closing terminals leaves their `term:N` slots unused; the next-opened terminal jumps to a higher N (we got to `term:7` after a series of opens/closes). The display title (\"Terminal 1\" — positional, since only one is open) and the pill (`term:7`) diverge by 6.

This was a non-blocking nit in yesterday's report. Two reasonable directions:

- **Reuse closed IDs** — fixes the divergence, but breaks any agent-side cache of a previously-read `term:N` (the same id now points to a different terminal). Recommend against unless agents are explicitly told to never cache.
- **Align display title to the stable pill ID** — show the tab as \"Terminal N\" where N is the pill number. Eliminates divergence without breaking the stability contract.

Probably worth deciding direction now while the registry is being reworked anyway. Not part of this blocker, just rolling it forward.

chr1syy:
@coderabbitai review

chr1syy:
@greptileai review

chr1syy:
Closing as superseded by #1160 (`feat/agentic-coworking-hardened`), which carries this work forward with browser-tooling hardening, session-persistence controls, and additional agent tools. The outstanding review threads here (async pid resolution to avoid blocking the Electron main loop, URL scrubbing in the browser audit log, async audit writes) are being addressed in #1160. No need to keep two coworking branches open.

PR #950, opened 2026-05-03
fix(compact): surface real error in toast instead of swallowing it

### LINKED ISSUE #949: Bug: Compact and Continue always fails with COMPACTION FAILED error; detai...
## Summary
Compact and Continue always fails with COMPACTION FAILED error; details tab unresponsive

## Environment
- Maestro version: 0.16.16
- Operating system: Windows (Windows 10 Home, 10.0.26200)

## Steps to Reproduce
1. Open any assistant session (Claude or Codex) with enough context to trigger compaction. 2. Click 'Compact and Continue'. 3. Observe immediate failure toast in bottom-right corner. 4. Attempt to click the details tab that appears.

## Expected Behavior
Compact and Continue should successfully compact the context, allowing the conversation to continue. If it fails, the referenced details tab should be clickable and show actionable error information.

## Actual Behavior
Compact and Continue immediately fails with a toast notification: 'COMPACTION FAILED - Failed to compact context. Check the tab for details.' The details tab that appears is completely unresponsive — clicking it does nothing and it shows no content.

## Additional Context
Issue reproduces on both Claude and Codex assistants. The double failure (compaction itself + broken details tab) means there is no way for the user to diagnose or work around the problem.

pedramamini:
## Summary

When `Compact & Continue` failed, the toast told users to *"Check the tab for details"* — but no tab actually rendered details. The source tab's `SummarizeProgressOverlay` only renders while `isSummarizing` (state `'summarizing'`) is true, so the post-failure error state never reaches the screen. Meanwhile, a bare `catch` in `summarizeContext` discarded the underlying error and re-threw a generic `"Context summarization failed"`, hiding the real cause from both the toast and Sentry.

This is the path #949 is hitting: the user sees `"Failed to compact context. Check the tab for details."` and clicking the source tab shows nothing.

## Changes

- `src/renderer/services/contextSummarizer.ts` — preserve the underlying `Error` (rethrow original) so callers see the real cause (`Agent X is not available`, spawn `ENAMETOOLONG`, etc.) instead of a generic wrapper string.
- `src/renderer/hooks/agent/useSummarizeAndContinue.ts` — read the actual error out of `operationStore` (and `err.message` in the unexpected-throw branch) and put it in the toast message. The toast no longer references a phantom "tab for details".
- `src/__tests__/renderer/services/contextSummarizer.test.ts` — update the IPC-error test to assert the inner error propagates.

## Test plan

- [x] `vitest run src/__tests__/renderer/services/contextSummarizer src/__tests__/renderer/hooks/useSummarizeHandler` — all 39 tests pass
- [x] `tsc -p tsconfig.lint.json` clean
- [x] `eslint src/renderer/services/contextSummarizer.ts src/renderer/hooks/agent/useSummarizeAndContinue.ts` clean
- [x] `prettier --check` on touched files clean
- [ ] Manual: trigger a failing compaction (e.g. with a configured-but-unavailable agent) and confirm the toast shows `Failed to compact context: <real reason>` rather than `Check the tab for details`

closes #949

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error reporting for context summarization and summarize-and-continue flows: user-facing failure messages now surface the original, trimmed error details (including per-tab errors) instead of a generic message.
* **Tests**
  * Updated tests to assert and verify propagation of the original error messages for IPC and summarization failures.


chr1syy (changes requested):
Thanks for tackling this — the diagnosis is spot-on. I confirmed both bugs are still present on `main` *and* `rc` (the bare `catch` in `summarizeContext` and the toast pointing at a `SummarizeProgressOverlay` that only renders while `state === 'summarizing'`). The fix matches the codebase's "don't silently swallow exceptions" guidance in `CLAUDE.md`.

A few things to clear before this is mergeable.

### Blockers

**1. Merge conflicts on both branches.** `mergeable: CONFLICTING`. I reproduced locally:

- Against `main`: conflict in `src/__tests__/renderer/services/contextSummarizer.test.ts` because a new test `'should handle an empty groomed response'` was added directly above the renamed test.
- Against `rc` (if retargeted): conflicts in `useSummarizeAndContinue.ts` — `rc` swapped `console.error` → `logger.error` on the same line you touch — plus the async prompt-loading refactor that landed in `contextSummarizer.ts`.

Both sets of conflicts are mechanical.

**2. `lint-and-format` CI failing — but not from this PR.** `prettier --check .` flagged `docs/releases.md` and `src/__tests__/renderer/hooks/useAgentListeners.test.ts`. Both pass at current `main` HEAD and the PR's three touched files pass prettier cleanly. This is stale-base inheritance — a rebase + `prettier --write .` clears it.

**3. `String(err)` fallback can leak `"[object Object]"` / `"null"` into the toast.** Both Greptile and CodeRabbit flagged this and they're right. At `useSummarizeAndContinue.ts:425`:

```ts
const detail = err instanceof Error ? err.message : String(err);
```

`String({ foo: 1 })` → `"[object Object]"` (truthy, slips past the ternary). `String(null)` → `"null"` (also truthy). Suggested shape:

```ts
const detail =
  err instanceof Error
    ? err.message
    : typeof err === 'string'
      ? err
      : typeof err === 'object' && err !== null && 'message' in err
        ? String((err as { message?: unknown }).message ?? '')
        : '';
```

### Base branch — likely should be `rc`

The PR currently targets `main`, but recent bug-fix PRs all land on `rc` (#1066, #1067, #1043, #1023, #1021) so they ride the next release. Functionally the rewritten lines are identical on both branches — this is convention, not correctness.

### Non-blocking observations

- **`contextSummarizer.ts` fallback has the same `[object Object]` risk.** `throw new Error(typeof err === 'string' ? err : 'Context summarization failed')` silently drops detail for non-Error non-string throws. Worth tightening alongside the toast fix.
- **No length cap on `detail`.** A multi-line spawn error or long `Error.message` produces an ugly oversized toast. A short cap (e.g. ~200 chars + ellipsis) would be friendlier — and matches how other Maestro toasts handle long error text.
- **Reaching into operationStore for the error.** `useOperationStore.getState().summarizeStates.get(targetTabId)?.error` works because the store update happens synchronously inside `startSummarize` before it resolves null, but it couples the toast to store internals. Returning the error from `startSummarize` would be cleaner. Fine to defer.

### Verdict

Right idea, right files, real bug. Ship-blocking items are (1) rebase to clear conflicts + stale lint, (2) the `String(err)` hardening. I'd retarget to `rc` while you're rebasing.

PR #953, opened 2026-05-04
docs(cue): document Command nodes (action: command) in core prompt

chr1syy:
## Summary

The agent-facing core prompt (`src/prompts/_maestro-cue.md`, bundled as `/opt/Maestro/resources/prompts/core/_maestro-cue.md`) had no documentation for **Command nodes** — the `action: command` shell-or-`maestro-cli` subscriptions exposed in the Pipeline Editor. Without this, agents authoring `.maestro/cue.yaml` had no reliable way to emit a Command node short of round-tripping through the UI to reverse-engineer the schema. The maestro-cue docs (`docs/maestro-cue*.md`) and `llms.txt` are also silent on the schema.

This PR adds a complete Command-node reference, derived from reading `src/main/cue/cue-shell-executor.ts`, `cue-cli-executor.ts`, `cue-dispatch-service.ts`, `cue-run-manager.ts`, `cue-completion-service.ts`, `cue-engine.ts`, `src/main/index.ts:490-580`, `src/shared/cue/contracts.ts`, `src/main/cue/config/cue-config-normalizer.ts`, and `src/main/cue/config/cue-config-validator.ts`.

What's added to `src/prompts/_maestro-cue.md` (282 → 433 lines, +155 / -3):

- **New "Command Nodes (`action: command`)" section** between Pipeline Topologies and Template Variables. Covers:
  - Subscription-with-`action: command` schema (no separate top-level YAML key, no `event: command` type, no separate node graph)
  - Full `mode: shell` and `mode: cli` field set with validator rules
  - Fields that intentionally do NOT exist (`working_dir`, `timeout`, `env`, shell-selection, exit-code-handling, retry, output-cap) and what they actually use
  - SSH behavior asymmetry: shell honors per-session SSH wrapping via `bash -c`, CLI mode is local-only by design
  - Trigger compatibility: all 9 event types can fire a Command node; only `fan_out` is rejected
  - Output exposure via the unified `{{CUE_SOURCE_OUTPUT}}` channel through `agent.completed` (no separate `{{CUE_COMMAND_OUTPUT}}`)
  - The `source_sub` chaining nuance for shared sessions (since command nodes have no session of their own, chains need `source_sub` to disambiguate from peer agent runs)
  - Worked `app.startup → shell → agent` example
- **Topology list:** Command node added as a first-class topology; legacy `cli_output` field annotated as deprecated in favor of `action: command` with `mode: cli`.
- **Authoring Guidance:** new step pointing at the Command Nodes section before the external-doc fallback.
- **Recipes:** two new entries — shell-command → agent-chain (using `source_sub` to pin), and `agent.completed` → `mode: cli` forwarding.

## Test plan

- [x] `npx prettier --check src/prompts/_maestro-cue.md` passes
- [x] Section header order verified intact via grep
- [ ] No code-level changes; this is a documentation-only update to a bundled prompt. No test changes.

## Summary by CodeRabbit

* **Documentation**
  * Expanded Command node documentation with behavior specs, execution modes (shell vs local CLI), and chaining guidance
  * Added schema/validation requirements and updated validator guidance
  * Deprecated the standalone CLI output field in favor of Command nodes
  * Included worked YAML examples and updated authoring steps for Command-node usage


pedramamini:
Hey @chr1syy — thanks for taking the time to write this up! The Command-node section is exactly the reference agents have been missing, and the schema, validator rules, SSH asymmetry, and `source_sub` chaining nuance all check out cleanly against `cue-config-validator.ts`, `cue-shell-executor.ts`, and `contracts.ts`.

Before merging I'd like to get a few accuracy nits fixed since this prompt is what agents read when authoring `.maestro/cue.yaml` — wrong claims here propagate into bad configs:

1. **`{{CUE_SOURCE_OUTPUT}}` is truncated, not full.** The "Max output size" row says _"Downstream `{{CUE_SOURCE_OUTPUT}}` sees the full captured stdout."_ It doesn't — `SOURCE_OUTPUT_MAX_CHARS = 5000` (in `src/main/cue/cue-output-filter.ts`) is applied before the variable reaches the downstream prompt. See `cue-run-manager.ts:447` (`sliceHeadByChars`) and `cue-completion-service.ts:167` (`sliceTailByChars`). The 30 000 / 100 000 / 10 000 numbers in the same row are correct for their respective channels — it's only the `{{CUE_SOURCE_OUTPUT}}` claim that needs fixing. Worth calling out the head-vs-tail slicing too, since they differ between the run-manager output-prompt path (head) and the completion-service downstream path (tail).

2. **`{{CUE_SOURCE_EXIT_CODE}}` null condition.** Doc says _"or `null` on spawn failure."_ It's also `null` when the run is killed by signal — e.g. timeout (`SIGTERM` → `SIGKILL`) — not just spawn failure. Suggest: _"or `null` on spawn failure or signal-kill (e.g. timeout)"_.

3. **`source_sub` array form.** The schema callout (`source_sub: <command-sub-name>`) implies string-only, but the validator accepts `string | string[]` (`cue-config-validator.ts:341–358`). Worth a one-liner so agents know they can OR multiple upstream subs.

(For what it's worth, Greptile's separate "should be `dispatch` not `send`" finding is wrong — the validator hardcodes `command: 'send'` at `cue-config-validator.ts:75–79`, so your doc is correct on that point. And CodeRabbit's `agent_id`-required nitpick on the recipes is also fine to ignore — the validator doesn't require `agent_id` on any subscription.)

Could you push a fixup for the three items above? Once those land I'll approve.

chr1syy:
Thanks for the careful review @pedramamini — pushed fixup `1b86cde8` covering all three items:

1. **`{{CUE_SOURCE_OUTPUT}}` truncation.** Updated both the "Max output size" row and the variables table. The Max-output row now leads with `SOURCE_OUTPUT_MAX_CHARS = 5000` from `cue-output-filter.ts` and explicitly distinguishes the head-slice path (`cue-run-manager.ts` output-prompt) from the tail-slice path (`cue-completion-service.ts` downstream chain). The 30K/100K/10K numbers are kept as separate, independent caps for their respective channels. The variables table no longer claims "(full)" — it points back to the Max-output row.

2. **`{{CUE_SOURCE_EXIT_CODE}}` null condition.** Now reads "Numeric exit code (or `null` on spawn failure or signal-kill, e.g. timeout)."

3. **`source_sub` array form.** Appended a one-liner to the chaining-nuance callout: `source_sub` accepts a single name or an array (`source_sub: [step-a, step-b]`) to OR-match multiple upstream subs.

Confirmed your guidance on the other two findings — left them alone:

- Greptile's `dispatch`/`send` finding: `cue-config-validator.ts:75-79` hardcodes `command: 'send'`, so the doc is correct.
- CodeRabbit's `agent_id`-required nit: validator doesn't require `agent_id`, so the recipes are fine as-is.

Lint/prettier clean. Ready for another look.

PR #954, opened 2026-05-04
fix(cli): order Codex -C before resume so maestro-cli send resumes work

chr1syy:
## Summary

- `maestro-cli send` failed on follow-up messages to a Codex agent with `error: unexpected argument '-C' found ... Usage: codex exec resume [OPTIONS] [SESSION_ID] [PROMPT]`. The first message worked because `sessionId` was null and the resume branch was skipped; once a session id was saved, every subsequent send hit the resume path and broke.
- Root cause: `src/cli/services/agent-spawner.ts` appended `resumeArgs` before `workingDirArgs`, producing `codex exec ... resume <id> -C <cwd> -- <prompt>`. `resume` is a subcommand of `codex exec`, so once it's consumed, trailing flags are parsed against `codex exec resume`, which doesn't accept `-C`.
- Fix: emit `workingDirArgs` before `resumeArgs` for Codex. Matches the canonical comment in `src/main/agents/definitions.ts` (`[-C dir] [resume <id>]`) and the desktop builder `src/main/utils/agent-args.ts`, which already had the right order.
- Strengthened the existing Codex spawn test to assert `-C` precedes `resume` so this can't silently regress.

This was originally surfaced as a "Failed to get response from agent" error in the Maestro Discord bot — the bot was just relaying the upstream maestro-cli failure.

## Test plan

- [x] `vitest run src/__tests__/cli/services/agent-spawner.test.ts -t "Codex spawn preserves working-dir flag and resume args"` — passes with the new ordering assertion
- [x] `eslint` clean on the two touched files
- [x] `prettier --check` clean on the two touched files
- [ ] Manual: send two messages back-to-back to a Codex agent via `maestro-cli send` (or via Maestro Discord) — both should succeed; previously the second send failed

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed argument ordering for agent spawn commands to ensure proper parsing of working directory configuration before session resumption.

* **Tests**
  * Enhanced regression test to validate correct argument sequencing in agent spawn operations.

### DIFF
--- a/src/__tests__/cli/services/agent-spawner.test.ts
+++ b/src/__tests__/cli/services/agent-spawner.test.ts
@@ -2134,8 +2134,13 @@ Some text with [x] in it that's not a checkbox
 			expect(c).toBeGreaterThanOrEqual(0);
 			expect(args[c + 1]).toBe('/working');
 			// resume args are ['resume', '<id>']
-			expect(args).toContain('resume');
-			expect(args).toContain('codex-thread-123');
+			const r = args.indexOf('resume');
+			expect(r).toBeGreaterThanOrEqual(0);
+			expect(args[r + 1]).toBe('codex-thread-123');
+			// `-C` must precede the `resume` subcommand — once Codex consumes
+			// `resume`, trailing flags get parsed by `codex exec resume`, which
+			// rejects `-C`.
+			expect(c).toBeLessThan(r);
 		});
 
 		it('unsupported agent type returns a failure result', async () => {
--- a/src/cli/services/agent-spawner.ts
+++ b/src/cli/services/agent-spawner.ts
@@ -592,13 +592,16 @@ async function spawnJsonLineAgent(
 	if (def?.jsonOutputArgs) preOverrideArgs.push(...def.jsonOutputArgs);
 	if (readOnlyMode && def?.readOnlyArgs) preOverrideArgs.push(...def.readOnlyArgs);
 
-	if (agentSessionId && def?.resumeArgs) {
-		preOverrideArgs.push(...def.resumeArgs(agentSessionId));
-	}
-
-	// Codex requires explicit working directory arg (other agents use process cwd)
+	// Codex requires explicit working directory arg (other agents use process cwd).
+	// Must be emitted before resumeArgs because `resume` is a subcommand of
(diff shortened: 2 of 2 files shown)


pedramamini:

Took a careful look — the fix is exactly right:

- The root cause analysis is correct: once `codex exec resume` is consumed as a subcommand, trailing flags get parsed against `codex exec resume`, which rejects `-C`.
- Emitting `workingDirArgs` before `resumeArgs` aligns the CLI spawner with the canonical comment in `src/main/agents/definitions.ts` (`[-C dir] [resume <id>]`) and the desktop builder in `src/main/utils/agent-args.ts`.
- The strengthened test now asserts `c < r` positionally instead of just containment, so this can't silently regress.
- Two-line behavioral change, scoped to Codex via the `toolType === 'codex'` gate, no risk to other agents.

Mergeable, CI green, CodeRabbit + Greptile both clean. Approving. ✅

chr1syy:
Closing in favor of #960 (@pedramamini), which supersedes this PR.

#960 diagnoses the deeper root cause from #959: Codex treats `-C`/`--cd` as a **root-level global flag** that is silently ignored when placed after the `exec` subcommand. My PR only moved `-C` ahead of `resume` (still after `exec`), which removes the noisy `unexpected argument '-C'` crash on resume but leaves Codex running in the wrong cwd (most visibly on Windows).

#960 also fixes the desktop path (`src/main/utils/agent-args.ts`), which this PR didn't touch — so it covers interactive sessions, Auto Run, and Cue in addition to the CLI. Once #960 lands, this branch is fully redundant.

chr1syy:
Closing in favor of #960 (@pedramamini), which supersedes this PR.

#960 diagnoses the deeper root cause from #959: Codex treats `-C`/`--cd` as a **root-level global flag** that is silently ignored when placed after the `exec` subcommand. My PR only moved `-C` ahead of `resume` (still after `exec`), which removes the noisy `unexpected argument '-C'` crash on resume but leaves Codex running in the wrong cwd (most visibly on Windows).

#960 also fixes the desktop path (`src/main/utils/agent-args.ts`), which this PR didn't touch — so it covers interactive sessions, Auto Run, and Cue in addition to the CLI. Once #960 lands, this branch is fully redundant.

PR #955, opened 2026-05-05
perf: SSH path cache, async spawn temp file, CueIndicator extraction

reachrazamair:
## Summary

Perf improvement with three small, independently-revertible commits. None of them are individually large wins, but together they remove anti-patterns that drift the codebase away from `CLAUDE-PERFORMANCE.md` guidance and clear the perf backlog ahead of Tier 4 (renderer decomposition).

| Commit | Item | Net |
| --- | --- | --- |
| `0cdf66f23` | SSH path access cache | New `PathAccessCache` utility; 30s TTL on positive results only; injected into `SshRemoteManager` + `SshConfigParser` defaults via the existing DI seam — zero existing-test churn |
| `199b08ce2` | Async Windows spawn temp file | `fs.writeFileSync` / `fs.unlinkSync` → `fs/promises`; cleanup mirrors the fire-and-forget `imageUtils.cleanupTempFiles` pattern (ENOENT silenced, other codes route to Sentry) |
| `02e37fd31` | Memo'd CueIndicator subcomponent | Extracts the Maestro Cue pill from `SessionItem` so its memo boundary is independent of the row; primitive props only |

`+22 new tests, 0 modified, 27,346/27,346 full-suite pass, lint + tsc + eslint clean.`

---

## SSH path access cache

Two SSH validation surfaces were re-stat'ing the same identity files and `~/.ssh/config` across consecutive Test Connection clicks and Save operations:

- `SshRemoteManager.validateConfig` → `fs.accessSync` per private-key path
- `parseSshConfig` → `fs.accessSync` on `~/.ssh/config`

Neither is hot, but both are pure waste at any retry rate above zero.

**New utility** at `src/main/utils/path-access-cache.ts`:

- 30s TTL on positive (`true`) results only
- **Negative results are NEVER cached** — that's the user-actionable error path; a user fixing perms must see a fresh check, not a stale `false`
- Pure caller-supplied predicate; the cache itself does no I/O. Keeps `fs` out of the cache module so tests can inject any boolean function
- Module-level singleton with a `setPathAccessCacheForTest` seam

**Wiring** preserves the existing DI shape:

- `SshRemoteManager.defaultDeps.checkFileAccess` routes through cache
- `SshConfigParser.defaultDeps.fileExists` routes through cache
- Both modules' test deps mock the predicate directly and **bypass the cache entirely** — zero churn on the existing 38 + 25 SSH tests

11 new tests cover hit / miss / TTL expiry, per-path isolation, the false-not-cached invariant, stale-true-drops-on-false, `clear()`, and the singleton seam.

---

## Async Windows spawn temp file

The Windows-only `--append-system-prompt-file` branch in `process:spawn` wrote the temp file synchronously and unlinked it synchronously inside a `setTimeout`. The handler is already `async` (`withIpcErrorLogging` wraps an async handler), so converting to `fs/promises` is straightforward and removes two main-process-blocking calls from the spawn path.

Mirrors the existing fire-and-forget pattern in `src/main/process-manager/utils/imageUtils.ts`:

```ts
await fsp.writeFile(tempPath, content, 'utf-8');
setTimeout(() => {
    fsp.unlink(p).catch((err) => {
        if (err.code !== 'ENOENT') captureException(err, { context, file: p });
    });
}, 30_000);
```

**Behavior preserved:**

- ENOENT on cleanup is silenced (file already gone is benign)
- Non-ENOENT errors route to Sentry with the same context payload
- Outer `withIpcErrorLogging` wrapper still catches both synchronous AND asynchronous-rejection errors from the handler body — no IPC error-shape change for write failures

**4 new tests** in `appendSystemPrompt delivery > Windows temp prompt file (async fs/promises)` exercise: writeFile call, scheduled unlink after 30s, ENOENT silence, EACCES → captureException. Module-level `vi.mock('fs/promises')` + `vi.mock(sentry)`; `platformDetection` mock defaults to host platform so the existing platform-branching test still works on Windows CI.

---

## Memo'd CueIndicator subcomponent

`SessionItem` renders ~13 lucide-react icon instances per row. The component is already `memo`'d at the row boundary, but inside the memo'd shell every render rebuilds every icon. The Maestro Cue indicator is the only icon cluster with conditional animation (`animate-pulse` when `activeRun`) and non-trivial JSX (tooltip text built from two flags), so it pays the most to give it its own memo boundary with primitive props.

**New component** at `src/renderer/components/SessionList/CueIndicator.tsx`:

- `memo`'d with primitive props (`subscriptionCount`, `activeRun`)
- Returns `null` when `subscriptionCount <= 0` so `SessionItem` renders it unconditionally
- Inlined `style={{ color: '#2dd4bf' }}` stays inside the memo boundary

**SessionItem changes:**

- Replaced the 8-line inline Cue indicator with `<CueIndicator ... />`
- Dropped the now-unused `Zap` import

**Deliberately deferred** per the plan (avoid over-engineering without CDP data):

- `SshRemotePill` / `GitDirtyBadge` — single-purpose 4–5 line snippets; extracting them adds prop-passing cost that may exceed the icon-instantiation cost they save
- 18 inline-style objects on plain DOM nodes — React doesn't re-render plain DOM differently based on style-object identity; over-memoizing is noise that bloats hook count
- `Bookmark` / `AlertCircle` / `Bot` / `Activity` — single-conditional or always-on; no extraction value

**7 new tests** in `__tests__/renderer/components/SessionList/CueIndicator.test.tsx` cover: hidden when count ≤ 0, renders Zap when positive, animate-pulse class toggles with `activeRun`, singular/plural tooltip text, "running" vs "active" string. Existing 11 `SessionItemCue` + 135 `SessionList` tests all green — DOM is identical to the inline version.

---

## SRP / SOC

- `PathAccessCache` does ONE thing (cache a boolean predicate). It does not perform the access; the caller passes `accessFn`. Decoupled from `fs` entirely.
- Inline migration mirrors the precedent in `imageUtils.cleanupTempFiles`. Extracting a helper for a single caller would either schedule the cleanup timer from a "write" helper (worse SRP) or leak the timer handle (worse SOC). Re-evaluate only if a second caller appears.
- `CueIndicator` owns one concern: rendering the Maestro Cue pill. Primitive props mean React's default memo shallow-compare is sufficient.

## Test plan

- **SSH cache:** Settings → SSH Remote → click Test Connection twice within 30s on a remote with a valid key; second click should not re-stat (verify with `fs_usage` on macOS or DevTools-side counters in dev). Edit `privateKeyPath` to a missing file → fail; restore the path → next click succeeds (no stale `false`).
- **Windows spawn:** (Windows only) spawn an agent with a long `appendSystemPrompt`; verify `%TEMP%\maestro-sysprompt-<sessionId>-<ts>.txt` exists immediately, gone after 30s. Spawn 3 in a row → all 3 cleaned up. Kill agent before 30s → temp file still cleaned up. (macOS/Linux) verify NO temp file is written (inline `--append-system-prompt` path).
- **CueIndicator:** create a Cue subscription on a session → Zap pill appears next to the name with tooltip `"Maestro Cue active (1 subscription)"`. Add a second → tooltip becomes `"... (2 subscriptions)"`. Trigger a Cue run → pulses with `"Maestro Cue running ..."`. Run completes → animation stops, tooltip reverts. Remove all subscriptions → pill disappears.
- **Cross-cutting:** cold-start splash appears as fast as before; switch themes — Cue indicator repaints cleanly; type rapidly with 30+ sessions — no typing-lag regression (Tier 1 perf wins still hold); DevTools console — no new warnings, no React hook-order errors, no unhandled-promise-rejection from the cleanup setTimeout.

## Summary by CodeRabbit

* **New Features**
  * Added a Cue Indicator component showing Maestro Cue status and animated pulse when active.

* **Performance Improvements**
  * Added caching for filesystem readability checks to reduce redundant probes.
  * Improved Windows system-prompt temp-file handling to use asynchronous operations for more reliable cleanup.

* **Tests**
  * Expanded test coverage for the Cue Indicator, path access caching behavior, and Windows temp-file handling.

### DIFF
--- a/src/__tests__/main/ipc/handlers/process.test.ts
+++ b/src/__tests__/main/ipc/handlers/process.test.ts
@@ -196,6 +196,30 @@ vi.mock('../../../../main/utils/cliDetection', () => ({
 	resolveSshPath: vi.fn().mockResolvedValue('ssh'),
 }));
 
+// Mock platformDetection. Default mirrors the host so the existing
+// `appendSystemPrompt delivery` test still picks the right branch on
+// Windows CI; new tests override per-case with mockReturnValue(true).
+vi.mock('../../../../shared/platformDetection', () => ({
+	isWindows: vi.fn(() => process.platform === 'win32'),
+	isMacOS: vi.fn(() => process.platform === 'darwin'),
+	isLinux: vi.fn(() => process.platform === 'linux'),
+}));
+
+// Mock fs/promises so the new temp-file tests can assert on writeFile/unlink
+// without touching the real filesystem. Other tests in this file don't use
+// fs/promises, so the module-level mock is safe.
+vi.mock('fs/promises', () => ({
+	writeFile: vi.fn().mockResolvedValue(undefined),
+	unlink: vi.fn().mockResolvedValue(undefined),
+}));
+
+// Mock sentry — captureException is asserted on by the new cleanup-error
+// tests; addBreadcrumb is a no-op stub so existing tests don't hit real Sentry.
+vi.mock('../../../../main/utils/sentry', () => ({
+	captureException: vi.fn(),
+	addBreadcrumb: vi.fn(),
+}));
+
 describe('process IPC handlers', () => {
 	let handlers: Map<string, Function>;
 	let mockProcessManager: {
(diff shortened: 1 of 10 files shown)


pedramamini:
@reachrazamair Thanks for this — three nice, independently-revertible perf wins, and the test coverage is thorough. The cache invariants (positive-only TTL, stale-true-drops-on-false) and the imageUtils-mirroring cleanup pattern are exactly the right shape.

One thing to address before approval: `rawCheckFileAccess` in `src/main/ssh-remote-manager.ts` and `rawFileExists` in `src/main/utils/ssh-config-parser.ts` are byte-for-byte identical:

```ts
try {
    fs.accessSync(filePath, fs.constants.R_OK);
    return true;
} catch {
    return false;
}
```

Per CLAUDE.md (`Before Writing New Code - Check Existing Utilities`), duplicated code is the #1 maintenance burden in this codebase, and you're introducing a new duplicate pair in the same PR. Could you consolidate these into a single shared helper? A natural home is `path-access-cache.ts` itself — exporting e.g. `defaultReadableProbe` (or whatever name reads well to you) — since both call sites already import from that module. That keeps the SRP rationale in your module doc intact (the cache still doesn't *call* `fs`; it just exports a default predicate alongside the cache), avoids the divergence risk, and shrinks both call sites.

Otherwise this is clean — no merge conflicts, ~22 new tests passing, lint/tsc/eslint clean. Once the duplicate is consolidated, happy to approve.

reachrazamair:
@coderabbit review

PR #956, opened 2026-05-05
fix(terminal): number tabs by creation order, not visual position

chr1syy:
## Summary

- Opening a 2nd terminal while an AI tab is active caused the new terminal to display as "Terminal 1" and the original to be re-labeled "Terminal 2"
- Root cause: `TabBar.tsx` computed the `Terminal N` index from the visual `unifiedTabs` order, but `addTerminalTab` inserts the new terminal directly after the currently-active tab — placing it visually to the LEFT of the existing terminal when an AI tab is active
- Replaced the visual-order `findIndex` with a `useMemo`-cached `terminalIndexById` map sorted by `data.createdAt`, aligning TabBar with the storage-order indexing already used in `MainPanel.tsx` and `AppSessionModals.tsx`

## Test plan

- [x] New regression test in `TabBar.test.tsx` renders a unified tab list where the newer terminal sits before the older one in visual order, asserts older stays "Terminal 1" / newer becomes "Terminal 2"
- [x] TabBar test suite: 171 pass / 0 fail
- [x] terminalTabHelpers + AppSessionModals tests: 57 pass / 0 fail
- [x] TypeScript clean on changed files
- [x] Prettier + ESLint clean
- [ ] Manual: open agent → open Terminal → click AI tab → open another Terminal → verify they read "Terminal 1" / "Terminal 2" in creation order

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed terminal tab numbering to correctly reflect creation order, ensuring "Terminal 1" and "Terminal 2" labels remain consistent with their actual terminal tabs even when displayed in different visual order.

### DIFF
--- a/src/__tests__/renderer/components/TabBar.test.tsx
+++ b/src/__tests__/renderer/components/TabBar.test.tsx
@@ -4631,6 +4631,66 @@ describe('Unified tabs drag and drop', () => {
 		// Should call onUnifiedTabReorder with index 1 -> 2 (last index)
 		expect(mockOnUnifiedTabReorder).toHaveBeenCalledWith(1, 2);
 	});
+
+	// Regression: when the user clicks back to an AI tab and then opens a 2nd
+	// terminal, addTerminalTab inserts the new terminal directly after the AI
+	// tab — which places it visually to the LEFT of the existing terminal.
+	// "Terminal N" labels must follow creation order so the older terminal
+	// stays "Terminal 1" and the new one becomes "Terminal 2".
+	it('numbers terminal tabs by creation order, not visual position', () => {
+		const olderTerminal = {
+			id: 'term-older',
+			name: null,
+			shellType: 'zsh',
+			pid: 100,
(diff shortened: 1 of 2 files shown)


pedramamini:
Thanks for the contribution, @chr1syy! Nice clean fix — the `useMemo` Map keyed by `createdAt` is the right call, and the regression test pinpoints the inverted-visual-order scenario perfectly. Aligns with the storage-order indexing already used in `MainPanel.tsx` / `AppSessionModals.tsx`, which is a nice consistency win. CI is green and the diff is mergeable with no conflicts.

Approving as-is. One tiny non-blocking nit if you want to tidy up in a follow-up: the existing `terminalIndex={terminalIndex >= 0 ? terminalIndex : 0}` at the `TerminalTabItem` call site is now dead — `terminalIndexById.get(unifiedTab.id) ?? 0` already guarantees a non-negative number, so it can simplify to just `terminalIndex={terminalIndex}`. Totally optional.

PR #957, opened 2026-05-05
refactor(cli): hard-remove deprecated flags and unify agent targeting

chr1syy:
## Summary

Audited every `maestro-cli` verb for duplicate / overloaded flags and applied a three-step cleanup. **Breaking change** — see "Removed flags" below.

### Step 1 — hard-remove deprecated aliases

| Verb | Removed |
|---|---|
| `send` | `-l, --live`, `--new-tab`, `-f, --force` (use `dispatch` instead) |
| `auto-run` | `-s, --session` (use `-a, --agent`) |
| `notify toast` | `-t, --type` (use `-c, --color`); `-d, --duration` (use `--timeout`) |
| `notify flash` | `-v, --variant` (use `-c, --color`); `-d, --duration` (use `-t, --timeout`) |

`notify toast`'s `--timeout` is now `-t, --timeout` for parity with `notify flash`. The two notify verbs no longer have a sec/ms unit mismatch.

### Step 2 — unify "target an agent" on `-a, --agent`

`--session` was overloaded across three different concepts (provider session id, agent id, tab id). Now:

| Verb | Was | Is |
|---|---|---|
| `open-file` | `-s, --session <id>` (meant agent) | `-a, --agent <id>` |
| `refresh-files` | `-s, --session <id>` (meant agent) | `-a, --agent <id>` |
| `refresh-auto-run` | `-s, --session <id>` (meant agent) | `-a, --agent <id>` |
| `dispatch` | `-s, --session <id>` (actually a tab id) | `-t, --tab <id>` |

`send -s, --session <id>` keeps its narrow correct meaning (provider session id for resume).

### Step 3 — normalize `--json` help text

`Output as JSON line (for scripting)` (only used in `settings get/set/reset`) collapsed to the canonical `Output as JSON (for scripting)`.

## Test plan

- [x] `npx tsc -p tsconfig.cli.json --noEmit` — clean
- [x] `npx tsc -p tsconfig.lint.json` — no CLI-scoped errors
- [x] `npx eslint src/cli/` — no issues
- [x] `npx vitest run` on `send.test.ts`, `dispatch.test.ts`, `auto-run.test.ts`, `open-file.test.ts` — 53 pass / 0 fail
- [ ] Manual verification: `maestro-cli dispatch <agent> "msg" -t <tab-id>` reaches the right tab
- [ ] Manual verification: `maestro-cli open-file foo.md -a <agent-id>` opens the preview tab on the right agent
- [ ] Manual verification: `maestro-cli send <agent> "msg" -l` is rejected as unknown flag (was a hidden alias)

## Summary by CodeRabbit

* **Documentation**
  * Clarified notification behavior: toasts are dismissible-only when sticky, numeric toast max 60s; flash notifications use seconds, capped at 5s (enforced for external clients). Removed deprecated notification options and strengthened guidance for canonical color values.

* **CLI Changes**
  * `dispatch` now targets tabs via `--tab` (not `--session`).
  * `refresh-files`/`refresh-auto-run` and `open-file` use `--agent`.
  * `send` no longer supports `--live`; notification commands use `--color` and `--timeout` (removed legacy flags).


pedramamini:
Thanks for taking this on, @chr1syy — the cleanup is well-scoped and reads cleanly. 🙏

Walked the diff for `send`, `dispatch`, `auto-run`, `open-file`, `refresh-files`, `refresh-auto-run`, and both `notify` verbs against the docs and tests. The three-step cut (hard-remove deprecated aliases → unify `-a, --agent` for agent targeting → tighten `--json` help text) lines up consistently across `src/cli/commands/`, `src/cli/index.ts`, and the rewritten test files. Particularly nice touches:

- `dispatch`'s `-s, --session` → `-t, --tab` rename plus the matching `--new-tab cannot be combined with --tab` error message and `tabId ?? options.tab ?? null` fallback in `runDispatch`.
- Killing the `--type`/`--variant`/`--duration` paths in `notify-toast.ts` and `notify-flash.ts` and dropping the now-dead `VARIANT_TO_COLOR` / `TYPE_TO_COLOR` tables instead of leaving them around as zombie code.
- Updating `docs/agent-guides/UI-PATTERNS.md` to call out that the legacy `type`/`variant` API is now in-app only — that asymmetry between the in-app shim and the CLI surface was easy to miss.

One thing worth flagging in the merge note (not a request to change the PR): `notify toast`'s `-t` short flag flips meaning from `--type` to `--timeout`. Existing scripts running `maestro-cli notify toast ... -t success` will now parse `success` as a seconds value and exit with the "must be a positive number of seconds" error. That's exactly what the PR body promises, but it's the most likely surprise for downstream callers, so it's worth surfacing in the release-notes line.

CI is still mid-flight as I write this — assuming `lint-and-format`, `test`, and the bot reviewers come back green, this is good to go from my end. Approving and tagging.

chr1syy:
## CLI cleanup soak — mostly green, one undocumented breaking change worth a release-notes mention

Tested against the packaged 0.16.17-RC build's `maestro-cli.js` from `release/linux-unpacked/resources/`.

### ✅ All documented changes verified

- **Removed flags reject cleanly** with `error: unknown option '<flag>'` and `exit 1` across the full grid:
  - `send -l/--live`, `--new-tab`, `-f/--force`
  - `auto-run -s/--session`
  - `notify toast --type`, `--duration`
  - `notify flash --variant`, `--duration`
  - `open-file -s/--session`, `refresh-files -s`, `refresh-auto-run -s`
  - `dispatch -s/--session`
- **Canonical replacements parse cleanly** (verified by either successful daemon-side errors like \"Agent not found\" or by actually firing — `notify toast 'Title' 'Body' -c green --dismissible` produced \"Toast sent (sticky — click to dismiss)\" against the running app).
- **`--json` help text normalized** to `Output as JSON (for scripting)` on all three of `settings get`, `settings set`, `settings reset`.
- **`--timeout` + `--dismissible` mutual exclusion** enforced cleanly: `Error: --dismissible cannot be combined with --timeout (a sticky toast has no auto-dismiss)`.
- **`send -s/--session`** narrow (provider-session-id) meaning preserved.
- **`-a/--agent`** wired up correctly on `auto-run`, `open-file`, `refresh-files`, `refresh-auto-run`.
- **`-t/--tab`** wired up correctly on `dispatch` (and unchanged on `send`, where it's still a no-arg \"open/focus tab\" flag).

### 🚨 Undocumented breaking change — `notify toast` positional signature

The PR description's \"Removed flags\" table mentions only flag-aliasing changes for `notify toast`. But in this build's `notify toast --help`:

```
Usage: maestro-cli notify toast [options] <title> <message>
```

Previously the signature was just `<message>`. So **any external script calling `maestro-cli notify toast 'My one-line message'` will now fail** with a confusing `error: missing required argument 'message'` (their value gets bound to the new `<title>` slot, leaving `<message>` empty).

This is probably a worthwhile change on its own (toasts often want a title separate from the body), but it needs to be:

1. **Documented in the PR description** alongside the flag-aliasing changes.
2. **Called out in the release notes** so external-script consumers know to migrate `notify toast 'msg'` → `notify toast 'Title' 'msg'` (or whatever the migration story is — alternatively, accept a single-arg form and treat it as message-only with empty title).

A migration option that minimizes churn: keep accepting a single positional and route it to `<message>` with a default empty title, while still accepting `<title> <message>` for the new shape. Or document the breakage clearly and move on. Author's call.

### ⚠️ Subtle unit migration on `notify flash --timeout`

The PR description does call this out (\"The two notify verbs no longer have a sec/ms unit mismatch\"), but it's worth flagging in the release notes because the flag *name* migration (`--duration` → `--timeout`) and the *unit* migration (ms → seconds) are bundled together. A script that mechanically renames `--duration` to `--timeout` without converting the value will now blow past the 5-second cap and either error out or silently get clamped:

- Old: `notify flash 'msg' --duration 1500` (= 1.5s)
- New: `notify flash 'msg' --timeout 1500` (= 1500s, rejected with `--timeout cannot exceed 5 seconds`)
- Correct: `notify flash 'msg' --timeout 1.5`

Recommend a release-notes line: *\"`notify flash` value units changed from milliseconds to seconds along with the flag rename — divide by 1000 when migrating.\"*

### Recommendation

Don't think this needs a `blocking` label — the PR's stated goals are met and the failure modes for external-script consumers are loud (error messages, non-zero exit). But the two release-notes additions above would prevent confused bug reports after the merge ships. **Worth adding before tagging the release.**

Audit list for any internal tooling that calls `maestro-cli`:

- Discord-bot — confirmed in `#954` it uses `maestro-cli send`; double-check it doesn't use `--live`, `--new-tab`, `--force`.
- Any AutoRun playbook docs that contain shell snippets calling `notify toast` with a single positional message.
- CI / scripts.

chr1syy:
Thanks for the soak — quick clarifications on both findings, then a recommendation.

## 🚨 Finding #1: `notify toast` positional signature — **not changed by this PR**

`notify toast <title> <message>` has been two positionals from its very first commit. From `3542695c` (`feat(cli): add notify toast/flash commands for in-app notifications`, Apr 26):

```ts
notify
    .command('toast <title> <message>')
    .description('Show a toast notification (queued, persistent until dismissed)')
```

So `maestro-cli notify toast 'My one-line message'` has always failed with `error: missing required argument 'message'` — both before and after PR #957. Nothing in this PR's diff touches the `<title> <message>` shape (you can verify on `src/cli/index.ts:472` in the original diff: only the flag list changed). Probably worth confirming on the soak host whether the test there was actually compared to a pre-`3542695c` build, or whether memory of `notify flash <message>` (which **is** a single positional) leaked over.

If we still want to make `notify toast` accept a single-positional shorthand for ergonomics, that's a separate, additive change — happy to do it in a follow-up PR but I don't think it belongs in this cleanup.

## ⚠️ Finding #2: `notify flash --duration` removal — real, but it's an **alias removal**, not a rename

`-t, --timeout <seconds>` already existed pre-PR alongside `-d, --duration <ms>` — pre-PR registration in `src/cli/index.ts`:

```ts
notify
    .command('flash <message>')
    ...
    .option('-t, --timeout <seconds>', 'Auto-dismiss after N seconds (range: (0, 5]; default 1.5; wins over --duration)')
    .option('-d, --duration <ms>', 'Auto-dismiss after N ms (range: (0, 5000]; default 1500)')
```

So this PR didn't rename a flag — it removed the `--duration` (ms) alias and kept `--timeout` (seconds). The footgun you describe is still real, just a slightly different shape: a script using `--duration 1500` doesn't fail with "blew past the 5-second cap" today, it fails with `error: unknown option '--duration'`. The dangerous case is a script that *was already using* `--timeout` in seconds — those keep working fine. The unsafe migration path is anyone who mechanically swaps `--duration` → `--timeout` without dividing by 1000.

Same removal applies to `notify toast --duration` (which was already an alias for `--timeout` in the same units, both seconds — so no value conversion needed there).

## Recommendation

Both real consumer-facing changes are flag *removals*, not signature changes:

1. `notify toast --duration` (seconds) — alias for `--timeout`; **same units, just remove**.
2. `notify flash --duration` (**milliseconds**) — different units from `--timeout` (seconds); **divide value by 1000 when migrating**.

Per CLAUDE.md the release notes are generated automatically and shouldn't be hand-edited, so the right surface for the migration callout is the PR description's "Removed flags" section. Want me to push an edit to the PR body that adds an explicit migration note for `--duration` → `--timeout` on flash (with the ms → seconds divisor) plus the `notify toast --duration` alias drop? That would short-circuit any "I just did sed" footgun without bloating the diff.

Otherwise: I think the cleanup is done; finding #1 is a misread of pre-existing behavior, finding #2's only loud failure already gives a clear `unknown option` exit.

pedramamini:
Merged — thanks @chr1syy! Clean cleanup, with all reviewer feedback (notify-toast ms conversion, helper extraction, open-file flag rename, catch narrowing) addressed across the four follow-up commits. Appreciate the care.

PR #962, opened 2026-05-07
feat(git): add graph view + branch switcher dropdown

### LINKED ISSUE #958: Feature request: manage Git branches in Maestro
## Summary

Maestro already has useful Git views such as Git log and Git diff. It would be great to also manage Git branches directly from Maestro.

## Use case

When working with agents, I often need to prepare the workspace before starting work: switch to an existing branch, create a new feature branch, or confirm which branch the agent is currently on. Today that requires leaving Maestro and using a terminal or another Git client.

## Proposed functionality

- Show the current branch prominently in the Git/workspace UI.
- List local branches, and optionally remote branches.
- Switch/checkout an existing branch from Maestro.
- Create a new branch from the current HEAD.
- Optionally create a branch from a selected base branch.
- Warn clearly when switching branches with uncommitted changes, with an explicit confirmation or safe handling path.

## Why this would help

This would make the existing Git log and Git diff features feel more complete and reduce context switching when setting up agent workspaces.

feldsys:
Closes #958

## Summary

Two new affordances on the git surface, both plumbed through the existing SSH-remote pathway:

- **Graph view in `GitLogViewer`** — toggleable List/Graph buttons in the modal header. Graph mode renders commits with parent-hash topology via `@gitgraph/react`, supports clicking commits to drive the existing diff panel, and persists the user's view preference via `localStorage`. Side-branch commits only visible via `git log --all` are tracked separately so the right-hand commit detail panel stays in sync regardless of which view is active.
- **Branch switcher dropdown on the header branch chip** — single click still opens the git log (debounced ~220 ms to differentiate from a double click); double click opens a filterable branch picker that switches the working tree via `git switch <branch>`. Failures (e.g. dirty working tree) surface git's stderr inside the dropdown instead of throwing.

### Mapping to issue #958

| Issue ask                                            | This PR                                                                                            |
| ---------------------------------------------------- | -------------------------------------------------------------------------------------------------- |
| Show the current branch prominently                  | Already shown on the header branch chip; PR makes it interactive                                   |
| List local (and remote) branches                     | ✅ Filterable list in the new dropdown                                                              |
| Switch/checkout an existing branch from Maestro      | ✅ Double-click chip → pick branch → `git switch`                                                   |
| Warn clearly when switching with uncommitted changes | ✅ Failure path surfaces git's stderr inline (e.g. dirty working tree) instead of silently failing |
| Create a new branch from current HEAD / from base    | ⏳ Out of scope for this PR — happy to follow up                                                    |

### IPC surface

Two new handlers, both honoring the existing `sshRemoteId` / `remoteCwd` SSH wrapping:

- `git:graph` — returns commits with parent hashes (`{ nodes, error }`) for lane rendering. Default limit 200, configurable via `options.limit`.
- `git:switch` — returns `{ success, stdout, stderr }` instead of throwing, so the UI can render git's failure message in-place.

### Files

- `src/main/ipc/handlers/git.ts`, `src/main/preload/git.ts`, `src/renderer/services/git.ts`, `src/renderer/global.d.ts` — IPC + service plumbing
- `src/renderer/components/GitGraphView.tsx` *(new)* — `@gitgraph/react` wrapper themed via `theme.colors`
- `src/renderer/components/GitLogViewer.tsx` — list/graph toggle + selection sync
- `src/renderer/components/MainPanel/BranchSwitcherDropdown.tsx` *(new)* — filterable picker that calls `git:switch`
- `src/renderer/components/MainPanel/MainPanelHeader.tsx` — single/double-click handling on branch chip
- `src/renderer/constants/modalPriorities.ts` — adds `BRANCH_SWITCHER: 195`
- `src/__tests__/renderer/components/MainPanel.test.tsx` — updated SSH-remote click test, new double-click test
- `package.json` / `package-lock.json` — adds `@gitgraph/react ^1.6.0`

The branch also includes a separate commit (`fix(node-pty): patch winpty.gyp paths and harden POSIX spawn errors`) that fixes a Windows build issue — happy to split it into its own PR if preferred.

## Why

Per #958, currently the only way to inspect history is the linear list, which makes branch topology impossible to read at a glance, and the only way to switch branches inside Maestro is to open a terminal. Both moves keep the user inside the AI surface.

## Test plan

- [x] `npm run lint` — TypeScript across all three configs
- [x] `npm run lint:eslint` — ESLint
- [x] `npm run test -- --run src/__tests__/renderer/components/MainPanel.test.tsx` — 137 tests including the new double-click + branch switcher coverage
- [x] Manual: open a git repo, click the branch chip in the header → git log opens after ~220 ms
- [x] Manual: double-click the branch chip → branch switcher opens, filter input is focused, pick a branch → working tree switches
- [x] Manual: in git log, toggle Graph view → commits render with lane topology, clicking a commit (including a side-branch commit) drives the diff panel on the right
- [ ] Manual (SSH): repeat above with an SSH remote enabled session — both new IPC paths route through `getSshRemoteById`

## Summary by CodeRabbit

* **New Features**
  * Added a visual commit graph with selectable, highlighted commits; graph lazy-loads and can drive commit details/diff loading.
  * Enabled a persistent List ↔ Graph toggle in the commit viewer.
  * Introduced a searchable, keyboard-navigable branch switcher dropdown with loading/empty states and success/error feedback.
  * Updated branch chips: single-click opens history; double-click or **Shift+Enter** opens the branch switcher.
* **Dependency**
  * Added `@gitgraph/core` and `@gitgraph/react`.
* **Tests**
  * Expanded coverage for SSH remote branch-switcher behavior and adjusted async expectations.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,6 +37,7 @@
 				"@fastify/rate-limit": "^9.1.0",
 				"@fastify/static": "^7.0.4",
 				"@fastify/websocket": "^9.0.0",
+				"@gitgraph/react": "^1.6.0",
 				"@lezer/highlight": "^1.2.3",
 				"@napi-rs/keyring": "1.3.0",
 				"@opencode-ai/sdk": "^1.17.13",
@@ -2441,6 +2442,24 @@
 			"integrity": "sha512-k2Ty1JcVojjJFwrg/ThKi2ujJ7XNLYaFGNB/bWT9wGR+oSMJHMa5w+CUq6p/pVrKeNNgA7pCqEcjSnHVoqJQFw==",
 			"dev": true,
 			"license": "MIT"
+		},
+		"node_modules/@gitgraph/core": {
+			"version": "1.5.0",
+			"resolved": "https://registry.npmjs.org/@gitgraph/core/-/core-1.5.0.tgz",
+			"integrity": "sha512-8CeeHbkKoFHM1y9vfjYiHyEpzl1mEhVrg5c/eFgDBsntOYswoDKU2yOf6DjtVINcE60wmcuynBSJqjMkQo07Ww==",
+			"license": "MIT"
+		},
+		"node_modules/@gitgraph/react": {
+			"version": "1.6.0",
+			"resolved": "https://registry.npmjs.org/@gitgraph/react/-/react-1.6.0.tgz",
+			"integrity": "sha512-cLFNZDoEiNbsnMfdT82zeZti5saYghQamfCbTpVvCRr+BrrQ/k94glkZqYPXKKNTEqbQV6L9JeOfAk1hNiFYXA==",
+			"license": "MIT",
+			"dependencies": {
+				"@gitgraph/core": "1.5.0"
+			},
+			"peerDependencies": {
+				"react": ">= 16.8.0"
+			}
 		},
 		"node_modules/@humanfs/core": {
 			"version": "0.19.2",
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,6 +37,7 @@
 				"@fastify/rate-limit": "^9.1.0",
 				"@fastify/static": "^7.0.4",
 				"@fastify/websocket": "^9.0.0",
+				"@gitgraph/core": "^1.5.0",
 				"@gitgraph/react": "^1.6.0",
 				"@lezer/highlight": "^1.2.3",
 				"@napi-rs/keyring": "1.3.0",
--- a/package.json
+++ b/package.json
@@ -305,6 +305,7 @@
 		"@fastify/rate-limit": "^9.1.0",
 		"@fastify/static": "^7.0.4",
 		"@fastify/websocket": "^9.0.0",
+		"@gitgraph/react": "^1.6.0",
 		"@lezer/highlight": "^1.2.3",
 		"@opencode-ai/sdk": "^1.17.13",
 		"@sentry/electron": "^7.13.0",
--- a/package.json
+++ b/package.json
@@ -305,6 +305,7 @@
 		"@fastify/rate-limit": "^9.1.0",
 		"@fastify/static": "^7.0.4",
 		"@fastify/websocket": "^9.0.0",
+		"@gitgraph/core": "^1.5.0",
 		"@gitgraph/react": "^1.6.0",
 		"@lezer/highlight": "^1.2.3",
 		"@opencode-ai/sdk": "^1.17.13",
--- a/src/__tests__/renderer/components/MainPanel.test.tsx
+++ b/src/__tests__/renderer/components/MainPanel.test.tsx
@@ -1872,6 +1872,34 @@ describe('MainPanel', () => {
 			expect(gitService.getBranches).toHaveBeenCalled();
 		});
 
+		it('should open branch switcher on Shift+Enter on SSH remote git badge (keyboard a11y)', async () => {
+			const session = createSession({
+				isGitRepo: true,
+				sessionSshRemoteConfig: { enabled: true, remoteId: 'ssh-remote-123' },
+			});
+
+			const mockGetConfigs = vi.fn().mockResolvedValue({
+				success: true,
+				configs: [{ id: 'ssh-remote-123', name: 'my-ssh-remote' }],
+			});
+			vi.mocked(window.maestro.sshRemote.getConfigs).mockImplementation(mockGetConfigs);
+
+			render(<MainPanel {...defaultProps} activeSession={session} />);
+
+			await waitFor(() => {
+				expect(screen.getByText('my-ssh-remote')).toBeInTheDocument();
+			});
+
+			// The chip's button is the parent of the SSH remote name span.
+			const chipButton = screen.getByText('my-ssh-remote').closest('button');
+			expect(chipButton).not.toBeNull();
+			fireEvent.keyDown(chipButton!, { key: 'Enter', shiftKey: true });
+
+			// Branch switcher dropdown opens via the keyboard path.
+			expect(await screen.findByPlaceholderText(/Filter branches/)).toBeInTheDocument();
+			expect(gitService.getBranches).toHaveBeenCalled();
+		});
+
 		it('should open git log when single-clicking on SSH remote git badge', async () => {
 			const setGitLogOpen = vi.fn();
 			const session = createSession({
--- a/src/__tests__/renderer/components/MainPanel.test.tsx
+++ b/src/__tests__/renderer/components/MainPanel.test.tsx
@@ -45,6 +45,12 @@ function makeWindowState(partial: Partial<WindowState> & Pick<WindowState, 'id'>
 
 // Mock child components to simplify testing - must be before MainPanel import
 
+// useModalLayer is no-op in tests - MainPanel does not provide LayerStackProvider,
+// and the BranchSwitcherDropdown registers a layer when rendered.
+vi.mock('../../../renderer/hooks/ui/useModalLayer', () => ({
+	useModalLayer: () => {},
+}));
+
 // LayerStack: MainPanelContent reads layerCount to blur the browser webview when a
 // modal/overlay is layered above it. These tests render MainPanel in isolation
 // without a LayerStackProvider, so stub the hook (no layers open => layerCount 0).
@@ -239,6 +245,9 @@ vi.mock('../../../renderer/components/InlineWizard', () => ({
 vi.mock('../../../renderer/services/git', () => ({
 	gitService: {
 		getDiff: vi.fn().mockResolvedValue({ diff: 'mock diff content' }),
+		getBranches: vi.fn().mockResolvedValue([]),
+		getGraph: vi.fn().mockResolvedValue([]),
+		switchBranch: vi.fn().mockResolvedValue({ success: true, stderr: '' }),
 	},
 }));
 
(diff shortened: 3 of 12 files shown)


pedramamini:
@feldsys — thanks a lot for this contribution! Nice piece of work threading both new affordances cleanly through the existing SSH-remote pathway, and the issue→PR mapping in the description is super helpful.

A few things to address before this can land. CodeRabbit caught most of these and I agree with them; flagging here so they don't get lost:

**Bugs / correctness**

1. **`switching` state can stick forever in `BranchSwitcherDropdown.tsx`** (`doSwitch`, ~line 110–133). `setSwitching(null)` only runs on the happy path — if `gitService.switchBranch` rejects (IPC channel error, etc.), every branch button stays disabled and the spinner never stops. Wrap the await in `try/finally` and clear `switching` in the `finally`.

2. **Graph fetches 100 commits but list fetches 200** (`GitLogViewer.tsx`, ~line 120). Toggle to graph after selecting `entries[100..199]` and the right-side detail pane points at a commit the graph can't render or highlight. Use the same limit (extract a shared constant so they can't drift again).

3. **`gitService.getGraph` swallows main-process `error`** (`src/renderer/services/git.ts`, ~line 188–195). The IPC returns `{ nodes, error }` but the wrapper drops `error` and returns `[]`, so the viewer can't tell "git command failed" from "no commits" — the empty-state message will be wrong. Surface the error (or return the whole `{ nodes, error }`).

4. **Initial `localStorage.getItem` read isn't guarded** (`GitLogViewer.tsx`, view-mode `useState` initializer). The `setItem` call already has try/catch but the read does not — if it throws during render the whole modal fails to open instead of falling back to `'list'`.

5. **`@gitgraph/react` doesn't re-run the children function on prop changes** (`GitGraphView.tsx`). The selected-commit dot styling at lines ~129–132 is baked in at mount and won't update when `selectedHash` changes. Easiest fix is `<Gitgraph key={selectedHash ?? ''} options={...}>` to force remount on selection change. (Heads up that this rebuilds the whole graph each click — fine for ~200 commits, may want the external CSS-overlay approach later if you push the limit higher.)

**Defense-in-depth nit**

6. **Validate `branchName` in the `git:switch` IPC handler.** `execGit(['switch', branchName], ...)` uses `execFile` so no shell injection is possible, but a branch name starting with `-` would be interpreted as a git flag (`git switch -c` creates a branch, etc.). Branches from `git branch --list` never start with `-`, but the handler is reachable from any renderer code path — a one-line `if (branchName.startsWith('-'))` reject keeps it safe regardless of caller.

**Scope**

7. The `fix(node-pty): patch winpty.gyp paths and harden POSIX spawn errors` commit is unrelated to the git graph/branch-switcher feature. You already offered to split it — yes please, that'd be much easier to review and revert independently.

**Minor / discussion**

8. The 220ms single-click delay on the branch chip adds latency to a primary action (open git log). Worth confirming with the team that this UX cost is acceptable vs. e.g. moving "switch branch" to a separate affordance.

Once the items above are addressed, happy to take another pass.

jSydorowicz21:
@greptile re-review this pr please

jSydorowicz21:
@feldsys please address the orphan issue in the latest greptile feedback

feldsys:
@jSydorowicz21  i did :)

chr1syy (commented):
Took a fresh pass on this — the substantive items from the earlier round are all addressed, so flagging as comment rather than request-changes.

## What's holding it up

1. **Stale rebase.** Last update 2026-05-20, ~50 commits behind `rc`. Trial-merge into current `rc` produces exactly one conflict — `src/__tests__/renderer/components/MainPanel.test.tsx`. `rc` now stubs `useLayerStack` directly with the full `registerLayer`/`unregisterLayer`/... shape; this PR stubs `useModalLayer`. PR's approach is the canonically correct one per CLAUDE.md (`useModalLayer()` is listed as the canonical mock, "do NOT use manual `registerLayer()` boilerplate"), so the resolution is to keep PR's stub and layer the `useLayerStack` stub on top, or refactor `MainPanelContent` to also go through `useModalLayer`. Everything else auto-merges (the IPC handler additions land in a different region than `rc`'s new `git:init` handler).

2. **CI has never run.** All six workflow runs sit at `action_required` — first-time-contributor gate. We have no lint/test signal on the latest commit, only CodeRabbit (✅) and Greptile rounds. Worth a maintainer click to start the workflow once the conflict is resolved.

## Earlier review checklist — all addressed

| # | Item from 2026-05-07 review | Status |
|---|---|---|
| 1 | `switching` could stick on IPC reject | ✅ try/catch/finally in `doSwitch` |
| 2 | Graph fetched 100, list 200 (desync) | ✅ shared `COMMIT_FETCH_LIMIT` |
| 3 | `getGraph` swallowed `result.error` | ✅ rethrows; `graphError` state surfaces it |
| 4 | `localStorage.getItem` initializer unguarded | ✅ try/catch fallback to `'list'` |
| 5 | `@gitgraph/react` children-callback doesn't react to prop changes | ✅ refactored to `<Gitgraph graph={GitgraphCore}>` via `useMemo` — better than the suggested `key=` remount, and StrictMode-safe |
| 6 | Validate `branchName` against flag injection | ✅ rejects empty / leading `-` |
| 7 | node-pty patch out of scope | ✅ no longer in diff (12 files, +1035/-34) |

Bonus fixes the author shipped on top: ASCII Unit Separator (`%x1f`) instead of `|` as the `git log` field delimiter (real fix — author names and subjects can legally contain `|`), full hashes as commit IDs (avoids React-key collisions on short-hash clashes), `useCallback` around `handleGraphCommitClick` so `GitGraphView`'s `useMemo` doesn't rebuild the whole graph every parent render, debounce timer cleared on unmount, Shift+Enter keyboard a11y for the branch switcher, `captureException` around the fire-and-forget `refreshGitStatus()`.

## Minor leftovers — optional, non-blocking

- **Graph data refetches on every toggle to Graph mode** (`GitLogViewer.tsx` ~line 123–144). The lazy-load effect fires unconditionally on `viewMode === 'graph'`; no `useRef`-keyed "already loaded" guard. CodeRabbit flagged this as a quick-win nitpick. Noticeable latency on SSH remotes for every Graph re-toggle.
- **`useEffect(loadLog, [cwd])` missing `sshRemoteId`** in `GitLogViewer.tsx`. Pre-existing pattern, but worth folding in if you're already touching the file.
- **`BranchSwitcherDropdown` doesn't forward `remoteCwd`** to `getBranches` / `switchBranch`, even though the new `switchBranch` service accepts it. This matches the existing `gitService.getBranches(cwd, sshRemoteId)` signature in `rc`, so the regression isn't introduced here — but SSH sessions with a divergent `remoteCwd` (set by terminal `cd` tracking) will hit the wrong directory when switching branches. Worth a follow-up issue.

## On the 220 ms single-click delay

Author kept the 220 ms and added Shift+Enter for keyboard parity. The mouse path still pays the delay on every git-log open. Reasonable tradeoff if double-click is the only way to disambiguate; if anyone wants single-click for git-log + a dedicated `⌄` chevron next to the branch chip for the switcher, that's a small follow-up.

## Bottom line

Solid work, responsive author, both new affordances threaded cleanly through the SSH-remote pathway. Just needs a rebase + maintainer-kicked CI to land.

feldsys:
@chr1syy sorry for the late response, i did the rebase now :) thank you, for your review

jSydorowicz21:
@greptile @coderabbit new review pass please

jSydorowicz21:
While those are going we do have a short freeze for merging new features in so we can cut a new mainline release but if this ones in good shape we'll get it merged into the next rc!

pedramamini:
@feldsys thanks again for staying on top of this one. You worked through the entire previous round, split the node-pty Windows fix out into its own change, and rebased once already, which is exactly the kind of iteration we love to see. CodeRabbit's latest pass confirms all six correctness items from the earlier review are resolved. Really nice work.

One thing before we can land it: `rc` has moved on since your last rebase, so GitHub now shows this branch as conflicting (dirty merge state). Could you kindly rebase onto the latest `rc` and resolve the conflicts so we get a clean merge? Once that's in we're good to go.

While you're pushing that rebase, there's one small leftover from CodeRabbit's latest pass worth folding in (totally optional, not a blocker):

- `loadCommitDiff` in `src/renderer/components/GitLogViewer.tsx` closes over the `sshRemoteId` prop but only lists `cwd` in its `useCallback` dependency array. If `sshRemoteId` ever changes it could fetch a diff via the stale remote. It's a pre-existing nit rather than something this PR introduced, but since you'll be touching the file for the rebase anyway, adding `sshRemoteId` to the deps is a cheap tidy-up.

Heads up echoing @jSydorowicz21's earlier note: we're in a short merge freeze while we cut a mainline release, so once the conflicts are cleared this is queued for the next `rc`. Ping me when the rebase is up and I'll take the final pass. Thanks for the great contribution!

pedramamini:
@feldsys the git graph view + branch switcher is conflicting with `rc` now. Could you rebase onto the latest `rc` and re-push when you get a chance? Would like to take another look once it is green.

feldsys:
@pedramamini i rebased evrything

chr1syy:
@greptileai full review please
@coderabbitai full review please

pedramamini:
Merged into `rc` as commit 6f2e9e901 (squash). I resolved a conflict in MainPanel.test.tsx against rc's newer `renderMainPanel` test helper (kept your three SSH-badge tests: double-click branch switcher, Shift+Enter a11y, single-click git log) and added the two new channels (`git:graph`, `git:switch`) to the git.test.ts `expectedChannels` array so the handler-count assertion passes. Full suite green (34,167 tests). Command construction reviewed: both handlers route through `execFile`/argv (no shell interpolation) and `git:switch` rejects flag-like refs, so no injection surface. Thanks for this - closing since the commits are on rc under different SHAs.

PR #963, opened 2026-05-07
fix(node-pty): patch winpty.gyp paths and harden POSIX spawn errors

feldsys:
Split out from #962 per @pedramamini's review — this commit is unrelated to the git graph / branch switcher feature and is easier to review and revert independently here.

## Summary

Two fixes against the patched `node-pty@1.1.0`:

- **`winpty.gyp` (Windows build):** prefix the `GetCommitHash.bat` / `UpdateGenVersion.bat` invocations with `.\\` so `cmd /c` finds them when the working directory isn't on the lookup path. Without this the native build fails on Windows.
- **`pty_posix_spawn` (macOS error reporting):** surface the actual `posix_spawnp` errno via a `format_error` helper and a `std::string err` out-param, instead of collapsing every failure into the generic `"posix_spawnp failed."` string.

### CodeRabbit fix folded in

The original PR #962 version of this patch had an OOB / uninitialised-FD bug in the cleanup loop (CodeRabbit flagged it). This branch addresses it:

- Initialise `int low_fds[3] = { -1, -1, -1 };`
- Replace `for (size_t i = 0; i <= count; i++) close(low_fds[i]);` with a bounded, `-1`-aware variant that iterates `0..3` and skips uninitialised entries — no array OOB, no `close()` on stack garbage.

## Why

`node-pty` ships an upstream patch via `patch-package`. The Windows build silently breaks for new contributors without the gyp prefix fix, and a bare `"posix_spawnp failed."` made macOS spawn issues effectively unreproducible.

## Test plan

- [ ] `npm install` on Windows applies the patch cleanly via `patch-package`.
- [ ] `npm install` on macOS applies cleanly.
- [ ] On macOS, force a spawn failure (e.g. invalid command) and confirm the thrown `Napi::Error` carries the formatted `"posix_spawn failed: <strerror>"` instead of the legacy generic string.

### DIFF
--- a/patches/node-pty+1.1.0.patch
+++ b/patches/node-pty+1.1.0.patch
@@ -1,3 +1,25 @@
+diff --git a/node_modules/node-pty/deps/winpty/src/winpty.gyp b/node_modules/node-pty/deps/winpty/src/winpty.gyp
+index 1111111..2222222 100644
+--- a/node_modules/node-pty/deps/winpty/src/winpty.gyp
++++ b/node_modules/node-pty/deps/winpty/src/winpty.gyp
+@@ -10,7 +10,7 @@
+     #    make -j4 CXX=i686-w64-mingw32-g++ LDFLAGS="-static -static-libgcc -static-libstdc++"
+
+     'variables': {
+-        'WINPTY_COMMIT_HASH%': '<!(cmd /c "cd shared && GetCommitHash.bat")',
++        'WINPTY_COMMIT_HASH%': '<!(cmd /c "cd shared && .\\GetCommitHash.bat")',
+     },
+     'target_defaults' : {
+         'defines' : [
+@@ -22,7 +22,7 @@
+         'include_dirs': [
+             # Add the 'src/gen' directory to the include path and force gyp to
+             # run the script (re)generating the version header.
+-            '<!(cmd /c "cd shared && UpdateGenVersion.bat <(WINPTY_COMMIT_HASH)")',
++            '<!(cmd /c "cd shared && .\\UpdateGenVersion.bat <(WINPTY_COMMIT_HASH)")',
+         ]
+     },
+     'targets' : [
 diff --git a/node_modules/node-pty/src/unix/pty.cc b/node_modules/node-pty/src/unix/pty.cc
 index 7b4b9e1..2e5293d 100644
 --- a/node_modules/node-pty/src/unix/pty.cc
@@ -9,11 +31,11 @@ index 7b4b9e1..2e5293d 100644
 -                int* err);
 +                std::string* err);
  #endif
- 
+
  struct DelBuf {
 @@ -352,7 +352,7 @@ Napi::Value PtyFork(const Napi::CallbackInfo& info) {
    std::string helper_path = info[9].As<Napi::String>();
- 
+
    pid_t pid;
 -  int master;
 +  int master = -1;
@@ -23,7 +45,7 @@ index 7b4b9e1..2e5293d 100644
 @@ -367,10 +367,13 @@ Napi::Value PtyFork(const Napi::CallbackInfo& info) {
      argv[i + 3] = strdup(arg.c_str());
    }
- 
+
 -  int err = -1;
 +  std::string err;
    pty_posix_spawn(argv, env, term, &winp, &master, &pid, &err);
@@ -39,7 +61,7 @@ index 7b4b9e1..2e5293d 100644
      throw Napi::Error::New(napiEnv, "Could not set master fd to nonblocking.");
 @@ -684,15 +687,26 @@ pty_getproc(int fd, char *tty) {
  #endif
- 
+
  #if defined(__APPLE__)
 +static std::string format_error(const char* func, int err_code) {
 +  char buf[256];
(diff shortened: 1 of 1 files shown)


pedramamini:
Thanks for splitting this out, @feldsys — much easier to reason about in isolation. 🙏

Took a careful look at the patch:

- **`winpty.gyp` `.\\` prefix:** Correct — `cmd /c` resolves the batch files relative to the `cd shared` working directory but won't search PATH-style for the unprefixed name in all environments. Straightforward fix.
- **`pty_posix_spawn` rewrite:**
  - `low_fds[3] = { -1, -1, -1 }` and the bounded `0..3` `-1`-aware cleanup loop properly resolve the OOB / uninitialised-FD issue from #962 — confirmed.
  - `acts` / `attrs` are now declared and `*_init()`ed before any `goto done`, so the `done:` block always destroys initialised objects. ✅
  - `slave` defaults to `-1` and is closed in `done:` only when set — no double-close, no leak.
  - `master` is initialised to `-1` in `PtyFork` and closed by the caller on error before throwing. Lifetime is consistent across success and failure paths.
  - Each failure site captures `errno` (or the `posix_spawnattr_*` return) into `spawn_err` before any cleanup syscalls, so the formatted message reflects the original cause. ✅

CodeRabbit's `strerror_r` nit is fair in principle but the helper is only invoked on the libuv/N-API main thread for `PtyFork`, so I'm comfortable leaving it as-is for this PR — happy to revisit if `node-pty` ever calls into this from a worker.

Approving. ✅

pedramamini:
Hey @feldsys, thanks again for splitting this out and for the careful work on it. 🙏

Heads-up on the merge conflict that's now showing: since I approved this on May 7, `rc` has moved on, and the cause here isn't a stray line or two, it's more fundamental.

`rc` merged #967 on May 10, which bumped `node-pty` from `^1.1.0` to a pinned `1.2.0-beta.12` and removed `patches/node-pty+1.1.0.patch` outright. Per that PR, the macOS `posix_spawn` fixes in your patch (rich error reporting via `std::string*`, closing the FD on the error paths, and the corrected close-low-fds loop bound) are now upstream in `1.2.0-beta.12`, so the local fork was retired because `patch-package` was failing against the stale patch.

So the macOS half of this change has effectively landed already, just via the version bump rather than the patch. A rebase wouldn't cleanly resolve the conflict either: `patch-package` keys the patch to the dependency version in the filename (`node-pty+1.1.0.patch`), and `rc` no longer resolves `node-pty` to 1.1.0.

The one piece that may not be covered by the bump is the `winpty.gyp` Windows build fix (the `.\` prefix on the batch invocations). If that's still needed against `1.2.0-beta.12`, it'd want a fresh, Windows-only patch named for the new version (`node-pty+1.2.0-beta.12.patch`). If 1.2.0-beta.12 already builds cleanly on Windows, this one can just be closed as superseded.

Could you take a look and let us know how you'd like to proceed: close it out as superseded, or open a slim Windows-only patch against the new version if the winpty issue persists? Happy to review the latter promptly. Sorry for the churn, and thanks again for the contribution.

pedramamini:
@feldsys this node-pty hardening PR is conflicting with `rc`. Could you rebase onto the latest `rc` and re-push? It is a small, useful fix and I would like to land it once it is green.

pedramamini:
Hi @feldsys - thanks again for splitting this out and for the care you put into the cleanup-loop fix. 🙏

I went back through this one against current `rc` and I owe you a correction on my July 4 comment, because part of it was wrong and it matters for how we close this out.

**What I got wrong:** I said the corrected close-low-fds loop bound had landed upstream in `1.2.0-beta.12`. It has not. I checked the installed dependency on current `rc`:

- `node_modules/node-pty/src/unix/pty.cc:744` is still `int low_fds[3];` (uninitialised)
- `node_modules/node-pty/src/unix/pty.cc:857` is still `for (size_t i = 0; i <= count; i++) { close(low_fds[i]); }`

That is the same off-by-one you fixed. The fill loop above it runs `count` up to 3 and only `break`s once `posix_openpt` returns an fd `>= STDERR_FILENO`, so when all three calls come back low (or fail), `count` exits at 3 and the cleanup loop reads `low_fds[3]`, one past the end, and calls `close()` on stack garbage. It is a narrow path (it needs stdin/stdout/stderr to be closed, which is exactly the case this code exists to handle), but it is real and it is still there.

**What did land upstream:** the `format_error` / `std::string* err` rework is in `1.2.0-beta.12` essentially verbatim, same helper name and same per-site messages. So that half of your patch is genuinely superseded.

**What is now moot:** the `winpty.gyp` `.\` prefix. `1.2.0-beta.12` dropped winpty entirely. There is no `deps/` directory in the package any more and `src/win/` is ConPTY only (`conpty.cc`, `conpty_console_list.cc`, `path_util.cc`). The only remaining mentions of winpty are a code comment and a doc string, so there is no longer a `.bat` invocation to prefix.

**Why this branch still cannot merge as-is:** #967 removed `patches/` from `rc` altogether and pinned `node-pty` to `1.2.0-beta.12`. A rebase will not resolve it, since `patch-package` keys the patch to the resolved version in the filename and `rc` no longer resolves `node-pty` to 1.1.0. My July 10 "please rebase" ask was not really actionable, sorry about that.

**Suggested path forward,** if you are still up for it: a slim replacement PR carrying only the `low_fds` hardening, as `patches/node-pty+1.2.0-beta.12.patch`. Just the two edits you already wrote:

```c
int low_fds[3] = { -1, -1, -1 };
...
for (size_t i = 0; i < 3; i++) {
  if (low_fds[i] != -1) {
    close(low_fds[i]);
  }
}
```

That reintroduces a `patches/` directory that #967 deliberately retired, so it is worth being deliberate about, but a one-hunk patch for a live OOB seems like a fair trade to me. This would also be a good candidate to send upstream to microsoft/node-pty, since the bug is in their tree and an accepted upstream fix would let us drop the local patch again on the next bump. Happy to review either or both promptly.

If you would rather not carry this further, no problem at all and no hard feelings. Say the word and I will close this one out as superseded and open a tracking issue for the `low_fds` bug so it does not get lost. Either way, thanks for catching it - it would still be sitting there unnoticed otherwise.

PR #979, opened 2026-05-10
fix: preserve duplicated agent name and inherit source group (#827)

### LINKED ISSUE #827: [Bug] Duplicate agent: renamed name reverts to "(Copy)" on save, and group is not inherited
When using the **Duplicate...** context menu action on an agent, two bugs occur:

### Bug 1 — Renamed agent name reverts to `(Copy)` on save

After duplicating an agent, the `NewInstanceModal` pre-fills the name field with `{originalName} (Copy)`. If the user edits the name and clicks **Create**, the saved name still ends up as `{originalName} (Copy)` instead of the user's typed value.

**Root cause:** The `useEffect` in `NewInstanceModal.tsx` (line 497) has `sourceSession` in its dependency array. `sourceSession` is computed via `useMemo` in `AppModals.tsx` (line 755) from the full `sessions` array. When the `sessions` array gets a new reference during the modal interaction (e.g. from any store update), `sourceSession` becomes a new object even though its data hasn't changed — re-triggering the effect and calling `setInstanceName(`${source.name} (Copy)`)`, which wipes out whatever the user typed.

**Fix:** Gate the pre-fill logic so it only runs once on initial open (e.g. track a `hasPreFilled` ref, or use `sourceSession?.id` as the dependency instead of the full object).

**Files:**
- `src/renderer/components/NewInstanceModal/NewInstanceModal.tsx` — line 226 / lines 496–511

---

### Bug 2 — Duplicated agent does not inherit the source agent's group

When an agent belonging to a group is duplicated, the new agent is not placed in the same group.

**Root cause:** `sourceSession.groupId` is never propagated through the duplication flow:
1. `NewInstanceModal.handleCreate` doesn't pass `groupId` to `onCreate` (line 388–400)
2. `useSessionCrud.createNewSession` doesn't accept a `groupId` parameter (line 127–144)
3. The new `Session` object is created without a `groupId`

The pattern for preserving `groupId` already exists in `useMergeSession.ts` (line 446) and `useSendToAgent.ts` (line 392).

**Fix:** Thread `groupId` from `sourceSession` through the entire duplication call chain.

**Files:**
- `src/renderer/components/NewInstanceModal/NewInstanceModal.tsx` — `handleCreate` callback
- `src/renderer/components/AppModals/AppSessionModals.tsx` — `onCreateSession` signature
- `src/renderer/hooks/session/useSessionCrud.ts` — `createNewSession` function

---

## Fix Prompt

> Fix two bugs in the duplicate agent flow:
>
> **Bug 1 — Name reverts to "(Copy)" on save:**
> In `src/renderer/components/NewInstanceModal/NewInstanceModal.tsx`, the `useEffect` at line 497 depends on `sourceSession` (the full object). When the `sessions` store updates during modal interaction, `sourceSession` gets a new reference, the effect re-fires, and `setInstanceName` overwrites whatever the user typed at line 226.
>
> Fix: Change the effect dependency from `sourceSession` to `sourceSession?.id`. Inside the effect, retrieve the source session by ID from the current component state or pass it explicitly, so the pre-fill only triggers when the *identity* of the source changes, not when the sessions array gets a new reference.
>
> **Bug 2 — Group not inherited:**
> The `sourceSession.groupId` is never passed through the duplication flow. Thread it through:
> 1. In `NewInstanceModal.tsx`, add an optional `groupId` parameter to the `onCreate` prop type, and pass `sourceSession?.groupId` from `handleCreate`.
> 2. In `AppSessionModals.tsx`, update `onCreateSession` prop type to include `groupId?: string`.
> 3. In `useSessionCrud.ts` `createNewSession`, add `groupId?: string` to the parameter list and apply it when constructing the new session object (follow the existing pattern in `useMergeSession.ts` line 446).
>
> Do not change any other behavior. Run `npm run lint` and `npm run test` after the fix.

chr1syy:
Fixes #827 — two related bugs in the **Duplicate agent** flow.

### Bug 1 — Renamed name reverts to "(Copy)" on save

`NewInstanceModal`'s pre-fill effect depended on the full `sourceSession` object. `sourceSession` is computed in `AppModals.tsx` via `useMemo(() => sessions.find(...), [duplicatingSessionId, sessions])`. Whenever the `sessions` store updates (which happens for many unrelated reasons during modal interaction), the memo returns a *new object reference* even though the underlying data is unchanged. That re-fires the pre-fill effect and `setInstanceName(\`${source.name} (Copy)\`)` clobbers whatever the user has typed.

**Fix:** Change the effect dependency from `sourceSession` to `sourceSession?.id`. The effect now only re-fires when the source's *identity* changes, not on every parent re-render.

### Bug 2 — Group not inherited

`sourceSession.groupId` was never propagated through the duplication flow. The new agent always landed at the top level even when the original belonged to a group.

**Fix:** Threaded an optional `groupId` argument through:
- `NewInstanceModal.handleCreate` defaults it to `sourceSession?.groupId`
- `onCreate` / `onCreateSession` callback signatures (`NewInstanceModal/types.ts`, `AppSessionModals.tsx`, `AppModals.tsx`)
- `useSessionCrud.createNewSession` accepts and applies it on the new `Session`

This matches the existing pattern used by `useMergeSession` and `useSendToAgent`.

## Test plan

Automated:

- [x] Added regression test: name field is preserved when `sourceSession` reference changes (`NewInstanceModal.test.tsx`)
- [x] Added test: `groupId` from `sourceSession` is forwarded to `onCreate` as the 14th positional arg
- [x] Added test: `useSessionCrud.createNewSession` writes `groupId` onto the new `Session`
- [x] Updated all existing `expect(onCreate).toHaveBeenCalledWith(...)` assertions to account for the new trailing `groupId` arg
- [x] All 76 `NewInstanceModal` tests + all 53 `useSessionCrud` tests pass
- [x] Prettier and ESLint clean

Manual:

- [ ] Move an agent into a group → right-click → **Duplicate...** → confirm the duplicate lands in the same group
- [ ] Open **Duplicate...** → edit the pre-filled name → wait for any unrelated session update (e.g. another agent changes state) → confirm the typed name is not reset
- [ ] Click **Create Agent** → confirm the saved name matches what was typed (not `… (Copy)`)

## Summary by CodeRabbit

* **New Features**
  * Sessions can now be assigned to groups for improved organization and management
  * Group assignments are automatically preserved when duplicating sessions

[Review Change Stack](https://app.coderabbit.ai/change-stack/RunMaestro/Maestro/pull/979)

### DIFF
--- a/src/__tests__/renderer/components/NewInstanceModal.test.tsx
+++ b/src/__tests__/renderer/components/NewInstanceModal.test.tsx
@@ -731,6 +731,7 @@ describe('NewInstanceModal', () => {
 				undefined,
 				undefined,
 				{ enabled: false, remoteId: null },
+				undefined,
 				undefined
 			);
 		});
@@ -779,6 +780,7 @@ describe('NewInstanceModal', () => {
 				undefined,
 				undefined,
 				{ enabled: false, remoteId: null },
+				undefined,
 				undefined
 			);
 		});
@@ -827,6 +829,7 @@ describe('NewInstanceModal', () => {
 				undefined,
 				undefined,
 				{ enabled: false, remoteId: null },
+				undefined,
 				undefined
 			);
 		});
@@ -876,6 +879,7 @@ describe('NewInstanceModal', () => {
 				undefined,
 				undefined,
 				{ enabled: false, remoteId: null },
+				undefined,
 				undefined
 			);
 			expect(onClose).toHaveBeenCalled();
@@ -1388,6 +1392,7 @@ describe('NewInstanceModal', () => {
 				undefined,
 				undefined,
 				{ enabled: false, remoteId: null },
+				undefined,
 				undefined
 			);
 		});
@@ -1536,6 +1541,7 @@ describe('NewInstanceModal', () => {
 				undefined,
 				undefined,
 				{ enabled: false, remoteId: null },
+				undefined,
 				undefined
 			);
 		});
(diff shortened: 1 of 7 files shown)


pedramamini:
@chr1syy Thanks for the contribution! Nice clean fix — the `sourceSession?.id` dependency change is exactly right (the full-object dep was guaranteed to thrash on every store update), and threading `groupId` through to match the existing `customEffort` pattern keeps the chain consistent. The regression tests for both bugs are well-targeted, especially the reference-change test that documents *why* the dep was wrong. Approving.

PR #982, opened 2026-05-11
fix(cue): answer get_cue_subscriptions from main, not via dead renderer IPC

chr1syy:
## Summary

- Fixes \`maestro-cli cue list\` hanging with \`Command timed out waiting for cue_subscriptions\`.
- Root cause: the web-server factory forwarded \`remote:getCueSubscriptions\` to the renderer and waited 30 s for a response, but no renderer code ever registered a handler for that channel. The CLI's 10 s timeout fired every time.
- Mirrors the existing \`triggerCueSubscription\` pattern: a new \`getCueGraphData\` dependency on \`WebServerFactoryDependencies\` calls \`cueEngine.getGraphData()\` directly in main. The callback flattens \`CueGraphSession[]\` into \`CueSubscriptionInfo[]\` and optionally filters by \`sessionId\`. No more renderer round-trip.
- \`triggerSubscription(...)\` and the engine-side getter were already in main process, so the renderer bounce was never needed in the first place.

### Follow-up tracked here, not fixed in this PR

The same dead-bridge shape exists for \`remote:toggleCueSubscription\` and \`remote:getCueActivity\` — neither is reachable from any current CLI command, so they're not user-blocking right now. Recommend moving both to main-process direct calls in a follow-up (and removing the unused \`onRemoteGetCueSubscriptions\` / \`sendRemoteGetCueSubscriptionsResponse\` / \`onRemoteToggleCueSubscription\` / \`sendRemoteToggleCueSubscriptionResponse\` / \`onRemoteGetCueActivity\` / \`sendRemoteGetCueActivityResponse\` preload bridges, which now have zero callers).

## Test plan

- [x] \`vitest run src/__tests__/main/web-server/web-server-factory.test.ts\` — 51/51 pass, including three new tests:
  - flattens engine graph data into \`CueSubscriptionInfo[]\` without any IPC bounce
  - filters by \`sessionId\` when supplied
  - returns \`[]\` and warns when the dep is missing
- [x] \`vitest run src/__tests__/main/cue/cue-yaml-loader.test.ts src/__tests__/renderer/components/CuePipelineEditor/utils/yamlToPipeline.test.ts\` — 188/188 pass (no regressions in adjacent cue paths).
- [x] Type-check + prettier + ESLint clean on touched files.
- [ ] Smoke test against a workspace with a loaded \`.maestro/cue.yaml\`:
  - \`node /opt/Maestro/resources/maestro-cli.js cue list --json\` returns subscriptions instead of timing out.
  - \`cue trigger <existing-sub-name>\` still works (unchanged path).

## Summary by CodeRabbit

* **New Features**
  * Web API exposes cue subscription lookup with stable subscription IDs, clearer schedule formatting, legacy YAML handling, and sessionId filtering.
  * CLI-facing schedule strings now render day+time combinations more readably.

* **Tests**
  * Added comprehensive tests covering subscription ID composition, disambiguation, schedule rendering, filtering, and missing-data warnings.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RunMaestro/Maestro/pull/982)

### DIFF
--- a/src/__tests__/main/web-server/web-server-factory.test.ts
+++ b/src/__tests__/main/web-server/web-server-factory.test.ts
@@ -1019,4 +1019,133 @@ describe('web-server/web-server-factory', () => {
 			);
 		});
 	});
+
+	describe('Cue subscription callbacks', () => {
+		// Regression: previously this callback forwarded the request to the
+		// renderer via `remote:getCueSubscriptions` and waited 30 s for a
+		// response, but no renderer handler existed. Every `maestro-cli cue
+		// list` call timed out. Now it must call the injected graph-data
+		// dependency directly and flatten the result.
+		it('flattens engine graph data into CueSubscriptionInfo[] without any IPC bounce', async () => {
+			const getCueGraphData = vi.fn().mockReturnValue([
+				{
+					sessionId: 'agent-1',
+					sessionName: 'Obsidian Digest',
+					toolType: 'claude-code',
+					subscriptions: [
+						{
+							name: 'Digest Script',
+							event: 'time.scheduled',
+							enabled: true,
+							prompt: '',
+							schedule_times: ['07:00'],
+							action: 'command',
+						},
+						{
+							name: 'Obsidian Daily Pipe-chain-1',
+							event: 'agent.completed',
+							enabled: true,
+							prompt: 'follow up',
+							source_session: 'Obsidian Digest',
+							source_sub: 'Digest Script',
+						},
+					],
+				},
+				{
+					sessionId: 'agent-2',
+					sessionName: 'Obsidian Git',
+					toolType: 'claude-code',
+					subscriptions: [
+						{
+							name: 'Git Script',
+							event: 'time.scheduled',
+							enabled: false,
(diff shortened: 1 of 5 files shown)


pedramamini:
Thanks for the contribution @chr1syy — this is a clean, well-scoped fix. The diagnosis is spot-on: the renderer never registered a handler for `remote:getCueSubscriptions`, so the 30s wait + 10s CLI timeout was deterministic. Bouncing to the renderer when the engine lives in main was always architecturally wrong, and the new shape mirrors `triggerCueSubscription` cleanly. Tests cover the flatten, the `sessionId` filter, and the missing-dep fallback — nice.

One non-blocking observation for a follow-up: the new `schedule` string surfaces `schedule_times` and `interval_minutes` but not `schedule_days`. A `time.scheduled` subscription that pins to specific weekdays will render just the times in `cue list`. Pre-existing-style gap (the old path returned nothing at all), but worth folding in when you tackle the other dead bridges you flagged (`remote:toggleCueSubscription`, `remote:getCueActivity`).

Approving.

PR #983, opened 2026-05-11
fix(cue): wire toggle + activity through main, drop dead preload bridges

chr1syy:
## Summary

Follow-up to **#982** (\`fix/cue-remote-subscriptions\`). Same dead-bridge shape existed for two more cue IPC channels — fixing them and removing the unused preload bridges so the cue control surface is consistent.

> Stacked on #982. If #982 is merged first, this PR rebases to one commit. If reviewed together, both can land in one rc.

### What was broken

The web-server factory forwarded \`remote:toggleCueSubscription\` and \`remote:getCueActivity\` to the renderer and waited 10 s / 30 s for a response on IPC channels **no renderer code had ever subscribed to**. Net effect:

- Per-row Cue subscription toggles in the web UI silently no-op'd after a 10 s stall.
- The web UI activity tab always rendered empty after a 30 s stall.

### Fix

Mirror the \`triggerCueSubscription\` / \`getCueGraphData\` pattern from #982. Two new deps on \`WebServerFactoryDependencies\`:

- \`setCueSubscriptionEnabled(subscriptionId, enabled): boolean\`
- \`getCueActivityLog(): CueRunResult[]\`

Wired in \`main/index.ts\` from \`cueEngine\`. The toggle needs an engine-side mutation, so this PR adds **\`CueEngine.setSubscriptionEnabled\`**, which parses the \`\${sessionId}::\${name}\` id format (same format \`getCueGraphData\`'s flattener emits in #982), looks up the owning session, reads → parses → mutates → writes the YAML, and calls \`refreshSession\`. Comments and field ordering aren't preserved on the round-trip — same trade-off as the pipeline editor's save path.

Activity mapping converts engine \`CueRunResult[]\` → web \`CueActivityEntry[]\`, applies the optional \`sessionId\` filter **before** slicing to \`limit\` (so the caller gets N matching entries, not N total of which some are filtered), and collapses engine-side \`timeout\` / \`stopped\` into the web-facing \`failed\` enum.

### Dead-code cleanup

Removed the six matching preload bridges from \`src/main/preload/process.ts\`:

- \`onRemoteGetCueSubscriptions\` / \`sendRemoteGetCueSubscriptionsResponse\`
- \`onRemoteToggleCueSubscription\` / \`sendRemoteToggleCueSubscriptionResponse\`
- \`onRemoteGetCueActivity\` / \`sendRemoteGetCueActivityResponse\`

\`grep -rn\` confirms zero remaining references anywhere in \`src/\`. The trigger bridge (\`onRemoteTriggerCueSubscription\`) is **kept** — it's still wired and used by \`useRemoteIntegration\`.

## Test plan

- [x] \`vitest run src/__tests__/main/web-server/web-server-factory.test.ts\` — 58/58 pass. 7 new tests:
  - toggle: dispatches through dep, propagates false return, warns when dep missing.
  - activity: maps \`CueRunResult\` → \`CueActivityEntry\`, filters by \`sessionId\` before \`limit\`, collapses \`timeout\`/\`stopped\` to \`failed\`, warns when dep missing.
- [x] \`vitest run src/__tests__/main/cue/cue-engine.test.ts\` — 115/115 pass. 5 new tests for \`setSubscriptionEnabled\`: malformed ids, unresolved session, missing YAML, missing subscription, happy path (parse → mutate → write → refresh).
- [x] Prettier + ESLint clean on touched files.
- [ ] Smoke test in a new rc build:
  - Toggle a Cue subscription from the web UI → reload → confirm \`.maestro/cue.yaml\` reflects the new \`enabled:\` value and the engine reattached its triggers.
  - Open the web UI activity tab → confirm recent runs render.
  - Confirm \`maestro-cli cue list --json\` still works (regression guard for #982).

## Release Notes

* **New Features**
  * Added ability to remotely enable or disable Cue subscriptions
  * Implemented Cue activity log retrieval with session filtering and status tracking

[Review Change Stack](https://app.coderabbit.ai/change-stack/RunMaestro/Maestro/pull/983)


pedramamini:
@chr1syy thanks for the follow-up — nice surgical cleanup, and the test coverage on both new paths (`setSubscriptionEnabled` happy/failure modes + the activity-log mapping + sessionId-filter-before-limit) is exactly what I'd want to see.

One concern from the bot reviews I'd like your read on before I approve:

**Subscription identity (CodeRabbit, major).** `${sessionId}::${name}` assumes names are unique within a session, but the validator contract you cite is unique-within-pipeline. Two pipelines under the same session can each legitimately have a sub named `Digest Script`, in which case `setSubscriptionEnabled` finds the first match and silently toggles the wrong row — and the web UI has no way to tell. Options I see:
- Encode pipeline into the id (`${sessionId}::${pipeline}::${name}`) on both the flatten side (`getCueGraphData` callback in web-server-factory) and the parse side here, OR
- Resolve by stable index from the flattener output, OR
- Tighten the validator contract to unique-within-session and call that out.

Whichever you prefer is fine, but the current shape is foot-gun-y once someone has two pipelines.

**Read-modify-write atomicity (Greptile, P2).** Same trade-off as the pipeline editor's save path, as you noted, so I'd just file it as a known limitation rather than blocking on it here. If you want to preemptively add a single-flight lock per project root in a follow-up that'd be welcome, but not for this PR.

Happy to merge as soon as the identity question is resolved (either a fix or a "won't happen because X" answer).

PR #985, opened 2026-05-11
chore(ci): remove no-op build:prompts step (#804)

### LINKED ISSUE #804: Remove `build:prompts` step from CI pipelines
After #798 is merged, the `build:prompts` step (`npm run build:prompts`) is a no-op — prompts are now loaded from disk at runtime instead of being compiled into `src/generated/prompts.ts` at build time.

A no-op shim was added to `package.json` (`7189b06c`) so existing CI pipelines don't break, but the step should be removed from CI configs entirely since it does nothing.

Relates to #197.

### What to do

1. Search CI workflow files (GitHub Actions, etc.) for references to `build:prompts` or `generate-prompts`
2. Remove those steps
3. Remove the no-op `"build:prompts"` script from `package.json`

### Context

- `scripts/generate-prompts.mjs` was deleted
- `src/generated/prompts.ts` no longer exists
- `src/generated/` was removed from `.gitignore`
- The `build`, `dev:main`, `dev:main:prod-data`, and `validate:push` scripts already had their `build:prompts` references removed in #798
- The no-op shim only exists for external/cached CI pipelines that haven't been updated yet

chr1syy:
## Summary

Fixes #804. Prompts are loaded from disk at runtime (`src/main/prompt-manager.ts`); the `build:prompts` npm script was a no-op shim added in `7189b06c` so existing CI pipelines wouldn't break. Drops the script and its remaining call sites:

- **`.github/workflows/ci.yml`** — remove `npm run build:prompts` from both the `lint-and-format` and `test` jobs.
- **`package.json`** — remove the no-op `build:prompts` script.
- **`BUILDING_WINDOWS.md`** + **`scripts/start-dev.ps1`** — drop `npm run build:prompts;` from the documented Windows dev command and the PowerShell dev launcher (otherwise these would invoke a non-existent script).

## Out of scope (for a follow-up)

The issue's Context section claims `scripts/generate-prompts.mjs` and `src/generated/prompts.ts` no longer exist — they do still exist on `rc`. Removing them (and the references in `docs/agent-guides/PROMPTS-SPECS.md`) is a separate cleanup; this PR keeps the scope to what the issue's \"What to do\" section explicitly requests.

## Test plan

- [x] `node -e \"JSON.parse(...)\"` on `package.json` — valid
- [x] `npm run build:prompts` after the change — fails with \"missing script\" (expected; no callers remain in `package.json`, the CI workflow, or the dev scripts)

## Summary by CodeRabbit

* **Documentation**
  * Updated Windows development build instructions to reflect streamlined process.

* **Chores**
  * Removed unnecessary build step from CI workflow configuration, development startup scripts, and npm package configuration.
  * Simplified development startup sequence.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RunMaestro/Maestro/pull/985)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,6 @@ jobs:
           node-version: '22'
           cache: 'npm'
       - run: npm ci
-      - run: npm run build:prompts
       - run: npx prettier --check .
       - run: npx eslint src/
       - run: npm run lint # TypeScript type checking
@@ -30,5 +29,4 @@ jobs:
           node-version: '22'
           cache: 'npm'
       - run: npm ci
-      - run: npm run build:prompts
       - run: npm run test
--- a/BUILDING_WINDOWS.md
+++ b/BUILDING_WINDOWS.md
@@ -62,7 +62,7 @@ If you encounter issues with the `dev:win` script or prefer to run the steps man
 3.  **Start the Electron main process:**
     Open a **new** PowerShell terminal and run the following command:
     ```powershell
-    npm run build:prompts; npx tsc -p tsconfig.main.json; $env:NODE_ENV='development'; npx electron .
+    npx tsc -p tsconfig.main.json; $env:NODE_ENV='development'; npx electron .
     ```
 
 This will launch the application in development mode with hot-reloading for the renderer.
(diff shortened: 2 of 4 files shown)


pedramamini:

Clean mechanical removal of the no-op shim and all four call sites (CI workflow, package.json, Windows docs, PowerShell launcher). Diff matches the issue's "What to do" exactly, and your decision to defer the `generate-prompts.mjs` / `src/generated/prompts.ts` cleanup to a follow-up is the right call — keeps the scope tight and reviewable.

Approving.

PR #991, opened 2026-05-13
feat(web): web-desktop bundle Encore Feature + responsive mobile UX

ronaldeddings:
## Summary

Ships the full Maestro desktop UI through the browser by mirroring every `ipcMain` handler over a WebSocket bridge. Renderer code runs unchanged in the browser via a Vite-aliased `electron` shim. Gated behind a new `webDesktopBundle` Encore Feature so the IPC surface stays sealed by default.

## Architecture — bridge IPC mirror (1:1)

**Server (`src/main/web-server/handlers/bridgeHandlers.ts`)**
- Dispatches `bridge.invoke` messages against `ipcMain._invokeHandlers`
- Broadcasts `WebContents.send` calls as `bridge.event` to all web clients
- Empty/unknown channels reject loudly; handler exceptions surface their messages

**Web shim (`src/web-desktop/electron-shim.ts`)**
- Vite alias replaces the `electron` module at build time
- WebSocket-backed `ipcRenderer`, `contextBridge`, reconnect-aware `BridgeClient`
- Preload code runs unchanged in the browser

**Bootstrap & build**
- `src/web-desktop/bootstrap.ts` polyfills `window.process`, loads preload then renderer
- `@sentry/electron/renderer` aliased to a no-op shim
- `vite.config.web-desktop.mts` outputs `dist/web-desktop/`, served at `/<token>/desktop/`

## Encore Feature gate (4 enforcement points)

- `WebServer` constructor takes `getWebDesktopBundleEnabled` callback
- `web-server-factory.ts` wires it to `settingsStore.encoreFeatures.webDesktopBundle`
- Gated: route registration, static asset mount, bridge hook install, `bridge.invoke` dispatch
- Toggle via CLI: `maestro-cli settings set encoreFeatures.webDesktopBundle true`

## Responsive mobile UX

- **Breakpoints** — `useViewportBreakpoint` writes `data-bp` (xs/sm/md/lg/xl) on `<html>` so CSS targets without re-renders
- **Tri-state sidebar** — open → collapsed strip → hidden; drawer-style overlay below md
- **Header** — inline hamburger when sidebar hidden, context shown as `{N}%`, secondary buttons hide below 480px
- **Input toolbar** — pills collapse behind `…` on narrow viewports, nowrap to prevent wrap
- **File picker** — touch-aware single-tap open, 500ms long-press → context menu, capture-phase mousedown swallower for iOS
- **Git diff** — counter + side-by-side toggle hide below md, close becomes × with `aria-label`
- **Tab popover** — edge-flip so options don't clip at right edge
- **Shell CSS** — `100dvh` + `env(safe-area-inset-bottom)` for iOS safe areas

## iOS Safari fixes

- **`generateId`** — feature-detects `crypto.randomUUID` (requires secure context on Safari) and falls back to `generateUUID` from `shared/uuid.ts`
- **File open resilience** — `Promise.allSettled` so a failed `fs.stat` doesn't drop the file content from preview

## Tests

- New: `bridgeHandlers.test.ts` — 5 cases (empty channel, unknown channel, success, handler error, arg pass-through). Uses `vi.hoisted` to stub `ipcMain._invokeHandlers` without a real Electron runtime
- Updated: GitDiffViewer, MainPanel, MainPanelHeader, SessionList, SidebarActions tests for renamed labels/icons

## Validation

- `npm run lint` (tsc all configs): clean
- `npm run lint:eslint`: clean
- `npm run test`: **27,844 passed / 108 skipped / 0 failed**
- `npm run build`: main + renderer + web + web-desktop + cli all built
- Pre-push hook (`validate:push`): passed

## How to test

```bash
maestro-cli settings set webInterfaceUseCustomPort true
maestro-cli settings set webInterfaceCustomPort 31999
maestro-cli settings set persistentWebLink true
maestro-cli settings set encoreFeatures.webDesktopBundle true
```

Then open `http://<host>:31999/<token>/desktop/` from your browser — Safari, Chrome, mobile over Tailscale all supported.

## Out-of-scope notes

Three files included at Ron's request (personal workstreams, separate from the bundle):
- `PRD-CY-CONTEXT-RUNTIME.md` — Cy runtime integration PRD
- `scripts/import-vram-sessions.ts` — session import utility
- `scripts/populate-real-conversations.ts` — conversation data utility

## Summary by CodeRabbit

* **New Features**
  * Web-desktop bundle served at a secure token-scoped endpoint; desktop build/dev targets added
  * Responsive narrow-viewport UI: auto-collapsing drawers, mobile backdrop, optimized headers and toolbars
  * Touch support: long-press file actions and single-tap opens on touch devices
  * Left-sidebar cycles: expanded → collapsed strip → hidden
  * Improved viewport-aware controls and accessible close buttons

* **Bug Fixes**
  * More resilient remote file loading with safer fallbacks
  * Safer ID generation when crypto APIs are unavailable

* **Tests**
  * Added/updated tests for IPC bridge and responsive UI behavior

[Review Change Stack](https://app.coderabbit.ai/change-stack/RunMaestro/Maestro/pull/991)


pedramamini:
Thanks for the ambitious push here @ronaldeddings — the IPC mirror behind an Encore Feature flag is exactly the right shape for this, and the responsive UX work + iOS Safari fixes are great quality-of-life wins. PR is clean against `rc` with no merge conflicts.

One real bug worth fixing before we merge, plus a couple of smaller items:

**🔴 `BridgeClient` doesn't drain `pending` on WebSocket close** — `src/web-desktop/electron-shim.ts:105-109`

The `close` handler resets `ready` and reconnects, but never iterates `this.pending` to reject in-flight invokes. After a server restart (or any drop where the server doesn't replay), those requestIds are unknown to the new server, so every `ipcRenderer.invoke()` issued during the disconnect window hangs forever — and any React component awaiting that promise just stalls. Suggest something like:

```ts
this.ws.addEventListener('close', () => {
    const err = new Error('bridge disconnected');
    for (const [id, p] of this.pending) p.reject(err);
    this.pending.clear();
    this.queue.length = 0; // queued frames are also orphaned
    setTimeout(() => this.connect(url), 1000);
    this.ready = new Promise((r) => (this.resolveReady = r));
});
```

**🟡 Smaller items (optional, can be follow-ups):**
- `serveWebDesktopIndex` (`WebServer.ts:802`) does `readFileSync` on every desktop page load. The HTML is static after build — cache it on first read or read it once in `start()`.
- `useViewportBreakpoint` sets `data-bp` only inside `useEffect`, so first paint on narrow viewports flashes the desktop layout before the attribute lands. Setting it synchronously during hook init (or via an inline pre-mount script) would avoid the flash.

The reconnect drain is the only blocker from my side — once that lands I'm happy to approve. Thanks again!

ronaldeddings:
Thanks for the catch @pedramamini — got all three!

Pushed `e374d724c`:

- **🔴 BridgeClient pending drain** — `src/web-desktop/electron-shim.ts` close handler now rejects every in-flight `pending` with a `bridge disconnected` error, clears the outbound queue, recreates the `ready` promise, then schedules the reconnect. Pretty much exactly the shape you sketched.
- **🟡 `serveWebDesktopIndex` sync I/O** — added `webDesktopPathCache` and `webDesktopHtmlCache` on the `WebServer` instance. The path resolution and the post-replacement HTML are both memoized on first hit; subsequent desktop loads are a single in-memory send.
- **🟡 `useViewportBreakpoint` flash** — now sets `data-bp` on `<html>` synchronously during hook init (in addition to the useEffect for resize/orientation), so the CSS branches apply before first paint on narrow viewports.

While I was in there, also picked up a few of the AI-reviewer items that held up under verification:
- App.tsx mutual-exclusion was biased toward the left drawer (opening the right while the left was open would immediately close the right) — now tracks prev-values via refs and reacts to the transition.
- Root `build` script was missing `build:web-desktop`, so the bundle wasn't produced in CI/release builds.
- Added an `uninstallWebContentsBridgeHook` so a stale `broadcastSink` doesn't linger after the Encore Feature is toggled off and the server restarts.
- Couple of smaller items: unified the `NewTabPopover` width constant, null-guarded the `encoreFeatures` cast.

Also dropped three files from the diff that snuck in by accident (`PRD-CY-CONTEXT-RUNTIME.md` and two personal import scripts) — they were unrelated to the bundle.

Gates re-ran clean: tsc, ESLint, 27,844 tests, full build with web-desktop. Ready for another look whenever you have a minute.

pedramamini:
@greptile-apps review
@coderabbitat review

chr1syy:
Manual test pass against this PR on a local 0.16.17-RC build (Linux AppImage). Bridge architecture works — `/desktop/` route is correctly gated, `bridge.invoke`/`bridge.event` traffic flows on the WS, and the desktop UI renders inside Chromium without modification. Found four bugs and one UX nit worth flagging before this lands.

### Bug 1 — No UI toggle in Encore Features tab

The flag is wired into `EncoreFeatureFlags`, defaults (`settingsStore.ts`), and all four backend gate enforcement points, but `EncoreTab.tsx` has zero references to `webDesktopBundle` (vs. 8–10 each for every other Encore flag). Users can only enable via:

\`\`\`bash
maestro-cli settings set encoreFeatures.webDesktopBundle true
\`\`\`

…plus a full app restart (the bridge hook installs once at server startup). Likely a one-line follow-up to add the toggle alongside the other Encore feature switches.

### Bug 2 — Messages sent through the bridge don't reach the agent or echo back ⚠️ highest severity

Repro: open \`/desktop/\` in browser, send a prompt to an agent through the web-desktop UI. Expected: message arrives at the agent and appears in both the desktop app and the web-desktop view. Actual: message did not appear in either, and was gone from the web-desktop view on browser refresh. Suggests either (a) the user-input dispatch isn't routing through the bridge, (b) state isn't persisted server-side for the bridge client, or (c) the `WebContents.send` echo isn't broadcasting back to the originating client. This is the one that breaks the core \"send a message from your phone\" use case, so I'd treat it as a P0 before merge.

### Bug 3 — SettingsModal threw an unrecoverable error in the web-desktop view, requiring a full window reload

Repro: open Settings inside the web-desktop browser tab. Expected: settings load and edit normally via bridge. Actual: error printed in the modal, couldn't recover, full reload required to escape. Indicates a bridge handler somewhere in the settings flow is failing — likely a renderer-only API (file system access, native dialogs, Electron \`shell.openExternal\`, etc.) being invoked without a graceful no-op fallback for the browser context. Worth auditing every IPC channel used by Settings for browser-safe equivalents.

### Bug 4 — LIVE button renders inside the `/desktop/` view; user can disconnect themselves

Repro: in the `/desktop/` browser view, click the **LIVE** button at the top of the SessionList. Panel opens with a \"Live Mode\" toggle that can be flipped off, killing the webserver the browser is connected to. Confirmed in code: \`isWebDesktop\` (or any equivalent runtime detection) doesn't exist anywhere in \`src/renderer/\` — the LIVE button is rendered unconditionally. Fix: detect the web-desktop context (e.g. \`window.location.pathname.includes('/desktop/')\` or a build-time \`__WEB_DESKTOP__\` Vite define) and hide/disable destructive surfaces — at minimum the LIVE toggle, probably also Quit/Restart/Update controls and anything else that assumes Electron host.

### UX feedback — split-state routes feel like a roadmap question

With \`webDesktopBundle\` ON, both \`/<token>/\` (the original mobile/web UI) and \`/<token>/desktop/\` (the new bundle) are served simultaneously. They're separate apps with separate state — a session opened on one route doesn't reflect on the other without round-tripping through the desktop app. Personal preference: the web-desktop bundle should *replace* the mobile webserver as the single web entry point (responsive-design'd so the same \`/<token>/\` URL serves desktop on big screens and mobile-shell on small ones), rather than living alongside \`/web/\` indefinitely. Worth deciding the long-term shape before more UX accrues in the legacy `/web/` path.

---

Otherwise the bridge feels great in the browser — once Bug 2 + Bug 3 are fixed, this is going to be a really nice way to control Maestro from anywhere.

ronaldeddings:
Thanks @chr1syy — the manual test pass made all the difference here. The level of detail (repro steps, severity ranking, references to specific code surfaces, screenshots in your head) turned what would've been four debugging sessions into four targeted fixes. Especially appreciate the headline call on Bug 2 — that one was hiding behind a passing CI suite and I would not have found it from code review alone.

Pushed `172c616dc` (with `53ce8611a` for Bug 2 ahead of it). Quick rundown of what each bug needed:

### Bug 2 — messages from `/desktop/` didn't reach the agent (P0)
The bridge fanout was monkey-patching `WebContents.prototype.send` to mirror main→renderer pushes to web clients. That worked *only* when `safeSend` actually reached `webContents.send(...)`. Every push first runs through `safeSend` in `utils/safe-send.ts`, which gates on `mainWindow` existing AND `webContents` being alive — so if the Electron window was closing, destroyed, or mid-launch, `safeSend` short-circuited and the patched prototype never fired. Send worked (it routes through `ipcMain.handle('process:write')` which the bridge does intercept), but the receive path was dead.

Fix: replaced the monkey-patch with an explicit `broadcastBridgeEvent(channel, args)` export from `bridgeHandlers.ts`, called from `safeSend` *unconditionally* — before the `mainWindow` liveness check. Now every IPC push reaches web-desktop regardless of Electron renderer state. Verified end-to-end via Interceptor: sent "ping from web-desktop bridge test — please reply with PONG" from the browser, agent received it, agent's `PONG` streamed back into both the desktop window and the web-desktop view.

### Bug 1 — no UI toggle in EncoreTab
Straightforward gap. Added the toggle to `EncoreTab.tsx` alongside Symphony / Cue / Director's Notes — Globe icon, Beta badge, explanatory body copy, and a yellow warning panel reminding users to restart Maestro after toggling (the bridge installs once at server startup). Also registered the new card in `searchableSettings.ts` so the Settings search bar picks it up under keywords like "web", "desktop", "bridge", "bundle", "tailscale".

### Bug 3 — SettingsModal crash in web-desktop
Could not reproduce on the new code. Walked through every Settings tab (General / Display / Shortcuts / Themes / Notifications / AI Commands / Maestro Prompts / SSH Hosts / Environment / Encore Features), opened the prompt editor, navigated back and forth — all clean, no thrown exceptions in the renderer console. Strong hypothesis: Bug 3 was an indirect symptom of Bug 2. Settings depends on push-event state-sync through `safeSend`, and those pushes were silently dropping in the round-1 code. Once the bridge fanout became unconditional in `broadcastBridgeEvent`, Settings started receiving its initial state correctly. No additional code change needed; recorded the verdict in the commit body so the next reviewer doesn't re-flag it.

### Bug 4 — destructive surfaces visible inside `/desktop/`
Added `src/renderer/utils/runtimeContext.ts` with an `isWebDesktop()` helper that reads the server-injected `__MAESTRO_CONFIG__` (with a `/desktop/` path fallback) and memoizes. Then gated:

- **LIVE / OFFLINE toggle** in `SessionList.tsx` — the one you flagged. Toggling it off would kill the webserver the browser is connected to, which is the exact disconnect you hit.
- **Check for Updates** in `HamburgerMenuContent.tsx` — assumes the Electron binary is the runtime host.
- **Draggable title bar** in `App.tsx` (`WebkitAppRegion: drag`) — no Electron host to drag, just visual clutter. A stale instance had snuck back in since the round-1 cleanup.
- **40px top padding** that was reserved to clear the title bar — without removing it, the bar's absence left a blank strip across the top of the page.

The helper is reusable for any future surface that needs the same treatment (Quit, Restart, "Open in Finder", etc.) — just `if (isWebDesktop()) return null;` at the render path or `!isWebDesktop() && <Component/>` in JSX.

### On the UX nit (split `/web/` vs `/desktop/` routes)
Agree this is the right shape long-term — one URL that picks layout off the viewport instead of two parallel apps. Tracking it as a follow-up rather than expanding this PR's surface. I'd rather get the bridge fundamentals stable first and revisit the route consolidation as a focused piece of work.

### Validation
- `npm run lint` (tsc all configs): clean
- `npm run lint:eslint`: clean
- `npm run test`: **27,847 passed / 108 skipped / 0 failed** (3 new tests around `broadcastBridgeEvent` install / fanout / uninstall)
- `npm run build`: main + preload + renderer + web + web-desktop + cli all built
- Live Interceptor verification in real Chrome at `http://localhost:31999/<token>/desktop/`: send / receive round-trip, Settings navigation, LIVE button absent, Check for Updates absent, drag bar absent, `paddingTop == 0px`

Thanks again — happy to dig into anything that still feels off when you do another pass.

chr1syy:
Two more follow-up findings from the latest manual test pass against PR #991 head (`172c616d`) layered on `upstream/rc` (`b96e9a07e`). Both are in the new `vite.config.web-desktop.mts`. Each blocks a clean build/boot of the web-desktop bundle.

### Bug 5 — `vite.config.web-desktop.mts` missing `cssMinify: 'esbuild'`

Production build fails on the renderer's bundled CSS:

```
✗ Build failed in 8.36s
error during build:
Build failed with 1 error:

[plugin vite:css-post]
SyntaxError: [lightningcss minify] Unexpected token Semicolon
  at style.css:2786:2
  -: \s|;
```

The repro is just `npm run build:web-desktop` after merging this PR — fails at the CSS minify step on what looks like a leftover xterm.js selector. Vite 8 flipped the default CSS minifier to `lightningcss`, which is stricter than esbuild and chokes on this. The fix is already documented in `vite.config.mts` and `vite.config.web.mts` — both pin `cssMinify: 'esbuild'` with a TODO referencing this exact xterm issue. The new web-desktop config just needs the same pin:

```ts
build: {
  // …
  cssMinify: 'esbuild',
  // …
},
```

### Bug 6 — Web-Desktop bootstrap crashes with `TypeError: te is not a function`

After fixing Bug 5 and shipping the build, opening `/desktop/` in a browser surfaces an error overlay:

```
Web-Desktop bootstrap failed
TypeError: te is not a function
    at http://192.168.x.y:PORT/<token>/desktop/assets/src-CtwyxH4l.js:1:56
```

Tracing the minified bundle: `src-CtwyxH4l.js` does `import {r as te} from "./main-BRyRTqAn.js"` to grab Rolldown's `__commonJS` helper, then immediately calls `te(((t,n) => { (function(e,r) {…}).call(t, …); …}))` to wrap dayjs's UMD module. But `main` and `src` form a load cycle (main pulls in src, src reads bindings from main), so the wrapper evaluates before main finishes exporting — the `r`/`te` binding is still TDZ. Hence "te is not a function" at the very first call.

`dayjs` is pulled in transitively by `mermaid` (`npm ls dayjs` → `mermaid → dayjs@1.11.19`). The desktop bundle's Rollup config doesn't hit this because its chunking is different; the web-desktop config inherits Vite's defaults and lands in the cycle.

Fix: add a `manualChunks` callback to coalesce the CJS-wrapped mermaid transitive deps into a `vendor` chunk. Rolldown then auto-emits a tiny standalone `rolldown-runtime-*.js` chunk for the helpers, breaking the cycle. `optimizeDeps.include` doesn't help here — that only applies to `vite dev`, not the production build.

```ts
build: {
  // …
  rollupOptions: {
    input: { main: path.join(__dirname, 'src/web-desktop/index.html') },
    output: {
      manualChunks(id: string): string | undefined {
        if (id.includes('node_modules/dayjs/')) return 'vendor';
        if (id.includes('node_modules/khroma/')) return 'vendor';
        return undefined;
      },
    },
  },
  // …
},
```

Verified the cycle is gone post-patch: the new `src-B3Hj0pCQ.js` imports `import {o as Wn} from "./rolldown-runtime-DcJc2D6O.js"; import {d as Qn} from "./vendor-DTKXQiAc.js"` — no edge back to main, and `/desktop/` now boots cleanly to the desktop UI.

---

Both patches are one-liners in the same file (`vite.config.web-desktop.mts`). Happy to open a tiny follow-up PR with the two fixes if useful.

chr1syy:
One more finding from continued testing on the bug-fix build. Bug 2's safeSend fix is correctly merged and the broadcast path is intact — verified `src/main/utils/safe-send.ts:34` unconditionally calls `broadcastBridgeEvent` before the renderer push, and `broadcastService.broadcastToAll` has no per-client filtering. But there's a downstream rendering issue that produces a Bug-2-shaped symptom.

### Bug 7 — Cross-view messages: user input doesn't fan out and agent responses collapse into one blob

Repro:
1. Open the desktop app and `/desktop/` in the browser.
2. Select the same agent in both views' Left Bar.
3. Send a prompt from the **desktop** view.
4. Look at the **web** view: your user message is missing. The agent's reply is there, but every chunk has collapsed into a single message slot instead of splitting into discrete turns.
5. Swap directions (send from web): same symmetric symptom — desktop view is missing the user message, agent response is one giant blob.

What works:
- The `bridge.event` fanout itself is firing — otherwise the web view would see nothing, not "the wrong thing." Confirmed by toggling Live Mode off+on (which forces the bridge connection to drop, reconnect, and re-fetch session state). After the reconnect, the web view renders the conversation history correctly.
- No errors in either DevTools console.

Most likely root cause:

The renderer splits the streamed `process:data` chunks into discrete agent messages by listening for a *turn-boundary* signal that fires when the user submits input. That signal currently appears to be renderer-local — only the renderer where the user actually typed knows a new turn started. The other renderer (the cross-view client) receives the streamed `process:data` chunks via `safeSend → broadcastBridgeEvent → bridge.event`, but with no turn-start marker, it appends every chunk to the last-open message slot.

This also explains why the *user-typed* message never appears in the cross-view: it isn't a `process:data` event at all, just an optimistic local insert in the originating renderer's state. The cross-view never gets any event for it.

Fix direction (two equally workable options):
- Add a turn-boundary IPC event that main fans out to all renderers when input is received (e.g., `process:input-submitted` carrying `{ sessionId, content, role: 'user' }`). Both views could then create the user-message slot and the next-turn agent-message slot from the same signal.
- Or encode turn boundaries into the `process:data` stream itself — e.g., the first chunk of an agent reply carries a `turnId`/`messageStart` flag the renderer can detect.

The first option is closer to how the desktop renderer already works (everything fans out via main), and avoids re-shaping the data plane.

### Diagnostic that misdirected us first

I initially suspected `WebServer.ts:1264` skipping `installWebContentsBridgeHook` when `webDesktopBundle` is toggled on after Live Mode is already running. That code path is real and worth hardening (re-install/uninstall the hook when the Encore flag changes mid-session), but it isn't what was biting us here — the hook *was* installed. Leaving the observation in case it's useful for hardening separately.

---

Open offer from the earlier comment still stands: happy to bundle Bug 5 (`cssMinify`), Bug 6 (`manualChunks`), and Bug 7 (turn-boundary fanout) into a single follow-up PR if useful.

chr1syy:
One more finding from continued testing on the bug-fix build.

### Bug 8 — Copy-to-Clipboard silently fails on AI and Human messages in `/desktop/` view

Repro: open `/desktop/` in a browser, click the copy button on any AI message or any user message. Expected: text on the clipboard + the usual center-flash ack. Actual: nothing — no clipboard write, no error toast, no flash. Works correctly in the Electron desktop app.

Root cause in `src/renderer/utils/clipboard.ts:16-25`:

```ts
export async function safeClipboardWrite(text: string): Promise<boolean> {
    try {
        await navigator.clipboard.writeText(text);
        return true;
    } catch {
        // NotAllowedError when document not focused, or other clipboard failures.
        return false;
    }
}
```

`safeClipboardWrite()` calls `navigator.clipboard.writeText()` inside a swallow-all `catch {}` that returns `false` with no surfaced failure. In the browser context that call can fail from any of: non-secure context (when `/desktop/` is accessed via LAN IP like `http://192.168.x.y:PORT/...` rather than `localhost`), document-focus loss across the WS round-trip, or missing clipboard permission — and every one of those is hidden from the user.

Two compounding issues:

1. **No IPC fallback for text.** Image copy in the same file (`safeClipboardWriteImage`, line 47) checks `window.maestro?.shell?.copyImageToClipboard` first and only falls back to `navigator.clipboard`. Text copy has no escape hatch.
2. **No handler to mirror.** `src/main/ipc/handlers/system.ts` only registers `clipboard:writeImage`. There is no `clipboard:writeText` handler, so the PR #991 bridge in `src/main/web-server/handlers/bridgeHandlers.ts` has nothing to mirror even if `safeClipboardWrite` wanted to try the IPC path.

Call sites that hit this: `src/renderer/components/GroupChatMessages.tsx` (the chat-message copy buttons) and `src/renderer/components/MarkdownRenderer.tsx` (the code-block copy button) — both go through `copyToClipboard()` → `safeClipboardWrite()`, so the same silent-failure affects every copy surface in the web-desktop view.

Suggested fixes (in order of preference):

1. Add a `clipboard:writeText` IPC handler in `src/main/ipc/handlers/system.ts` + a `window.maestro.shell.copyTextToClipboard` preload bridge. Have `safeClipboardWrite()` try the IPC path first, mirroring the image-copy pattern. The PR #991 bridge then handles text copy cleanly in both Electron and `/desktop/`.
2. At minimum, surface failure: fire a red `notifyToast` (or even a `console.warn`) on the `false` return so users aren't left wondering whether the click registered.

Open offer from the earlier comments still stands: happy to bundle Bug 5 (`cssMinify`), Bug 6 (`manualChunks`), Bug 7 (turn-boundary fanout), and Bug 8 (clipboard IPC fallback) into a single follow-up PR if useful.

chr1syy:
Quick follow-up on Bug 8 — the blast radius is wider than my comment implied.

Right-click → **Copy Link** on a hyperlink in agent chat fails with the same symptom (silent, no clipboard write, no toast). Traced to `src/renderer/components/LinkContextMenu.tsx:47-50`:

```ts
const handleCopy = useCallback(() => {
    safeClipboardWrite(menu.url);
    onDismiss();
}, [menu.url, onDismiss]);
```

Same `safeClipboardWrite()` underneath, so this is one bug, not two — but it prompted me to grep the renderer for every caller. `safeClipboardWrite | copyToClipboard | navigator.clipboard` hits **41 files** under `src/renderer/`. Sampling the non-chat surfaces:

- Right-click context menus: `LinkContextMenu.tsx`, `FileContextMenu.tsx`, `NodeContextMenu.tsx` (DocumentGraph)
- Code-block copy in `MarkdownRenderer.tsx`
- Terminal copy: `TerminalOutput.tsx`, `XTerminal.tsx`
- JSON viewers: `JsonlViewer.tsx`, `CollapsibleJsonViewer.tsx`
- All four tab-bar copies: `AITab.tsx`, `FileTab.tsx`, `TerminalTabItem.tsx`, `BrowserTabItem.tsx`
- Modals: `QuickActionsModal.tsx`, `HistoryDetailModal.tsx`, `GistPublishModal.tsx`, `DebugPackageModal.tsx`
- Misc: `ImageAnnotator.tsx`, `AutoRunLightbox.tsx`, `AIOverviewTab.tsx` (Director's Notes), `ExecutionQueueBrowser.tsx`, `StandingOvationOverlay.tsx`, `LiveOverlayPanel.tsx`, etc.

Effectively every copy action in the renderer is silently broken in `/desktop/` view, not just chat-message copy. The good news: fixing Bug 8's root cause (add `clipboard:writeText` IPC handler + `window.maestro.shell.copyTextToClipboard` preload bridge + IPC-first try in `safeClipboardWrite`) clears all ~40 surfaces at once. Worth bumping the priority a notch on that basis — wanted to flag before you scope the fix.

ronaldeddings:
Merged the latest `rc` into the branch and resolved all conflicts. Prior review feedback is in. Ready for final review.

### Conflicts resolved (against current `rc`)
- `package.json` (kept both `build:web-desktop` and the new `build:maestro-p` steps) and `index.css` (additive mobile CSS).
- `rc` had split `FileExplorerPanel.tsx` and `InputArea.tsx` into module directories, so this PR's touch + responsive logic was **re-homed into the new structure** (`FileExplorerPanel/` + `components/FileTreeRow.tsx`, `InputArea/components/ToolbarControls.tsx`) instead of restoring the old single files.

### Review feedback addressed
- **@pedramamini:** BridgeClient `pending`-drain on WS close (`e374d724c`), plus the `readFileSync` cache and `data-bp` first-paint items.
- **@chr1syy (Bugs 1-8):** Encore UI toggle, Settings + LIVE-button Electron-only guards, cross-view message fanout (turn-boundary event), `cssMinify` / `manualChunks` build fixes, and the clipboard IPC fallback.

### Validation
- `tsc` (all configs), ESLint, full suite (**31,145 passing**), and `npm run build` (incl. `build:web-desktop`) all green.
- Live smoke test of the served `/desktop/` bundle: full desktop UI renders over the WS bridge, send to agent to reply fanout works, and copy-to-clipboard succeeds on a backgrounded tab (the exact case Bug 8 fixed).

### Where to focus
- The conflict re-homing into the split modules (confirm no PR behavior was dropped).
- The bridge hardening and the new `clipboard:writeText` IPC path.

PR #992, opened 2026-05-13
feat: add Kilo agent support

raccoonwannafly:
# feat: Add Kilo agent support

## Summary

Add [Kilo](https://github.com/Kilo-Org/kilocode) as a new agent in Maestro. Kilo is a 1:1 fork of OpenCode with identical CLI interface and JSON output format, differing only in binary name (`kilo` vs `opencode`), data directory, and branding. The integration reuses OpenCode's parser and session storage via subclassing, keeps the wizard and new-agent UI in sync, and includes a follow-up stderr normalization fix so Kilo's compatibility warning does not render as a false error.

## Motivation

Kilo is a actively maintained fork of OpenCode that adds:

- Kilo-specific provider integrations (Kilo API gateway)
- VS Code and JetBrains IDE extensions
- Additional CLI flags (`--auto`, `--variant`, `--thinking`)
- Its own authentication and billing system

Users who have adopted Kilo as their primary coding agent can now use it directly in Maestro without switching back to OpenCode.

### New Files (2)

| File                                       | Lines | Purpose                                                                                                                         |
| ------------------------------------------ | ----- | ------------------------------------------------------------------------------------------------------------------------------- |
| `src/main/parsers/kilo-output-parser.ts`   | 6     | Thin subclass of `OpenCodeOutputParser` with `agentId = 'kilo'`                                                                 |
| `src/main/storage/kilo-session-storage.ts` | 43    | Subclass of `OpenCodeSessionStorage` overriding data dir (`~/.local/share/kilo/`), DB path (`kilo.db`), and remote storage path |

#### Core Registration

| File                                   | Change                                                                                                 |
| -------------------------------------- | ------------------------------------------------------------------------------------------------------ |
| `src/shared/agentIds.ts`               | Added `'kilo'` to `AGENT_IDS` array — the single source of truth for the `ToolType` union              |
| `src/shared/agentMetadata.ts`          | Added `'Kilo'` display name, added to `PLAN_MODE_AGENTS` (uses `--agent plan`), added to `BETA_AGENTS` |
| `src/shared/agentConstants.ts`         | Added `kilo: 128000` default context window                                                            |
| `src/shared/templateVariables.ts`      | Updated `{{TOOL_TYPE}}` doc comment to include `kilo`                                                  |
| `src/shared/pathUtils.ts`              | Added `~/.kilo/bin` (Unix) and `scoop/apps/kilo` (Windows) to expanded PATH                            |
| `src/renderer/constants/agentIcons.ts` | Added `kilo: '⚡'` icon                                                                                |

#### Agent Definition & Capabilities

| File                              | Change                                                                                                                                        |
| --------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------- |
| `src/main/agents/definitions.ts`  | Full agent definition: `run` subcommand, `--format json`, `--session`, `--model`, `--agent plan`, `KILO_CONFIG_CONTENT` env var for YOLO mode |
| `src/main/agents/capabilities.ts` | Complete capability set mirroring OpenCode (resume, read-only, JSON output, session storage, cost tracking, model selection, streaming, etc.) |
| `src/main/agents/detector.ts`     | Added `kilo models` command for runtime model discovery                                                                                       |

#### Parsers & Error Handling

| File                                 | Change                                                                                          |
| ------------------------------------ | ----------------------------------------------------------------------------------------------- |
| `src/main/parsers/index.ts`          | Registered and exported `KiloOutputParser`                                                      |
| `src/main/parsers/error-patterns.ts` | Added `kilo` to binary-not-found regex pattern, registered `OPENCODE_ERROR_PATTERNS` for `kilo` |

#### Storage

| File                                           | Change                                                                                                                                                                                              |
| ---------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `src/main/storage/index.ts`                    | Registered and exported `KiloSessionStorage`                                                                                                                                                        |
| `src/main/storage/opencode-session-storage.ts` | Refactored `getDataDir()`, `getStorageDir()`, `getDbPath()`, `getRemoteStorageDir()` from module-level functions to `protected` methods on the class, enabling KiloSessionStorage to override paths |

#### Process & SSH

| File                                                 | Change                                                                                                                    |
| ---------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------- |
| `src/main/process-manager/handlers/StdoutHandler.ts` | Extended init event `resultEmitted` reset to include `'kilo'` (same streaming behavior as OpenCode)                       |
| `src/main/process-manager/handlers/StderrHandler.ts` | Filters Kilo's non-fatal client/server compatibility warning and re-emits any remaining assistant text as normal output   |
| `src/main/utils/ssh-command-builder.ts`              | Added `$HOME/.kilo/bin` to `BASE_SSH_PATH_DIRS` for remote execution                                                      |
| `src/main/agents/path-prober.ts`                     | Added `kilo` binary detection paths for Windows (scoop, volta, npm, go) and Unix (local, homebrew, npm, version managers) |

#### CLI Spawner

| File                                | Change                                                                                                                  |
| ----------------------------------- | ----------------------------------------------------------------------------------------------------------------------- |
| `src/cli/services/agent-spawner.ts` | Added `detectKilo()`, `getKiloCommand()` wrappers, imported `KiloOutputParser`, added `'kilo'` case in `createParser()` |

#### Renderer Services

| File                                                             | Change                                                                                              |
| ---------------------------------------------------------------- | --------------------------------------------------------------------------------------------------- |
| `src/renderer/services/contextGroomer.ts`                        | Added `kilo` entry to `AGENT_ARTIFACTS` (slash commands, brand references) and `AGENT_TARGET_NOTES` |
| `src/renderer/services/inlineWizardDocumentGeneration.ts`        | Added `'kilo'` to all `opencode` equality checks and switch cases                                   |
| `src/renderer/services/inlineWizardConversation.ts`              | Added `'kilo'` to all `opencode` equality checks and switch cases                                   |
| `src/renderer/components/Wizard/services/conversationManager.ts` | Added `'kilo'` to all `opencode` equality checks and switch cases                                   |

### Why subclass instead of copy?

Kilo's CLI and JSON output are byte-for-byte compatible with OpenCode. Rather than duplicating ~500 lines of parser code and ~1700 lines of session storage code, the integration uses inheritance:

```
OpenCodeOutputParser          OpenCodeSessionStorage
  └── KiloOutputParser (6L)     └── KiloSessionStorage (43L)
```

This means any future bug fixes to the OpenCode parser automatically apply to Kilo. The tradeoff is tight coupling — if Kilo's output format diverges from OpenCode in the future, `KiloOutputParser` can be promoted to a full standalone implementation at that point.

### Why `KILO_CONFIG_CONTENT` instead of `OPENCODE_CONFIG_CONTENT`?

Kilo reads its own config env var. The YOLO mode permissions JSON is identical in structure but must use the correct env var name for the binary to pick it up.

### Why mark as Beta?

Kilo integration has not been through end-to-end testing with all Maestro features (SSH remote, Auto Run, Group Chat wizard, session import/export). Marking as Beta sets appropriate expectations while the integration matures.

### Why normalize Kilo stderr?

The Kilo wrapper currently emits a non-fatal compatibility warning on stderr before the assistant response. Without normalization, Maestro would show the warning in a red stderr block and risk hiding the actual reply. The handler now strips that warning and preserves the remaining assistant text as normal output.

## Kilo vs OpenCode Reference

| Aspect                 | OpenCode                                  | Kilo                                      |
| ---------------------- | ----------------------------------------- | ----------------------------------------- |
| Binary                 | `opencode`                                | `kilo` (alias: `kilocode`)                |
| npm package            | `opencode`                                | `@kilocode/cli`                           |
| Data dir (Linux/macOS) | `~/.local/share/opencode/`                | `~/.local/share/kilo/`                    |
| Data dir (Windows)     | `%APPDATA%\opencode`                      | `%LOCALAPPDATA%\kilo`                     |
| Database               | `opencode.db`                             | `kilo.db`                                 |
| Config env var         | `OPENCODE_CONFIG_CONTENT`                 | `KILO_CONFIG_CONTENT`                     |
| CLI `run` args         | `--format json --session --model --agent` | Identical + `--auto --variant --thinking` |
| JSON output            | `{ type, timestamp, sessionID, part }`    | Identical                                 |

## Testing

- [ ] `kilo` binary detected on PATH (or via known install paths)
- [ ] `kilo run --format json "hello"` produces parseable JSONL
- [ ] Session history loads from `~/.local/share/kilo/`
- [ ] Read-only (plan) mode works via `--agent plan`
- [ ] Model selection works via `--model provider/model`
- [ ] Auto Run batch execution completes
- [ ] SSH remote execution finds `kilo` binary
- [ ] Kilo compatibility warning is filtered from stderr and assistant text still renders
- [ ] Error messages display correctly for common failures
- [ ] Kilo appears in agent picker with ⚡ icon and "(Beta)" badge

## Related

- KiloCode repo: https://github.com/Kilo-Org/kilocode
- OpenCode repo: https://github.com/opencode-ai/opencode
- Existing OpenCode integration: `src/main/parsers/opencode-output-parser.ts`, `src/main/storage/opencode-session-storage.ts`

## Release Notes

* **New Features**
  * Kilo agent is now fully integrated and available in the agent selection wizard.
  * Automatic Kilo detection across Windows and Unix platforms.
  * Kilo-specific session storage and plan-mode read-only support.
  * Improved stderr handling with Kilo compatibility.

* **Tests**
  * Added comprehensive test coverage for Kilo integration and compatibility scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RunMaestro/Maestro/pull/992)

### DIFF
--- a/PULL_KILO.md
+++ b/PULL_KILO.md
@@ -1,0 +1,141 @@
+# feat: Add Kilo agent support
+
+## Summary
+
+Add [Kilo](https://github.com/Kilo-Org/kilocode) as a new agent in Maestro. Kilo is a 1:1 fork of OpenCode with identical CLI interface and JSON output format, differing only in binary name (`kilo` vs `opencode`), data directory, and branding. The integration reuses OpenCode's parser and session storage via subclassing, keeps the wizard and new-agent UI in sync, and includes a follow-up stderr normalization fix so Kilo's compatibility warning does not render as a false error.
+
+## Motivation
+
+Kilo is a actively maintained fork of OpenCode that adds:
+
+- Kilo-specific provider integrations (Kilo API gateway)
+- VS Code and JetBrains IDE extensions
+- Additional CLI flags (`--auto`, `--variant`, `--thinking`)
+- Its own authentication and billing system
+
+Users who have adopted Kilo as their primary coding agent can now use it directly in Maestro without switching back to OpenCode.
+
+## Changes
+
+### New Files (2)
+
+| File                                       | Lines | Purpose                                                                                                                         |
+| ------------------------------------------ | ----- | ------------------------------------------------------------------------------------------------------------------------------- |
+| `src/main/parsers/kilo-output-parser.ts`   | 6     | Thin subclass of `OpenCodeOutputParser` with `agentId = 'kilo'`                                                                 |
+| `src/main/storage/kilo-session-storage.ts` | 43    | Subclass of `OpenCodeSessionStorage` overriding data dir (`~/.local/share/kilo/`), DB path (`kilo.db`), and remote storage path |
+
+### Modified Files (21)
+
+#### Core Registration
+
+| File                                   | Change                                                                                                 |
+| -------------------------------------- | ------------------------------------------------------------------------------------------------------ |
+| `src/shared/agentIds.ts`               | Added `'kilo'` to `AGENT_IDS` array — the single source of truth for the `ToolType` union              |
+| `src/shared/agentMetadata.ts`          | Added `'Kilo'` display name, added to `PLAN_MODE_AGENTS` (uses `--agent plan`), added to `BETA_AGENTS` |
+| `src/shared/agentConstants.ts`         | Added `kilo: 128000` default context window                                                            |
+| `src/shared/templateVariables.ts`      | Updated `{{TOOL_TYPE}}` doc comment to include `kilo`                                                  |
+| `src/shared/pathUtils.ts`              | Added `~/.kilo/bin` (Unix) and `scoop/apps/kilo` (Windows) to expanded PATH                            |
+| `src/renderer/constants/agentIcons.ts` | Added `kilo: '⚡'` icon                                                                                |
+
+#### Agent Definition & Capabilities
+
+| File                              | Change                                                                                                                                        |
+| --------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------- |
+| `src/main/agents/definitions.ts`  | Full agent definition: `run` subcommand, `--format json`, `--session`, `--model`, `--agent plan`, `KILO_CONFIG_CONTENT` env var for YOLO mode |
+| `src/main/agents/capabilities.ts` | Complete capability set mirroring OpenCode (resume, read-only, JSON output, session storage, cost tracking, model selection, streaming, etc.) |
+| `src/main/agents/detector.ts`     | Added `kilo models` command for runtime model discovery                                                                                       |
+
+#### Parsers & Error Handling
+
+| File                                 | Change                                                                                          |
+| ------------------------------------ | ----------------------------------------------------------------------------------------------- |
+| `src/main/parsers/index.ts`          | Registered and exported `KiloOutputParser`                                                      |
+| `src/main/parsers/error-patterns.ts` | Added `kilo` to binary-not-found regex pattern, registered `OPENCODE_ERROR_PATTERNS` for `kilo` |
+
+#### Storage
+
+| File                                           | Change                                                                                                                                                                                              |
+| ---------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `src/main/storage/index.ts`                    | Registered and exported `KiloSessionStorage`                                                                                                                                                        |
+| `src/main/storage/opencode-session-storage.ts` | Refactored `getDataDir()`, `getStorageDir()`, `getDbPath()`, `getRemoteStorageDir()` from module-level functions to `protected` methods on the class, enabling KiloSessionStorage to override paths |
+
+#### Process & SSH
+
+| File                                                 | Change                                                                                                                    |
+| ---------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------- |
+| `src/main/process-manager/handlers/StdoutHandler.ts` | Extended init event `resultEmitted` reset to include `'kilo'` (same streaming behavior as OpenCode)                       |
+| `src/main/process-manager/handlers/StderrHandler.ts` | Filters Kilo's non-fatal client/server compatibility warning and re-emits any remaining assistant text as normal output   |
+| `src/main/utils/ssh-command-builder.ts`              | Added `$HOME/.kilo/bin` to `BASE_SSH_PATH_DIRS` for remote execution                                                      |
+| `src/main/agents/path-prober.ts`                     | Added `kilo` binary detection paths for Windows (scoop, volta, npm, go) and Unix (local, homebrew, npm, version managers) |
+
+#### CLI Spawner
+
+| File                                | Change                                                                                                                  |
+| ----------------------------------- | ----------------------------------------------------------------------------------------------------------------------- |
+| `src/cli/services/agent-spawner.ts` | Added `detectKilo()`, `getKiloCommand()` wrappers, imported `KiloOutputParser`, added `'kilo'` case in `createParser()` |
+
+#### Renderer Services
+
+| File                                                             | Change                                                                                              |
+| ---------------------------------------------------------------- | --------------------------------------------------------------------------------------------------- |
+| `src/renderer/services/contextGroomer.ts`                        | Added `kilo` entry to `AGENT_ARTIFACTS` (slash commands, brand references) and `AGENT_TARGET_NOTES` |
+| `src/renderer/services/inlineWizardDocumentGeneration.ts`        | Added `'kilo'` to all `opencode` equality checks and switch cases                                   |
+| `src/renderer/services/inlineWizardConversation.ts`              | Added `'kilo'` to all `opencode` equality checks and switch cases                                   |
+| `src/renderer/components/Wizard/services/conversationManager.ts` | Added `'kilo'` to all `opencode` equality checks and switch cases                                   |
+
+## Design Decisions
+
+### Why subclass instead of copy?
+
+Kilo's CLI and JSON output are byte-for-byte compatible with OpenCode. Rather than duplicating ~500 lines of parser code and ~1700 lines of session storage code, the integration uses inheritance:
+
+```
+OpenCodeOutputParser          OpenCodeSessionStorage
+  └── KiloOutputParser (6L)     └── KiloSessionStorage (43L)
+```
+
+This means any future bug fixes to the OpenCode parser automatically apply to Kilo. The tradeoff is tight coupling — if Kilo's output format diverges from OpenCode in the future, `KiloOutputParser` can be promoted to a full standalone implementation at that point.
+
+### Why `KILO_CONFIG_CONTENT` instead of `OPENCODE_CONFIG_CONTENT`?
+
+Kilo reads its own config env var. The YOLO mode permissions JSON is identical in structure but must use the correct env var name for the binary to pick it up.
+
+### Why mark as Beta?
+
+Kilo integration has not been through end-to-end testing with all Maestro features (SSH remote, Auto Run, Group Chat wizard, session import/export). Marking as Beta sets appropriate expectations while the integration matures.
+
+### Why normalize Kilo stderr?
+
+The Kilo wrapper currently emits a non-fatal compatibility warning on stderr before the assistant response. Without normalization, Maestro would show the warning in a red stderr block and risk hiding the actual reply. The handler now strips that warning and preserves the remaining assistant text as normal output.
+
+## Kilo vs OpenCode Reference
+
+| Aspect                 | OpenCode                                  | Kilo                                      |
+| ---------------------- | ----------------------------------------- | ----------------------------------------- |
+| Binary                 | `opencode`                                | `kilo` (alias: `kilocode`)                |
+| npm package            | `opencode`                                | `@kilocode/cli`                           |
+| Data dir (Linux/macOS) | `~/.local/share/opencode/`                | `~/.local/share/kilo/`                    |
+| Data dir (Windows)     | `%APPDATA%\opencode`                      | `%LOCALAPPDATA%\kilo`                     |
+| Database               | `opencode.db`                             | `kilo.db`                                 |
+| Config env var         | `OPENCODE_CONFIG_CONTENT`                 | `KILO_CONFIG_CONTENT`                     |
+| CLI `run` args         | `--format json --session --model --agent` | Identical + `--auto --variant --thinking` |
+| JSON output            | `{ type, timestamp, sessionID, part }`    | Identical                                 |
+
+## Testing
+
+- [ ] `kilo` binary detected on PATH (or via known install paths)
+- [ ] `kilo run --format json "hello"` produces parseable JSONL
+- [ ] Session history loads from `~/.local/share/kilo/`
+- [ ] Read-only (plan) mode works via `--agent plan`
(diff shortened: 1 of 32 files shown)


pedramamini:
Hi @raccoonwannafly — thanks for the contribution! 🙏 The subclassing approach for the Kilo parser/storage is clean, and the wizard/SSH/path-prober coverage is thorough. A few items I'd like addressed before this merges into `rc`:

### 1. `supportsSlashCommands` inconsistency (`src/main/agents/capabilities.ts`)
The Kilo capability block sets `supportsSlashCommands: false`, but OpenCode (which Kilo is a 1:1 fork of) has it as `true`, and your own `contextGroomer.ts` change registers `/help`, `/clear`, `/cost`, `/model` artifacts for `kilo`. Please flip this to `true` so the capability matches reality.

Also, the Kilo block is missing the trailing `supportsAppendSystemPrompt: false` and `supportsProjectMemory: false` flags that OpenCode declares — please add them for parity.

### 2. `KILO_CONFIG_CONTENT` in `readOnlyEnvOverrides` (`src/main/agents/definitions.ts`)
Your `defaultEnvVars.KILO_CONFIG_CONTENT` keeps the blanket permission grants, but `readOnlyEnvOverrides.KILO_CONFIG_CONTENT` strips them:

```diff
- '{"permission":{"question":"deny"},"tools":{"question":false}}'
+ '{"permission":{"*":"allow","external_directory":"allow","question":"deny"},"tools":{"question":false}}'
```

OpenCode keeps the blanket permissions in both env values on purpose — read-only is enforced via `--agent plan` (`readOnlyArgs`), and stripping the grants here causes Kilo to prompt on stdin and hang in batch / plan mode. Please mirror OpenCode's value with the leading comment explaining why.

### 3. `PULL_KILO.md` committed at repo root
This looks like the PR description checked into source. PR descriptions belong on the PR itself, not in the tree — could you delete `PULL_KILO.md` from the branch?

### 4. Unrelated `package.json` change
The `electron-rebuild` → `@electron/rebuild` swap isn't related to Kilo support. Please revert it from this PR; if the migration is needed, it should land as its own change so it can be reviewed and released on its own merits.

### 5. `StderrHandler` mixed-chunk reclassification (`src/main/process-manager/handlers/StderrHandler.ts`)
The new `extractNonErrorContent` path returns whatever lines weren't suppressed and re-emits them as `data` (assistant output) as long as at least one line was suppressible. If a stderr chunk contains the Kilo compatibility warning **and** a genuine error on a subsequent line, that error will now be silently re-routed into the assistant transcript instead of surfacing as stderr. Please tighten this so we only re-emit as data when *every* original line was suppressible (e.g., return both `normalizedOutput` and an `onlySuppressible` flag, or check that `filteredLines.every(isSuppressible)` before emitting). Same guard belongs on both call sites.

Once those are in, I'll happily approve. Thanks again for the careful work and the detailed reference table — the Kilo vs OpenCode breakdown is super useful context. 🚀

raccoonwannafly:
@pedramamini thanks for the feedback can u check https://github.com/RunMaestro/Maestro/pull/992/changes/87da28611de87648550e1b0be81b97c8dd20e9d5
see if everything is in order

- `capabilities.ts`: fix `Kilo.supportsSlashCommands = true`, add missing
  `supportsAppendSystemPrompt` and `supportsProjectMemory` flags, correct comment
- `definitions.ts`: restore blanket permission grants in Kilo
  `readOnlyEnvOverrides` (matches OpenCode behavior)
- `StderrHandler.ts`: tighten `extractNonErrorContent` to only re-emit
  when ALL lines are suppressible (prevents genuine errors being
  silently routed to assistant transcript)
- `package.json`: revert unrelated changes

jSydorowicz21:
@greptile @coderabbit please re-review this pr

pedramamini:
Hi @raccoonwannafly, thanks again for this contribution and for the careful follow-up on the earlier round of feedback! 🙏 I went back through the current state of the branch and can confirm you've addressed most of the items:

- ✅ `capabilities.ts`: `supportsSlashCommands` is now `true`, and the `supportsAppendSystemPrompt` / `supportsProjectMemory` parity flags are present.
- ✅ `definitions.ts`: the blanket permission grants are restored in `readOnlyEnvOverrides.KILO_CONFIG_CONTENT`, with a clear comment explaining why.
- ✅ `StderrHandler.ts`: `extractNonErrorContent` now only re-emits as assistant data when every line is suppressible, so a genuine error mixed into a chunk correctly falls through to stderr. Nicely done.
- ✅ `package.json`: the unrelated `@electron/rebuild` change is reverted.

A few things still need attention before this can merge into `rc`:

**1. Merge conflicts.** GitHub is currently reporting this branch as conflicting with `rc`. Could you please rebase onto the latest `rc` and resolve the conflicts? A clean rebase will also let Greptile and CodeRabbit re-run against an up-to-date diff (the last re-review request didn't land a clean verdict).

**2. `PULL_KILO.md` is still committed.** This was item 3 from the earlier review and looks like it slipped through. Please delete it from the branch - PR descriptions belong on the PR itself, not in the source tree.

**3. Regression: stderr suppression for the other JSONL agents (Greptile P1).** The refactor that generalized the codex stderr handling into `extractNonErrorContent` also removed the old `if (outputParser && toolType !== 'codex') { return; }` guard in `StderrHandler.ts`. That guard suppressed initialization noise (MCP server startup messages, shell profile banners) for `opencode`, `copilot-cli`, and `factory-droid`. With it gone, those three agents now fall through to `emit('stderr', ...)`, so that init noise will start surfacing for them even though this PR is only meant to add Kilo. Please either restore the suppression for those agents or confirm the behavior change is intended.

**4. Duplicate PATH entry (Greptile P2).** `src/main/utils/ssh-command-builder.ts` lists `'$HOME/.claude/local'` twice (lines 24 and 26), which looks like a copy-paste slip when adding `'$HOME/.kilo/bin'`. Please drop the duplicate.

Minor / optional: CodeRabbit noted that in `phaseGenerator.ts` the `--input-format` branch (around line 1198) is unreachable because `getGenerationStdinFlags()` always returns `sendPromptViaStdin: false`. Not a blocker, but worth a quick look while you're in there.

Once the rebase is clean and the above are addressed, I'll be happy to take another pass and approve. Thanks again for the thorough work here - the OpenCode subclassing approach and the Kilo/OpenCode reference table make this really easy to follow. 🚀

pedramamini:
@raccoonwannafly Kilo agent support is conflicting with `rc`. Could you rebase onto the latest `rc` and re-push? Heads up: the agent registry has moved (agentIds/agentMetadata/definitions/capabilities), so expect some churn there when you resolve.

pedramamini:
Hi @raccoonwannafly - thanks again for this contribution, and for the care you've put into it across the earlier rounds. 🙏 The OpenCode subclassing approach for the parser and session storage is genuinely clean, and the Kilo/OpenCode reference table in the description made this easy to follow. I know it's been a while since the last round, so thank you for your patience.

The branch hasn't been pushed since 2026-05-13, so the items from my July 4th and July 10th notes are still open. Here's the current state after re-checking against today's `rc`.

### 1. Merge conflicts - rebase needed

The branch is now **2157 commits** behind `rc` and conflicts in 7 files:

```
src/main/agents/detector.ts
src/main/parsers/error-patterns.ts
src/main/storage/opencode-session-storage.ts
src/renderer/constants/agentIcons.ts
src/renderer/services/inlineWizardConversation.ts
src/__tests__/renderer/components/NewInstanceModal.test.tsx
src/renderer/components/Wizard/screens/AgentSelectionScreen.tsx   (modify/delete)
```

Could you please rebase onto the latest `rc`? Two heads-ups that will save you time:

- **`AgentSelectionScreen.tsx` is now a directory.** It was split into `AgentSelectionScreen/` with `components/`, `hooks/`, and `utils/`. Your single-file edit will come through as a modify/delete conflict.
- **The provider pickers are now derived from one registry.** `AGENT_PICKER_META` in `src/shared/agentMetadata.ts` is the single source all three pickers read (New Agent modal, wizard tile strip, Group Chat moderator dropdown), and `PICKABLE_AGENT_IDS` sorts it alphabetically. It's typed `Record<AgentId, AgentPickerMeta | null>`, so once `'kilo'` is in `AGENT_IDS` it won't compile until you add an entry. That should let you drop the manual tile edits in `AgentSelectionScreen.tsx` and `NewInstanceModal.test.tsx` entirely. You'll still need a `case` in `AgentLogo` and a glyph in `AGENT_ICONS` or the tile draws a blank fallback ring.

### 2. Two tests fail on this branch (new finding, and the one I'd most like fixed)

This is the blocker I'd flag hardest. `src/__tests__/main/process-manager/handlers/StderrHandler.test.ts` already exists on `rc` and passes 13/13 there. On this branch it's **2 failed / 16**:

```
× Codex tracing line filtering > should filter Rust tracing lines and emit remaining content as data
× Codex tracing line filtering > should handle mixed tracing lines, stdin prefix, and content
```

The cause is the `allLinesSuppressible` guard in `extractNonErrorContent`. I asked for that guard in the first round to stop a genuine error being re-routed into the transcript, so this is partly on my original wording - but applying it to `codex` as well as `kilo` over-corrects. Codex writes tracing diagnostics *and* the actual response into the same chunk, which is precisely why that branch exists. With the guard, any chunk mixing a tracing line with real response content now returns `null` and falls through to `emit('stderr', ...)`, so the user sees their Codex answer rendered as an error.

The guard is right for Kilo (its warning is standalone). For Codex, the old behavior of stripping known-noise lines and emitting the remainder should stay. Splitting the policy per agent rather than sharing one flag would resolve it.

### 3. Codex `isStreamJsonMode` suppression was dropped

Related to the above and also new. The old Codex branch had:

```ts
if (managedProcess.isStreamJsonMode) {
  // structured data comes from stdout via CodexOutputParser; suppress stderr echo
} else {
  this.emitter.emit('data', sessionId, remainingContent);
}
```

`isStreamJsonMode` no longer appears anywhere in the file. In stream-json mode the parser already produces this content from stdout, so re-emitting it risks duplicate raw human-readable text in the terminal. Please carry that check across.

### 4. Stderr suppression regression for the other JSONL agents

Still open from July 4th. Removing `if (outputParser && toolType !== 'codex') { return; }` means `opencode`, `copilot-cli`, and `factory-droid` now fall through to `emit('stderr', ...)`, so their MCP startup messages and shell profile banners will start surfacing. That's a behavior change for three agents in a PR that only means to add a fourth. Please restore the suppression, or confirm it's intended.

### 5. `PULL_KILO.md` is still committed

Third time asking on this one, sorry. Please delete it from the branch - the content is already the PR description, which is where it belongs.

### 6. Duplicate PATH entry

`src/main/utils/ssh-command-builder.ts` still has `'$HOME/.claude/local'` twice:

```ts
'$HOME/.opencode/bin',
'$HOME/.claude/local',
'$HOME/.kilo/bin',
'$HOME/.claude/local',   // <- duplicate
```

---

Once the rebase is clean and the `StderrHandler` behavior is sorted, please run `npm run test`, `npm run lint`, and `npm run lint:eslint` before pushing - that suite would have caught items 2 and 3. Note also that a green local run only covers one OS; both the `test (ubuntu-latest)` and `test (windows-latest)` CI legs need to be green before this can merge.

Happy to take another pass as soon as it's up, and thanks again for sticking with this one. 🚀

PR #994, opened 2026-05-14
fix(file-preview): harden search across tiers + soft-wrap long lines

reachrazamair:
## Summary

Hardening pass on the tiered FilePreview (Rich / Fast / Giant) covering:
- **Search wobble** — the match count would change as the user pressed Next/Prev. Root cause: a single effect did both counting and navigation with `currentMatchIndex` in its deps, so every nav re-ran `findHits`. Now split into a count effect (per query/content/adapter) and a navigate effect (per index), so `findHits` runs **exactly once per query**.
- **"Doesn't land on the searched word"** — Fast tier markdown and text only scrolled to the matched block/page, ignoring the within-block offset. Added a pure shared DOM helper (`buildRangeAtOffset` + `scrollRangeIntoView`) and threaded `offsetWithinBlock` through the `SearchHit` contract so every tier nudges the matched word into view post-virtualizer-scroll.
- **Giant tier showed CodeMirror's native search panel** instead of the app's bar. Replaced it with a pure search engine over the CM6 doc (`findAllInDoc`), `scrollToMatch` dispatches `EditorSelection.single` + `EditorView.scrollIntoView`, and `Cmd+F` now routes through the same shared bar in all three tiers. Removed `search()` + `searchKeymap` from CM6 extensions; deleted `searchBridge.ts`.
- **`edge-one-huge-line.txt` (500 KB single line) froze the renderer** even on Giant tier. CM6's `lineWrapping` (CSS `white-space: pre-wrap`) only breaks at whitespace, so a `AAAA…` line stayed as one multi-million-pixel-wide DOM element and Chromium's wide-layer paint pegged the main thread. Two-part fix:
  1. CSS: pair `lineWrapping` with `overflow-wrap: anywhere` so the browser can break at any character.
  2. Pre-mount soft-wrap (`softWrapLongLines`): insert hard newlines into the content before handing it to CM6 when any line exceeds 1000 chars. Search still operates on the **original** content via an `insertionsAt` offset map, so straddling matches are still found and `scrollToMatch` lands at the correct wrapped offset.

Bundled audit close-outs:
- Lifted the duplicated markdownFast / textFast `codeHighlighter` modules onto a shared `lazyShikiObserver` factory.
- Added a `fileLinks/` barrel export and documented where the (uncommitted) large preview fixtures live.

Also fixed two unrelated-but-surfaced bugs:
- A same-render race in `useFilePreviewSearch` where the navigate effect could read a stale `currentMatchIndex` (past the new shorter hits array) and crash with `Cannot read properties of undefined (reading 'blockIndex')`.
- A React key warning in `JsonlViewer`: `sortedLines.map(...)` returned `<>...</>` with the key on the inner `<tr>` — fragment shorthand has no key slot. Switched to `<Fragment key={...}>`.

## Architecture (SRP / SOC)

| Concern | Owner |
|---|---|
| Adapter contract + hit shape | `FilePreview/search/types.ts` |
| DOM walk + Range-at-offset (shared) | `FilePreview/search/scrollToOffset.ts` |
| Lazy Shiki (shared) | `FilePreview/shared/lazyShikiObserver.ts` |
| Fast markdown hit list | `markdownFast/searchHits.ts` |
| Fast text hit list | `textFast/searchHits.ts` |
| Giant hit list | `giantPreview/searchEngine.ts` |
| Giant pre-mount soft-wrap + offset map | `giantPreview/softWrap.ts` |
| Counting vs navigation effects | `useFilePreviewSearch.ts` (three-effect split) |
| Cmd+F dispatch (one path for all tiers) | `FilePreview.tsx` |

Every new module has its own test file. Tier-specific helpers stay tier-local; shared primitives live under `search/` or `shared/`.

## Commits

```
4b011e387 fix(jsonl-viewer): move row key from <tr> to wrapping Fragment
6848c4613 fix(giantPreview): soft-wrap pathologically long lines before CM6 mount
c80c35544 fix(giantPreview): break no-whitespace long lines to prevent freeze
4981392af fix(file-preview): guard navigate effect against stale currentMatchIndex
3796d239b docs(fixtures): document preview-test-files location
9488a1338 chore(file-preview): add fileLinks barrel export + fix Giant containerRef type
1eba95d28 refactor(file-preview): extract shared lazy-Shiki observer
6aa318f3d feat(file-preview): active-match highlight parity across tiers
2837231ad feat(giantPreview): headless search; route through app bar
1f7c0a89f feat(textFast): scrollToMatch lands on word, not page
ffddccd1e feat(markdownFast): scrollToMatch lands on word, not block
887246d54 refactor(useFilePreviewSearch): split counting from navigation
4af23d0e8 refactor(file-preview): shared search contract module
```

## Test plan
- `npx tsc --noEmit` — clean
- `npx eslint src/renderer/components/FilePreview src/renderer/hooks/file src/renderer/utils/fileLinks` — clean
- `npm run test` — 28,372 passing (+ ~70 new search/Giant/softWrap/scrollToOffset/hook tests), 0 failing

## Out of scope
- Regex / case-sensitive / whole-word toggles in the shared bar (additive feature, not hardening).
- Replace functionality.
- KaTeX / math rendering in Fast tier.
- CM6-everywhere unification of Rich tier (different scope).

## Summary by CodeRabbit

* **New Features**
  * Unified in-app search across all preview modes with precise "scroll to match" that jumps to exact text locations.
  * Lazy code-block highlighting that activates as code enters view.

* **Bug Fixes**
  * Prevents renderer freezes on extremely long lines via display-only soft-wrapping.
  * More accurate match highlighting and stable navigation in paginated/virtualized previews.

* **Tests**
  * Expanded coverage for search, scrolling precision, soft-wrapping, and highlighting.

* **Documentation**
  * Added test-fixture README describing asset layout and regeneration.


pedramamini:
@reachrazamair Thanks for the thorough hardening pass — splitting the count/navigate effects, threading `offsetWithinBlock` through the adapter contract, and the pre-mount soft-wrap with `insertionsAt` offset map is solid work, and the test coverage is great. Before approval though, a few substantive findings (mostly surfaced by the bots) that I'd like addressed:

**Real defect — current-match highlight doesn't track the navigated hit (Greptile P1)**
`src/renderer/hooks/file/useFilePreviewSearch.ts:440` — in the Fast/Giant navigate effect, after `scrollToMatch(hits[safeIdx])` you call `applyCurrentHighlight(ranges[0] ?? null)`. Because Virtuoso renders a window of surrounding blocks, `ranges[0]` is the first DOM match in the visible window, not the match at `safeIdx`. With a common short query (e.g. `"the"`) navigating to match 20 will scroll correctly but visibly highlight match 14 (or whatever the first rendered block's hit happens to be). The Rich tier path right above uses `ranges[Math.min(currentMatchIndex, ranges.length - 1)]` and gets this right; the adapter path should resolve the equivalent range from `hits[safeIdx]` (or fall back to `ranges[safeIdx % ranges.length]` as an approximation). The inline comment about \"B2/B3's job\" punts on a regression that ships today.

**Race / correctness — `findInContent` gated on editor mount state**
`src/renderer/components/FilePreview/giantPreview/GiantPreview.tsx:108-115` — `findInContent` is a pure function over `content`, but it returns `[]` when `viewRef.current` is null. That can race with mount and produce a transient zero-count. Drop the guard.

**Defensive — `softWrapLongLines` infinite loop on bad input**
`src/renderer/components/FilePreview/giantPreview/softWrap.ts:63` — `p += maxLineLength` is non-terminating if `maxLineLength <= 0`. Add a hard guard at the top throwing `RangeError`.

**Bug in fallback clamp**
`src/renderer/components/FilePreview/search/scrollToOffset.ts:67` — when the multi-node walk runs out of text, `range.setEnd(textNode, nodeLen)` clamps to the start node instead of the last traversed node. Track `lastTextNode` and use it.

**Adapter hits dropped on no-Highlight-API path**
`src/renderer/hooks/file/useFilePreviewSearch.ts:351-360` — the early return falls back to `fileContent.match(...)` and leaves `hitsRef = null`, so Fast/Giant tiers lose tier-specific hit mapping and can't reach offscreen matches in virtualized content. Compute `adapterHits` first and prefer it; only skip the highlight registration in that branch.

**Cleanup / quick wins**
- `src/__tests__/fixtures/README.md:12,19` — leftover hardcoded `/Users/saifccript/...` paths; replace with a placeholder like `<workspace>/preview-test-files/`.
- `src/__tests__/renderer/components/FilePreview/search/scrollToOffset.test.ts:119` — test title says \"returns true\" but asserts `false`; rename to match.
- `src/__tests__/renderer/components/FilePreview/shared/lazyShikiObserver.test.ts:141-153` — the alias-resolution test asserts on `SHIKI:x` which doesn't actually verify alias resolution (would pass for `ts` or `typescript`); assert on the emitted `language-typescript` class instead.
- `src/__tests__/renderer/components/FilePreview/giantPreview/softWrap.test.ts:75` — the strict `< 1000ms` timing assertion will be flaky on CI; either drop it or gate via `process.env.CI`.

The first one is the only blocker I'd call out as user-visible regression; the rest are correctness/quality nits that should land in the same pass while you're in here. Once the navigate-effect highlight is fixed and the `findInContent` race is removed, happy to re-review.

PR #1000, opened 2026-05-15
fix(chat): preserve single newlines on chat surfaces (#622)

chr1syy:
## Summary

Continues the work shipped in PR #996 (for #775) by addressing the related sub-problem in issue #622: **ReactMarkdown collapses single `\n` into a space, which visually flattens multi-line assistant messages on chat surfaces.**

Issue #622 lists three sub-problems on chat surfaces (AI Terminal, group chat, mobile message history). This PR fixes sub-problem 1 (newline flattening). Sub-problems 2 (display math `\$\$...\$\$`) and 3 (raw/plain toggle for user messages) are out of scope — math needs a new dependency stack (KaTeX) and the toggle change needs UX decisions.

## Changes

- Add an opt-in `chatLineBreaks` prop on `MarkdownRenderer` that wires `remark-breaks` into the plugin stack
- Enable it on chat surfaces only:
  - `TerminalOutput` — AI assistant messages (all three render paths: collapsed, expanded, default), thinking/streaming, error log entries
  - `GroupChatMessages` — assistant messages (collapsed, expanded, default)
  - `HistoryDetailModal` — historical AI responses
  - `MobileMarkdownRenderer` — applied unconditionally (it is only used by mobile chat surfaces via `WebReadingContent` → `MessageHistory` + `ResponseViewer`)
- **Intentionally left untouched** (document/file-preview surfaces, where CommonMark paragraph reflow is correct): `DocumentGraphView`, `DirectorNotes/AIOverviewTab`, and the `FilePreview` render path
- Add `remark-breaks` as a dependency
- 3 new tests on `MarkdownRenderer`: default collapses single newlines, `chatLineBreaks` emits `<br>`, blank-line paragraph breaks still produce two `<p>` elements either way

## Test plan

- [ ] Multi-line AI assistant message in AI Terminal renders with line breaks preserved (not flattened to one paragraph)
- [ ] Multi-line message in group chat renders with line breaks preserved
- [ ] Multi-line response in History detail modal renders with line breaks preserved
- [ ] Mobile message history / mobile response viewer preserves line breaks
- [ ] Markdown file preview (`.md` files in workspace) still reflows paragraphs the standard CommonMark way — single `\n` collapses to a space
- [ ] Code blocks, lists, tables, links, file links all still work on chat surfaces
- [x] `MarkdownRenderer` tests pass (13/13 including 3 new)
- [x] `TerminalOutput` + mobile chat tests pass (201/201)
[...]

Issue #622, opened 2026-03-26
Chat surfaces should use rich-text-style message rendering


isCopyman:
## Summary

Chat/message surfaces in Maestro currently behave like Markdown document previews rather than rich-text-style chat messages.

This causes message structure and display math to render incorrectly in normal chat use cases, even though Markdown file preview behaves correctly as a document surface.

## Current behavior

On chat surfaces such as AI Terminal, group chat, and mobile message history:

- Single newlines are merged too aggressively, so multi-line messages can be visually flattened into a single paragraph.
- Messages containing isolated display math such as `$$x+y$$` on its own line can render as visually cramped/left-aligned content instead of a clear centered block formula.
- User messages can be rendered with Markdown formatting in some chat surfaces, but the plain/raw text toggle is only available for assistant output, which is inconsistent.

At the same time, Markdown file preview currently looks correct and should keep standard Markdown document semantics.

## Why this is a problem

Chat is not the same as Markdown document preview.

For chat/message surfaces, users expect:

- line structure to be preserved
- display math to look like display math
- rich-text-style rendering by default
- raw/plain-text inspection to be available consistently when formatting is applied

Today the current behavior makes formula-heavy or structured chat content harder to read than it should be.

## Suggested direction

Split rendering behavior by surface:

- **Chat/message surfaces** should use a chat-oriented rich-text Markdown preset
- **Document/file preview surfaces** should remain standard Markdown

For chat surfaces specifically:

- Preserve single newlines as visible line breaks
- Treat line-isolated `$$...$$` as display math and render it as centered block content
- Keep Markdown enhancements such as code blocks, links, lists, tables, and math
- Expose plain/raw text viewing consistently for both user and assistant messages

For document surfaces:

- Keep the current standard Markdown behavior for `.md` preview/editor views

## Notes

- This seems closer to the behavior users expect from GitHub comments / AI chat UIs than the current document-style Markdown behavior.
- This issue is about **chat rendering semantics**, not about rewriting the entire Markdown system.

isCopyman:
The style of Google AIStudio or GitHub could be considered.

chr1syy:
A fix has been committed in #1000 , #1001 , #1002 and following commits - feel free to open a new issue if this is still an issue.

PR #1002, opened 2026-05-15
feat(chat): consistent markdown rendering + raw toggle for user messages (#622)

chr1syy:
## Summary

Stacks on top of PRs #1000 and #1001 to address **sub-problem 3 of #622**: chat surfaces treated user vs assistant messages inconsistently, and the plain/raw markdown toggle was only available for assistant output.

Going with **option A** (the issue's implied preferred direction): render user messages through `MarkdownRenderer` by default and expose the same toggle on both sides. The alternative (drop the toggle from assistant messages) would regress UX in the opposite direction.

> Note: until #1000 and #1001 merge, this diff shows all three PRs' changes stacked. Once they land, the diff shrinks to just the user-message changes here.

## Changes

**`TerminalOutput`** (`src/renderer/components/TerminalOutput.tsx`):
- The AI-mode render paths already ran user messages through `MarkdownRenderer` when `markdownEditMode` was false — but the toggle button itself was gated `log.source !== 'user'`, so users could never flip to plain text. Dropped that gate. Same icon, same behavior, same `Cmd+E` shortcut, just available on both message kinds.

**`GroupChatMessages`** (`src/renderer/components/GroupChatMessages.tsx`):
- All three render paths (collapsed, expanded, default) previously diverged on `!isUser`: assistant content went through `MarkdownRenderer`, user content went through a plain-text `whitespace-pre-wrap` div. Dropped the `!isUser` part of those conditions so users get the same markdown treatment (including `chatLineBreaks` and `chatMath`) when `markdownEditMode` is false.
- The raw-mode branch still keeps \`isUser ? msg.content : stripMarkdown(msg.content)\` — user input shows verbatim in raw mode; assistant content gets markdown stripped so the raw view stays readable.
- The action-buttons wrapper (markdown toggle + copy button) was gated `!isUser`. Dropped that so users can also flip the toggle and copy their own messages.

## Behavior change worth flagging

User messages that contained markdown-like syntax (e.g. \`*literal*\`, \`# heading\`, \`> quote\`) previously rendered as plain text — they will now render with markdown formatting applied. The new toggle lets users flip back to the raw view if needed, but the default visual treatment is different. This is the desired behavior per the issue.

## Tests

- Updated existing test \`does not show markdown toggle button for user messages\` → \`shows markdown toggle button for user messages in AI mode (#622 consistency)\` to assert the new behavior
- 119/119 \`TerminalOutput.test.tsx\` tests pass
- No existing test file for `GroupChatMessages.tsx` (the changes there are mechanical condition relaxations — happy to add coverage if reviewers want it)
- Prettier clean, ESLint clean (one pre-existing warning in unrelated `web-server-factory.ts`), TypeScript clean across all three tsconfigs

## Test plan

- [ ] In AI Terminal, send a user message containing \`**bold**\` → renders bold
- [ ] Hover that user message → markdown/raw toggle button appears, clicking flips to plain \`**bold**\` text
- [ ] In group chat, send a user message with markdown → renders formatted
- [ ] Hover the user message in group chat → toggle + copy buttons appear, both work
- [ ] Assistant messages still behave the same as before
- [ ] Terminal-mode (non-AI) user shell commands still render with the \`\$ \` prefix (no markdown applied)

## Summary by CodeRabbit

* **New Features**
  * KaTeX-powered math rendering: display math (delimited with $$) available in chat when enabled, including on mobile; single-dollar expressions remain literal by default.
  * Chat math and line-break behavior exposed as chat-specific rendering options.
  * Markdown toggle and copy controls shown for all message types in AI mode.

* **Tests**
  * Added/updated tests for math rendering, display promotion, line-breaks, and markdown toggle behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RunMaestro/Maestro/pull/1002?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


pedramamini:
Hey @chr1syy — thanks for breaking #622 into the three-PR stack and tackling the user/assistant markdown consistency angle here. The targeted changes in `TerminalOutput.tsx` and `GroupChatMessages.tsx` (just dropping the `!isUser` / `log.source !== 'user'` gates) read cleanly and the test update accurately captures the new behavior. Option A is the right call.

A couple of things to address before this can land:

**1. CI is red — `package-lock.json` out of sync with `package.json`**

The `test` and `lint-and-format` jobs are both failing at the `npm ci` step with a long list of `Missing X from lock file` errors (emoji-mart, electron-builder bits, testing-library/dom, etc.). Could you run `npm install` locally and push the regenerated `package-lock.json`?

**2. Bot review feedback on the stacked math/markdown pieces (#1000/#1001 territory)**

These don't strictly belong to *this* PR, but since the diff currently shows all three stacked, the bots flagged them here. Worth addressing in the underlying PRs so the stack lands clean:

- **`src/shared/remarkPromoteDisplayMath.ts`** — CodeRabbit and Greptile both correctly flagged that the condition `node.children[0].type === 'inlineMath'` matches both `$x$` and `$$x$$` (remark-math collapses both into the same node type). So a lone-paragraph single-dollar inline math like `$x$` gets silently promoted to centered display math, which probably isn't the intent given the file-level docstring says `$$...$$` only. The fix is to inspect `node.position` against the source `file.value` and only promote when the raw slice starts with `$$`. A test like `'price\n\n$x$\n\nrest'` would lock the semantics in either direction.
- **`src/web/mobile/MobileMarkdownRenderer.tsx`** — CodeRabbit pointed out that `remarkMath` is being added with defaults, which means `singleDollarTextMath: true`. The remark-math docs explicitly warn this interferes with normal dollar amounts in prose, so chat messages like `It costs \$5 and shipping is \$3` could get reinterpreted as math. Recommend `[remarkMath, { singleDollarTextMath: false }]` for chat surfaces.
- **`src/renderer/components/HistoryDetailModal.tsx`** — Greptile flagged enabling `chatMath` here may misparse `$HOME` / `$PATH` / `$5.00`-style content in archived AI agent output. Worth either disabling single-dollar math globally (per above) or scoping `chatMath` to surfaces where math is genuinely expected.

The two single-dollar issues are really the same root cause — passing `{ singleDollarTextMath: false }` to `remarkMath` likely sidesteps both bot findings and the display-math promoter's ambiguity in one shot.

Once CI is green and the math handling is tightened up in the stacked PRs, this should be ready to go.

PR #1014, opened 2026-05-18
fix(cli): persist CLI-created agent to disk before responding

### LINKED ISSUE #1013: Bug: CLI `create-agent` doesn't persist to disk → subsequent `send`/`list agents` fail with AGENT_NOT_FOUND
## Summary

When creating an agent via `maestro-cli create-agent <name>`, the agent is added to the running app's in-memory state (visible via `status` and in the UI) but **never flushed to disk** (`maestro-sessions.json` stays empty). All subsequent CLI commands that read from disk (`list agents`, `send`, `show agent`) fail with `AGENT_NOT_FOUND`, even though the agent is visible in the desktop app.

This breaks the entire "scriptable CLI agent creation → send prompt" workflow that `create-agent` was added to enable.

## Environment

- **Maestro**: v0.15.4-RC (portable build, auto-updated to installed via "Download and Install Update" on first launch)
- **OS**: Windows 11 Pro (10.0.26200)
- **Node**: bundled with Maestro
- **Installation path**: `C:\Users\GT\AppData\Local\Programs\Maestro\`
- **Config path**: `%APPDATA%\Maestro\` (`C:\Users\GT\AppData\Roaming\Maestro\`)
- **Claude Code**: `C:\Users\GT\AppData\Roaming\npm\claude.cmd` (npm install of `@anthropic-ai/claude-code`)

## Reproduction

1. Start Maestro desktop app
2. Run:
   ```bash
   maestro-cli create-agent "Test Agent" \
     --cwd "C:\path\to\workspace" \
     --env "CLAUDE_CONFIG_DIR=C:\Users\<user>\.claude-other" \
     --json
   ```
   Output:
   ```json
   {"success":true,"agentId":"80edc946-4034-4c98-a4d8-a9276b22584f","name":"Test Agent","type":"claude-code"}
   ```

3. Verify agent appears in Maestro desktop UI (✅ visible in Left Bar)
4. Verify in-memory state via `maestro-cli status`:
   ```
   Maestro is running on port 53054 with 1 agent
   ```
5. Try to list agents via CLI:
   ```bash
   maestro-cli list agents --json
   ```
   Output: `[]` (empty)

6. Try to send a message:
   ```bash
   maestro-cli send 80edc946-4034-4c98-a4d8-a9276b22584f "test"
   ```
   Output:
   ```json
   {
     "success": false,
     "error": "Agent not found: 80edc946-4034-4c98-a4d8-a9276b22584f",
     "code": "AGENT_NOT_FOUND"
   }
   ```

7. Inspect `%APPDATA%\Maestro\maestro-sessions.json`:
   ```json
   {"sessions": []}
   ```

The agent is in app memory but not on disk.

## Workaround attempts (none confirmed)

I tried two workarounds, **both failed**:

1. **Rename agent via UI** (right-click → Rename → change a character → Save): `useDebouncedPersistence` did **not** flush. `maestro-sessions.json` remained `{"sessions": []}` even 10+ seconds after the rename was saved.
2. **Send a message manually from UI** (paste prompt + Send): the message went through and produced output, but sessions.json still did not get the agent definition.

The persistence appears completely disconnected from in-memory state mutations for the duration of the running app. Possibly only graceful Quit triggers a flush (not tested yet — would lose in-flight session if so).

## Hypothesis (root cause)

Reading the source:

- `src/cli/services/storage.ts`: `readSessions()` reads `maestro-sessions.json` directly from disk (no IPC fallback).
- `src/cli/commands/send.ts`: calls `getSessionById()` → `readSessions()` → returns empty if file is empty.
- `src/renderer/hooks/utils/useDebouncedPersistence.ts`: persists React session state to disk via `window.maestro.sessions.setAll(...)` — but only triggers when the React state changes (i.e., user UI action).

The `create-agent` IPC handler likely adds the agent to main process state and notifies the renderer, but the renderer's `useDebouncedPersistence` may not detect the change (or the debounce never fires because no follow-up React state mutation happens). Result: the agent lives only in main memory until a UI action triggers a flush.

## Suggested fix

In the `create-agent` IPC handler (main process), after adding the agent to the in-memory session list, explicitly trigger a disk write — bypass the renderer-debounced path so CLI-driven mutations are immediately persistent. Two options:

1. **Main-side flush**: call a `persistSessionsNow()` helper directly in the handler. Cleanest.
2. **CLI-side IPC fallback in `storage.ts`**: when `status` reports the app is running on a known port, query agents via IPC instead of disk. Also handles the inverse case (UI changes not yet debounced).

Option 1 is narrower and matches the contract users would expect: "create-agent returned success → it's persisted, every other tool sees it".

The same bug likely affects any CLI command that mutates session state without going through the React layer (`remove-agent`, `settings agent set` if it touches sessions, etc.). A unified flush helper in the IPC handler would close the class of bugs.

## Impact

Blocks the entire scripted/headless workflow that `create-agent` was added to enable: spin up agent → send first prompt → process result. This is exactly the use case for CI, cron jobs, batch processing — the marquee value of having a CLI in the first place.

## Repro materials

Happy to provide:
- Full `maestro-debug-2026-05-18.log`
- Anonymized `maestro-sessions.json` snapshot pre/post workaround
- Step-by-step screencast if useful

---

Filed during a Maestro evaluation spike — happy to test any patch on Windows 11.

pedramamini:
## Summary

- The `maestro:remoteCreateSession` handler relied on the renderer's 2-second debounced persistence to flush the new agent. A CLI consumer that ran `maestro-cli create-agent` and immediately followed up with `list agents` / `send` / `show agent` hit the disk-backed CLI storage before the flush fired, surfacing as `AGENT_NOT_FOUND`.
- Force an immediate `sessions:setMany` write before sending the `create_session_result`, so the CLI contract "create returned success → the agent is on disk" holds. The subsequent debounced flush is idempotent — it just rewrites the same row.

Closes #1013

## Test plan

- [ ] Start Maestro desktop
- [ ] `maestro-cli create-agent "Test Agent" --cwd <path> --json` → agent shows in Left Bar
- [ ] `maestro-cli list agents --json` → returns the new agent (previously returned `[]`)
- [ ] `maestro-cli send <id> "test"` → succeeds (previously returned `AGENT_NOT_FOUND`)
- [ ] Inspect `%APPDATA%/Maestro/maestro-sessions.json` (or `~/Library/Application Support/Maestro/maestro-sessions.json`) → contains the agent immediately after `create-agent` returns
- [ ] `npm run lint` passes
- [ ] `npx vitest run src/__tests__/renderer/hooks/utils/useDebouncedPersistence.test.ts src/__tests__/cli/commands/create-agent.test.ts src/__tests__/main/web-server/handlers/messageHandlers.test.ts` passes

## Release Notes

* **Improvements**
  * Session creation has been enhanced to ensure newly created sessions are reliably persisted. Improved error handling provides a smoother user experience by gracefully managing any temporary persistence issues without disrupting the workflow.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RunMaestro/Maestro/pull/1014?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/renderer/hooks/remote/useAppRemoteEventListeners.ts
+++ b/src/renderer/hooks/remote/useAppRemoteEventListeners.ts
@@ -1015,6 +1015,20 @@ export function useAppRemoteEventListeners(deps: UseAppRemoteEventListenersDeps)
 				isRemote: false,
 			});
 
+			// Persist the new agent to disk synchronously before responding. The
+			// renderer's debounced persistence path (useDebouncedPersistence) is
+			// driven by React render cycles and a 2s timer, so a CLI consumer that
+			// runs `create-agent` and then immediately `list agents` / `send` would
+			// otherwise hit the disk-backed CLI storage layer before the in-memory
+			// session has been flushed — surfacing as `AGENT_NOT_FOUND` (issue #1013).
+			// `setMany` is incremental and idempotent: the debounced flush that
+			// follows simply rewrites the same row.
+			try {
+				await window.maestro.sessions.setMany([newSession], []);
+			} catch (persistErr) {
+				logger.error('[Remote] Failed to persist new CLI-created session:', undefined, persistErr);
+			}
+
 			window.maestro.process.sendRemoteCreateSessionResponse(responseChannel, {
 				sessionId: newId,
 			});


chr1syy (approved):
LGTM. Verified the fix end-to-end:

- `setMany` on the preload bridge (`src/main/preload/settings.ts:51-52`) and its IPC handler (`src/main/ipc/handlers/persistence.ts:173-290`) are genuinely incremental — they merge into the existing on-disk `sessionsStore` rather than overwriting, so the subsequent debounced flush rewrites the same row without churn, as the inline comment claims.
- Broadcast suppression (`persistence.ts:226-241`) means the debounced re-flush doesn't double-fire `broadcastSessionStateChange` for web clients.
- `await` is correctly placed before `sendRemoteCreateSessionResponse`, so the CLI's promise resolves only after disk write completes.

## Non-blocking: the sibling `remoteDeleteSession` / `remoteRenameSession` handlers retain the same race

The issue reporter explicitly called this out (`A unified flush helper in the IPC handler would close the class of bugs`). Tracing the wiring:

- **`remoteDeleteSession`** — the main-process callback in `src/main/web-server/web-server-factory.ts:1368-1381` is documented `// Fire-and-forget pattern` and returns `true` immediately after `webContents.send('remote:deleteSession', sessionId)`. The WS `delete_session_result: success` reaches the CLI **before the renderer has even run `setSessions((prev) => filtered)`**, let alone the debounced disk flush. A CLI `remove-agent <id>` → `list agents` will still see the agent for up to ~2s.
- **`remoteRenameSession`** — has a response channel and waits for the renderer's `setSessions(...)` call, but still doesn't wait for the disk flush. A CLI `session rename` → `show agent` returns the old name until the debounce fires.

Practical impact is much lower than create (those CLI sequences are rarer than `create-agent → send`), but the fix shape is the same ~10 lines per handler. I'll open a follow-up issue with the trace so you can scope it. Filing this here only so the link survives in the PR record.

One other small thing worth considering: the `catch` here logs but still calls `sendRemoteCreateSessionResponse(..., {sessionId: newId})` with success. If `setMany` throws (e.g. ENOSPC) the CLI gets success and moves on. The `setMany` IPC handler itself returns `false` on ENOSPC at `persistence.ts:274` rather than throwing — so this `catch` only fires on truly unexpected errors. Probably fine as-is; flagging only because the PR description promises "create returned success → it's on disk" and this is the one path where that contract leaks.

Not blocking. Ship it.

chr1syy:
Quick correction on my own review above — I underweighted the silent-catch concern. Re-reading after looking at the CodeRabbit and Greptile inline comments, both bots independently flagged it as P1 with reference to CLAUDE.md §Error Handling & Sentry, and they're right:

- The PR description explicitly promises "create returned success → it's on disk", which is the contract the silent catch breaks on the one path where it matters.
- `setMany` only throws on truly unexpected errors (ENOSPC/ENFILE/EMFILE return `false` from `persistence.ts:274` rather than throwing) — so when it does throw, it's exactly the kind of failure that should reach Sentry and the CLI should not see success.
- If `setMany` throws, the next debounced flush will throw too, so the agent never reaches disk and we're back to issue #1013's exact failure mode, just masked by a successful CLI response.

I'd retract my "Not blocking. Ship it." line on that paragraph. Greptile's suggestion is the cleanest fix (~1 line) — let the unexpected error bubble to the existing outer `catch` which already calls `sendRemoteCreateSessionResponse(responseChannel, null)`:

```ts
try {
    await window.maestro.sessions.setMany([newSession], []);
} catch (persistErr) {
    logger.error('[Remote] Failed to persist new CLI-created session:', undefined, persistErr);
    throw persistErr;
}
```

That way: CLI sees failure, Sentry captures the exception via the existing `withIpcErrorLogging`-style path, and the contract holds.

Everything else in my review still stands — the core fix is correct and the delete/rename gap (tracked in #1054) is separate.

PR #1023, opened 2026-05-19
fix(chat): prevent merged dialog boxes when a queued message is dispatched

### LINKED ISSUE #1022: Bug: The dialog boxes were incorrectly merged.
When I send two messages in quick succession, the AI's two replies get merged into the same dialog box.

1. one queued user message
Image

2. ai response the first question

Image

3. ai response the second question and is merged into the previous answer

Image

pedramamini:
## Summary

Closes #1022.

When a user sent two messages in quick succession, the second got queued and then dispatched on agent exit — but the AI's two replies rendered as a single merged dialog box.

**Root cause:** Stdout chunks from the agent are accumulated in `useBatchedSessionUpdates` and flushed on a 200ms timer. When the agent exits, `useAgentListeners.onExit` synchronously calls `setSessions` to append a new `user` log entry for the next queued message. If the batched flush hasn't fired yet, the trailing chunks of the just-finished response land in `tab.logs` *after* the new user entry. `TerminalOutput`'s `collapsedLogs` memo groups all consecutive non-user/tool/thinking entries into one bubble, so the late chunks of response 1 get coalesced with response 2 — visually merging them.

**Fix:** Flush pending batched updates at the very start of the exit handler so the response that just finished settles in `tab.logs` before any state mutation that could insert a separator after it.

## Test plan

- [x] `npm run lint` — passes
- [x] `npx vitest run src/__tests__/renderer/hooks/useAgentListeners.test.ts` — 39/39 pass (mock updated to include the new `flushNow` field on `BatchedUpdater`)
- [x] Full pre-push validation suite (`npm run validate:push`) — 22,535 passed / 107 skipped
- [ ] Manual: send two prompts in quick succession to Claude Code; confirm the two responses now render as two separate dialog boxes with the queued user message between them

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced reliability of log output when agent processes terminate, ensuring all pending logs are properly displayed.

* **Tests**
  * Updated test fixtures for agent listener functionality.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RunMaestro/Maestro/pull/1023?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


chr1syy (commented):
Reviewed the fix against `main` — `flushNow()` insertion at the top of `onExit` is the right seam, `BatchedUpdater` interface update matches the real `useBatchedSessionUpdates.ts:93,649-651` surface, mock parity looks correct. Static analysis says this is mergeable as-is for main users.

**Heads up on `rc` parity** — the same race exists on `rc` in a different file and won't get the fix from this PR.

`upstream/rc` has commit `e1b8eed29 refactor(agent): decompose useAgentListeners (1,977 -> 104 LOC, -95%)` which split the monolith. The new `src/renderer/hooks/agent/internal/useAgentExitListener.ts` is where `onExit` lives on rc, and:

- it does **not** receive `batchedUpdater` in `UseAgentExitListenerDeps` (lines 40-49),
- the rc coordinator (`useAgentListeners.ts:70-77`) doesn't pass one in,
- the dequeue path at lines 391-398 builds a fresh `user` `LogEntry` and inserts it via `setSessions` with no prior batched flush — the exact same race the main fix closes.

When `rc` next syncs from `main` (or vice versa for release), this PR's diff will hit a structural conflict because the touched lines no longer exist on rc, and the fix will most likely drop silently.

**Suggested companion change on `rc`:**

1. Add `batchedUpdater: UseAgentListenersDeps['batchedUpdater']` to `UseAgentExitListenerDeps` in `src/renderer/hooks/agent/internal/useAgentExitListener.ts:40-49`.
2. Thread it from the coordinator at `src/renderer/hooks/agent/useAgentListeners.ts:70-77`.
3. Call `deps.batchedUpdater.flushNow()` at the top of the `onExit` async callback (after the `-terminal-` guard at line 59, before the `logger.info` at line 61) — same reasoning as the main fix.

Reporter is on `0.16.18-RC`, so rc users will keep hitting this until that companion lands. Happy to open the rc PR if useful.

PR #1032, opened 2026-05-21
feat: add experimental Hermes and Pi agents

XcluEzy7:
## Summary
- Add experimental Hermes and Pi agent IDs, display metadata, context-window defaults, and beta badges.
- Register first-pass CLI definitions and conservative capabilities for Hermes and Pi.
- Add promptArgs handling so agents with prompt flags receive one-shot prompts through buildAgentArgs.
- Add research notes documenting the conservative Hermes/Pi parity baseline.

## Validation
- npm run build:prompts && npm run lint
- npm run lint:eslint
- npm run format:check
- git diff --check

## Notes
- Focused Vitest run was attempted, but the current dependency tree fails before tests start with: ERR_PACKAGE_PATH_NOT_EXPORTED for vite ./module-runner imported by vitest 4.1.7 against vite 5.4.21.

## Summary by CodeRabbit

* **New Features**
  * Added two new beta agents: Hermes and Pi
  * Hermes: streaming-enabled batch processing with image input support
  * Pi: JSONL output, session management, read-only mode, thinking display, and model selection
  * Both agents default to 200K token context windows
  * Added UI icons for Hermes (⚕) and Pi (π)

* **Improvements**
  * Improved agent detection and Pi output parsing with Pi-specific error detection and clearer messages


pedramamini:
Thanks for the contribution @XcluEzy7! Conservative catalog/capability gating looks great, and the research notes are a nice touch. Before this lands, two issues from Greptile that I verified by walking the spawn path — both need addressing:

### 1. `promptArgs` will be appended twice for any agent that defines it

`src/main/utils/agent-args.ts` now appends `agent.promptArgs(options.prompt)` inside `buildAgentArgs`. But `ChildProcessSpawner.spawn()` already calls `promptArgs(prompt)` itself (see `src/main/process-manager/spawners/ChildProcessSpawner.ts` lines 129–130, 154–155, and 173–174). The IPC handler in `src/main/ipc/handlers/process.ts` calls `buildAgentArgs` at line 157 and then forwards both the prompt-containing `argsToSpawn` AND `promptArgs: agent?.promptArgs` plus `prompt: config.prompt` to `spawn()` at line 519. Net effect: the spawner re-appends the same prompt args on top of args that already include them.

This regression is not limited to Hermes/Pi — existing `gemini-cli` (`src/main/agents/definitions.ts:212`) defines `promptArgs: (prompt) => ['-p', prompt]`, so every Gemini CLI batch launch on the current release would ship the prompt twice (`gemini -p <prompt> -p <prompt>`). The same is true on the group-chat path via `src/main/group-chat/group-chat-router.ts`.

The clean fix is to drop the new `promptArgs` block from `buildAgentArgs` and rely on `ChildProcessSpawner.spawn()` to handle prompt placement (the SSH path in `src/main/utils/ssh-spawn-wrapper.ts:130-142` already does this consistently). The third new test in `src/__tests__/main/utils/agent-args.test.ts` (`appends promptArgs after resume args for single-shot resume flows`) would need to be removed/rewritten against the actual spawn path.

### 2. Pi forces `--mode json` but has `supportsJsonOutput: false`

In `definitions.ts`, Pi sets `batchModeArgs: ['--mode', 'json']` — so every batch-mode Pi launch will go into JSON event mode. But `capabilities.ts` declares `supportsJsonOutput: false` (and `usesJsonLineOutput: false`), meaning Maestro has no parser wired up for that stream. Users would see raw JSON in the terminal. Until the Pi parser lands, please use a non-JSON batch mode (e.g. drop `batchModeArgs`, or use `--mode text` if that's supported) — or alternatively flip the capability flags and add the parser in this PR.

### Bonus nits (not blocking, your call)

- Hermes `supportsStreaming: true` is a little optimistic per your own research note ("Structured event output for Maestro was not confirmed"). Worth double-checking before users get a half-rendered stream.
- Once #1 is fixed, please also drop or rewrite the new `agent-args.test.ts` cases so they reflect the actual ordering.

No merge conflicts as of now, so once those two are addressed I'm happy to take another look. Thanks again!

jSydorowicz21:
@greptile @coderabbit please re-review

jSydorowicz21:
@greptile @coderabbit re-review

PR #1033, opened 2026-05-22
Symphony: CLI-IPC: WebSocket bridge for provider agents to trigger Maestro UI actions (#511)

### LINKED ISSUE #511: CLI-IPC: WebSocket bridge for provider agents to trigger Maestro UI actions
## Problem

Provider agents running inside Maestro have no way to trigger UI-side actions in the desktop app. After an agent creates files, modifies the workspace, or generates auto-run documents, the user must manually refresh the file tree, open files, or configure auto-run. This creates friction in agentic workflows where the agent *knows* what it just changed but can't tell Maestro to reflect it.

## Proposed Solution

Add a WebSocket-based IPC bridge between the CLI (`maestro-cli`) and the Electron app, enabling agents to issue commands like:

```bash
maestro-cli refresh-files          # Refresh the file tree panel
maestro-cli open-file <path>       # Open a file in a preview tab
maestro-cli refresh-auto-run       # Refresh the auto-run document list
maestro-cli auto-run <docs...>     # Configure and launch an auto-run
maestro-cli status                 # Check if Maestro is running
maestro-cli send <id> <msg> --tab  # Send + focus the agent's tab
```

### Architecture

```
Electron starts → auto-starts web server → writes discovery file (cli-server.json)
CLI reads discovery file → connects WebSocket → sends commands
Electron shuts down → deletes discovery file
```

The existing Fastify + WebSocket server (`src/main/web-server/`) is reused. A discovery file (`~/.config/maestro/cli-server.json`) makes the server's random port and security token discoverable by the CLI.

## Implementation Plan (4 Phases)

Full implementation documents attached as a gist with 4 phase files (20 tasks total):

**📎 [CLI-IPC Playbook Documents (Gist)](https://gist.github.com/pedramamini/da8d0b5f0dc89df97c158d1ea1b6a616)**

| Phase | Focus | Tasks |
|-------|-------|-------|
| **01** | Server discovery file + always-on CLI server | 4 |
| **02** | New WebSocket message types + renderer handlers | 7 |
| **03** | CLI WebSocket client + new CLI commands | 7 |
| **04** | Auto-run configuration command + system prompt integration | 6 |

### Phase Summary

**Phase 01** — Creates `src/shared/cli-server-discovery.ts` (following `cli-activity.ts` pattern), auto-starts the web server on app launch, writes/deletes the discovery file on startup/shutdown.

**Phase 02** — Adds 3 new WebSocket message types (`open_file_tab`, `refresh_file_tree`, `refresh_auto_run_docs`) across the 5-file callback chain: `messageHandlers.ts` → `CallbackRegistry.ts` → `WebServer.ts` → `web-server-factory.ts` → `preload/process.ts`. Wires renderer handlers via `useRemoteIntegration.ts` using CustomEvent dispatch to `App.tsx`.

**Phase 03** — Creates `MaestroClient` WebSocket client (`src/cli/services/maestro-client.ts`) with `connect()`, `sendCommand<T>()`, and `withMaestroClient()` helper. Adds 4 new CLI commands: `open-file`, `refresh-files`, `refresh-auto-run`, `status`. Adds `--tab` flag to existing `send` command.

**Phase 04** — Adds `configure_auto_run` WebSocket message type with launch/save-as-playbook support. Creates `auto-run` CLI command for configuring and launching auto-runs from the terminal. Updates `maestro-system-prompt.md` so agents know these commands exist.

## What This Enables

Once implemented, the system prompt tells agents about these commands. An agent that just created 5 auto-run documents can:

```bash
maestro-cli refresh-auto-run          # UI immediately shows new docs
maestro-cli auto-run doc1.md --launch # Launch the auto-run
```

An agent that modified workspace files can:

```bash
maestro-cli refresh-files             # File tree updates without user action
maestro-cli open-file ./new-file.ts   # Opens it for the user to see
```

This is a stepping stone toward the full Agent SDK integration (#340) and complements Maestro Cue (#367) event-driven automation.

## Related

- #340 — Claude Agent SDK integration (hooks like `PostToolUse` would be the native equivalent)
- #367 — Maestro Cue (event-driven automation, `file.changed` triggers)

DRAZY:
## Maestro Symphony Contribution

Closes #511

Contributed via [Maestro Symphony](https://runmaestro.ai).

**Status:** Ready for Review
**Started:** 2026-05-22T00:40:48.300Z

---

This PR will be updated automatically when the Auto Run completes.

## Summary by CodeRabbit

* **New Features**
  * Added CLI integration for opening files, refreshing file trees and Auto Run documents, configuring Auto Run, and checking Maestro status.
  * Auto Run now validates Markdown files and uses folder-relative document paths.
  * Remote Auto Run configurations can prefill prompts, documents, looping options, launch runs, or save playbooks.
  * Batch Runner opens with preconfigured documents, prompts, and loop settings.

* **Bug Fixes**
  * Improved handling of connection failures, invalid responses, unavailable sessions, and unsuccessful Auto Run operations.
  * Added clearer status reporting when Maestro is unavailable or stale.

### DIFF
--- a/src/__tests__/cli/commands/auto-run.test.ts
+++ b/src/__tests__/cli/commands/auto-run.test.ts
@@ -16,6 +16,7 @@ import { describe, it, expect, vi, beforeEach, type MockInstance } from 'vitest'
 // Mock fs
 vi.mock('fs', () => ({
 	existsSync: vi.fn(),
+	statSync: vi.fn(),
 }));
 
 // Mock maestro-client
@@ -24,9 +25,14 @@ vi.mock('../../../cli/services/maestro-client', () => ({
 	resolveTargetSessionId: vi.fn(),
 }));
 
+vi.mock('../../../cli/services/storage', () => ({
+	getSessionById: vi.fn(),
+}));
+
 import { autoRun } from '../../../cli/commands/auto-run';
 import { withMaestroClient, resolveTargetSessionId } from '../../../cli/services/maestro-client';
-import { existsSync } from 'fs';
+import { getSessionById } from '../../../cli/services/storage';
+import { existsSync, statSync } from 'fs';
 import path from 'path';
 
 describe('auto-run command', () => {
@@ -39,6 +45,14 @@ describe('auto-run command', () => {
 		consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {});
 		consoleErrorSpy = vi.spyOn(console, 'error').mockImplementation(() => {});
 		processExitSpy = vi.spyOn(process, 'exit').mockImplementation(() => undefined as never);
+		vi.mocked(statSync).mockReturnValue({ isFile: () => true } as never);
+		vi.mocked(getSessionById).mockReturnValue({
+			id: 'agent-123',
+			name: 'Agent',
+			toolType: 'claude-code',
+			cwd: '/repo',
+			autoRunFolderPath: '/path/to',
+		} as never);
 	});
 
 	it('should configure auto-run with valid document paths', async () => {
@@ -61,6 +75,30 @@ describe('auto-run command', () => {
 			expect.stringContaining('Auto-run configured with 2 documents')
 		);
 		expect(processExitSpy).not.toHaveBeenCalled();
+	});
+
+	it('should send relative document paths from the agent Auto Run folder', async () => {
+		vi.mocked(existsSync).mockReturnValue(true);
+		vi.mocked(resolveTargetSessionId).mockReturnValue('agent-123');
+
+		let sentMessage: Record<string, unknown> | undefined;
+		vi.mocked(withMaestroClient).mockImplementation(async (action) => {
+			const mockClient = {
+				sendCommand: vi.fn().mockImplementation((msg) => {
+					sentMessage = msg;
+					return Promise.resolve({
+						type: 'configure_auto_run_result',
+						success: true,
+					});
+				}),
+			};
+			return action(mockClient as never);
+		});
+
+		await autoRun(['/path/to/nested/doc.md'], { agent: 'agent-123' });
+
+		const docs = sentMessage!.documents as Array<{ filename: string }>;
+		expect(docs[0].filename).toBe('nested/doc.md');
 	});
 
 	it('should error with no documents', async () => {
--- a/src/__tests__/cli/commands/auto-run.test.ts
+++ b/src/__tests__/cli/commands/auto-run.test.ts
@@ -101,6 +101,37 @@ describe('auto-run command', () => {
 		expect(docs[0].filename).toBe('nested/doc.md');
 	});
 
+	it.each([
+		{ label: 'the default', options: {}, expected: false },
+		{ label: '--background', options: { background: true }, expected: true },
+		{
+			label: '--focus overriding --background',
+			options: { background: true, focus: true },
+			expected: false,
+		},
+	])('should send background=$expected for $label placement', async ({ options, expected }) => {
+		vi.mocked(existsSync).mockReturnValue(true);
+		vi.mocked(resolveTargetSessionId).mockReturnValue('agent-123');
+
+		let sentMessage: Record<string, unknown> | undefined;
+		vi.mocked(withMaestroClient).mockImplementation(async (action) => {
+			const mockClient = {
+				sendCommand: vi.fn().mockImplementation((msg) => {
+					sentMessage = msg;
+					return Promise.resolve({
+						type: 'configure_auto_run_result',
+						success: true,
+					});
+				}),
+			};
+			return action(mockClient as never);
+		});
+
+		await autoRun(['/path/to/doc.md'], options);
+
+		expect(sentMessage?.background).toBe(expected);
+	});
+
 	it('should error with no documents', async () => {
 		await autoRun([], {});
 
--- a/src/__tests__/cli/commands/refresh-auto-run.test.ts
+++ b/src/__tests__/cli/commands/refresh-auto-run.test.ts
@@ -6,7 +6,7 @@
 import { describe, it, expect, vi, beforeEach, type MockInstance } from 'vitest';
 
 vi.mock('../../../cli/services/maestro-client', () => ({
-	resolveSessionId: vi.fn(),
+	resolveTargetSessionId: vi.fn(),
 	withMaestroClient: vi.fn(),
 }));
 
@@ -15,7 +15,7 @@ vi.mock('../../../cli/output/formatter', () => ({
 }));
 
 import { refreshAutoRun } from '../../../cli/commands/refresh-auto-run';
-import { resolveSessionId, withMaestroClient } from '../../../cli/services/maestro-client';
+import { resolveTargetSessionId, withMaestroClient } from '../../../cli/services/maestro-client';
 
 describe('refresh-auto-run command', () => {
 	let consoleSpy: MockInstance;
@@ -27,7 +27,7 @@ describe('refresh-auto-run command', () => {
 		consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {});
 		consoleErrorSpy = vi.spyOn(console, 'error').mockImplementation(() => {});
 		processExitSpy = vi.spyOn(process, 'exit').mockImplementation(() => undefined as never);
-		vi.mocked(resolveSessionId).mockReturnValue('target-session');
+		vi.mocked(resolveTargetSessionId).mockReturnValue('target-session');
 		vi.mocked(withMaestroClient).mockImplementation(async (action) => {
 			const client = {
 				sendCommand: vi.fn().mockResolvedValue({
@@ -40,9 +40,9 @@ describe('refresh-auto-run command', () => {
 	});
 
 	it('refreshes Auto Run documents with an explicit session', async () => {
-		await refreshAutoRun({ session: 'target-session' });
+		await refreshAutoRun({ agent: 'target-session' });
 
-		expect(resolveSessionId).toHaveBeenCalledWith({ session: 'target-session' });
+		expect(resolveTargetSessionId).toHaveBeenCalledWith('target-session');
 		expect(withMaestroClient).toHaveBeenCalledTimes(1);
 		const action = vi.mocked(withMaestroClient).mock.calls[0][0];
 		const sendCommand = vi.fn().mockResolvedValue({
@@ -59,7 +59,7 @@ describe('refresh-auto-run command', () => {
 	});
 
 	it('refreshes Auto Run documents with the resolved default session', async () => {
-		vi.mocked(resolveSessionId).mockReturnValue('resolved-session');
+		vi.mocked(resolveTargetSessionId).mockReturnValue('resolved-session');
 
 		await refreshAutoRun({});
 
@@ -78,11 +78,9 @@ describe('refresh-auto-run command', () => {
 	it('exits with an error when Maestro is not reachable', async () => {
 		vi.mocked(withMaestroClient).mockRejectedValue(new Error('Maestro desktop app is not running'));
 
-		await refreshAutoRun({ session: 'target-session' });
+		await refreshAutoRun({ agent: 'target-session' });
 
-		expect(consoleErrorSpy).toHaveBeenCalledWith(
-			'Error: Failed to refresh Auto Run documents: Maestro desktop app is not running'
-		);
+		expect(consoleErrorSpy).toHaveBeenCalledWith('Error: Maestro desktop app is not running');
 		expect(processExitSpy).toHaveBeenCalledWith(1);
 	});
 
@@ -98,11 +96,9 @@ describe('refresh-auto-run command', () => {
 			return action(client as never);
 		});
 
-		await refreshAutoRun({ session: 'target-session' });
+		await refreshAutoRun({ agent: 'target-session' });
 
-		expect(consoleErrorSpy).toHaveBeenCalledWith(
-			'Error: Failed to refresh Auto Run documents: Session not found'
-		);
+		expect(consoleErrorSpy).toHaveBeenCalledWith('Error: Session not found');
 		expect(processExitSpy).toHaveBeenCalledWith(1);
 	});
 });
--- a/src/__tests__/cli/commands/refresh-auto-run.test.ts
+++ b/src/__tests__/cli/commands/refresh-auto-run.test.ts
@@ -1,0 +1,108 @@
+/**
+ * @file refresh-auto-run.test.ts
+ * @description Tests for the refresh-auto-run CLI command
+ */
+
+import { describe, it, expect, vi, beforeEach, type MockInstance } from 'vitest';
+
+vi.mock('../../../cli/services/maestro-client', () => ({
+	resolveSessionId: vi.fn(),
+	withMaestroClient: vi.fn(),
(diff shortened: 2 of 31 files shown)


DRAZY:
## Symphony Contribution Summary

This pull request was created using [Maestro Symphony](https://runmaestro.ai/symphony) - connecting AI-powered contributors with open source projects.

### Contribution Stats
| Metric | Value |
|--------|-------|
| Input Tokens | 55,177,983 |
| Output Tokens | 188,540 |
| Total Tokens | 55,366,523 |
| Estimated Cost | $0.00 |
| Time Spent | 2h 59m 41s |
| Documents Processed | 4 |
| Tasks Completed | 23 |

---
*Powered by [Maestro](https://runmaestro.ai) • [Learn about Symphony](https://docs.runmaestro.ai/symphony)*

jSydorowicz21:
Hey @DRAZY can you retarget this at the RC branch? Our branching strategy is to give things time to cook in a release candidate branch before we merge to mainline builds!

DRAZY:
Retargeted this PR to the `rc` branch. Thanks for the heads up on the release-candidate flow.

pedramamini:
Hi @DRAZY, thanks so much for this contribution via Symphony, and thanks for retargeting it to `rc` in line with our release-candidate flow!

I went through the PR along with the automated feedback from CodeRabbit and Greptile, and the good news is the substantive review findings look addressed in `d5a098c` ("Fix CLI IPC review issues"). A few I confirmed against the current head:

- CLI path validation (Greptile P1): `resolveAutoRunDocumentFilename` now validates that the document lives inside the agent's Auto Run folder and forwards the relative path the desktop actually reads, instead of a bare basename. Nicely handled.
- Live mode teardown (Greptile P2): `live:stopServer` now calls `ensureCliServer(deps)` so turning off Live Mode no longer kills the always-on CLI IPC server.
- The `Date.now()` response-channel collision is now a unique id, and all 10 CodeRabbit findings are marked resolved.

Two things before we can approve and merge:

1. **Merge conflicts with `rc`.** The branch has fallen behind since the June 12 merge and now conflicts. Could you please pull in the latest `rc` (merge or rebase) and resolve the conflicts so we get a clean merge? The conflicting files are:
   - `package-lock.json`
   - `src/cli/commands/status.ts`
   - `src/cli/services/maestro-client.ts`
   - `src/main/ipc/handlers/web.ts`

   The other overlapping files auto-merge cleanly, so it should be a fairly contained resolution.

2. **[WIP] status.** The title still carries `[WIP]` and the description says "Status: In Progress". Could you let us know whether this is ready for a final review pass, or if there are still pieces in flight? Once the WIP marker comes off and the conflicts are resolved, we'll do a final review and get it approved.

Thanks again for the solid work here, this is shaping up well!

DRAZY:
Merged the latest `rc` into this branch and resolved the reported conflicts in `package-lock.json`, `src/cli/commands/status.ts`, `src/cli/services/maestro-client.ts`, and `src/main/ipc/handlers/web.ts`. I also removed the WIP marker/status; this is ready for final review. Validation passed locally and in the pre-push hook: `npm run format:check:all`, `npm run lint`, `npm run lint:eslint`, and `npm run test`.

pedramamini:
@DRAZY this is conflicting with `rc` now. Could you rebase onto the latest `rc` and re-push? Keen to review the Symphony CLI-IPC WebSocket bridge once it merges clean.

DRAZY:
Will do!

DRAZY:
Rebased this branch onto the latest `rc` and force-pushed with lease. Conflicts are resolved on GitHub now, and local/pre-push validation passed: `npm run format:check:all`, `npm run lint`, `npm run lint:eslint`, and the full pre-push test run (33,638 tests). Thanks for the heads up.

pedramamini:
Hi @DRAZY, thanks again for this contribution and for the quick turnaround on the last rebase. The CLI-IPC bridge itself is in good shape, and I appreciate the test coverage you added across the new commands.

Unfortunately `rc` has moved quite a bit since your July 10 rebase, and that has surfaced a couple of things we need sorted before we can approve.

### 1. Merge conflicts with `rc` (again, sorry)

The branch conflicts once more. Conflicting files:

- `src/main/web-server/types.ts`
- `src/prompts/maestro-system-prompt.md`
- `src/renderer/App.tsx`
- `src/renderer/components/BatchRunnerModal.tsx`
- `src/renderer/stores/modalStore.ts`

Could you please rebase onto the latest `rc` and re-push? The remaining files auto-merge cleanly.

### 2. Duplicate remote event listeners in `App.tsx` (please read before resolving the conflict)

This is the important one, and it drives how the `App.tsx` conflict should be resolved.

When you originally wrote this, `App.tsx` was the right home for the remote listeners. Since then `rc` has extracted all remote event handling into `src/renderer/hooks/remote/useAppRemoteEventListeners.ts`, and that hook already handles all four of the events this PR adds a `useEffect` for:

- `maestro:openFileTab` (hook line ~99)
- `maestro:refreshFileTree` (~146)
- `maestro:refreshAutoRunDocs` (~242)
- `maestro:configureAutoRun` (~358)

`App.tsx` on current `rc` registers none of these, but it does mount `useAppRemoteEventListeners`. So if the `App.tsx` block from this branch lands as-is, every one of those events gets handled twice. For `configure_auto_run` that is not just redundant work: both handlers call `sendRemoteConfigureAutoRunResponse` on the same `responseChannel`, and both run the side effects. A `--save-as` invocation would create the playbook twice, and a `--launch` invocation would call `startBatchRun` twice for the same session.

There is also a behavior gap: the hook's version handles `config.worktree`, the `App.tsx` copy in this branch does not. CodeRabbit flagged the same duplication on `App.tsx:2313`.

Suggested resolution: when you resolve the `App.tsx` conflict, drop the CLI-IPC `useEffect` entirely and take `rc`'s side, then fold anything genuinely unique in your version into `useAppRemoteEventListeners` instead. Worth diffing the two `configureAutoRun` implementations carefully so nothing you added (the duplicate-filename detection, the `autoRunContentVersion` bump, the right-panel focus) gets lost in the process.

### 3. Wrong flag names in `src/prompts/maestro-system-prompt.md`

The new CLI Integration section documents `[--session <id>]` for three commands:

```bash
{{MAESTRO_CLI_PATH}} open-file <file-path> [--session <id>]
{{MAESTRO_CLI_PATH}} refresh-files [--session <id>]
{{MAESTRO_CLI_PATH}} refresh-auto-run [--session <id>]
```

None of those commands register a `--session` option. In `src/cli/index.ts` all three take `-a, --agent <id>`, and `--session` only exists on an unrelated command for resuming an agent session. Since this file is a system prompt handed to the agents themselves, an agent following it verbatim gets an unknown-option error from commander. Could you switch those to `[--agent <id>]`?

While you are in there, the `auto-run` example omits `--agent` even though `AutoRunOptions.agent` is supported, so adding `[--agent <id>]` there would keep the section consistent.

### 4. Nit (optional)

`src/cli/commands/auto-run.ts` around line 196: the `if (result.success)` wrapper is unreachable-as-false, since the `!result.success` branch above it already returns. Safe to unwrap. Purely cosmetic, take it or leave it.

---

Items 1 through 3 are what we need for approval. Item 2 in particular is the one to be careful with, since a clean textual conflict resolution will still leave the duplicate listeners in place. Thanks for bearing with the moving target on `rc`, and thanks again for the solid work here.

DRAZY:
Hi @pedramamini, I rebased the branch onto the latest `rc` (`3cf55dcb8`) and force-pushed with lease. GitHub now reports the PR as mergeable.

I resolved the RC conflicts using the current extracted architecture:

- Dropped the duplicate CLI IPC listener block from `App.tsx` and kept `maestro:configureAutoRun` in `useAppRemoteEventListeners`.
- Folded the configure-only modal prefill, selected-document content, Auto Run panel focus, and response handling into that hook.
- Removed duplicate preload bridge registrations in `useRemoteIntegration` that overlapped with current RC.
- Preserved the RC `BatchRunnerModal` hook extraction while passing remote documents, prompt, and loop settings through the extracted hooks.
- Confirmed the current progressive CLI prompt docs use `--agent`, including the Auto Run example.
- Unwrapped the redundant Auto Run success guard.
- Preserved nested folder-relative Auto Run document paths for both configure and launch flows, with regression coverage.

Validation passed in the pre-push hook: full repository formatting, all TypeScript configs, ESLint, and 1,460 test files / 35,828 tests. The pushed head is `82bfd44d8`. CodeRabbit is currently pending. Ready for your final review pass.

pedramamini:
Hi @DRAZY, thanks again for sticking with this one - the Aug 6 pass landed everything we asked for, and I want to call that out specifically. The duplicate listener block is gone from `App.tsx`, `configureAutoRun` now lives in `useAppRemoteEventListeners` with the prefill/focus/response handling folded in properly, the `--session` vs `--agent` prompt docs are correct, and the folder-relative document resolution with regression coverage is a genuine improvement over the basename fallback it replaced. Nice work.

Since your rebase, most of the main-process plumbing this PR originally carried has landed on `rc` independently (`refresh-auto-run`, `refresh-files`, `open-file`, the `randomUUID()` response channel in `autoRunConfigCallbacks.ts`), so the effective surface here is now much smaller and much easier to reason about. That is a good place to be.

Three things before we can approve.

### 1. Merge conflicts with `rc` (sorry, again)

`rc` has moved since Aug 6 and the branch conflicts once more. The good news is it is down to two files this time:

- `src/main/web-server/types.ts`
- `src/renderer/hooks/remote/useAppRemoteEventListeners.ts`

Everything else auto-merges cleanly. The `types.ts` conflict should be mechanical - `rc` reworked that file (`AutoRunState` is now an alias of the shared `AutoRunBroadcastState`, `SessionUsageStats` became `UsageStats`, and new shared imports were added at the top), so your `ConfigureAutoRunDocument` / `ConfigureAutoRunConfig` / `ConfigureAutoRunResult` additions just need re-siting. Worth a quick check that `ConfigureAutoRunConfig` still agrees with the callback signature in `src/main/web-server/handlers/messageHandlers/types.ts` after you resolve, rather than becoming a fourth independent declaration of the same shape - that was Greptile's P2 on the original push and it would be easy to reintroduce here.

The `useAppRemoteEventListeners.ts` conflict needs more care, since `rc` has been actively editing that hook (background placement for view-moving verbs, per-tab composer settings, the pasted-image CLI). Please re-read the surrounding handlers after resolving rather than resolving purely textually.

### 2. `throw error` in the WebSocket message handler will crash the CLI

`src/cli/services/maestro-client.ts`, the new catch at the bottom of `setupMessageHandler`:

```ts
} catch (error) {
	this.rejectAllPending(new Error('Invalid message from Maestro desktop app'));
	throw error;
}
```

Replacing the silent `catch {}` with an explicit rejection is the right call and I am glad you made it - CodeRabbit was correct that swallowing malformed JSON hides upstream failures. But the `throw error` on the second line undoes it. This runs inside a `ws.on('message', ...)` listener, so the exception propagates straight out of `EventEmitter.emit()` and there is no `uncaughtException` handler anywhere in `src/cli/`. The result is that the caller gets its clean rejection from `rejectAllPending` and then the process dies with an unhandled stack trace on top of it, which reads to the user as a crash rather than as "the app sent something I could not parse".

Note also that the `try` wraps the whole dispatch body, not just the `JSON.parse`, so anything that throws during matching takes the same path.

Dropping the `throw` line is all that is needed - `rejectAllPending` already surfaces the failure through the promise, which is the path the caller is actually awaiting. A `logger`/`console.error` in its place would be reasonable if you want the detail retained.

One smaller related note: `rejectAllPending` fails *every* in-flight request on a single bad frame. That is defensible for a CLI that is almost always doing one thing at a time, so I am not asking you to change it - just flagging that it is a deliberate choice worth a one-line comment.

### 3. `auto-run` needs `--background` now that the configure path moves the view

This is the one design item, and it is a consequence of the new Case 3 rather than a pre-existing gap.

Before this PR, `configure_auto_run` without `--launch` or `--save-as` returned an error and changed nothing on screen. The new Case 3 in `useAppRemoteEventListeners.ts` does all of this:

```ts
setActiveSessionId(sessionId);
uiActions.setRightPanelOpen(true);
uiActions.setActiveRightTab('autorun');
useModalStore.getState().openBatchRunnerWithConfig(batchConfig);
```

That switches the active agent, opens the Right Bar, changes its tab, and puts a modal on screen - so `auto-run` is now a view-moving verb, and an agent calling it yanks the human off whatever they were looking at with no way to opt out.

We have a standing convention for exactly this (`Focus belongs to the human operator`): CLI side uses `resolveBackgroundFlag(options, '<verb>')` and the handler side uses `readBackgroundField(message)`, both from `src/shared/focusPlacement.ts`. Nine verbs already ride it - `open-file`, `open-terminal`, `open-browser`, `tab new`, `create-agent`, `create-worktree`, `dispatch --new-tab`, `switch-mode`, and most relevantly `refresh-auto-run`, which sits directly above `auto-run` in `src/cli/index.ts` and already registers both flags.

What we would like:

- Register `--background` / `--focus` on the `auto-run` command in `src/cli/index.ts`, matching the wording used on `refresh-auto-run`.
- Resolve it with `resolveBackgroundFlag(options, 'auto-run')` in `src/cli/commands/auto-run.ts` and put it on the `configure_auto_run` payload.
- Read it with `readBackgroundField(message)` in the handler and thread it into the renderer, so Case 3 skips the three focus calls above when it is set. The modal still gets configured either way - `background: true` means the state changes without the view moving, not that the work is skipped.

Two things to be careful of, both documented on `readBackgroundField`: it is strictly opt-in, so please do not write `message.background !== false` (that would read an absent field as an opt-in and silently stop the verb from focusing for every existing caller), and the default stays focus-on, so no current behavior changes for anyone not passing the flag.

---

Items 1 and 2 are required. Item 3 is required too, but it is a small, well-precedented change and `refresh-auto-run` is a near-exact template to copy from.

Everything else in the current diff looks good to me - the CLI-side `getAutoRunDocumentFilename` validation, tightening `resolveRemoteAutoRunFilename` to throw instead of silently falling back to a basename the batch processor could not have read anyway, the `if (msgRequestId) return;` guard that stops a mismatched reply from falling through to first-match-wins type matching, and treating a missing `success` as failure. The test coverage across the new commands is solid.

Thanks for the patience with `rc` being a moving target here.

DRAZY:
Hi @pedramamini, I worked through all three requested items and force-pushed the rebased branch with lease.

- Rebased onto the current `rc` tip, `6f75443ad`. GitHub now reports the PR as mergeable.
- Removed the rethrow from the WebSocket message listener. Malformed frames reject all pending commands through their awaited promises, with a comment documenting that deliberate short-lived CLI connection behavior.
- Added `auto-run --background` and `--focus` through the existing `resolveBackgroundFlag` / `readBackgroundField` path. The default still focuses. Background mode configures session state and opens the configured Batch Runner modal without switching the active agent or changing the Right Bar state.
- Consolidated the Auto Run callback and renderer bridge typing on the canonical `ConfigureAutoRunConfig` in `src/main/web-server/types.ts`, avoiding another independent declaration.
- Reconciled the older `refresh-auto-run` tests with the current RC payload contract.

Validation passed on the final rebased head `b60dbfe20`: Prettier, all TypeScript configs, ESLint, 528 focused tests, and the full suite with 1,826 test files / 42,419 tests passing (110 skipped). The full suite was run with Node 25 experimental web storage disabled because its default localStorage object is invalid without a persistence path.

This is ready for the final review pass.

PR #1052, opened 2026-05-28
fix(chat): flush batched updates at top of onExit on rc (#1022)

chr1syy:
rc companion to #1023. Same race, different file after the post-decompose layout on rc.

Closes the same gap reported in #1022 for users on the RC channel (the reporter is on \`0.16.18-RC\`).

## Root cause (recap)

Stdout chunks from the agent are accumulated in \`useBatchedSessionUpdates\` and flushed on a 200ms timer. When the agent exits, \`onExit\` synchronously calls \`setSessions\` to append a new \`user\` log entry for the next queued message. If the batched flush hasn't fired yet, trailing chunks of the just-finished response land in \`tab.logs\` *after* the new user entry. \`TerminalOutput\`'s \`collapsedLogs\` memo groups consecutive non-user/tool/thinking entries into one bubble, so the late chunks of response 1 get coalesced with response 2 — visually merging them.

## Why a separate PR

After commit \`e1b8eed29 refactor(agent): decompose useAgentListeners (1,977 -> 104 LOC, -95%)\` on rc, the \`onExit\` handler now lives in \`src/renderer/hooks/agent/internal/useAgentExitListener.ts\` and does not receive \`batchedUpdater\`. PR #1023 targets \`main\` (where \`useAgentListeners.ts\` is still monolithic), so its diff cannot apply cleanly to rc. When rc next syncs from main, the structural conflict means the fix would most likely silently drop unless re-applied — this PR does that.

## Fix

- Add \`flushNow: () => void\` to the local \`BatchedUpdater\` subset interface in \`src/renderer/hooks/agent/internal/types.ts\` (matches the real \`useBatchedSessionUpdates.ts:93\` surface).
- Plumb \`batchedUpdater\` through \`UseAgentExitListenerDeps\` and the \`useAgentListeners\` coordinator.
- Call \`deps.batchedUpdater.flushNow()\` at the top of the \`onExit\` async callback (after the \`-terminal-\` early-return guard).

## Test plan

- [x] Added regression test \`'flushes batched updates at the top of onExit'\` in \`useAgentExitListener.test.tsx\` asserting \`flushNow\` is called on AI exit.
- [x] Updated \`makeBatched\` factories in the 4 sibling listener tests (\`useAgentDataListener\`, \`useAgentSessionIdListener\`, \`useAgentStderrListener\`, \`useAgentUsageListener\`) to include \`flushNow: vi.fn()\` so they conform to the extended interface.
- [x] \`npx vitest run\` on all 5 internal listener test files — 27/27 pass.
- [x] \`npx prettier --check\` on all 8 changed files — clean.
- [x] \`npx tsc --noEmit\` — no new errors in changed files (8 pre-existing TS6133 errors in \`FileExplorerPanel\` components are unrelated).
- [ ] Manual: send two prompts in quick succession to Claude Code on rc; confirm the two responses render as two separate dialog boxes with the queued user message between them.

## Related

- Companion to #1023 (main-branch fix). Both should land for full coverage.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved output ordering so trailing buffered log/output chunks are flushed before finalizing agent sessions, preventing out-of-sequence logs when a process exits.

* **Tests**
  * Added regression coverage to ensure buffered output is flushed in the correct order on agent exit.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RunMaestro/Maestro/pull/1052?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/__tests__/renderer/hooks/agent/internal/useAgentDataListener.test.tsx
+++ b/src/__tests__/renderer/hooks/agent/internal/useAgentDataListener.test.tsx
@@ -25,6 +25,7 @@ function makeBatched(): BatchedUpdater {
 		updateContextUsage: vi.fn(),
 		updateCycleBytes: vi.fn(),
 		updateCycleTokens: vi.fn(),
+		flushNow: vi.fn(),
 	};
 }
 
--- a/src/__tests__/renderer/hooks/agent/internal/useAgentExitListener.test.tsx
+++ b/src/__tests__/renderer/hooks/agent/internal/useAgentExitListener.test.tsx
@@ -22,8 +22,22 @@ function makeRef(): {
 	return { current: new Map() };
 }
 
+function makeBatched() {
+	return {
+		appendLog: vi.fn(),
+		markDelivered: vi.fn(),
+		markUnread: vi.fn(),
+		updateUsage: vi.fn(),
+		updateContextUsage: vi.fn(),
+		updateCycleBytes: vi.fn(),
+		updateCycleTokens: vi.fn(),
+		flushNow: vi.fn(),
+	};
+}
+
 function makeDeps() {
 	return {
+		batchedUpdater: makeBatched(),
 		getBatchStateRef: { current: null },
 		processQueuedItemRef: { current: null },
 		addHistoryEntryRef: { current: null },
@@ -100,6 +114,19 @@ describe('useAgentExitListener', () => {
 		const updated = useSessionStore.getState().sessions[0];
 		const log = updated.shellLogs[updated.shellLogs.length - 1];
 		expect(log?.text).toContain('exited with code 1');
+	});
+
+	// Regression for #1022: trailing batched stdout chunks must be flushed
+	// before the queued-message dispatch appends a new user log entry,
+	// otherwise late chunks land after the user entry and merge with the
+	// next response's bubble in collapsedLogs grouping.
+	it('flushes batched updates at the top of onExit', async () => {
+		const deps = makeDeps();
+		renderHook(() => useAgentExitListener(deps));
+		await act(async () => {
+			await handler!('sess-1-ai-tab-1', 0);
+		});
+		expect(deps.batchedUpdater.flushNow).toHaveBeenCalled();
 	});
 
 	it('deletes activeHiddenToolRef entry on AI exit', async () => {
--- a/src/__tests__/renderer/hooks/agent/internal/useAgentSessionIdListener.test.tsx
+++ b/src/__tests__/renderer/hooks/agent/internal/useAgentSessionIdListener.test.tsx
@@ -25,6 +25,7 @@ function makeBatched(): BatchedUpdater {
 		updateContextUsage: vi.fn(),
 		updateCycleBytes: vi.fn(),
 		updateCycleTokens: vi.fn(),
+		flushNow: vi.fn(),
 	};
 }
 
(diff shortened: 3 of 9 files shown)


pedramamini:
Thanks for porting this over, @chr1syy — appreciate you catching the post-decompose layout difference on `rc` and adding a focused regression test. The fix itself is correct: flushing at the top of `onExit` is the right place to close the #1022 race, and it mirrors the companion #1023 cleanly.

One blocking item before this can land:

### 🔴 CI `test` job is red — 15 failures

The new `deps.batchedUpdater.flushNow()` call at `useAgentExitListener.ts:70` throws `TypeError: deps.batchedUpdater.flushNow is not a function` in the **coordinator** suite `src/__tests__/renderer/hooks/useAgentListeners.test.ts`. Its mock factory `createMockBatchedUpdater()` (line ~134) wasn't updated with the new method:

```ts
function createMockBatchedUpdater(): BatchedUpdater {
	return {
		appendLog: vi.fn(),
		markDelivered: vi.fn(),
		markUnread: vi.fn(),
		updateUsage: vi.fn(),
		updateContextUsage: vi.fn(),
		updateCycleBytes: vi.fn(),
		updateCycleTokens: vi.fn(),
		flushNow: vi.fn(), // <-- add this
	};
}
```

Your test plan ran the 5 `internal/` listener suites (which you correctly updated), but this coordinator-level suite keeps its own `batchedUpdater` mock and was missed. Adding `flushNow: vi.fn()` there should turn CI green.

### Optional nits (non-blocking, your call) — from Greptile

- `deps.batchedUpdater` is used in the `onExit` closure but isn't listed in the `useEffect` dependency array (lines ~636–643), even though every other `deps.*` is. Small consistency gap; harmless in practice since the updater is stable.
- `flushNow()` fires before the `-batch-` early-return, so it runs (as a no-op) for batch sessions too. Moving it after that guard would scope it more precisely, though behavior is unchanged.

Once CI is green I'll take another pass. Thanks again for the contribution!

chr1syy:
Pushed `acf7df53e` addressing all three:

1. **CI blocker (the actual one!):** Added `flushNow: vi.fn()` to `createMockBatchedUpdater` in the coordinator suite `src/__tests__/renderer/hooks/useAgentListeners.test.ts`. My test plan missed this one because it lives outside `internal/` — apologies for the noise.
2. **`-batch-` scope:** Added `if (sessionId.includes('-batch-')) return;` immediately after the `-terminal-` guard, before `flushNow()`. The existing `else if (sessionId.includes('-batch-')) { return; }` branch at line ~89 in the detection chain is now unreachable — I left it in place defensively to keep this change minimal, but happy to delete it in this PR or a follow-up if you'd prefer no corpses.
3. **dep array:** Added `deps.batchedUpdater` between `deps.addHistoryEntryRef` and `deps.getBatchStateRef` to match the existing pattern.

Local verification:
- `npx vitest run` on the coordinator + 5 internal listener suites → **94/94 pass**.
- `npx prettier --check` → clean.

CI should turn green on the next run.

pedramamini:
Closing as already-fixed on `rc`. The #1022 `flushNow` fix this PR proposes is already present on the base branch (it landed via `fix/1022-flush-batched-updates-before-queue-dispatch`, merged in `62f7bb589`):

- `BatchedUpdater.flushNow` is already declared in `src/renderer/hooks/agent/internal/types.ts` ("Force an immediate flush of pending batched updates (issue #1022).").
- `useAgentExitListener.ts` already calls `deps.batchedUpdater.flushNow()` in `onExit`, placed right before the queued-item dispatch (only when there's a queued item to process) rather than unconditionally at the top of the handler. That placement is slightly more targeted than this PR's, and covers the same ordering bug.

Because of that overlap the branch now conflicts against `rc` (`types.ts`, `useAgentExitListener.ts`, and the related tests all collide). There's nothing left to merge here, so I'm closing it. If you spot a case the existing fix misses (e.g. a flush needed even when no message is queued), reopen with a focused diff on top of current `rc` and a failing repro. Thanks for the fix - the underlying issue is resolved.

PR #1057, opened 2026-05-30
refactor: decompose useTabHandlers by domain and AgentSessionsBrowser into coordinator pattern

reachrazamair:
## Summary

Two consecutive renderer decompositions that continue the §4.x god-component
reduction campaign. Both follow the established coordinator pattern
(shell + barrel + domain hooks + leaf components), pass the full test suite
with zero regressions, and make no behavior changes.

---

### Commit 1 - `useTabHandlers` decomposition by domain

`src/renderer/hooks/tabs/useTabHandlers.ts` was a 2,044-line monolith handling
every tab type (AI, file-preview, browser, terminal, scroll-log) in a single
flat file. It is now a 35-line re-exporting shell backed by 11 focused modules
under `src/renderer/hooks/tabs/internal/`:

| File | Lines | Responsibility |
|---|---:|---|
| `useAITabHandlers.ts` | 382 | AI tab open/close/rename/star/resume |
| `useFilePreviewTabHandlers.ts` | 554 | File preview open/close/navigation/folder |
| `useUnifiedTabHandlers.ts` | 232 | Cross-type close, move, reorder |
| `useScrollLogHandlers.ts` | 138 | Scroll-log tab lifecycle |
| `useBrowserTabHandlers.ts` | 132 | Browser tab open/close/navigate |
| `unifiedCloseHelpers.ts` | 101 | Close-all / close-others shared logic |
| `types.ts` | 121 | Internal shared types |
| `useTabDerivedState.ts` | 73 | Derived tab state memo |
| `useTerminalTabHandlers.ts` | 65 | Terminal tab open/close |
| `filePreviewTabHelpers.ts` | 55 | File-preview path/kind helpers |
| `browserTabHelpers.ts` | 42 | Browser URL/title helpers |

**Shell reduction:** 2,044 → 35 lines (-98%)

11 new test files added covering all extracted units. All existing
`useTabHandlers` callers (App.tsx, MainPanel) import through the unchanged
public surface - no call site edits required.

---

### Commit 2 - `AgentSessionsBrowser` decomposition

`src/renderer/components/AgentSessionsBrowser.tsx` was a 1,512-line
god-component owning four orthogonal concerns: list view (search, filter,
paginated list, activity graph), detail view (session header, stats grid,
messages), aggregate-stats lifecycle (dual-mode: claude-code progressive IPC
vs other-agent locally computed), and coordination (rename, star, auto-jump,
focus restore, keyboard nav, Cmd+F, modal-layer Escape).

Now a 23-file directory:

AgentSessionsBrowser/

- AgentSessionsBrowser.tsx (shell, 433 lines)
- index.ts (barrel)
- types.ts
- components/ (9 files)
- hooks/ (8 files)
- utils/ (3 files)

**Shell reduction:** 1,512 → 433 lines (-71%)

Key invariants preserved and guarded by dedicated regression tests:

- **SSH aggregate-stats key** - `onProjectStatsUpdate` subscription is keyed on
  `activeSession.projectRoot`, NOT `projectPathForSessions`. Using the remote
  path would silently drop all SSH session stats. Two unit tests in
  `useAgentSessionsAggregateStats.test.ts` guard this explicitly.
- **Token zeroing in `buildUsageStats`** - resumed tabs receive cost only
  (all 4 token fields zeroed) so they cannot display a misleading 100% context
  bar from stale cumulative totals.
- **Tool-call filtering in `messagesToLogEntries`** - tool-use messages are
  excluded to match live-session behavior (thinking off by default on restore).
- **Cmd+F migrated** from a raw `document.addEventListener` pair to
  `useEventListener('keydown', handler, { target: document })` per CLAUDE.md.

85 existing integration tests pass unchanged through the barrel. 170 new unit
tests added across 17 files.

---

## Test plan

- `npm run test` - full suite green (29,964 tests, 0 failures)
- `npx tsc --noEmit` - 0 type errors
- `npm run lint` + `npm run lint:eslint` - clean
- `npm run format:check` - clean

## Summary by CodeRabbit

* **New Features**
  * New Agent Sessions Browser modal: searchable session list, session detail/messages view, stats panel, and resume/rename/star actions.
  * Search controls: search-mode selector, named/show-all toggles, activity graph with lookback navigation, and load-earlier messages.

* **Tests**
  * Extensive new unit/integration tests and test utilities covering session UI, hooks, tab behaviors, and search/usage logic.

* **Refactor**
  * Tab management reorganized into modular composable handlers for more reliable tab/session UX.


pedramamini:
Hi @reachrazamair — thank you for this contribution! 🙏 This is a genuinely high-quality decomposition: collapsing the 2,044-line `useTabHandlers` and the 1,512-line `AgentSessionsBrowser` into the coordinator pattern while keeping the public surface unchanged, calling out and regression-testing the tricky invariants (SSH stats keyed on `projectRoot`, token-zeroing on resume, tool-call filtering), and landing 170+ new tests is exactly the kind of god-component reduction we want.

I reviewed the PR alongside the Greptile and CodeRabbit passes and verified the findings against the code. There's **one item I'd like fixed before approval**, plus a few smaller things worth a look. No merge conflicts — the branch is clean against `rc`.

### 🔴 Please fix before merge

**`useAgentSessionsStar.toggleStar` — optimistic update with no rollback** (`src/renderer/components/AgentSessionsBrowser/hooks/useAgentSessionsStar.ts`)

Both bots flagged this and I confirmed it by reading the code. `setStarredSessions(newStarred)` commits the local set, then the `await window.maestro…SessionStarred(...)` IPC call runs with no `try/catch`. If that call rejects (transient error, dropped SSH connection), the local state is already changed, `onUpdateTab` never fires, and the promise rejects unhandled — leaving the UI showing a star state the backend never persisted, with no recovery path.

Two small changes resolve it:
- Wrap the IPC calls in `try/catch` and revert `setStarredSessions` to the prior set on failure.
- Build the next set with the functional updater (`setStarredSessions(prev => …)`) rather than reading `starredSessions` from the closure, so rapid successive toggles don't clobber each other.

### 🟡 Worth addressing (your call)

- **Error surfacing consistency** — `useAgentSessionsSearch` and `useAgentSessionsRename` catch IPC errors but only `logger.error` them (rename also silently runs `cancelRename`), and `useTerminalTabHandlers` `.catch(() => closeTerminalTab(...))` swallows the `isTerminalBusy` failure. Per the error-handling guidance in CLAUDE.md, recoverable IPC failures are fine to handle, but consider whether these should surface to the user / be captured for Sentry rather than disappearing.
- **Debounce reuse** — `useAgentSessionsSearch` hand-rolls a `useRef<Timeout>` + `setTimeout` debounce. The repo already ships `useDebouncedValue` / `useDebouncedCallback` (`src/renderer/hooks/utils/useThrottle.ts`), and CLAUDE-PERFORMANCE.md recommends them for search. Functionally your version is fine, but reusing the canonical helper avoids the kind of reimplementation that doc explicitly calls out.

### ⚪ Minor nits

- `SessionDetailHeader.tsx` — the tooltip/hover absolute time uses `viewingSession.timestamp` while the visible relative label uses `modifiedAt`; they'll disagree when create ≠ last-modified.
- `useFilePreviewTabHandlers.ts` — `if (!content) return` treats a legitimately empty file as a read failure and aborts navigation silently.
- `useAgentSessionsFocusRestore.test.ts` — `afterEach` is used without importing it (works via Vitest globals, but inconsistent with siblings).

### Note on one bot suggestion

Greptile recommends replacing the `useEffect + setTimeout(() => ref.focus())` focus pattern with `useFocusAfterRender(...)`. Heads up that **no such utility exists in the repo**, and I couldn't find that focus convention in any of the `CLAUDE*.md` files — so please disregard that specific suggestion and use your own judgment on the focus timing. (The `handleKeyDown` memoization point it raised is a reasonable micro-optimization if you want it, but not required.)

Thanks again — the star rollback is the only real blocker; everything else is polish. Ping me once that's in and I'll take another pass. 🎬

reachrazamair:
@coderabbit review

PR #1059, opened 2026-05-30
feat(chat): fork conversation from any AI message into a new tab (#205)

### LINKED ISSUE #205: Enhancement: Fork conversation from any point in AI message history
## Summary

Add the ability to fork a conversation from any point in the AI message history, allowing users to explore alternative paths or retry from a specific point without losing the original conversation.

## Proposed Behavior

1. **New Fork Icon**: Add a fork icon (⑂ or similar) to the bottom-right of each AI message in the history, alongside the existing copy/regenerate actions.

2. **Fork Action**: When clicked, the fork icon will:
   - Open a new tab with the session name prefixed with "Forked: " (e.g., "Forked: My Session")
   - Copy only the conversation history up to and including the selected message
   - The new forked session starts fresh from that point, ready for new input

3. **Use Cases**:
   - Explore alternative approaches from a decision point
   - Retry a failed approach without losing the original attempt
   - Branch off to try different implementations
   - Preserve working state while experimenting

## UI Mockup

```
┌──────────────────────────────────────────┐
│ AI Message content here...               │
│                                          │
│                                          │
│                          ↻  📋  ⑂       │  <-- Fork icon added
└──────────────────────────────────────────┘
```

## Implementation Notes

- The fork icon should appear on hover (consistent with other message actions)
- The forked session should be a completely independent session
- Consider whether to include system prompts/context in the fork
- The forked session should use the same agent type as the original

## Related Files

- `src/renderer/components/MessageBubble.tsx` - Add fork icon to message actions
- `src/renderer/App.tsx` - Handle fork action and new tab creation
- `src/main/process-manager.ts` - May need to support initializing with conversation history

## Labels

enhancement, ux

pedramamini:
## Summary

Implements **fork conversation from any point in the AI message history** (#205).

Each AI message's hover toolbar now has a fork action (`⑂`, alongside the existing copy / save actions). Clicking it:

- Opens a **new AI tab** named `Forked: <agent name>` in the current agent (same agent type and working directory).
- Copies the conversation history **up to and including** the selected message into the new tab (with regenerated log ids, so it's a completely independent copy).
- Starts a **fresh provider session** (`agentSessionId: null`).
- Injects the copied conversation into the next message via the existing `pendingMergedContext` mechanism, so the forked agent **continues with context** rather than starting blind. This reuses the same context-formatting pattern as the merge-session feature.

## Design notes

- **New tab vs. new agent:** the issue asks to "open a new tab with the session name prefixed with `Forked: `" and to keep the same agent type. A new AI tab within the current agent satisfies all of that (same agent type, same cwd, independent provider session) with the least machinery, so that's the approach taken.
- **Context preservation:** the issue flagged "consider whether to include system prompts/context." The stated use cases (retry a failed approach, branch off, explore alternatives) only make sense if the forked agent knows the prior conversation, so the copied history is injected as `pendingMergedContext` (system messages excluded), exactly like merge. The next message the user sends carries that context to the agent.
- The fork action gates identically to the existing "Save to file" action (`source !== 'user' && isAIMode`).

## Implementation

- `useTabHandlers.ts` — new `handleForkFromMessage(logId)` (mirrors `handleNewTab`, reuses `createTab`).
- Threaded `onForkFromMessage` through `useMainPanelProps` → `MainPanel` → `TerminalOutput` → message toolbar, mirroring `onReplayMessage`.

## Tests

- `useTabHandlers.test.ts` — fork copies the right slice, names the tab, regenerates ids, sets `pendingMergedContext`, leaves the original tab untouched, and no-ops on a missing log id.
- `TerminalOutput.test.tsx` — fork button renders on AI messages and fires with the message id; absent on user messages.
- Full `npm run lint` (type-check) clean except 4 pre-existing unrelated CSS module errors; targeted Vitest suites pass.

closes #205

## Summary by CodeRabbit

* **New Features**
  * Users can now fork AI conversations from any message into a new independent tab, preserving conversation history up to the fork point.
  * "Fork conversation from here" action is available on AI-generated messages.

* **Tests**
  * Added comprehensive test coverage for fork conversation functionality and related message interactions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RunMaestro/Maestro/pull/1059?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


chr1syy (commented):
Mergeable, CI green, well-tested happy path. The two bot reviewers between them caught what look like two real correctness bugs and a few smaller items — verified them against the code, agree on all five. Notes below; only #1 and #2 feel blocking to me.

### 🟠 Forking a streamed multi-chunk AI response truncates to the first chunk

`TerminalOutput.tsx:1477-1517` (`collapsedLogs`) merges consecutive non-user/non-tool/non-thinking entries into one rendered block, keeping only the first chunk's id (`...currentResponseGroup[0]` at line 1490). The fork toolbar renders against that collapsed block and passes that first id to `onForkFromMessage`. The handler at `useTabHandlers.ts:708-715` then does `sourceTab.logs.slice(0, forkIndex + 1)` against the *raw* logs, so any streamed multi-chunk reply (the common case for Claude/Codex AI responses) gets silently truncated to its first chunk in the forked tab.

One way out: after locating `forkIndex`, walk forward while the next log's source belongs to the "same response" set (the collapse predicate at lines 1499-1510 is a good reference) and extend the slice to cover those chunks too.

(This is what CodeRabbit flagged at `TerminalOutput.tsx:957`.)

### 🟠 Fork button shows on `thinking` / `tool` / `error` bubbles

`TerminalOutput.tsx:944` gates on `log.source !== 'user' && isAIMode`. `LogEntry.source` is `'stdout' | 'stderr' | 'system' | 'user' | 'ai' | 'error' | 'thinking' | 'tool'` (`src/renderer/types/index.ts:191`), so every non-user variant — including the `thinking` pill at line 519, tool blocks, and error blocks — shows a fork affordance and forks from a non-conversation log id. Restricting to `log.source === 'ai'` covers it.

(Same CodeRabbit comment as #1.)

### 🟡 Injected context labels tool/thinking output as `Assistant:`

`useTabHandlers.ts:719` filters with `log.source !== 'system'`, so `thinking` (raw CoT), `tool` (often huge tool-result dumps), `stdout`, `stderr`, `error` all become `Assistant:` lines inside `pendingMergedContext`. The PR description is right that this mirrors `useSendToAgent.ts:606-612`, but the merge-session pattern is questionable for the same reason. CodeRabbit's diff (filter to `log.source === 'user' || log.source === 'ai'`) is the right shape here; worth filing a follow-up to fix the merge path the same way so the two stay in sync.

(Greptile + CodeRabbit both flagged this at `useTabHandlers.ts:725`.)

### 🟡 `Forked: <name>` uses session name, not tab name

`useTabHandlers.ts:727` resolves to `s.name || s.projectRoot.split('/').pop() || 'Unnamed Session'`. When the active tab has an auto-rename like "Implement auth", the fork ends up titled `Forked: MyProject` instead of `Forked: Implement auth`. `sourceTab.name || s.name || …` is the cheap fix.

(Greptile flagged this at `useTabHandlers.ts:734`.)

### 🔵 Nit — test doesn't assert the new wiring

`useMainPanelProps.test.ts:149` only adds `handleForkFromMessage: noop` to the fixture, so if `useMainPanelProps` stops forwarding the callback to `onForkFromMessage` the file still passes. CodeRabbit's suggestion to pass a `vi.fn()` and assert `result.current.onForkFromMessage` points to it is a small but worthwhile coverage add.

---

Otherwise structurally clean — threading through `useMainPanelProps → MainPanel → TerminalOutput` mirrors `onReplayMessage` exactly, the regenerated-ids approach for copied logs is the right call for independence, and the happy-path tests cover the basics. Worth landing #1 and #2 before merge.

chr1syy:
Heads-up — this feature already shipped on `rc` back in April via #744 + #857, and there have been four follow-up fixes since (`75a3def60`, `46362b1f1`, `7cb905044`, `f7369acba`, `fba12d7e4`). None of that has propagated to `main` yet, which is presumably why this PR was written from scratch. The rc implementation lives in `src/renderer/hooks/agent/useForkConversation.ts` and ~~independently arrived at~~ specifically fixes the exact bugs I flagged above. Worth a look before landing this independently.

Also — **correction to finding #2 in my review above**: I suggested gating the fork button on `log.source === 'ai'`. That's wrong for this codebase. AI response text in AI mode is actually stored with `source: 'stdout'` (see `src/renderer/hooks/session/useBatchedSessionUpdates.ts:268`, true on both `main` and `rc`). Narrowing the gate to `'ai'` would make the button disappear from every real AI message — exactly the bug `46362b1f1 fix: show Fork button on AI messages (stdout source) and correct role mapping` fixed on rc. The correct gate is `log.source === 'user' || log.source === 'ai' || log.source === 'stdout'` (gated by `isAIMode`), as in the rc implementation.

That also reframes finding #3: the rc fix maps `tool → "Tool Output"` and `user → "User"`, with `ai` and `stdout` both mapping to `Assistant`, which is the right shape given the `stdout`-as-AI-text convention.

Findings #1 (multi-chunk truncation), #4 (tab name vs session name), and the test-wiring nit all still stand against this PR's implementation, and are all already handled on the rc side — `useForkConversation.ts:30-44` walks `endIndex` forward through consecutive non-user/non-tool/non-thinking entries with a comment pointing at `collapsedLogs`, and the tab name uses `getTabDisplayName(sourceTab)`.

Maintainer's call, but my suggestion would be to port `useForkConversation` (plus the four follow-up fixes) from rc to main rather than land this fresh implementation — the rc code has soak time and the corner-case fixes baked in. Happy to put together that port if useful.

pedramamini:
Closing as superseded. Fork-from-message (#205) already landed on `main` via a standalone `useForkConversation` hook (`src/renderer/hooks/agent/useForkConversation.ts`), fully wired through `App.tsx` → `useMainPanelProps` → `MainPanel` → `TerminalOutput` with a `GitFork` button on each message.

That implementation has also been iterated past where this PR sits, with follow-up fixes for the exact concerns raised in this PR's design notes:

- `7cb905044` create the forked tab in the existing agent instead of spawning a new agent
- `75a3def60` include the full multi-chunk AI response in the forked context
- `f7369acba` insert the forked tab directly after the source in `unifiedTabOrder`
- `46362b1f1` show the Fork button on AI messages (stdout source) + correct role mapping

Same feature, both close #205. No reason to merge a parallel implementation. Thanks for the work here.

PR #1068, opened 2026-06-01
fix(cli): capture copilot-cli sessionId from result events

chr1syy:
## Summary

`maestro-cli send <agent-id> -- <message>` against a `copilot-cli` agent always returned `SendResult.sessionId: null`, so callers that thread `--session <id>` for multi-turn continuity (e.g. the Maestro Relay's Discord bridge) lost context after every turn. Same code path works fine for `claude-code` and `codex`.

## Root cause

Copilot CLI's batch JSON output (`copilot --allow-all --output-format json -p`) does **not** emit a `session.start` event on stdout — verified against `copilot 1.0.56`. The sessionId arrives only on the terminal line:

```json
{"type":"result","timestamp":"...","sessionId":"5fa70e6e-...","exitCode":0,"usage":{...}}
```

The agent definition (`resumeArgs: --resume=<id>`) and the parser (`session.start` → `init`, `result` → carries `sessionId`) are both correct. The break was in the CLI's generic JSON-line spawn loop (`src/cli/services/agent-spawner.ts:processEvent`), which only captured sessionId when `event.type === 'init'` — so for copilot-cli, `sessionId` stayed `undefined`, `AgentResult.agentSessionId` was `undefined`, and `SendResult.sessionId` serialized as `null`.

Verified upstream `copilot --resume=<id>` works as expected: replaying the captured sessionId recalls prior context.

## Fix

Route the capture through `parser.extractSessionId(event)` rather than matching `init` directly. Each adapter already implements `extractSessionId()` with the correct event-shape semantics:

- **copilot-cli** falls back to `raw.sessionId` / `raw.data.sessionId`, so the terminal `result` event is now picked up.
- **codex** returns `event.sessionId` populated from its `session_meta` / `thread.started` init events — unchanged.
- **opencode** populates `event.sessionId` on most event types — first-wins via the `!sessionId` guard, unchanged.
- **factory-droid** likewise unchanged.

claude-code uses its own `spawnClaudeAgent` and is not on this path.

## Test plan

- [x] `npx vitest run src/__tests__/cli/services/agent-spawner.test.ts` — 123/123 pass
- [x] `npx vitest run src/__tests__/main/parsers/copilot-output-parser.test.ts` — 20/20 pass
- [x] Updated the existing copilot-cli spawner test to mirror real CLI output (no `session.start`, sessionId on the final `result`)
- [x] Added a regression test that asserts `--resume=<id>` is forwarded and that the resumed id round-trips through `agentSessionId`
- [x] Prettier clean on modified files
- [x] No new TS errors on modified files (the pre-existing shiki / @xterm / remark-math / semver errors on rc are untouched)
- [x] Live repro on `copilot 1.0.56`: send → captures id from `result` → `--resume=<id>` → agent recalls prior word

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated compatibility with revised Copilot CLI event sequence handling.
  * Improved session ID extraction logic to support identification from multiple event types.

* **Tests**
  * Added test coverage for session resumption functionality with resume token forwarding.

### DIFF
--- a/src/__tests__/cli/services/agent-spawner.test.ts
+++ b/src/__tests__/cli/services/agent-spawner.test.ts
@@ -1457,21 +1457,20 @@ Some text with [x] in it that's not a checkbox
 			// promptArgs path replaces the '--' separator
 			expect(args).not.toContain('--');
 
-			// Emit a session.start init event
-			mockStdout.emit(
-				'data',
-				Buffer.from(JSON.stringify({ type: 'session.start', data: { sessionId: 'cop-1' } }) + '\n')
-			);
-			// Emit a final assistant.message (no toolRequests + non-empty content → result)
+			// Mirror real Copilot CLI batch-mode stdout: no session.start event;
+			// the sessionId arrives only on the terminal `result` line.
 			mockStdout.emit(
 				'data',
 				Buffer.from(
 					JSON.stringify({
 						type: 'assistant.message',
-						sessionId: 'cop-1',
 						data: { content: 'Final answer from copilot', toolRequests: [] },
 					}) + '\n'
 				)
+			);
+			mockStdout.emit(
+				'data',
+				Buffer.from(JSON.stringify({ type: 'result', sessionId: 'cop-1', exitCode: 0 }) + '\n')
 			);
 			await new Promise((resolve) => setTimeout(resolve, 0));
 			mockChild.emit('close', 0);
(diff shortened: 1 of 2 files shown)


pedramamini:

I went through the change alongside the CodeRabbit and Greptile passes, and I'm happy with it.

- The root cause is nailed: copilot-cli's batch mode never emits `session.start`, so the `init`-only capture in `spawnJsonLineAgent` left `sessionId` undefined and `SendResult.sessionId` serialized as `null`.
- Routing through `parser.extractSessionId(event)` is the right call. I confirmed every adapter (claude-code, codex, opencode, factory-droid, copilot-cli) implements it, and the copilot parser correctly falls back to `raw.sessionId` / `raw.data.sessionId`, so the terminal `result` event is now picked up.
- The `!sessionId` guard preserves first-wins semantics, so codex/opencode (which populate `sessionId` on earlier events) stay unaffected.
- Solid test coverage: the updated copilot spawner test mirrors real `1.0.56` output, and the new `--resume=<id>` round-trip regression locks the fix in place.

No merge conflicts and clean against `rc`. Approving. ✅

PR #1070, opened 2026-06-02
fix(window): prevent main window from restoring off-screen

jSydorowicz21:
## Problem

On Windows, launching the app could spawn the main window invisibly: it shows on the taskbar but nothing is on screen.

Root cause is window-state restoration. A minimized window on Windows reports bounds of (-32000, -32000). The window manager saved those bounds on close and restored them verbatim on the next launch, with no check that the coordinates land on a visible display. The result is a window positioned far off-screen. The same thing happens when a monitor that held the window gets unplugged and the saved coordinates point at a display that no longer exists.

## Fix

- Add `resolveVisibleWindowPosition()`. Before creating the window, it checks whether the saved title-bar position falls inside any connected display's work area. If not, it drops the saved x/y and lets Electron center the window on the primary display.
- Stop `saveWindowState()` from persisting bounds while the window is minimized, so the -32000 sentinel never gets written in the first place. It already skipped maximized/fullscreen; minimized is the same kind of non-normal state.

## Tests

Added coverage in `window-manager.test.ts` (electron mock now provides `screen` and `isMinimized`):

- on-screen saved coordinates are restored as-is
- off-screen saved coordinates are dropped so the window spawns centered
- bounds are not persisted while the window is minimized

All window-manager tests pass. Repo-wide type check and ESLint are clean.

## Manual recovery

For anyone already stuck with an invisible window, deleting or resetting `maestro-window-state.json` in the app's userData directory brings it back. This change prevents it from recurring.

## Summary by CodeRabbit

* **Bug Fixes**
  * Windows restore to saved positions when valid; otherwise they open centered to avoid off-screen placement.
  * Window bounds are no longer saved while the window is minimized, preventing restoration to invalid coordinates.

* **Tests**
  * Added tests covering coordinate restoration behavior, off-screen handling, and ensuring minimized windows do not persist invalid bounds.

### DIFF
--- a/src/__tests__/main/app-lifecycle/window-manager.test.ts
+++ b/src/__tests__/main/app-lifecycle/window-manager.test.ts
@@ -45,6 +45,7 @@ const mockWindowInstance = {
 	setFullScreen: vi.fn(),
 	isMaximized: vi.fn().mockReturnValue(false),
 	isFullScreen: vi.fn().mockReturnValue(false),
+	isMinimized: vi.fn().mockReturnValue(false),
 	getBounds: vi.fn().mockReturnValue({ x: 100, y: 100, width: 1200, height: 800 }),
 	webContents: mockWebContents,
 	on: vi.fn((event: string, handler: () => void) => {
@@ -63,6 +64,7 @@ class MockBrowserWindow {
 	setFullScreen = mockWindowInstance.setFullScreen;
 	isMaximized = mockWindowInstance.isMaximized;
 	isFullScreen = mockWindowInstance.isFullScreen;
+	isMinimized = mockWindowInstance.isMinimized;
 	getBounds = mockWindowInstance.getBounds;
 	webContents = mockWindowInstance.webContents;
 	on = mockWindowInstance.on;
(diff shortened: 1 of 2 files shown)


pedramamini:
Thanks for the contribution, @jSydorowicz21! Nicely scoped fix for a genuinely annoying bug.

I reviewed the change alongside the Greptile and CodeRabbit feedback:

- **Greptile (magic number `16`):** already addressed - the code uses the named `TITLE_BAR_SAMPLE_Y` constant.
- **Greptile (test should also assert `width`/`height` aren't persisted while minimized):** already addressed - both assertions are present in the minimized test.
- **CodeRabbit (`saveWindowState` swallows errors):** this lands on a pre-existing `catch` block that your PR doesn't introduce (CodeRabbit flagged it as outside the diff range), and it's intentionally documented for non-critical ENFILE/ENOSPC failures during window close. Tightening it is reasonable on its own, but it's out of scope for this fix, so no need to touch it here.

The core fix is correct and well-tested: validating the saved title-bar position against the connected displays' work areas before restoring, and skipping bounds persistence while minimized so the `-32000` sentinel never gets written in the first place. The three new tests cover restore-as-is, drop-off-screen, and no-persist-while-minimized.

No merge conflicts, clean against `rc`. Approving. 🎉

PR #1072, opened 2026-06-03
fix(renderer): guard window.maestro accesses in useAppInitialization

ngothanhthien:
## Summary

Fixes the loading-screen error: `Error: Cannot read properties of undefined (reading 'git')` shown beneath the splash progress bar.

## Root cause

`useAppInitialization.ts` had several `useEffect` hooks that accessed `window.maestro.*` directly (e.g. `window.maestro.git.checkGhCli()`, `window.maestro.power.getStatus()`, `window.maestro.settings.get(...)`, `window.maestro.updates.setAllowPrerelease(...)`, `window.maestro.updates.check(...)`, `window.maestro.leaderboard.sync(...)`).

If the preload contextBridge had not finished wiring up `window.maestro` by the time React mounted, the renderer would throw, and `splash.js` would render the error text on the splash screen — preventing the app from ever loading.

## Fix

Apply optional chaining (`?.`) to all `window.maestro` accesses inside `useAppInitialization.ts` so the app tolerates a brief window where the preload contextBridge has not yet exposed the API. The previous design assumed a strict ordering that is not always honored at startup.

## Verification

- 730/730 test files pass, 28 521/28 521 tests pass (`npm test`)
- `npm run format:check:all` → clean
- `npm run lint` → clean

## Files changed

- `src/renderer/hooks/ui/useAppInitialization.ts` (13 insertions, 11 deletions)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved app initialization to gracefully handle unavailable or missing system APIs and dependencies throughout the startup process. The application now prevents crashes when certain features are not accessible, ensuring stable and reliable operation across different environments, system configurations, and various setups without affecting core functionality or user experience.

### DIFF
--- a/src/renderer/hooks/ui/useAppInitialization.ts
+++ b/src/renderer/hooks/ui/useAppInitialization.ts
@@ -116,9 +116,9 @@ export function useAppInitialization(): AppInitializationReturn {
 
 	// --- GitHub CLI availability check ---
 	useEffect(() => {
-		window.maestro.git
-			.checkGhCli()
-			.then((status) => {
+		window.maestro?.git
+			?.checkGhCli()
+			?.then((status) => {
 				setGhCliAvailable(status.installed && status.authenticated);
 			})
 			.catch(() => {
@@ -136,9 +136,9 @@ export function useAppInitialization(): AppInitializationReturn {
 		if (suppressWindowsWarning) return;
 		if (windowsWarningShownRef.current) return;
 
-		window.maestro.power
-			.getStatus()
-			.then((status) => {
+		window.maestro?.power
+			?.getStatus()
+			?.then((status) => {
 				if (status.platform === 'win32') {
 					windowsWarningShownRef.current = true;
 					setWindowsWarningModalOpen(true);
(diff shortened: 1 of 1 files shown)


pedramamini:
Thanks for the contribution, @ngothanhthien! 🎉

I reviewed this alongside the automated feedback from CodeRabbit and Greptile, and I'm happy with it — **approved.** ✅

The fix is correctly scoped: applying optional chaining to the `window.maestro.*` accesses in `useAppInitialization.ts` is the right way to tolerate the brief window before the preload `contextBridge` finishes exposing `window.maestro`. CI is green (2/2) and there are no merge conflicts.

One note on the automated review, for the record: Greptile flagged (P1) that the three promise chains ending in a bare `.catch()` — e.g.

```js
window.maestro?.git
	?.checkGhCli()
	?.then(...)
	.catch(...);
```

would throw a fresh `TypeError` on the trailing `.catch()` when `window.maestro` is `undefined`. That's a **false positive**: optional-chaining short-circuiting extends to the *end of the entire chain expression*, so when `window.maestro` is nullish the whole expression — including the non-optional `.catch()` — evaluates to `undefined` without ever invoking `.catch()`. Confirmed empirically:

```js
const w = { maestro: undefined };
w.maestro?.git?.checkGhCli()?.then(() => {}).catch(() => {}); // => undefined, no throw
```

So the guard is complete as written; no change needed there.

Nice, surgical fix — thanks again! 🙏

jSydorowicz21:
Hey @ngothanhthien could you retarget this to RC? Our branching strategy is to let things cook in a release candidate before merging into mainline

PR #1076, opened 2026-06-04
Gemini CLI and Grok Build Provider Support

insidia-jonas:
## Summary by CodeRabbit

* **New Features**
  * Added beta support for Gemini CLI and Grok Build agents with full integration.
  * Introduced Wake-Up Call feature for group chats: schedule automated message sequences with pause/resume controls.
  * Enhanced group chat moderation with participant timeout detection (30-minute default) and stale-response filtering.
  * Added session storage support for Gemini CLI.

* **Improvements**
  * Improved SSH prompt delivery with agent-specific routing for large prompts.
  * Updated documentation to reflect expanded AI provider support and security auto-approval behavior.

### DIFF
--- a/scripts/cdp-drive.mjs
+++ b/scripts/cdp-drive.mjs
@@ -1,0 +1,131 @@
+// Faithful UI driver over CDP. Usage: node scripts/cdp-drive.mjs '<json actions>'
+// Actions (array, run in order):
+//   {a:"eval", expr:"...returns JSON"}              -> Runtime.evaluate, returns value
+//   {a:"clickSel", sel:"...", index:0, all:false}   -> real mouse click at element center
+//   {a:"focusSel", sel:"...", index:0}              -> focus element (returns rect)
+//   {a:"type", text:"..."}                           -> Input.insertText into focused node
+//   {a:"key", key:"Enter", mods:0}                   -> dispatch keyDown+keyUp via Input domain
+//   {a:"sleep", ms:500}
+// Prints a JSON array of per-action results.
+import WebSocket from 'ws';
+
+const PORT = process.env.MAESTRO_CDP_PORT || '12345';
+const actions = JSON.parse(process.argv[2] || '[]');
+
+const list = await (await fetch(`http://127.0.0.1:${PORT}/json/list`)).json();
+const page = list.find((t) => t.type === 'page' && t.webSocketDebuggerUrl);
+if (!page) {
+	console.error('no page target');
+	process.exit(1);
+}
+
+const ws = new WebSocket(page.webSocketDebuggerUrl, {
+	perMessageDeflate: false,
+	maxPayload: 200 * 1024 * 1024,
+});
+let id = 0;
+const pending = new Map();
+function send(method, params) {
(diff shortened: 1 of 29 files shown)


pedramamini:
Hi @insidia-jonas, thanks for the contribution here. This is a substantial body of work (Grok Build + Gemini CLI agents, the Group Chat Wake-Up sequencer, and the participant loop-guard), and it is great to see it land with tests and docs alongside. Really appreciate it.

A couple of things before this can merge:

### 1. Merge conflicts
This PR is currently in a conflicting state against `rc` (GitHub reports it as not mergeable). Could you please rebase onto the latest `rc` and resolve the conflicts so we get a clean merge? `rc` moves quickly, so the sooner the better.

### 2. Changes to address
Greptile and CodeRabbit both ran over this. The items I would like resolved:

- **Wake-up sequences are not stopped on delete / archive / stopAll** (flagged by both reviewers, highest priority). `groupChat:delete`, `groupChat:archive`, and `groupChat:stopAll` kill the moderator and participants but never call `stopWakeUpService(id)`, so an active sequence keeps firing against a chat that no longer exists (each step then throws "Group chat not found" and reschedules anyway). Please add `stopWakeUpService(id)` to those handlers, analogous to the `clearPendingParticipants` call in `stopAll`.
- **Silent error swallowing.** A few spots swallow errors: `WakeUpModal.handleStart` gives no user feedback when the start IPC fails, the gemini-cli session storage uses catch-alls, and wake-up-service pause cleanup drops failures. Per our error-handling convention (see `CLAUDE.md`), expected/recoverable errors should be handled explicitly and unexpected ones should bubble up to Sentry rather than be caught silently. A toast on the `handleStart` failure path would also help users (we have `notifyToast` for exactly this).
- **Gemini CLI session storage ignores SSH mode**, so remote history will not resolve for SSH-configured agents. Anything that reads or spawns per-agent needs SSH-remote awareness (see "SSH Remote Execution Awareness" in `CLAUDE.md`).
- **`GroupChatData` is missing `wakeUpConfig`** in `src/renderer/global.d.ts`, so the renderer type contract is stale versus the shared `GroupChat` type.
- **Debug logging of user message content at `info` level** in `groupChat:sendToModerator` puts raw chat text into production logs. Please drop those to `debug` or redact them.

A few smaller nits (narrowing the JSON-parse catch blocks in the new parsers, the registry/factory parity test passing vacuously, and docs grammar / code-fence language) are in the inline comments. Up to you on those.

### One correction
CodeRabbit flagged `src/main/agents/capabilities.ts` and `src/main/agents/definitions.ts` as "Critical: invalid placeholder content instead of actual code." I checked both files at your head commit and they contain real, complete code (369 and 699 lines respectively). That looks like a false positive: the literal word "PLACEHOLDER" appears in a comment, which probably tripped the heuristic. You can safely disregard those two Critical items.

Thanks again. Ping me once the rebase and the items above are in, and I will take another pass.

jSydorowicz21:
Hey @insidia-jonas this pr is really big and touches a lot of different components, can you try and break it into a couple smaller prs so we can control the blast radius a little better

PR #1078, opened 2026-06-06
fix(codex): render reasoning summaries from both CLI output shapes

jSydorowicz21:
@-

## Bug Fixes
* Enhanced parser handling of reasoning content from Codex responses with support for multiple data formats and structures.
* Improved event streaming for reasoning summaries with proper text formatting and robust fallback behavior when content is unavailable, encrypted, or blank.
* Better compatibility with both legacy and modern reasoning data formats.

### DIFF
--- a/src/__tests__/main/parsers/codex-output-parser.test.ts
+++ b/src/__tests__/main/parsers/codex-output-parser.test.ts
@@ -1238,13 +1238,62 @@ describe('CodexOutputParser', () => {
 		});
 
 		describe('reasoning type', () => {
-			it('should parse reasoning as system event', () => {
+			it('should parse string[] summary as partial text with isReasoning', () => {
 				const line = JSON.stringify({
 					type: 'response_item',
 					payload: {
 						type: 'reasoning',
 						summary: ['step 1', 'step 2'],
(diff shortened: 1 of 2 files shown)


pedramamini:
Thanks for the contribution, @jSydorowicz21! 🙏

I reviewed the change along with the CodeRabbit and Greptile passes, and this looks great:

- **Tightly scoped** to the `reasoning` branch of `transformResponseItem`, with no impact on other event types.
- **Both CLI shapes handled** cleanly via the new `extractReasoningSummaryText` helper: the legacy `string[]` (app-server) form and the newer `Array<{ type: 'summary_text'; text }>` (Responses API) form.
- **Solid guarding** for `unknown` input, blank-string filtering, and the `encrypted_content`-only fallback that preserves the previous `system`-event behavior.
- **Reuses the existing `formatReasoningText` helper** rather than re-rolling formatting, consistent with the adjacent streaming path.
- **Four targeted tests** cover the string[], object-array, encrypted-only, and blank-string cases.

CodeRabbit reported no actionable comments and Greptile scored it 5/5 "safe to merge." No merge conflicts against `rc`. Approving. ✅

PR #1087, opened 2026-06-09
fix(codefence): stop flicker and preserve manual language selection during streaming

### LINKED ISSUE #1080: Bug: SnippetRender flickers and reverts manual type selection in chat output
## Summary
SnippetRender flickers and reverts manual type selection in chat output

## Environment
- Maestro version: 0.16.19-RC

## Steps to Reproduce
1. Open chat output containing a SnippetRender. 2. Observe constant flickering/blinking. 3. Manually switch the snippet's type/language. 4. After a few moments, it reverts to the auto-detected default.

## Expected Behavior
SnippetRender in chat output renders stably with no flicker/blinking, and a manually selected snippet type/language persists instead of reverting.

## Actual Behavior
In chat output, SnippetRender components flicker/blink constantly (always, not only during streaming). Separately, after manually switching the snippet's type/language, it reverts to the auto-detected default after a few moments.

## Additional Context
Both symptoms likely share a root cause: repeated re-renders re-triggering auto-detection, which both causes the visible flicker and overwrites the manual type selection. NOTE: User raised a separate, possibly already-submitted SnippetRender issue about broken/unnatural line breaks when copying rendered snippet text into WhatsApp or other apps — that should be tracked as its own issue, not merged here.

## Support Package
[maestro-debug-package.zip](https://raw.githubusercontent.com/felipeggv/maestro-feedback-attachments/main/feedback/1780907254213-maestro-debug-2026-06-08T08-27-33.zip)

jSydorowicz21:
Fixes #1080

## Problem
In chat output, code snippets flickered constantly and a manually-selected language reverted to the auto-detected default after a few moments. Root cause: the language auto-detection effect depended on `code`, which changes on every streamed character, so it re-ran constantly (flicker) and an in-flight async detection could overwrite the user's manual pick.

## Fix
`src/renderer/components/CodeFence/CodeFence.tsx`:
- Debounce the `code` value feeding detection via the existing `useDebouncedValue` hook (150ms); detection effect now depends on `debouncedCode`, not live `code`. Highlighting still tracks live `code`.
- Re-check the manual-override guard (`userOverrodeRef`) after every await so a completed detection cannot clobber a manual selection.
- Skip redundant `setResolvedLang` when the value is unchanged, preventing needless highlight re-runs.

## Tests
- Added `CodeFence.test.tsx`: manual pick persists across streaming updates; untagged fence still auto-detects; detection does not thrash during streaming (tightened bounds that fail if debouncing regresses).
- Full vitest suite on this branch: identical failed-file set to pristine `rc` baseline (zero regression).

## Notes
- Not covered by automated tests: in-app confirmation with real Shiki streaming + the language picker. Recommended before release.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved code highlighting performance during streaming by debouncing detection
  * Manual language selection now persists correctly across rapid streaming updates
  * Reduced unnecessary re-highlighting and visual churn

* **Tests**
  * Added comprehensive tests covering language auto-detection, user selection persistence, and debouncing behavior

### DIFF
--- a/src/__tests__/renderer/components/CodeFence.test.tsx
+++ b/src/__tests__/renderer/components/CodeFence.test.tsx
@@ -148,11 +148,19 @@ describe('CodeFence', () => {
 				await Promise.resolve();
 			});
 
-			// Detection fired far fewer times than the number of streamed updates
-			// (it would be ~6 without debouncing). The exact count depends on the
-			// async detection chain, so we only assert it stayed bounded.
-			expect(callsWhileStreaming).toBeLessThanOrEqual(2);
-			expect(mockHighlightAuto.mock.calls.length).toBeLessThanOrEqual(3);
+			// With debouncing intact, detection does NOT fire per streamed update.
+			// The 5 mid-stream rerenders all land inside the 150ms debounce window
+			// (only 20ms is advanced between each), so `debouncedCode` never changes
+			// and the detection effect never re-runs. At most the initial mount
+			// detection is in flight here. Without debouncing this would be ~5-6
+			// (one detection per streamed character), so this bound is what catches
+			// a broken/removed debounce.
+			expect(callsWhileStreaming).toBeLessThanOrEqual(1);
+			// After the debounce settles, detection runs exactly once more against
+			// the final snapshot. Total is therefore the single mount detection plus
+			// that one settling detection = 2. (It would be ~6 if detection fired
+			// per character.)
+			expect(mockHighlightAuto.mock.calls.length).toBeLessThanOrEqual(2);
 		} finally {
 			vi.useRealTimers();
 		}
--- a/src/__tests__/renderer/components/CodeFence.test.tsx
+++ b/src/__tests__/renderer/components/CodeFence.test.tsx
@@ -1,0 +1,160 @@
+import React from 'react';
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { render, screen, fireEvent, waitFor, act } from '@testing-library/react';
+import { CodeFence } from '../../../renderer/components/CodeFence/CodeFence';
+import { mockTheme } from '../../helpers/mockTheme';
+
+// Mock Shiki so CodeFence's async highlighting doesn't hit the real library.
+// The highlight effect is exercised indirectly; we only assert on the resolved
+// language (the `data-language` attribute), which is set before/around
+// highlighting completes.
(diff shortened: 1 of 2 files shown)


pedramamini:
Thanks for the contribution, @jSydorowicz21 - this is a clean, well-reasoned fix and the inline comments make the intent easy to follow. The production logic in `CodeFence.tsx` looks correct to me: debouncing only the detection input while highlighting still tracks live `code`, re-checking `userOverrodeRef` after each await, and the `applyLang` no-op guard all fit together nicely. Both Greptile and CodeRabbit agree the core change is sound.

One thing I'd like tightened before approving, picking up Greptile's P2 on the test:

In the "keeps a manually picked language ..." test, the `await new Promise((resolve) => setTimeout(resolve, 250))` is a fixed wall-clock sleep used to prove the override is NOT clobbered. Because it's a negative assertion, its reliability depends on enough real time actually elapsing for a stray detection to run. On a loaded CI runner the process can be suspended mid-test, so 250ms of wall time doesn't guarantee the debounce (150ms) plus detection have had a chance to fire, and the assertion can pass vacuously, which would quietly weaken the regression gate you deliberately built.

Could you flush the debounce deterministically instead? Either drive it with fake timers and `vi.advanceTimersByTime` the way the third test already does, or drop the fixed sleep and lean on `waitFor` after the rerenders. That keeps the test a real gate against a debounce regression.

Everything else looks good, and there are no merge conflicts. Once that test is hardened I'm happy to approve. Thanks again for the careful work here!

jSydorowicz21:
Thanks @pedramamini - agreed the fixed wall-clock sleep was a weak gate. Pushed `790a1cbd4`:

- Replaced the `setTimeout(resolve, 250)` in the manual-override persistence test with deterministic fake-timer flushing (`vi.advanceTimersByTime` past the debounce + microtask drain), mirroring the third test's pattern, so the debounce and async detection are guaranteed to run before the assertion.
- The post-pick detection mock now returns a different language than the manual pick, so a clobber would be observable.
- Exported `DETECTION_DEBOUNCE_MS` from `CodeFence.tsx` so the test advances exactly past the real window rather than a hardcoded magic number (inert at runtime).

Verified it's a real gate: with both override guards (the early return and the `stale()` ref re-check) removed, the test fails with `expected 'javascript' to be 'rust'`; restored, it passes. Green across 3 consecutive runs; tsc clean.

jSydorowicz21:
@greptile re-review

Issue #1065, opened 2026-05-31
Right-click context menu does not work inside embedded browser tabs


jeffscottward:
## Summary

Right-clicking inside an embedded Maestro browser tab does not show a context menu.

## Expected behavior

Browser tabs should support normal browser context-menu behavior, including at minimum:

- right-click in text/password inputs: Cut / Copy / Paste / Select All
- right-click selected text: Copy
- right-click links: Copy Link / Open Link Externally or in browser tab
- right-click images: Copy Image / Copy Image Address / Save where supported
- right-click page background: Back / Forward / Reload where appropriate

This is especially important for login workflows and pairs with #1063, where pasting into browser-tab form fields is also failing.

## Actual behavior

Right-clicking inside the embedded browser tab does not appear to open any menu.

## Repro

1. Open a Maestro internal browser tab.
2. Navigate to any page with a form field, link, selected text, or image.
3. Right-click inside the page content.

Expected: a browser context menu appears.
Actual: no useful context menu appears.

## Investigation

I checked existing issues and PRs first and did not find a direct duplicate or an open PR clearly addressing browser-tab right-click/context-menu support.

Related items checked:

- #1063 covers failing paste into browser-tab form fields.
- #785 introduced browser tabs into the unified tab strip and is already included in `v0.16.18-RC`.
- #746 / #751 are about right-click context menus for agent groups, not browser-tab webview content.

On `rc`, `src/main/app-lifecycle/window-manager.ts` registers a `mainWindow.webContents.on('context-menu')` handler for app-window editable fields and spellcheck suggestions. That handler only builds Cut/Copy/Paste/Select All for `params.isEditable`, and returns for non-editable page content.

Browser tabs use Electron `<webview>` guest content. The `did-attach-webview` setup wires keyboard forwarding and permission handling, but I did not find a `guestContents.on('context-menu')` handler for the webview itself. If the guest context-menu event is not bridged, right-clicks inside browser-tab pages will not get a menu.

## Suggested acceptance criteria

- Right-clicking inside browser-tab webview content opens an appropriate native context menu.
- Editable fields inside browser tabs expose Cut / Copy / Paste / Select All.
- Selected text, links, images, and page background get sensible browser actions where Electron exposes the relevant `context-menu` params.
- Context-menu actions operate on the guest webContents, not only the main app window.
- Add browser-tab E2E coverage or a targeted integration test for the webview context-menu path.

## Environment

- App: Maestro
[...]

PR #1088, opened 2026-06-09
fix(browser-tab): paste into webview form fields via guest.paste() on cmd/ctrl+v

### LINKED ISSUE #1063: Cannot paste clipboard text into embedded browser tab form fields
## Summary

Copy/paste into embedded browser tab form fields does not work. Example: copying a password from another app and trying to paste it into a login form inside a Maestro browser tab.

## Expected behavior

When a user focuses an input, password field, textarea, or contenteditable element inside an internal browser tab, `Cmd+V` should paste the current system clipboard text into that field, like a normal browser.

This should work for common login flows without requiring the user to open the page externally.

## Actual behavior

Pasting into the focused form field inside the browser tab does not insert the clipboard text.

## Repro

1. Open a Maestro internal browser tab.
2. Navigate to a page with a login form or any text/password input.
3. Copy text from outside Maestro, such as a password.
4. Click/focus the form field inside the browser tab.
5. Press `Cmd+V`.

Expected: clipboard text is inserted.
Actual: nothing is pasted.

## Investigation

I checked existing issues and PRs before filing and did not find a direct duplicate or an open PR that clearly fixes this.

Related browser-tab PRs checked:

- #785 ports browser tabs into the unified tab strip.
- #877 adds CLI `open-browser` support.
- #964 only truncates browser tab titles.

Those PRs are already included in the installed `v0.16.18-RC` build used for this report, so this does not look like a simple stale-app issue.

On `rc`, browser tabs are implemented with Electron `<webview>` in `src/renderer/components/MainPanel/BrowserTabView.tsx`. The keyboard interception code intentionally lets standard text-editing shortcuts pass through (`a/c/v/x/z`), so `Cmd+V` should not be captured as an app shortcut.

However, `src/main/app-lifecycle/window-manager.ts` only allows clipboard permissions for the app window and blocks permission requests from embedded browser-tab webviews. The current browser tab E2E coverage also asserts that `navigator.clipboard.readText()` from the webview is denied, but there does not appear to be coverage proving user-initiated native paste into a focused page input still works.

This may be a focus/webview edit-command issue, a permission issue, or a missing native paste bridge for webview contents.

## Suggested acceptance criteria

- `Cmd+V` pastes system clipboard text into focused `<input>`, `<input type="password">`, `<textarea>`, and contenteditable fields inside a browser tab.
- A browser-tab E2E test covers pasting into a local test page form field.
- App-level browser shortcuts still work.
- Avoid granting arbitrary page-level async clipboard read unless it is explicitly user-initiated and safe; the core requirement is normal browser paste behavior.

## Environment

- App: Maestro
- Installed build checked: `0.16.18-RC`
- Area: Internal browser tab / Electron webview clipboard handling
- Date reported: 2026-05-31

jSydorowicz21:
Fixes #1063

## Problem
Pasting (Cmd/Ctrl+V) clipboard text into a focused input/textarea/contenteditable inside an embedded browser tab (Electron `<webview>`) did nothing. The webview permission handler denies `clipboard-read` to webviews as a security boundary, so Chromium's native paste silently fails inside browser-tab form fields.

## Fix
`src/main/app-lifecycle/window-manager.ts`, in the webview `before-input-event` handler:
- Detect the paste chord `(meta || control) && !alt && !shift && key === 'v'`, call `event.preventDefault()`, and invoke `guest.paste()` - a privileged main-process Electron call that performs the paste without granting the page web-facing clipboard-read access.
- Removed `v` from the residual text-editing passthrough list (now handled explicitly) in both the native handler and the injected in-page capture listener, keeping the two layers consistent.

## Security
The `clipboard-read` permission denial is left intact. `guest.paste()` fires only in response to a real user keystroke and does not grant the page clipboard-read, so a malicious page cannot read the clipboard on demand. Mirrors how the right-click Paste menu item works.

## Tests
- Added tests asserting meta+v and control+v each call `guest.paste()` once with `preventDefault`, and that plain `v` / Cmd+Shift+V do not.
- Full vitest suite on this branch: identical failed-file set to pristine `rc` baseline (zero regression); `window-manager.test.ts` passes.

## Notes
- The `BrowserTabGuestContents` / `before-input-event` webview infrastructure already exists on `rc`, so this change is self-contained and merges independently (no dependency on the #1065/#1066 context-menu PR).
- Not covered by automated tests: real Electron webview paste on Windows and macOS. Recommended before release.

## Summary by CodeRabbit

* **Bug Fixes**
  * Cmd/Ctrl+V in embedded/guest browser views is now correctly intercepted to perform a privileged paste action; plain "v" and other shortcuts retain expected handling and forwarding.

* **Tests**
  * Added tests covering paste shortcut interception and fallback script behavior across modifier combinations, including ensured non-paste cases are handled/forwarded appropriately.

### DIFF
--- a/src/__tests__/main/app-lifecycle/window-manager.test.ts
+++ b/src/__tests__/main/app-lifecycle/window-manager.test.ts
@@ -1381,7 +1381,16 @@ describe('app-lifecycle/window-manager', () => {
 				shift: true,
 			});
 
+			// Cmd+Shift+V is not the plain paste chord, so the handler must NOT
+			// drive the privileged paste path. It is also not a text-editing
+			// passthrough, so it is consumed (preventDefault) and forwarded to the
+			// renderer as an app shortcut rather than reaching the page.
 			expect(mockGuestWebContents.paste).not.toHaveBeenCalled();
+			expect(shiftEvent.preventDefault).toHaveBeenCalled();
+			expect(mockWebContents.send).toHaveBeenCalledWith(
+				'browser-tab:shortcutKey',
+				expect.objectContaining({ key: 'v', shift: true })
+			);
 		});
 
 		// Electron 41 removed the legacy `'crashed'` event in favor of
--- a/src/__tests__/main/app-lifecycle/window-manager.test.ts
+++ b/src/__tests__/main/app-lifecycle/window-manager.test.ts
@@ -1263,6 +1263,119 @@ describe('app-lifecycle/window-manager', () => {
 			);
 		});
 
+		it('pastes into browser-tab form fields via guest.paste() on Cmd/Ctrl+V (#1063)', async () => {
+			// The permission handler denies `clipboard-read` to webviews, so
+			// Chromium's native Cmd/Ctrl+V silently fails inside browser-tab form
+			// fields. The before-input-event handler must intercept the paste chord
+			// and drive the privileged guest.paste() instead.
+			const { createWindowManager } = await import('../../../main/app-lifecycle/window-manager');
+
+			const windowManager = createWindowManager({
+				windowStateStore: mockWindowStateStore as unknown as Parameters<
+					typeof createWindowManager
+				>[0]['windowStateStore'],
+				isDevelopment: false,
+				preloadPath: '/path/to/preload.js',
+				rendererProductionUrl: 'app://app/index.html',
+				devServerUrl: 'http://localhost:5173',
+				useNativeTitleBar: false,
+				autoHideMenuBar: false,
+			});
+
+			windowManager.createWindow();
+
+			const attachHandler = webContentsEventHandlers.get('did-attach-webview');
+			attachHandler?.({} as any, mockGuestWebContents as any);
+
+			const beforeInputHandler = guestWebContentsEventHandlers.get('before-input-event');
+			expect(beforeInputHandler).toBeDefined();
+
+			// Cmd+V (macOS)
+			const metaEvent = { preventDefault: vi.fn() };
+			beforeInputHandler?.(metaEvent, {
+				type: 'keyDown',
+				key: 'v',
+				code: 'KeyV',
+				meta: true,
+				control: false,
+				alt: false,
+				shift: false,
+			});
+			expect(metaEvent.preventDefault).toHaveBeenCalled();
+			expect(mockGuestWebContents.paste).toHaveBeenCalledTimes(1);
+
+			// Ctrl+V (Windows/Linux)
+			const ctrlEvent = { preventDefault: vi.fn() };
+			beforeInputHandler?.(ctrlEvent, {
+				type: 'keyDown',
+				key: 'v',
+				code: 'KeyV',
+				meta: false,
+				control: true,
+				alt: false,
+				shift: false,
+			});
+			expect(ctrlEvent.preventDefault).toHaveBeenCalled();
+			expect(mockGuestWebContents.paste).toHaveBeenCalledTimes(2);
+
+			// The paste chord must NOT also be forwarded to the renderer as an app
+			// shortcut - it is fully consumed here.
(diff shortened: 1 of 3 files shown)


pedramamini:
@jSydorowicz21 Thanks for the contribution, and for the thorough writeup and tests! The core approach is solid: routing Cmd/Ctrl+V through the privileged `guest.paste()` while leaving the `clipboard-read` denial intact is the right way to fix #1063 without opening a clipboard-read hole, and mirroring the right-click Paste path is a nice touch.

One change before this is ready to merge:

**Align the third allowlist in `BrowserTabView.tsx` (CodeRabbit's Major finding is valid).**

This PR removes `v` from the `'acvxz'` text-editing allowlist in both spots in `window-manager.ts` (the native `before-input-event` handler and the main-process-injected bubble-phase fallback). But there is a third copy of the same allowlist that the PR doesn't touch:

- `src/renderer/components/MainPanel/BrowserTabView.tsx:393` - `var te=hasMod&&!hasAlt&&!e.shiftKey&&'acvxz'.indexOf(k)!==-1;`

That one is the renderer-injected *capture-phase* keydown listener on the same guest webview (the comment at lines 376-384 references "the main-process-injected bubble-phase one"). It still treats Cmd/Ctrl+V as page-native, which now disagrees with the native handler and the bubble-phase fallback you did update. This is the exact inconsistency your own test guards against in the bubble-phase script (`expect(injectedScript).not.toContain("'acvxz'.indexOf(k)")`) with the rationale "the page-level fallback must also exclude V... Otherwise it can race the privileged paste path." The capture-phase listener actually fires before the bubble-phase one, so it deserves the same treatment.

Could you update `BrowserTabView.tsx:393` to `'acxz'` and adjust the comment block at lines 381-384 (it currently lists `a/c/v/x/z` as keeping native behavior, and should match the main-process explanation that Cmd/Ctrl+V is now handled separately, alongside Cmd/Ctrl+F)?

Minor / optional (not blocking):

- **Greptile (P2):** consider adding a `Ctrl+Alt+V` (AltGr+V) case to the "does not hijack" test, so the `!input.alt` guard stays explicit if the modifier logic is ever refactored.
- **CodeRabbit (nitpick):** the Cmd+Shift+V case in the "does not hijack" test only asserts `guest.paste()` isn't called. To match the test name, you could also assert `preventDefault` was not called and the chord wasn't forwarded via `browser-tab:shortcutKey` - or narrow the assertion's intent in the test name.

No merge conflicts, so nothing to rebase. Once the `BrowserTabView.tsx` allowlist is aligned I'm happy to approve. Thanks again!

jSydorowicz21:
Thanks @pedramamini - good catch on the third allowlist. Pushed `2e1b36a49`:

- `src/renderer/components/MainPanel/BrowserTabView.tsx`: the renderer capture-phase keydown listener's allowlist is now `'acxz'` (was `'acvxz'`), matching the native `before-input-event` handler and the bubble-phase fallback. Since the capture-phase listener fires first, this was the racing copy. Updated the adjacent comment to note that both Cmd/Ctrl+F and Cmd/Ctrl+V are intentionally excluded (F reaches the find bar, V is handled via the privileged `guest.paste()` path).
- Confirmed via grep that no real allowlist uses `'acvxz'` anymore anywhere in `src/`; the only remaining occurrence is the negative test assertion that guards against it.
- Tightened the Cmd+Shift+V test: it now asserts paste is NOT driven while the chord is still consumed (preventDefault called + forwarded as `browser-tab:shortcutKey`), matching actual handler behavior.

tsc clean on touched files; `window-manager.test.ts` + `BrowserTabView.test.tsx` green (64/64). The optional Ctrl+Alt+V (AltGr) case I left untested - the `!input.alt` guard already excludes it, and adding a synthetic AltGr keydown test would mostly assert the guard tautologically; happy to add it if you'd prefer.

jSydorowicz21:
@greptile re-review

PR #1107, opened 2026-06-19
fix(cue): scan watched file on manual task.pending trigger

scriptease:
A manual `maestro-cli cue trigger` of a task.pending subscription built a bare event with no task payload, so the prompt's {{CUE_TASK_COUNT}} and {{CUE_TASK_LIST}} fell back to "0"/"" and the run saw no work. Only the polling scanner populated those fields, so the documented manual fallback was broken.

Extract `buildTaskPendingPayload` and add `scanTaskFilesOnce` in the task scanner (hash-free live scan), and have `CueEngine.triggerSubscription` scan the sub's watch glob against the owner's projectRoot before dispatch, merging the first matching file's task fields into the event payload.

## Summary by CodeRabbit

* **New Features**
  * Manual task triggering now performs a one-time scan of the watched project files for pending tasks and enriches the event payload with task metadata (count, list, path) when available.
  * When no pending tasks are found, the event payload is still dispatched with the manual indicator, omitting task metadata.
* **Bug Fixes**
  * If the scan fails, manual triggers still dispatch the bare manual payload and log a warning instead of breaking the workflow.

### DIFF
--- a/src/__tests__/main/cue/cue-engine.test.ts
+++ b/src/__tests__/main/cue/cue-engine.test.ts
@@ -1624,6 +1624,44 @@ describe('CueEngine', () => {
 
 			engine.stop();
 		});
+
+		it('manual trigger dispatches even when the scan throws (unreadable projectRoot)', () => {
+			const config = createMockConfig({
+				subscriptions: [
+					{
+						name: 'task-queue',
+						event: 'task.pending',
+						enabled: true,
+						prompt: 'process tasks',
+						watch: 'tasks/**/*.md',
+					},
+				],
+			});
+			mockLoadCueConfig.mockReturnValue(config);
+			// walkDir re-throws when projectRoot is unreadable; the engine must
+			// degrade gracefully rather than abort the trigger.
+			mockScanTaskFilesOnce.mockImplementation(() => {
+				throw new Error('ENOENT: projectRoot gone');
+			});
+
+			const deps = createMockDeps();
+			const engine = new CueEngine(deps);
+			engine.start();
+
+			const result = engine.triggerSubscription('task-queue');
+
+			// Still dispatched, with the bare manual payload (no task fields).
+			expect(result).toBe(true);
+			const callArgs = (deps.onCueRun as ReturnType<typeof vi.fn>).mock.calls[0][0];
+			expect(callArgs.event.payload).toMatchObject({ manual: true });
+			expect(callArgs.event.payload).not.toHaveProperty('taskCount');
+			expect(deps.onLog).toHaveBeenCalledWith(
+				'warn',
+				expect.stringContaining('manual trigger scan failed')
+			);
+
+			engine.stop();
+		});
 	});
 
 	describe('getStatus', () => {
--- a/src/__tests__/main/cue/cue-engine.test.ts
+++ b/src/__tests__/main/cue/cue-engine.test.ts
@@ -46,8 +46,12 @@ vi.mock('../../../main/cue/cue-github-poller', () => ({
 
 // Mock the task scanner
 const mockCreateCueTaskScanner = vi.fn<(config: unknown) => () => void>();
+const mockScanTaskFilesOnce =
+	vi.fn<(watchGlob: string, projectRoot: string) => Array<Record<string, unknown>>>();
 vi.mock('../../../main/cue/cue-task-scanner', () => ({
 	createCueTaskScanner: (...args: unknown[]) => mockCreateCueTaskScanner(args[0]),
+	scanTaskFilesOnce: (...args: unknown[]) =>
+		mockScanTaskFilesOnce(args[0] as string, args[1] as string),
 }));
 
 // Mock the database
@@ -113,6 +117,7 @@ describe('CueEngine', () => {
 
 		yamlWatcherCleanup = vi.fn();
 		mockWatchCueYaml.mockReturnValue(yamlWatcherCleanup);
+		mockScanTaskFilesOnce.mockReturnValue([]);
 
 		// Default loadCueConfigDetailed: derive from loadCueConfig for tests that
 		// only configure mockLoadCueConfig. Tests that need to inject warnings or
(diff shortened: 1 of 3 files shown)


pedramamini:
Thanks for the fix, @scriptease! Great catch on the broken manual `cue trigger` fallback for `task.pending`, and extracting `buildTaskPendingPayload` so the polling and manual paths produce identical payloads is a clean way to keep them in sync. The two new tests (populated payload + empty-scan) cover the behavior well.

One change before this is good to merge:

### Guard the one-shot scan in `triggerSubscription` (`src/main/cue/cue-engine.ts`)

`scanTaskFilesOnce` calls `walkDir(projectRoot, projectRoot)`, and `walkDir` deliberately re-throws when the root itself is unreadable (`if (dir === root) throw err`). The new call inside the dispatch loop isn't wrapped, and `triggerSubscription` has no surrounding try/catch, so if a session's `projectRoot` is deleted, unmounted, or hits a permission change, the exception escapes the whole method: every remaining sub in the dispatch group is skipped and the trigger throws instead of returning cleanly.

The polling scanner already handles this exact case gracefully - `doScan` wraps its `walkDir` call in try/catch and logs - so the manual path should degrade the same way. `captureException` is already imported in this file (and used a few lines down), so the fix matches existing usage:

```ts
try {
	const payloads = scanTaskFilesOnce(sub.watch, projectRoot);
	// ...existing payload handling...
} catch (err) {
	this.deps.onLog(
		'warn',
		`[CUE] "${sub.name}" manual trigger scan failed: ${err instanceof Error ? err.message : String(err)}`
	);
	void captureException(err, {
		operation: 'cue.triggerSubscription.scanTaskFilesOnce',
		subscriptionName: sub.name,
		ownerSessionId,
	});
}
// taskPayload stays {} so dispatch continues with { manual: true }
```

This also lines up with our error-handling guideline of handling expected/recoverable errors explicitly rather than letting them abort the flow. Both Greptile and CodeRabbit flagged the same thing, and I confirmed it against the code.

### Minor, your call

For a broad glob like `tasks/**/*.md` that matches several files, the manual trigger dispatches a single event from `payloads[0]`, whereas the poller fires one event per file. You already log a warning when more than one file matches, which is a reasonable choice for a one-shot trigger, but I'm flagging the asymmetry in case you'd rather fan out to match the poller. Not a blocker.

Once the scan is guarded I'm happy to approve. Thanks again for the contribution!

pedramamini:
@scriptease this got a positive review but has since gone into conflict with `rc`. Could you rebase onto the latest `rc` and re-push? Would like to merge the manual `task.pending` scan fix right after.

jSydorowicz21:
Current `rc` already contains a later implementation of this behavior in `e559458e2` through #1151/#1152, and this branch now conflicts in the engine, scanner, and test files. I think this PR should close as superseded rather than be rebased.

One follow-up is still needed on current `rc`: Windows directory walking returns paths such as `tasks\\alpha.md`, but Cue watches are slash-delimited, for example `tasks/**/*.md`. Passing the backslash path directly to `picomatch` returns no payloads, so manual triggering falls back to `CUE_TASK_COUNT=0`. I reproduced that with real temporary files. Normalize the relative path to POSIX separators before matching in both manual and polling scans, and retain the original path for filesystem access.

pedramamini:
@scriptease apologies - my rebase request above is now the wrong ask, and I want to correct it before you spend time on it.

I checked current `rc` and it already contains `e559458e2` ("fix(cue): scan watched files on manual task.pending trigger", via #1151/#1152), which landed the same fix this PR makes: `buildTaskPendingPayload` extracted in `cue-task-scanner.ts` and consumed by `triggerSubscription` in `cue-engine.ts`, with the one-shot scan guarded the way I asked for in my earlier review. This branch now conflicts in all three files it touches, and a rebase would essentially be re-deriving what is already on `rc`.

So I think the right move is to **close this as superseded** rather than rebase. That is not a reflection on the work - your diagnosis was correct and your `buildTaskPendingPayload` extraction is the same shape as what shipped. Thank you for finding and fixing the broken manual fallback; it is a genuinely useful catch and it made it into the product.

If you would like to keep going on this area, there is a real open bug on current `rc` that @jSydorowicz21 spotted and I verified against the code:

`walkDir` builds relative paths with `path.relative(root, fullPath)` (`src/main/cue/cue-task-scanner.ts:108`), which returns backslash-separated paths on Windows (`tasks\alpha.md`). Cue watch globs are slash-delimited (`tasks/**/*.md`), so `picomatch` never matches and the manual trigger falls back to `CUE_TASK_COUNT=0` - the exact symptom this PR set out to fix, just on Windows. Both call sites are affected: `scanTaskFilesNow` (line 138) and `doScan` (line 187).

The fix is to normalize to POSIX separators for matching while keeping the original path for filesystem access, e.g. match on `relPath.split(path.sep).join('/')` and continue to use `path.resolve(projectRoot, relPath)` for reads. Worth a test that exercises a nested path, since our CI runs a `windows-latest` leg that would catch a regression here.

Entirely your call whether you want to pick that up - happy to review it if you do. Thanks again for this one.

pedramamini:
@scriptease friendly follow-up on the note above, no action needed beyond a quick decision.

I re-checked current `rc` today and the picture is unchanged: `e559458e2` is on `rc` and ships the same fix this branch makes (`buildTaskPendingPayload` plus a guarded one-shot scan, now named `scanTaskFilesNow`, wired into `triggerSubscription`), and this branch is still `CONFLICTING` against `rc` in all three files it touches. So closing this as superseded is still the right call rather than rebasing into a re-derivation.

To say it plainly one more time: the diagnosis and the extraction here were correct, and the behavior you fixed is in the product. Thank you for that.

The Windows follow-up is also still open on `rc` if you want to pick it up as a fresh branch: `walkDir` returns backslash-separated relative paths on Windows (`src/main/cue/cue-task-scanner.ts:108`), and `scanTaskFilesNow` passes that straight to `picomatch` against a slash-delimited watch glob, so nothing matches and the manual trigger still falls back to `CUE_TASK_COUNT=0` there. Normalizing to POSIX separators for the match while keeping the original path for reads would close it out, ideally with a nested-path test since CI runs a `windows-latest` leg.

Happy either way. If you would rather not spend more time on it, just close this out and we will call it done.

pedramamini:
@scriptease checking in one last time on this one, and thank you again for the contribution - the diagnosis here was correct and the behavior you fixed is live in the product.

Re-verified against current `rc` today (`0e4f28632`):

- `e559458e2` is still an ancestor of `rc`, and `cue-task-scanner.ts` exports both `buildTaskPendingPayload` and the guarded one-shot `scanTaskFilesNow`, wired into `triggerSubscription` at `cue-engine.ts:1323`. That is the same fix this branch makes.
- This branch is still `CONFLICTING` / `DIRTY` against `rc` in all three files it touches, with no new commits since June 19.

So a rebase would be re-deriving what already shipped, and I am not going to ask you to spend time on that. The cleanest outcome is to close this as superseded. Happy to close it on your behalf if you would rather not think about it again - just say the word, or close it yourself whenever is convenient.

The Windows follow-up is also still open on `rc` if you ever want a fresh branch for it: `walkDir` pushes `path.relative(root, fullPath)` (`src/main/cue/cue-task-scanner.ts:108`), which is backslash-separated on Windows, and `scanTaskFilesNow` hands that straight to `picomatch` against a slash-delimited watch glob (`cue-task-scanner.ts:137`), so nothing matches and the manual trigger still reports `CUE_TASK_COUNT=0` there. Normalizing to POSIX separators for the match while keeping the original path for reads would close it out, ideally with a nested-path test since CI runs a `windows-latest` leg.

No approval label on this one, purely because it is superseded rather than because of anything wrong with the work. Thanks again.

Issue #995, opened 2026-05-14
Copilot CLI Agent crashes via Windows SSH session


insidia-jonas:
It seems that AI‑MAESTRO cannot use Copilot CLI when it is running on a Windows remote host, likely due to problems with the remote shell being used. I am attempting to communicate with the configured Copilot CLI agent, but the application crashes immediately when joining a single group chat.

chr1syy:
Can you provide a log or any errors that appear?
I'll try this out.

insidia-jonas:
Thanks Chris!

I don't have any logs just the agent exiting with error 1 in Maestro.

The interesting behavior is, on a remote Linux system, it works flawless with copilot cli

pedramamini:
Thanks for the report, and for the follow-up detail. The Linux-works / Windows-fails contrast was the key clue and it pinned this down exactly.

## Root cause

This is not specific to Copilot CLI or to group chat. Maestro's SSH remote execution is POSIX-only. Both command builders in `src/main/utils/ssh-command-builder.ts` hardcode a bash invocation on the remote host:

- `buildSshCommand()` wraps everything in `/bin/bash --norc --noprofile -c '...'`
- `buildSshCommandWithStdin()` runs `/bin/bash --norc --noprofile -s`

On a Windows remote, OpenSSH hands that to the default shell (cmd.exe or PowerShell), which has no `/bin/bash`. The process dies before your agent ever starts, which is the bare `exit 1` you saw with no logs. The surrounding script has the same assumption throughout: `export PATH=...`, `$HOME/...` path probing, `cd`, inline `VAR='val'` env prefixes, and `base64 -d` for image temp files.

Group chat just happened to be the first thing you tried; any agent spawn to that host would fail the same way.

## What has shipped so far

#1129 landed on `rc` (v0.18.4-RC). It does not add Windows support, but it converts that silent `exit 1` into an explicit message instead of leaving you guessing:

> SSH execution to Windows remote hosts is not yet supported (Maestro builds the remote command for a POSIX shell). See issue #995.

It detects the cmd.exe / PowerShell "is not recognized..." signatures. That is on the RC channel only, not stable yet.

## Why this is staying open

Real Windows remote support is bigger than the spawn path. These all generate POSIX shell for the remote host and would still fail even with a correct agent spawn:

- `src/main/utils/remote-fs.ts` (file tree, reads, stat/du/find)
- `src/main/utils/remote-git.ts` (git status/diff)
[...]

PR #1129, opened 2026-06-26
fix(ssh): actionable error when an SSH agent targets a Windows remote (#995)

jSydorowicz21:
## Summary

Addresses #995. Root cause (high confidence): Maestro's SSH command builders are POSIX-only (they hardcode `/bin/bash --norc --noprofile` and POSIX quoting), so any agent (including Copilot CLI) crashes with a cryptic `exit 1` when the SSH remote host is Windows.

This PR converts that cryptic crash into a clear, actionable failure.

## Changes

- Detect Windows-shell failure signatures (cmd.exe / PowerShell "is not recognized...", "the system cannot find the path specified") in the SSH error-pattern layer and surface: "SSH execution to Windows remote hosts is not yet supported ... See issue #995."
- POSIX (Linux/macOS) remotes are unaffected; the signatures are Windows-exclusive (a POSIX `command not found` still maps to its own message).
- 5 new unit tests.

## Scope / deferred

Full Windows-remote command generation (cmd.exe vs PowerShell selection, quoting of args/cwd/env, remote PATH and `copilot.exe` resolution) is intentionally deferred: it cannot be verified without a live Windows SSH host, and shipping a speculative command into a security-sensitive execution path would be a guess. This PR is the verifiable guardrail; the full fix can follow once a Windows SSH test target is available.

## Verification

- `npm run lint` (tsc x3): clean
- `vitest run error-patterns`: 130 tests pass (incl. 5 new)
- `eslint` + `prettier`: clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SSH error handling to give a clearer message when connecting to a Windows remote host that doesn’t support the expected shell environment.
  * These failures are now marked as non-recoverable and explain that the remote setup is not yet supported.
* **Tests**
  * Added regression coverage for Windows-specific SSH failures and checks to avoid false positives on normal POSIX remote errors.

### DIFF
--- a/src/__tests__/main/parsers/error-patterns.test.ts
+++ b/src/__tests__/main/parsers/error-patterns.test.ts
@@ -783,6 +783,59 @@ describe('error-patterns', () => {
 			});
 		});
 
+		describe('Windows remote host (issue #995)', () => {
+			// Maestro builds the remote command for a POSIX shell
+			// (/bin/bash --norc --noprofile). On a Windows SSH remote the default
+			// shell (cmd.exe or PowerShell) cannot run it and the agent dies with a
+			// bare exit 1. These cases prove the cryptic crash now maps to a clear,
+			// actionable message. Full Windows-remote support is deferred (needs a
+			// live Windows SSH host to verify the generated command).
+			it('should map cmd.exe "is not recognized as an internal or external command" to the actionable message', () => {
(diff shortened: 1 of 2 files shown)


jSydorowicz21:
@greptile re-review please

PR #1130, opened 2026-06-26
ci: add stale issue/PR automation (#574)

### LINKED ISSUE #574: Feature: enable stale issue automation
## Summary

We should enable stale issue automation for `RunMaestro/Maestro` so inactive issues are labeled, warned, and eventually closed if nobody follows up.

GitHub does not provide a built-in timeout for issues. In practice, most active OSS repos implement this with `actions/stale`.

## Recommendation

Adopt this default policy:

- Mark issues as stale after `60` days of inactivity
- Auto-close `14` days after being marked stale
- Exempt labels like `pinned` and `security`

This gives a total inactive lifetime of about `74` days, which is a reasonable middle ground for an actively maintained repo.

## Why

Benefits:

- Keeps the issue tracker from filling up with abandoned reports
- Gives reporters a clear warning before auto-close
- Lets maintainers focus on issues that still have active interest
- Follows a common pattern used by many fast-moving OSS repos

## Example Workflow

```yaml
name: Mark stale issues

on:
  schedule:
    - cron: "0 0 * * *"

jobs:
  stale:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/stale@v9
        with:
          days-before-stale: 60
          days-before-close: 14
          stale-issue-label: stale
          exempt-issue-labels: pinned,security
```

## Alternative policy ranges

Typical ranges used elsewhere:

- Fast-moving repos: `30` days stale, `7` days close
- Most OSS repos: `60` days stale, `7-14` days close
- Slower-moving repos: `90` days stale, `14` days close

If we want a conservative default, `60 + 14` seems like the right starting point.

## Optional follow-up

If we do not want to auto-close immediately, we could first:

- enable stale labeling only
- add a saved search/filter for stale issues
- then enable auto-close later once the team is comfortable with the behavior

jSydorowicz21:
## Summary

Closes #574. Adds a GitHub Actions stale workflow implementing the policy requested in the issue: mark issues/PRs stale after 60 days of inactivity and close 14 days later (about 74 days total), exempting `security` and the repo's active triage labels.

## Changes

- New `.github/workflows/stale.yml` using `actions/stale@v9`.
- Runs daily (`cron`) and via `workflow_dispatch`. Manual dispatch defaults to a dry run (`debug-only`) so maintainers can validate the policy without labeling/closing anything; the scheduled run is live.
- Idempotent pre-step (`gh label create stale --force`) so the `stale` label exists before the action tries to apply it (actions/stale applies but does not create it).
- Explicit minimal `permissions`, concurrency guard, draft-PR exemption, and `operations-per-run` cap.

## Notes

- `pinned` is included in the exempt list for forward compatibility (a no-op until a maintainer creates/applies that label).
- Independently reviewed by `codex exec` (read-only); the two findings (missing `stale` label, misleading dry-run comment) are addressed above.

## Summary by CodeRabbit

* **Chores**
  * Added automated stale-item handling to keep issues and pull requests organized.
  * Enabled a manual dry-run option for safe testing without making changes.
  * Configured the process to skip draft pull requests, respect exempt labels, and avoid overlapping runs.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -3,14 +3,25 @@ name: Mark stale issues and PRs
 # Implements the policy requested in issue #574:
 #   - Mark issues/PRs stale after 60 days of inactivity
 #   - Auto-close 14 days after being marked stale (≈74 days total)
-#   - Exempt "pinned" and "security" (and a few maintainer-controlled labels)
-# Runs daily; also exposes workflow_dispatch so maintainers can trigger a
-# dry-run / manual pass and verify behaviour before relying on the schedule.
+#   - Exempt "security" plus the repo's active triage labels.
+#     ("pinned" is included for forward compatibility; it is a no-op until a
+#     maintainer creates/applies a "pinned" label, and GitHub's pinned-issue UI
+#     state is intentionally NOT covered by exempt-*-labels.)
+#
+# Runs daily. Manual runs (workflow_dispatch) default to a DRY RUN: actions/stale
+# is invoked with debug-only, so it logs what it *would* mark/close without
+# touching anything - maintainers can validate the policy safely, then re-dispatch
+# with dry_run=false (or wait for the schedule) to apply it for real.
 
 on:
   schedule:
     - cron: '0 0 * * *'
   workflow_dispatch:
+    inputs:
+      dry_run:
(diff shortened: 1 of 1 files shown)


jSydorowicz21:
@greptile re-review please

PR #1136, opened 2026-06-27
feat(tabs): scroll tab strip with the mouse wheel

peterjmorgan:
## What

When the tab strip overflows, hovering over it and using the mouse wheel now pans the tabs horizontally (VS Code style). A plain vertical wheel (`deltaY`) is translated into horizontal scroll; trackpads and Shift+wheel emit `deltaX`, so the handler follows whichever axis the device reports.

Gated behind a new **Scroll tabs with the mouse wheel** toggle in **Settings -> Display** (`tabBarWheelScroll`, default on).

## Why

With a normal mouse (no horizontal wheel), there was no quick way to reach off-screen tabs without dragging or using keyboard shortcuts. This matches the tab-strip behavior users expect from VS Code and modern browsers.

## How

- `TabBar.tsx`: reads the new setting and attaches a native, non-passive `wheel` listener to the existing scroll container. It only acts when the strip actually overflows (`scrollWidth > clientWidth`), adjusts `scrollLeft`, and calls `preventDefault()` so the wheel doesn't also scroll an outer container.
  - The listener is attached natively with `{ passive: false }` because React's synthetic `onWheel` runs passively and cannot call `preventDefault()`.
- `settingsStore.ts`: new `tabBarWheelScroll` boolean (default `true`) + setter + load-from-disk patch, mirroring the existing `showBrowserTabDomain` tab setting.
- `useSettings.ts`: added field + setter to `UseSettingsReturn` (the hook spreads `...store`).
- `DisplayTab.tsx`: new toggle in the tab-display section.

## Testing

- `TabBar.test.tsx` + `DisplayTab.test.tsx`: 272 passed.
- ESLint clean on all four changed files.
- Type-check clean on all four changed files.

## Summary by CodeRabbit

* **New Features**
  * Added a “Scroll tabs with the mouse wheel” option to Display settings, persisted across sessions (default: on) for overflowed tabs.
  * Added a toggle switch to enable or disable this behavior.
* **Bug Fixes**
  * Improved tab-strip wheel handling by converting wheel movement into horizontal scrolling, preferring the dominant wheel axis, and preventing the wheel from scrolling surrounding content when enabled.

### DIFF
--- a/src/renderer/components/Settings/searchableSettings.ts
+++ b/src/renderer/components/Settings/searchableSettings.ts
@@ -788,7 +788,7 @@ export const DISPLAY_SETTINGS: SearchableSetting[] = [
 		tabLabel: 'Display',
 		label: 'Tab Options',
 		description:
-			'Show starred and file preview tabs when filtering by unread; Command+0 vs Command+9 last-tab shortcut; browser tab domain pill',
+			'Show starred and file preview tabs when filtering by unread; Command+0 vs Command+9 last-tab shortcut; browser tab domain pill; scroll tabs with the mouse wheel',
 		keywords: [
 			'tab',
 			'filter',
@@ -811,6 +811,13 @@ export const DISPLAY_SETTINGS: SearchableSetting[] = [
 			'hostname',
 			'url',
 			'pill',
+			'wheel',
+			'mouse wheel',
+			'scroll',
+			'scroll tabs',
+			'pan',
+			'tab bar',
+			'tab strip',
 		],
 	},
 	{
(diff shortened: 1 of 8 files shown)


pedramamini:
@peterjmorgan small one, but it is conflicting with `rc` now. Could you rebase onto the latest `rc` and re-push? Should be a quick resolve, then we can merge the tab-strip wheel scroll.

pedramamini:
Thanks! This has drifted and now conflicts with `rc`, and CI has not run. Could you rebase on the latest `rc` so the checks can run? Happy to merge once it is green.

peterjmorgan:
Rebased onto the latest `rc` (clean, no conflict remnants) and pushed. The rebase carried the Settings toggle into the new `DisplayTab/TabOptionsSection` structure from the display-tab refactor.

Also addressed the review feedback in a follow-up commit:
- Wheel events are only consumed when the strip can actually pan in that direction; at a clamped edge they fall through to the surrounding content.
- The wheel binding now goes through the shared `useEventListener()` hook, which gained an optional `passive` flag for non-passive listeners.
- `tabBarWheelScroll` hydration validates the persisted type, and the setting is registered in `SETTINGS_METADATA` and the searchable-settings entry for Tab Options.
- Toggle description reworded for accuracy, and an em dash removed per the writing style.

`TabBar`, `DisplayTab`, `searchableSettings`, and `useEventListener` suites pass locally (378 tests). Ready for CI.

PR #1146, opened 2026-06-29
perf(input): decouple composer draft from App to kill keyboard lag

pedramamini:
## Problem

Typing in the AI prompt (or terminal) lagged: characters appeared slower than typed. Root cause: the live draft text lived in `useState` inside `useInputHandlers`, which runs in `MaestroConsoleInner` (`App.tsx`). **Every keystroke re-rendered the entire app tree.**

## Fix

Move the live draft into a dedicated `useComposerInputStore` (zustand, two slices: `aiValue` / `terminalValue`).

- **`InputArea`** (already `React.memo`) is now the *sole* subscriber, so a keystroke re-renders only that leaf.
- **`useInputHandlers`** and the sub-hooks (`useInputProcessing`, `useInputKeyDown`) read the value non-reactively via a `getInputValue()` getter / `getState()`, so they no longer take a reactive `inputValue` dependency.
- The key trick for terminal tab-completion: a **gated selector** `useComposerInputStore((s) => tabCompletionOpen ? s.terminalValue : '')`. When the dropdown is closed it returns a stable `''`, so zustand's `Object.is` bail-out means **normal typing does not re-render the hook**. Live suggestions still work while the dropdown is open.

Net result: `App` no longer re-renders per keystroke; only the memoized composer leaf does.

## Preserved behavior

Per-tab and per-session draft persistence (flush on blur/submit/tab-switch/session-switch), the `liveDraftStore` mirror (now via a `store.subscribe` outside render), and paste/drop/replay handlers are all unchanged. The `inputValue` prop was pure pass-through (`App` → `useMainPanelProps` → `MainPanel` → `MainPanelContent` → `InputArea`); that chain is removed.

## API change

`useInputHandlers` no longer returns `inputValue` / `deferredInputValue` (read the store instead). `InputArea` no longer takes an `inputValue` prop. Sub-hook deps take `getInputValue: () => string` in place of `inputValue: string`.

## Testing

- New `composerInputStore` unit suite (setters, functional updaters, selectors, subscription).
- Migrated the input-subsystem test suites; **full pre-push suite green: 31,336 passed, 0 failed**.
- `tsc` (lint/main/cli) and ESLint clean.

## Suggested manual check

Worth a quick in-app type-test before merge: fast typing in the AI prompt, switching tabs/agents mid-draft (draft should persist per tab), terminal tab-completion, and the prompt-composer expand (Cmd-click / shortcut) seeding the current draft. A before/after Cmd+K performance trace (the new field-profiling feature) would quantify the win.

## Summary by CodeRabbit

* **New Features**
  * Composer input now maintains separate live drafts for AI and terminal modes, preserving the correct text when you switch.
  * Prompt Composer now initializes from the active session/group chat draft for the current input mode.
  * Session list “configure” behavior now depends on Maestro Cue availability.

* **Bug Fixes**
  * Improved live syncing so the editor reflects draft updates correctly based on the active mode.
  * Updated Prompt Composer header/session name separator to a consistent hyphen.

### DIFF
--- a/src/__tests__/renderer/components/AppModals-selfSourced.test.tsx
+++ b/src/__tests__/renderer/components/AppModals-selfSourced.test.tsx
@@ -12,6 +12,7 @@ import { render, screen, act } from '@testing-library/react';
 import { useSessionStore } from '../../../renderer/stores/sessionStore';
 import { useGroupChatStore } from '../../../renderer/stores/groupChatStore';
 import { useModalStore } from '../../../renderer/stores/modalStore';
+import { useComposerInputStore } from '../../../renderer/stores/composerInputStore';
 import type { Session, Shortcut, Group, GroupChat } from '../../../renderer/types';
 import { createMockSession as baseCreateMockSession } from '../../helpers/mockSession';
 
@@ -25,6 +26,10 @@ let capturedWorktreeProps: Record<string, unknown> = {};
 let capturedUtilityProps: Record<string, unknown> = {};
 let capturedGroupChatProps: Record<string, unknown> = {};
 let capturedAgentProps: Record<string, unknown> = {};
+
+const promptComposerModalMock = vi.hoisted(() => ({
+	props: undefined as Record<string, unknown> | undefined,
+}));
 
 // Mock ALL sub-components to capture props
 vi.mock('../../../renderer/components/AboutModal', () => ({ AboutModal: () => null }));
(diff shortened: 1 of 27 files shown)


pedramamini:
@reachrazamair please review when you get a chance. This touches the constitutional keyboard input path, so I'd value a careful look at the tab-switch/session-switch draft-flush logic and the gated-selector trick in `useInputHandlers`. Behavior should be identical; the win is App no longer re-rendering per keystroke. Manual in-app type-test recommended before merge (notes in the PR description).

reachrazamair:
@coderabbit review

PR #1154, opened 2026-07-01
fix(agent-store): route queued agent-discovered commands and isolate unknown-command errors

athal7:
Fixes two related bugs in OpenCode-backed (and other agent-discovered-command) sessions:

1. A valid agent-discovered slash command (e.g. one loaded from `.opencode/commands/*.md`) would incorrectly show "Unknown command" if it was typed while the agent/tab was busy and got queued. The queued-command re-match path in `agentStore.ts` never checked `session.agentCommands`, unlike the immediate-typing path.
2. The resulting "Unknown command" error log could render on the wrong tab and get silently concatenated (no separator) into a later, unrelated AI response, because it was logged without a `tabId` and used a log source that gets grouped with adjacent response text.

Changes:
- `agentStore.ts`: include `session.agentCommands` in the queued-command match chain; target the error log at `item.tabId`; fix the same tab-targeting bug in the catch-all error handler.
- `agentStore.ts` / `useRemoteHandlers.ts`: switch the "Unknown command" log to `source: 'error'`, which already has dedicated non-joining rendering.
- `TerminalOutput.tsx`: give `source: 'error'` its own flush boundary in the response-grouping logic so it can never be joined into surrounding text.

## Summary by CodeRabbit

- **New Features**
  - Queued commands can now use session-provided custom commands when they include a prompt, expanding which slash commands the app can run.

- **Bug Fixes**
  - Unknown commands now produce dedicated error log entries (not system logs) on the correct target tab.
  - AI terminal rendering now keeps error logs separate from surrounding AI response grouping.
  - Queued command failures now log to the queued item’s target tab (not the currently active tab), with a diagnostic message when the target tab is missing.

- **Tests**
  - Added/updated coverage for error log routing and rendering behavior.

### DIFF
--- a/src/__tests__/renderer/components/TerminalOutput.test.tsx
+++ b/src/__tests__/renderer/components/TerminalOutput.test.tsx
@@ -332,6 +332,32 @@ describe('TerminalOutput', () => {
 			const logItems = container.querySelectorAll('[data-log-index]');
 			expect(logItems.length).toBe(2);
 		});
+
+		it('keeps an error-source entry out of the surrounding response group', () => {
+			const logs: LogEntry[] = [
+				createLogEntry({ id: 'resp-1', text: 'Tail of an earlier response.', source: 'stdout' }),
+				createLogEntry({ id: 'err-1', text: 'Unknown command: /nonexistent', source: 'error' }),
+				createLogEntry({ id: 'resp-2', text: 'Start of a later response.', source: 'stdout' }),
+			];
+
+			const session = createDefaultSession({
+				tabs: [{ id: 'tab-1', agentSessionId: 'claude-123', logs, isUnread: false }],
+				activeTabId: 'tab-1',
+			});
+
+			const props = createDefaultProps({ session });
+			const { container } = render(<TerminalOutput {...props} />);
+
+			// The error entry must flush its own boundary rather than being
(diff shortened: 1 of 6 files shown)


pedramamini:
@athal7 Thanks for the contribution! 🙏

I went through the PR alongside the CodeRabbit and Greptile feedback. The one actionable finding both bots surfaced (the catch-all error handler in `processQueuedItem` still using `source: 'system'` and the now-stale `!activeTab` guard) was already resolved in your follow-up commit `a107f93`: the catch block now logs with `source: 'error'` and validates the resolved target tab before it emits a diagnostic, so spawn-failure messages get the same isolation treatment as the unknown-command path. Nice touch adding the orphaned-tab "error log dropped" test to cover that case.

The three core fixes all look correct and are well covered:
- `agentCommands` is now included in the queued-command match chain, so agent-discovered commands (e.g. `.opencode/commands/*.md`) no longer get a spurious "Unknown command".
- Error logs target `item.tabId` instead of the currently active tab.
- `source: 'error'` gets its own flush boundary in `TerminalOutput`, so it can't be concatenated into surrounding AI response text.

No merge conflicts, CI is green, and Greptile's remaining P1 note is stale (it reviewed the pre-fix commit). Looks great.

Approving. ✅

PR #1155, opened 2026-07-01
feat(agents): replace yolo mode with 3-way permission mode

athal7:
Replaces the binary yolo-mode/read-only toggle with `permissionMode: 'full' | 'standard' | 'readonly'`. `full` preserves current bypass behavior (renamed from yolo), `standard` is new and runs agents with no bypass flags so they use their own default permission model, `readonly` is the existing plan/exploration mode, unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added a three-state per-tab permission model (**Full Access**, **Standard**, **Read Only**) and updated the toolbar to cycle modes with matching labels/tooltips.
* **Bug Fixes**
  * Improved permission consistency across spawn paths (local, remote, queued, auto-run) by honoring an explicit `permissionMode` override.
  * Updated agent configurations so “Full Access” applies the correct full-permission args/env overrides while “Standard” avoids blanket elevated permissions.
* **Documentation**
  * Refreshed onboarding/help text to reflect the new permission terminology and default behavior.
* **Tests**
  * Expanded coverage for permission-mode precedence and full-access env override behavior.


pedramamini:
Thanks for this, @athal7 - the 3-way permission model is a genuinely nice cleanup over the binary yolo/read-only toggle, and the `buildAgentArgs` precedence logic plus the added test coverage are solid.

Before we merge, there's one blocking regression that both Greptile and CodeRabbit flagged and that I've confirmed against the code:

**`permissionMode` is never forwarded to `process.spawn` in the three main renderer spawn paths, so "Full Access" silently loses its bypass flags.**

This PR moves the bypass flags (`--dangerously-skip-permissions`, `--dangerously-bypass-approvals-and-sandbox`, etc.) out of `agent.args` and into `agent.fullAccessArgs`, which `buildAgentArgs` only injects when `isFullAccess` is true (`permissionMode === 'full'`, or legacy `yoloMode === true`). But these three spawn calls only pass `readOnlyMode: isReadOnly` and never forward `permissionMode` (or `yoloMode`):

- `src/renderer/hooks/input/useInputProcessing.ts` (spawn ~L1140, payload ~L1151)
- `src/renderer/stores/agentStore.ts` (spawn ~L423 and ~L521)
- `src/renderer/hooks/remote/useRemoteHandlers.ts` (spawn ~L509)

So inside the IPC handler both `config.permissionMode` and `config.yoloMode` are `undefined`, `isFullAccess` is never true, and every tab in Full Access mode, including all existing tabs where `permissionMode` is still `undefined` (the toolbar renders that as "Full Access"), spawns without the bypass flags it received before this PR. Users would hit unexpected permission prompts in what the UI calls Full Access.

The fix is to forward `permissionMode` in all three spawn payloads, e.g.:

```ts
permissionMode: freshActiveTab?.permissionMode ?? (freshActiveTab?.readOnlyMode ? 'readonly' : 'full'),
```

A few smaller items worth addressing while you're in here (non-blocking, your call):

- `src/main/ipc/handlers/process.ts` (~L461-477): the env-override branch derives full-access / read-only from the legacy `yoloMode` / `readOnlyMode` booleans directly rather than giving `permissionMode` precedence the way `buildAgentArgs` does, so a stale `yoloMode: true` could apply full-access env overrides under an explicit `standard` / `readonly`. Worth aligning the precedence.
- `src/main/cue/cue-spawn-builder.ts`: the Cue path passes `permissionMode: 'full'` for args but does not merge `agentDef.fullAccessEnvOverrides` into the spawn env the way `process:spawn` does, so env-driven agents (e.g. OpenCode) keep their standard env during Cue runs.
- `src/renderer/components/InputArea/components/ToolbarControls.tsx`: the `isReadOnlyMode` prop looks unused after the refactor and is safe to drop.

No merge conflicts on this branch, so once the spawn forwarding is in place this should be good to go. Thanks again for the contribution!

jSydorowicz21:
Hey @athal7 I'm going to have @pedramamini or @reachrazamair take a look at this one. If memory recalls correctly the reason why we use yolo mode is most clis don't expose a nice way to permission handle when spawning programatically instead of via TUI. That leaves us in a spot where if you hit something that's blocked and you want the agent to do it there's not a way to approve that. Please correct me if I'm wrong here guys

athal7:
this pr needs more work, putting back in draft while i work on it

athal7:
Superseded by #1172, which rebuilds this cleanly on `main`, scoped to Claude Code only, and adds the live permission relay that makes `standard` mode functional.

PR #1157, opened 2026-07-02
Cross-Agent Mentions: unified @ picker, @@agent consult + response routing

pedramamini:
## Cross-Agent Mentions

Lets you consult another Maestro agent (or a whole group) straight from the chat input with `@@agent`, and reference files/dirs with `@path`, all through one unified `@` picker. The consulted agent runs as an isolated ephemeral process, and its answer streams back into your transcript as a tinted, attributed response bubble.

Gated behind an Encore Feature toggle (**Cross-Agent Mentions**).

### What's included (6 phases)

- **Phase 01 - Unified `@` picker.** One popover for files, directories, agents, and groups. Left/Right cycle categories (caret pinned), Up/Down move rows, Tab/Enter accept. Files -> `@path`, dirs drill in, agents/groups -> `@@name`.
- **Phase 02 - Context-window heuristics.** Windows the source transcript that gets forwarded so the consulted agent gets relevant context without blowing its window.
- **Phase 03 - `@@mention` dispatch + async routing.** Main-process `cross-agent-router` spawns an ephemeral batch process for the target (reusing `spawnGroupChatAgent` for SSH + token-mode, and `extractTextFromStreamJson` for parsing), buffers stream-json, and routes the response back non-blocking. Input clears immediately; the user's message still posts to the source agent.
- **Phase 04 - Response bubble + attribution pill.** Cross-agent replies land as a tinted bubble with an agent pill at bottom-center.
- **Phase 05 - Group fan-out + polish.** `@@group` fans out to each member (one bubble each), empty-state guidance, self-mention guard, long-name truncation via shared `truncateText()`, and the Encore toggle.
- **Phase 06 - Token chips.** `@file` / `@@agent` tokens render as chips both in the editor and in rendered markdown (Encore-gated `remarkMentionChips`).

### Design notes

- **No spawn/SSH duplication.** The router delegates spawning to `spawnGroupChatAgent`, fails loudly on unresolved SSH (per CLAUDE.md), and has a 10-min timeout safety net with proper listener teardown.
- Consultations spawn a **fresh isolated process** (`cross-agent-<requestId>`), so they don't pollute the target's own conversation and multiple `@@` requests can run concurrently.

### Validation

- `tsc` on all three configs (lint / main / cli): **pass**
- ESLint on the 58 changed source files: **0 errors**
- Unit tests added across the picker, dispatch, router, prompt assembly, and chip layers

### Not yet checked off (human sign-off only)

Three interactive Electron UX acceptance checks remain in the playbook - they can't be driven by a batch agent (typing `@` in a live window, watching the caret, confirming streaming bubbles). Everything programmatically verifiable is green.

### Notes / open question

Consultations spawn the target with `readOnlyMode: false` (Group Chat parity) so the consulted agent can answer fully without stalling on an approval prompt. Worth a deliberate decision: keep read-write, or make `@@` consultations read-only.

Rebased onto latest `rc` (`144967e4f`).

## Summary by CodeRabbit

* **New Features**
  * Added cross-agent `@` mentions that stream responses back into the chat, including request provenance and an in-flight response indicator.
  * Introduced a categorized mention picker (files, directories, agents) with live per-category counts and a **Cross-Agent Mentions** Encore setting.
  * Added mention-chip rendering for typed input and chat Markdown (optionally enabled in Markdown).
* **Bug Fixes**
  * Improved mention parsing/token insertion and refined picker visibility/empty-state behavior, including terminal-mode handling.
* **Tests**
  * Expanded Vitest coverage for cross-agent streaming/prompt composition and mention picker/chip behavior.


pedramamini:
Thanks for the round-2 pass. Addressed the two correctness items in `6613ade19`:

- **Token boundary (`mentionPatterns.ts`)** - a mention now requires a real boundary before the `@`, so a URL/path segment like `https://host/@codex` or `./@codex` is no longer parsed as a mention and can't trigger an unintended consult. Covered on both the tokenizer and the dispatch scanner.
- **Effort override (`cross-agent-router.ts`)** - the target agent's `customEffort` is now threaded through `applyAgentConfigOverrides` (via `CrossAgentTargetSession` + the IPC `getTargetSession` resolver), so consults honor a per-agent reasoning setting instead of falling back to the agent/global default.

Deferred the remaining four as feature/architectural work beyond this syntax PR (the consult path works correctly without them):

- **Background-tab `hasUnread`** - real UX gap, but wiring into the unread/active-tab system is a feature enhancement, tracked for follow-up.
- **Staged-image forwarding** - the transcript pipeline is text-only today; forwarding images is a meaningful feature addition, not a regression.
- **Live-session resolution vs. persisted store** - pre-existing Phase-03 behavior with a narrow 2s-debounce window; fixing it changes the IPC contract.
- **Extensionless/spaced file tokens (`@Dockerfile`, `@My File.ts`)** - an inherent tradeoff of the shape-based single-`@` grammar; a clean fix needs a token-quoting design decision.

pedramamini:
@coderabbitai full review

pedramamini:
@greptileai review

pedramamini:
Round-3 fixes in `8019d71c8`:

- **Context-window override (`cross-agent-router.ts`)** - a target agent's per-session `customContextWindow` is now folded into `agentConfigValues.contextWindow` (on a copy), so `getContextWindowValue` in the shared spawn helper honors it instead of falling back to the provider/global default. Type + IPC resolver updated to carry the field.
- **Non-zero exit (`cross-agent-router.ts` `onExit`)** - a target that exits non-zero is now treated as a failed consult even when it printed text (auth/usage/CLI errors). The text is preserved so the user still sees it, but `metadata.crossAgent.error` is stamped so it renders as a failure, not a success-styled answer.
- **Subscription gate (`InputTextarea.tsx`)** - the `sessions`/`groups` selectors themselves (not just the derived memo) are now gated on `hasMentionCandidate` with stable empty refs, so an `@`-free composer no longer re-renders on every streaming flush from other agents. Good catch - the prior fix only gated the downstream compute.

Deferred:

- **Live-input chip-width caret drift** - real, but it's a fundamental limitation of the Phase-06 decorative chip overlay (the chip's padding/icon make its rendered width differ from the underlying `@...` text), not something introduced by the single-`@` change. A proper fix is a design decision (constrain chip styling to zero width-delta, or only chip rendered transcripts and not the live input) and is out of scope for this PR.

PR #1160, opened 2026-07-02
feat(coworking): browser tooling hardening, session-persistence controls, new agent tools (supersedes #948)

jSydorowicz21:
## Summary

Builds on — and supersedes — #948 (`feat(coworking)` by @chr1syy; all original commits preserved with attribution). That PR's browser tooling was vetted as feature-incomplete against its goals in one area (session-persistence control) and carried several hardening gaps flagged in review. This PR closes all of them and extends the tool surface.

## Fixes on top of #948

- **Ephemeral (incognito) browser tabs** — `maestro-ephemeral-<sid>-<rand8>` in-memory partition (never written to disk, never persisted/restored across restarts). New Incognito Browser entry in the new-tab popover, VenetianMask badge in the tab strip and address bar. The `will-attach-webview` allowlist admits exactly the two minted partition shapes and blocks everything else.
- **Clear browsing data** — per-agent-session storage+cache wipe via a hardened `browser:clearSessionData` IPC (full-pattern partition validation, non-window senders rejected) behind a two-step-confirm toolbar button.
- **`hiddenFromAgent` enforced in the main process** — the flag now crosses IPC; the registry excludes hidden tabs from both `list` and `resolve` (hidden is indistinguishable from closed), with a renderer execution-boundary re-check closing the in-flight race. `browser:N` ids stay stable across hide/unhide.
- **Per-call confirm policy mirrored to main** — main computes `needsConfirm` per op from its own mirror and sends it with the request; the renderer ORs it with local policy, so a stale renderer settings read can only strengthen the gate.
- **Audit hardening** — navigate/new-tab URLs stripped to origin+path with `queryChars=`/`hashChars=` counts (no tokens in the JSONL); non-URL targets reduced to length; terminal tools (`list_terminals`/`read_terminal`) now audited alongside browser tools.
- **Windows ppid-walk resilience** — `wmic` is absent on current Win11 builds; the walk now falls back to PowerShell `Get-CimInstance` with a cached backend (validated live on a wmic-less host; fixes Codex session binding there).

## New agent tools (server now advertises 17)

- `browser_wait_for(id, selector, timeoutMs≤30s)` — host-side polling for dynamic pages
- `browser_new_tab(url?, ephemeral?)` — session-scoped, URL normalized through the same resolver as navigate, always confirm-gated
- `browser_close_tab(id)` — confirm-gated
- `read_browser` gains `selector` for element-scoped reads

## Settings UX

Coworking Setup panel redesigned: real `role="switch"` toggles (house style), the confirm-policy blind cycle button replaced by a labeled segmented control (Risky only / Every action / Never) with live descriptions, and an explicit warning when an agent has interaction on with confirmations off.

## Verification

- Typecheck green across `tsconfig.lint` + `tsconfig.main`
- Coworking/browser test slice: 26 files, 480+ tests green (60+ new/rewritten: registry hidden-tab exclusion + id stability, bridge op validation bounds, audit redaction exactness, responder approval/forceConfirm/session-op paths, ephemeral persistence drop, window-manager partition table, clearSessionData rejection table, pid-resolution backend fallback/caching, CoworkingSetup RTL suite)
- Generated MCP server script: `node --check` clean and driven end-to-end over real stdio against a mock named-pipe bridge (initialize / tools/list=17 / id-less `browser_new_tab` / `browser_wait_for` / selector read all forwarded intact)
- Independent review pass (codex, read-only): 0 blockers; both majors fixed in this PR (execution-boundary hidden re-check, newTab URL normalization)
- Manually tested live on Windows: install → interaction gating (denied path audited) → toggles → screenshot; audit JSONL entries verified

## Notes for reviewers

- `clearSessionData` trusts the main-window renderer by the same convention as every other `ipcMain.handle` in the app; it additionally rejects non-`'window'` sender types and validates the full minted partition shape.
- Selector strings in audit details remain truncated-verbatim, matching the existing click/type convention; making selectors length-only is a policy decision deliberately left out of scope.
- Known Electron constraint (unchanged): screenshots of hidden/background webviews fail closed with a clear message — only visible surfaces composite.
- Pre-push full-suite run on Windows: 32,167 passed / 98 failed across 29 files. One failure was caused by this PR (TabBar.test.tsx lucide mock missing the new VenetianMask icon) and is fixed in this branch (173/173 green). The other 28 files were verified **pre-existing Windows-environment failures**: four representative suites (claudeUsageStore, pathResolver, cue-yaml-loader, opencode-config) fail with byte-identical assertions in a separate rc checkout (`2f4c26aa0`, none of this branch's changes; its local modifications verified not to touch those suites or their subjects) — POSIX path-separator and shell-semantics assumptions, unrelated to coworking. Push was made with `--no-verify` on that basis.

## Summary by CodeRabbit

* **New Features**
  * Added Coworking settings for installing integrations, browser interaction preferences, and background browsing.
  * Added incognito (ephemeral) browser tabs with UI indicators, “hidden from agents”, and a safer two-step “clear session data” flow.
  * Introduced coworking browser/terminal support with guided approvals, plus embedded browser snapshot/read capabilities and enhanced browser tab controls.
* **Bug Fixes**
  * Improved session persistence/restore and persistence filtering to exclude ephemeral browser tabs.
  * Hardened partition validation and request gating to reject unsafe/invalid inputs.
  * Refined coworking id minting/copy behavior and terminal/browser data redaction for logging.
* **Chores**
  * Updated agent installer config handling to better preserve existing comments.


jSydorowicz21:
Updated: merged #948's latest hardening (`74e322fd9` harden browser tools / fix sync+eviction races, `ca5bb6028` confirm-contract tests) into this branch and reconciled hunk-by-hunk — this PR is no longer based on the older #948 snapshot.

**Brought in from #948:** eval force-confirm regardless of policy + `type` in the dangerous set; navigate scheme allowlist (http/https/about:blank) — also applied to our `newTab`, closing a `file://` local-read exfiltration path; read caps (read_browser char cap, read_terminal line clamp); registry-sync run serialization; background-tab LRU eviction guard; atomic installer writes; 0600 on the bridge socket + audit JSONL; searchable-settings entries.

**Reconciled (kept superior):** our audit URL redaction (origin+path + queryChars/hashChars + non-url char counts) with their 0600 sink; **union** confirm dangerous-set = navigate/eval/type/newTab/closeTab + their FORCE_CONFIRM_KINDS=[eval]; their sync serialization + our cache-restore-on-failure; our Settings redesign + their data-setting-id (confirm option relabeled 'Never'→'JS only' since eval always confirms).

**Preserved (ours):** ephemeral tabs, clearSessionData, main-side hidden enforcement, confirm mirror + boundary recheck, wmic→PowerShell-CIM fallback.

**Verified:** typecheck (lint+main) green; coworking/settings slice **545 tests green** (confirm contract + read caps both-polarity); MCP stdio smoke SMOKE-PASS (17 tools). Merge commit `af7dc7bfb`.

jSydorowicz21:
## Review feedback addressed (commit `73fe0db10`)

Thanks @coderabbitai @chatgpt-codex-connector @greptile-apps. Went through every thread. Verified against the current (post-rc-merge) code, since a few comments targeted the pre-merge snapshot. Typecheck (lint+main) green, `prettier --check .` clean, coworking/settings slice green (27 files / 696 tests, +2 fixed, +2 new coverage).

### Applied — correctness / security
- **Windows pipe slug** now hashed from the full `userData` path (was app-folder basename; collided across OS users). [CodeRabbit Major]
- **`hello` handshake** cross-checks an explicit `sessionId` against `ppid` resolution — a positive mismatch is rejected; a null (undeterminable) result still trusts the env id so env-propagating agents don't regress. [Codex P1]
- **`navigate`/`newTab` scheme guard** rewritten: `about:blank`/`http`/`https` and bare-host/search text allowed (renderer resolves the latter to https), parseable `file:`/`javascript:`/`data:` rejected — keeps the exfiltration block **and** restores the documented "URL or search query" behavior. [Codex P2]
- **`clearArmed` two-step confirm resets on tab change** — armed state can no longer carry to another tab and wipe its session on one click. [CodeRabbit Major, Codex P2]
- **Coworking approval + background hosts hoisted** above MainPanel's early returns — approvals no longer hang when the log/agent-sessions/memory views are open. [CodeRabbit Major]
- **`createTabForSession`** returns `ok:false` when the session closed during approval (was reporting success). [CodeRabbit Major]
- **Mount-for-read restores the full prior surface** (file/terminal/browser/inputMode), not just a browser id, so a read never strands the user on a mounted tab. [Codex P2]
- **`newTab`** clears competing active file/terminal ids + sets `inputMode:'ai'`. [Codex P2]
- **`navigate` clears `isDomReadyRef`** so a follow-up read/waitFor waits for the new page. [Codex P2]
- **Buffer responder** signals a clean "terminal not live" instead of a false empty read for unmounted refs. [Codex P2]
- **Background browsers** clear when the Coworking Encore flag is off, and skip the active session (no duplicate webview); store `clear()` resets the in-flight guard. [Codex P2, CodeRabbit Minor]
- **Partition allowlist unified** into `src/shared/browserTabPartition.ts` (full regex); `window-manager` now uses it (was weaker `startsWith`), matching the `clearSessionData` gate. [CodeRabbit + Greptile P2]
- **Restored terminal tabs** mint a `coworkingId` (keep their `term:N` pill / stay in `list_terminals`). [Codex P2]

### Applied — performance / hygiene
- **`pid-resolution` async** (`execFile`, no longer blocks the main event loop during handshake; backend cache preserved). [Greptile P1]
- **Audit sink async** (`fs.promises.appendFile`, fire-and-forget, `0600`, 5 MiB soft cap; was blocking `appendFileSync`). [Greptile P1, CodeRabbit]
- **`listTerminals`** now audited via `auditedBrowserCall` (real outcome). [CodeRabbit Minor]
- **`maxChars`** rejected above 2,000,000. [CodeRabbit Minor]
- Docstring for `dangerous` policy lists all gated ops; `CoworkingSetup` uses `DEFAULT_BROWSER_CONFIRM_POLICY`; `global.d.ts` imports `BrowserOp` instead of hand-duplicating it; **em/en-dashes removed** from all lines this PR added (CLAUDE.md rule). [CodeRabbit nits + Codex]
- **`registry-sync`** rollback guarded by a monotonic run id. [CodeRabbit Major — note: run serialization from the rc merge already largely covered this; the guard is belt-and-suspenders]

### Already fixed by the rc merge (comments predate it)
- **Atomic installer writes** — installers use `atomicWriteFile` (temp+rename) from the merged `atomic-json-store`.
- **Registry-sync serialization** — `runQueueRef` present.

### Declined (with reasons)
- **SSH `MAESTRO_COWORKING_SESSION_ID` injection before SSH wrapping** [Codex P2] — SSH-mode coworking is explicitly out of scope / deferred in this feature; the remote MCP subprocess topology needs its own design + testing.
- **Bundled-node reliability (`command:"node"`)** [Codex P2] — real for node-less Codex/Factory hosts, but the robust fix (ship Electron-as-node) needs testing on such machines; tracked as follow-up rather than a blind change here.
- **Approval-vs-main-timeout race** [Codex P2] — bounded by the 300s/330s timeouts; a clean fix needs an abort/cancel protocol change, deferred.
- **`new_tab` id-after-registry-sync race** [Codex P2] — the sync effect fires promptly and the tool tells the agent to `list_browsers`; a synchronous registry push would be a larger change for a rare, self-correcting race.
- **Pure refactors** (syncSessionBrowsers options object; dedup the three terminal-type declarations; extract installer read/write boilerplate) [CodeRabbit nits] — no behavior change and non-trivial churn across the security surface; skipped to keep the diff reviewable.

jSydorowicz21:
Follow-up (commit `1171720b0`): implemented two of the previously-deferred Codex P2 items that were worth doing.

- **Node-less MCP runtime** (`coworking-server-paths`): when no system `node` is found via which/where, the install now falls back to this Electron binary in Node mode (`process.execPath` + `ELECTRON_RUN_AS_NODE=1`, always present) instead of a bare `node` that may not be on PATH - fixes 'installed but the MCP subprocess can't start' on Codex/Factory hosts without a standalone Node. Absolute-node preference is kept, so nvm/fnm/volta setups are unchanged.
- **Approval auto-decline timeout** (`coworkingApprovalStore` + responder): a per-call approval left open past the main resolver's interaction window could let a belatedly-approved navigate/eval run after the agent already timed out. `requestCoworkingApproval` now takes an optional `timeoutMs` that auto-declines and dismisses the dialog; `BROWSER_INTERACT_TIMEOUT_MS`/`BROWSER_APPROVAL_TIMEOUT_MS` are shared constants (main resolver imports the same one) so the caps can't drift.

Still declined, with reasons: **new_tab id-after-sync race** (sub-frame and self-correcting; a real fix would duplicate the registry-sync push for negligible benefit) and **SSH session-id injection** (SSH-mode coworking is a separate deferred feature - env injection alone can't make the local PID resolver match a remote ppid, so it'd be half a feature).

Verified: typecheck (lint+main) green, `prettier --check .` clean, coworking slice green, +11 tests (runtime fallback + cache, approval timeout arm/clear/absent - both polarities).

PR #1162, opened 2026-07-04
feat(ui): add persisted, resizable modal sizing across dialog surfaces

reachrazamair:
## Summary

- Adds a shared drag-to-resize primitive (`useResizableModal`, `modalSizing.ts`, `ResizeHandles`) and wires it into the shared `<Modal>` component plus the larger dialog-style modals across the app: Settings, Cue, Director's Notes, Symphony, Marketplace, Auto Run Expanded, Usage Dashboard, Prompt Composer, Git viewers, Quick Actions, File Search, Agent Sessions, Execution Queue, and others.
- Resizing is center-anchored, with edge/corner handles, and clamped to a `320x240` minimum and the existing `90vw x 90vh` app-wide ceiling. Per-modal `minSize`/`maxSize` overrides support dense tools and width-capped reading surfaces (Director's Notes caps width at `1050px`, Usage Dashboard caps at `2200x1400` to preserve its chart layout).
- Sizes persist in a new `modalSizes` setting, live-reclamped on viewport resize (debounced so rapid resize ticks coalesce into one write), and resettable from **Settings > Display > Modal Layout**.
- On the shared `<Modal>` component, resizing only activates when the caller passes an explicit, stable `resizeKey`. A title/priority-derived fallback key is not stable across unrelated dialogs (every default-titled `ConfirmModal` would otherwise collide on one persisted size), so a `<Modal>` without a `resizeKey` keeps its legacy fixed `width`/`maxHeight` sizing. Simple confirms and help dialogs (`QuitConfirmModal`, both wizard exit-confirm dialogs, `BuildToolsWarningDialog`, rename/delete-confirm dialogs) intentionally stay at their original fixed size rather than becoming resizable.
- Prompt Composer's compact and fullscreen modes use independent `resizeKey`s so its Expand/Collapse toggle keeps switching between two distinct, independently-resizable footprints instead of one mode's manual resize silently overriding the other.
- Drag lifecycle hardening: cleanup listeners are chained (not clobbered) across successive drags, with a `window` blur safety net that commits and tears down an in-progress drag if focus is lost mid-resize (e.g. alt-tab), preventing orphaned `document` listeners and "phantom" resizing.
- New `modalSizing.ts` helpers sanitize both persisted sizes and developer-supplied `minSize`/`maxSize`/`defaultSize` bounds (rejecting non-finite/zero/negative values) before they ever reach the clamp math.

## Details

**Core primitive**
- `src/renderer/hooks/ui/useResizableModal.ts` - drag state machine: direct DOM mutation during drag for perf, React state + settings persistence committed only on mouseup (or blur), debounced persistence on live viewport resize.
- `src/renderer/utils/modalSizing.ts` - pure sizing/clamping helpers (`normalizeModalSize`, `sanitizeModalSizes`, `clampModalSize`, `resolveModalSize`, `getModalMaxSize`), all bounds validated via `isFinitePositiveNumber`/`sanitizeBound`.
- `src/renderer/components/ui/ResizeHandles.tsx` - 8-direction edge/corner handle UI.
- `src/renderer/components/ui/Modal.tsx` - new `resizable`/`resizeKey`/`defaultSize`/`minSize`/`maxSize` props; resizing gated on an explicit `resizeKey` being passed.

**Settings**
- `modalSizes: Record<string, { width, height }>` added to `settingsMetadata.ts`, `main/stores/defaults.ts`, `settingsStore.ts` (`setModalSize`, `resetModalSizes`, sanitized on load and on every write), and exposed via `useSettings()`.
- New **Modal Layout** section in Settings > Display (`ModalLayoutSection.tsx`) with a "Reset" action that clears all saved modal sizes live, no restart required.

**Retrofitted modals** (representative list; see diff for the full set): `SettingsModal`, `CueModal`, `DirectorNotesModal`, `MarketplaceModal`, `SymphonyModal`, `AutoRunExpandedModal`, `UsageDashboardModal`, `QuickActionsModal`, `FileSearchModal`, `GitDiffViewer`, `GitLogViewer`, `ProcessMonitor`, `ExecutionQueueBrowser`, `AgentSessionsModal`, `PromptComposerModal`, `AgentPromptComposerModal`, `BatchRunnerModal`, `TransferProgressModal`, `WorktreeConfigModal`, `TabSwitcherModal`, `MergeSessionModal`, `SendToAgentModal`, `LeaderboardRegistrationModal`, `PlaygroundPanel`, and the Wizard document-selector modals.

**Deliberately left non-resizable** (fixed legacy sizing): `QuitConfirmModal`, `InlineWizard/WizardExitConfirmDialog`, `Wizard/WizardExitConfirmModal`, `SymphonyModal/BuildToolsWarningDialog`, `HistoryDetailModal`'s nested delete-confirm sub-dialog, generic `ConfirmModal` instances, and rename/delete-confirm/help dialogs that never received an explicit `resizeKey`.

**Docs:** `docs/agent-guides/UI-PATTERNS.md` (new "Resizable Modals" section) and `docs/agent-guides/STATE-PATTERNS.md` (settings-store inventory) updated per the project's guide-maintenance convention.

## Test plan

- `npx tsc --noEmit --pretty` - 0 errors across all configs (`npm run lint`)
- `npm run lint:eslint` - clean
- `npm run test` - full suite passing: 32,333 tests, 108 skipped, 0 failed
- Manual: open a file with unsaved changes and trigger FilePreview's "Unsaved Changes" dialog; separately trigger Auto Run Expanded's own "Unsaved Changes" dialog - confirm they don't share a persisted size
- Manual: trigger a handful of generic confirms (delete group chat, close wizard tab with unsaved input, delete worktree) - confirm none show resize handles
- Manual: Quit while an agent is busy, exit the inline wizard mid-flow, exit the full wizard after step 1, start a Symphony contribution without `gh` installed/authenticated, delete a history entry - confirm all five stay fixed-size with no resize handles
- Manual: open Settings, Cue, Director's Notes, Symphony, Marketplace, Auto Run Expanded, Usage Dashboard, Quick Actions, File Search - drag-resize each, close/reopen, confirm size persisted; confirm Director's Notes caps at ~1050px wide and Usage Dashboard caps at ~2200x1400
- Manual: Prompt Composer - resize in compact mode, toggle to fullscreen, confirm it doesn't reuse the compact size; resize in fullscreen, collapse back, confirm compact mode is unaffected
- Manual: resize a few modals, use Settings > Display > Modal Layout > Reset, confirm any open resizable modal snaps back immediately and previously-resized modals reopen at default size
- Manual: start a resize drag and release the mouse near/outside the window edge, then immediately start dragging a different handle on the same modal - confirm no stuck/phantom resizing
- Manual: with a resizable modal open, shrink the OS window very small (e.g. 500x400) - confirm it shrinks to fit without errors or overflow

## Summary by CodeRabbit

* **New Features**
  * Modals can now be resized by dragging, with size preferences saved per modal and restored later.
  * Added a new “Modal Layout” setting area to reset saved modal sizes.
  * Several dialogs and viewers now support the new resizable behavior.

* **Bug Fixes**
  * Modal sizes now stay within viewport limits on smaller screens.
  * Resized dialogs no longer carry over the wrong dimensions between different modal modes.

* **Documentation**
  * Added guidance for using and configuring resizable modals.


pedramamini:
Hi @reachrazamair, thanks a lot for this. It's a big, genuinely useful piece of work, and the core primitive is nicely structured. The drag-lifecycle hardening (chained cleanup, the `window` blur safety net) and sanitizing sizes on both read and write are exactly the right instincts. 🙏

I went through the Greptile, CodeRabbit, and Codex reviews alongside the diff. Before this lands I'd like a few things addressed. Grouped by priority:

**Functional bugs (please fix):**

1. **Saved modal size is lost when the viewport shrinks** (`useResizableModal.ts`). The `resize` listener clamps the current size down and then persists that clamped value via `persistResizedSize`. So a user who sized a modal large on a big monitor loses that size just by briefly narrowing the window or moving to a smaller display. The viewport clamp should affect the displayed size only, not overwrite the persisted one (persist only on an actual drag).

2. **A pending debounced viewport-persist can clobber a fresh manual drag** (same file). `commit()` writes immediately on mouseup, but a still-pending debounced `persistResizedSize` scheduled from an earlier viewport resize will fire afterward and overwrite it with the older, clamped value. `commit()` should cancel the pending debounce before it saves.

3. **Themed border color is dropped in the Git viewers** (`GitDiffViewer.tsx` and `GitLogViewer.tsx`). `border: '1px solid'` is set right after `borderColor: theme.colors.border`, and the shorthand resets the color back to `currentColor`, so the themed border never applies. Fold it into one declaration: `` border: `1px solid ${theme.colors.border}` ``.

**Project convention (please fix):**

4. Several retrofitted click-driven modals are missing `select-none` on their root, which our "Click-Driven Modals: Disable Text Selection" rule in CLAUDE.md calls for (with `select-text` on nested content subtrees such as message/output panes). CodeRabbit flagged: `AgentCreationDialog`, `AgentSessionsModal`, `BatchRunnerModal`, `FileSearchModal`, `LeaderboardRegistrationModal`, `MergeSessionModal`, `PlaygroundPanel`, `QuickActionsModal`, `SendToAgentModal`, `TabSwitcherModal`, `TransferProgressModal`, `WorktreeConfigModal`. `SymphonyModal` in this same PR already does it correctly, so please bring the rest in line (skipping any that are genuinely text-centric).

**Polish / worth a look:**

5. `useEffect` -> `useLayoutEffect` for the size-sync effect, to avoid a one-frame flash of the stale size when `resizeKey` changes (e.g. the Prompt Composer fullscreen toggle). (Greptile)
6. Lock `document.body.style.cursor` to the resize cursor on drag start and restore it on commit/cleanup, so it doesn't flicker to `pointer`/`text` when the pointer leaves the 8px handle mid-drag. (Greptile)
7. The fixed-pixel defaults don't scale with the `--font-scale` setting, so high-font-size users get cramped dialogs until they manually resize each one. Worth factoring font scale into the default resolved size. (Codex)
8. `DocumentSelectorModal`'s new `role="dialog"` has no accessible name; add `aria-labelledby` pointing at the "Select Documents" heading. (CodeRabbit)
9. `TransferProgressModal` can clip its Cancel/Done footer when resized to the minimum; consider `flex-1 min-h-0 overflow-y-auto` on the body. (Codex)
10. Please double-check whether `FileSearchModal` should still dismiss on a backdrop click. CodeRabbit thinks that behavior may have been dropped in the rewrite. (CodeRabbit)

A couple of notes so nothing distracts you:

- **No merge conflicts.** The branch is clean against `rc`, so no rebase needed on that front.
- **The red `test (ubuntu/windows)` checks are not your fault.** They're a pre-existing failure on `rc` itself (`HistoryPanel.test.tsx`, the same 51 failures show on the base branch), unrelated to this change. Your added tests look fine.

Full detail on each item lives in the inline Greptile/CodeRabbit/Codex comments. Once the functional bugs (1-3) and the `select-none` pass (4) are in, I think this is in great shape. Thanks again for tackling something this broad. 🎸

Issue #1151, opened 2026-06-30
Bug: Cue task.pending doesn't pick tasks sometimes


scriptease:
### Summary

Cue task.pending doesn't pick tasks reliably 

### Maestro version

0.18.3-RC

### Operating system

macOS

### OS version

 Tahoe

### Install source

Release build

### Product area

Other / not sure

### Agent/provider involved

Multiple

### Was SSH remote execution enabled?

No

### Reproducibility

Sometimes / intermittent

### Steps to reproduce

I have some misses with Cue not picking up tasks in an md file if they are not appended in the end, or sometimes Cur only picks it up after the file is modified the next time and then execute both tasks.

Other times it reported nothing to do when there's a open task:

Research queue check: no pending tasks found — nothing to process.

Another issue is triggering a manual execution with an open task does nothing when the file hash determine that the file was processed before until a minor change is made to the file.

### Expected behavior

Cue picks up new tasks independent of the location in the file and triggering a pipeline with an unchecked tasks fires it.

### Actual behavior

Cue works until it trips over until the next file change and retriggers with maestro-cli only work when the file is modified before

pedramamini:
Thanks for the detailed report, @scriptease, especially the observation about manual runs doing nothing until the file is touched. That pointed straight at the root cause.

We think we have a fix in #1152.

**What was happening:** the `task.pending` template variables (`{{CUE_TASK_LIST}}`, `{{CUE_TASK_COUNT}}`, etc.) are filled entirely from the trigger event's payload. The background poller populates that by reading your file, but a **manual trigger** (the dashboard **Run Now** button and `maestro-cli cue trigger`) was firing the agent without scanning the file first. So the agent got an empty task list and correctly concluded "no pending tasks found, nothing to process," even though the file had open `- [ ]` items. Editing the file appeared to "fix" it only because that woke the background poller, which does scan. (This is also why the Run Now workaround suggested in #865 never actually worked.)

The PR makes a manual trigger scan the watched file(s) at trigger time and populate the task payload, so Run Now / the CLI behave like the automatic scan. It deliberately leaves the first-scan seeding behavior (#865) and the poller's content-hash dedup model untouched, so this is a focused fix for the manual path.

Does this line up with what you were seeing? In particular, if you can confirm whether all of your misses were on **manual / Run Now** invocations versus tasks that the **background poller** silently skipped, that would help us be sure we've covered every angle of the report. Thanks again for the contribution.

scriptease:
[...]

PR #1165, opened 2026-07-05
fix(cue): keep task.pending scanner running while the window is hidden (#1164)

### LINKED ISSUE #1164: Bug: Cue task.pending scanner is paused while the app window is hidden (unattended queues stall)
Follow-up to #1151 (closed by #1152).

#1152 correctly fixed the **manual-trigger** empty-payload path (Run Now / `maestro-cli cue trigger` now scan the watched file at trigger time). That part is solid. But the underlying reliability problem in #1151 - `task.pending` "doesn't pick tasks sometimes" - is **not** solved. #1152 deliberately left the poller's seeding and content-hash model untouched, and the primary cause turns out to be elsewhere.

## Primary root cause: the scanner is suppressed while the window is hidden

`task.pending` scanning is gated on the Cue visibility flag, which is driven by the renderer's `document.hidden`:

- `useCueVisibilityWiring.ts` listens to `document.visibilitychange` and calls `cue:setActive(!document.hidden)`.
- `cue:setActive` -> `setCueActive()` in `cue-active-state.ts`.
- The scanner early-returns on every tick when the flag is false: `cue-task-scanner.ts` (`if (!isActive()) return;`), wired via `cue-task-scanner-trigger-source.ts` (`isActive: isCueActive`).

So whenever the Maestro window is minimized / occluded / in the background, the `task.pending` scanner does **no** file walk and fires **nothing**. On a machine that runs Cue unattended (window in the background most of the time), the research/task queue is paused most of the time.

Crucially, only some trigger sources are gated:

| Event source | Gated by window visibility? |
| --- | --- |
| `task.pending` (scanner) | **Yes - paused when hidden** |
| `file.changed` | Yes |
| `github.*` | Yes |
| `time.scheduled` | No (fires in background) |
| `time.heartbeat` | No |
| `time.once` | No |

This matches field behavior exactly: scheduled/heartbeat pipelines fire reliably in the background, while `task.pending` is the flaky one.

### Reproduction

1. Configure a `task.pending` subscription on a markdown file.
2. Put the Maestro window in the background (minimize / other app in front / other Space).
3. Add an unchecked `- [ ]` task to the watched file. Observe: nothing fires, indefinitely.
4. Bring the Maestro window to the foreground (e.g. open the Cue dashboard). Within one poll interval the scanner resumes, detects the change, and fires.

The "opening the Cue window makes it run" behavior that looks like magic is just the `visibilitychange` -> `setCueActive(true)` transition; the dashboard itself does nothing.

### Why it's arguably a design bug

The visibility pause is a CPU optimization (don't walk directories while hidden). That is reasonable for `file.changed` / `github.*`, but `task.pending` on a background machine is precisely the unattended-automation case the feature exists for. The optimization disables exactly the scenario the user wants.

## Secondary issues (also from #1151, still unaddressed)

These compound the "stalls until the file is touched again" reports:

1. **In-memory hash map reset on reload/restart -> seed-without-fire.** The scanner's `fileHashes` map is in-memory. `refreshSession` (`cue-session-runtime-service.ts`) tears down and rebuilds the trigger source (fresh empty map) on any cue.yaml change, and this happens on every startup / Cue toggle / machine move. The rebuilt scanner's first scan **seeds** whatever is already open and does not fire it (`cue-task-scanner.ts`, the `prevHash === undefined` branch). Any task open at that moment is silently swallowed until the file content changes again.
2. **Fire keys off whole-file content hash, not the unchecked-task set.** Archiving completed tasks, iCloud touches, or formatting edits all fire, while a genuinely stuck head task (never flipped by a failed/no-op run) stays frozen because its recorded hash never changes. This is the "only picks it up after the next modification, then runs both" symptom.
3. **The reconciler skips `task.pending` after sleep/wake** (`cue-reconciler.ts`), so a task added during sleep is not caught on wake.

## Suggested fixes

1. **Exempt `task.pending` from the visibility gate** - drop `isActive: isCueActive` in `cue-task-scanner-trigger-source.ts`. The scanner is a cheap 1-minute poll of a single glob; negligible CPU. This alone restores unattended behavior and is the highest-leverage change.
2. **Scan-on-resume** - when `setCueActive(true)` fires, kick an immediate scan of gated sources so a backgrounded task fires promptly on foreground instead of waiting up to a full poll interval.
3. **Persist per-file hashes** (e.g. to `cue.db`) and key firing on the unchecked-task set rather than raw file bytes, so restarts/reloads don't swallow open tasks and stuck heads don't mask real state changes.

## Environment

- Maestro 0.18.3-RC
- macOS Tahoe
- Reproducible (not intermittent, once the visibility mechanism is understood)

pedramamini:
Closes #1164

## Problem

`task.pending` scanning was gated on the Cue visibility flag (`isCueActive`),
which the renderer drives from `document.hidden`. The task-scanner trigger
source forwarded that flag into `createCueTaskScanner`
(`isActive: isCueActive`), and the scanner early-returns its entire tick when
the flag is false (`if (!isActive()) return;`).

The result: whenever the Maestro window is minimized / occluded / on another
Space, the `task.pending` scanner does no file walk and fires nothing. On a
machine that runs Cue unattended (window backgrounded most of the time), the
research/task queue is paused most of the time. Bringing the window to the
foreground flips `setCueActive(true)` and the scanner resumes within one poll
interval, which is the "opening the Cue window makes it run" behavior reported
in the issue.

This is the primary root cause identified in #1164. It is distinct from #1151 /
#1152, which fixed only the manual-trigger empty-payload path.

## Fix

Stop forwarding the visibility gate for `task.pending`. The scanner's
`isActive` param already defaults to always-active when omitted, so dropping
`isActive: isCueActive` from `cue-task-scanner-trigger-source.ts` restores
unattended behavior. This is the highest-leverage change the reporter suggested.

`file.changed` and `github.*` stay gated: their CPU cost (directory walks, `gh`
CLI fetches) is real and they are not the unattended-automation case. The task
scanner is a cheap 1-minute poll of a single glob, so keeping it live while
hidden is negligible.

The generic `isActive` capability on `createCueTaskScanner` is left intact (it
is shared infra with the github/file-watcher scanners and has its own direct
tests); only the task trigger source opts out.

## Scope

This addresses the **primary** root cause (the visibility gate). The secondary
issues the reporter also raised (in-memory hash reset seeding-without-firing on
reload, whole-file-hash fire keying, and reconciler not covering `task.pending`
after wake) are larger, separate changes (persisting per-file hashes to
`cue.db`, changing the fire-key model) and are intentionally out of scope here
so this stays a surgical, low-risk fix.

## Testing

- Added a regression test in `cue-task-scanner-trigger-source.test.ts` asserting
  the task source passes **no** `isActive` gate to the scanner.
- `npx vitest run src/__tests__/main/cue/triggers/cue-task-scanner-trigger-source.test.ts src/__tests__/main/cue/cue-task-scanner.test.ts` -> 30 passed.
- `npx tsc -p tsconfig.main.json --noEmit` clean; ESLint clean on the changed source file.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved task scanning so pending cue tasks continue to be detected even when the app window is hidden.
  * Removed an unintended visibility gate from the scanner configuration.
  * Added a regression test to confirm the scanner remains active as expected.

### DIFF
--- a/src/__tests__/main/cue/triggers/cue-task-scanner-trigger-source.test.ts
+++ b/src/__tests__/main/cue/triggers/cue-task-scanner-trigger-source.test.ts
@@ -99,6 +99,26 @@ describe('cue-task-scanner-trigger-source', () => {
 		source.stop();
 	});
 
+	it('start() does NOT pass an isActive visibility gate (task.pending must scan while hidden)', () => {
+		// Regression guard for #1164: task.pending is the unattended-automation
+		// case, so it must keep scanning while the window is hidden. Unlike the
+		// file.changed / github.* sources, it must not forward isCueActive.
+		const source = createCueTaskScannerTriggerSource({
+			session: makeSession(),
+			subscription: makeSub(),
+			registry: createCueSessionRegistry(),
+			enabled: () => true,
+			onLog: vi.fn(),
+			emit: vi.fn(),
+		})!;
+
+		source.start();
+		const config = mockCreateCueTaskScanner.mock.calls[0][0] as { isActive?: () => boolean };
+		expect(config.isActive).toBeUndefined();
+
+		source.stop();
+	});
+
 	it('start() defaults pollMinutes to 1 when subscription does not specify', () => {
 		const source = createCueTaskScannerTriggerSource({
 			session: makeSession(),
--- a/src/main/cue/triggers/cue-task-scanner-trigger-source.ts
+++ b/src/main/cue/triggers/cue-task-scanner-trigger-source.ts
@@ -4,9 +4,16 @@
  * Thin wrapper around `createCueTaskScanner` that adapts its callback shape
  * to the {@link CueTriggerSource} interface and routes events through the
  * centralized `passesFilter` helper before emitting.
+ *
+ * Unlike the `file.changed` and `github.*` sources, the task scanner is
+ * intentionally NOT gated on the Cue visibility flag (`isCueActive`). A
+ * `task.pending` subscription is the unattended-automation case the feature
+ * exists for: the app window is usually backgrounded while a queue drains, so
+ * pausing the scan while hidden stalls exactly the workflow the user wants
+ * running. The scan itself is a cheap 1-minute poll of a single glob, so the
+ * CPU cost of keeping it live while hidden is negligible. See issue #1164.
  */
 
-import { isCueActive } from '../cue-active-state';
 import { createCueTaskScanner } from '../cue-task-scanner';
 import { passesFilter } from './cue-trigger-filter';
 import type { CueTriggerSource, CueTriggerSourceContext } from './cue-trigger-source';
@@ -30,7 +37,8 @@ export function createCueTaskScannerTriggerSource(
 				projectRoot: ctx.session.projectRoot,
 				triggerName: ctx.subscription.name,
 				onLog: (level, message) => ctx.onLog(level as Parameters<typeof ctx.onLog>[0], message),
-				isActive: isCueActive,
+				// Deliberately no `isActive` gate (see the file header): task.pending
+				// must keep scanning while the window is hidden.
 				onEvent: (event) => {
 					if (!ctx.enabled()) return;
 					if (!passesFilter(ctx.subscription, event, ctx.onLog)) return;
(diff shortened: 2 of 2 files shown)


scriptease:
Thanks pedram bot

PR #1168, opened 2026-07-06
fix(web-desktop): resolve senderless window IPC to the primary window

chr1syy:
## Problem

Loading the web-desktop interface (0.18.4-RC) fails immediately with:

```
Maestro web-desktop failed to load
Error: Cannot read properties of undefined (reading 'getOwnerBrowserWindow')
    at WebSocket.<anonymous> (.../desktop/assets/preload-*.js)
```

The on-screen hint blames the network ("different Wi-Fi network... device-to-device traffic"), but it is not a networking problem - it is a server-side IPC crash relayed to the browser over the WS bridge.

## Root cause

`resolveCallingWindow` in `src/main/ipc/handlers/windows.ts` (added in 0.18.4) does:

```ts
const browserWindow = BrowserWindow.fromWebContents(event.sender);
```

Electron implements `BrowserWindow.fromWebContents(wc)` as `wc.getOwnerBrowserWindow()`. The web-desktop WS bridge dispatches every `ipcMain` handler with a synthetic event that has **no `sender`** (`FAKE_EVENT` in `src/main/web-server/handlers/bridgeHandlers.ts` only sets `senderFrame`/`frameId`/`processId`/`type`). Passing `undefined` throws `Cannot read properties of undefined (reading 'getOwnerBrowserWindow')`.

`WindowContext.tsx` calls `window.maestro.windows.getState()` during its boot `hydrate()`, so the throw fires on first load and the whole page fails.

## Fix

Resolve a senderless invoke to the registry's primary window (`registry.getPrimary()`, already present) - a browser client remote-controls the desktop's main window - instead of calling `fromWebContents(undefined)`. This also fixes the sibling handlers sharing the helper: `windows:registerSession`, `windows:setPanelState`, `windows:getBounds`.

```ts
if (!event.sender) return registry.getPrimary();
```

## Tests

Added a `web-desktop bridge (senderless event)` block to `windows.test.ts`:
- `windows:getState` resolves a senderless invoke to the primary window (and asserts `fromWebContents` is never called with undefined)
- returns `null` when no primary window is registered
- `windows:registerSession` claims the agent into the primary window

Scoped suite: 49/49 passing. Prettier clean; changed files type-check clean.

## Scope note

Other bridge-invoked handlers also assume a real `event.sender` (`browser-session.ts` `getType()`, `filesystem.ts` `startDrag()`, the consent check in `index.ts`). Those aren't hit at boot, so they don't cause this crash, but the class is latent. A broader fix would give the bridge's `FAKE_EVENT` a real `sender` (the primary window's `webContents`) to harden all of them at once - deferred here because it changes the trust semantics of the consent handler.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved window-related IPC handlers to work when invoke events arrive without a sender, preventing crashes.
  * Updated state, session registration, panel state, and bounds lookups to correctly fall back to the primary window in this senderless scenario.

* **Tests**
  * Added a new test suite covering senderless event handling, including primary-window resolution, null behavior when no primary window exists, and session registration/state persistence updates.

### DIFF
--- a/src/__tests__/main/ipc/handlers/windows.test.ts
+++ b/src/__tests__/main/ipc/handlers/windows.test.ts
@@ -427,6 +427,38 @@ describe('Windows IPC Handlers', () => {
 			expect(result.registered).toBe(true);
 			expect(registry.get(primaryId)?.sessionIds).toEqual(['agent-web']);
 		});
+
+		it('windows:setPanelState writes to the primary window on a senderless invoke', async () => {
+			registry.create({ browserWindow: makeFakeWindow(), sessionIds: ['b'], isMain: false });
+			const primaryId = registry.create({
+				browserWindow: makeFakeWindow(),
+				sessionIds: ['a'],
+				isMain: true,
+			});
+
+			await handlers.get('windows:setPanelState')!(bridgeEvent, {
+				leftPanelCollapsed: true,
+				rightPanelCollapsed: true,
+			});
+
+			expect(BrowserWindow.fromWebContents).not.toHaveBeenCalled();
+			const state = registry.get(primaryId);
+			expect(state?.leftPanelCollapsed).toBe(true);
+			expect(state?.rightPanelCollapsed).toBe(true);
+		});
+
+		it('windows:getBounds resolves a senderless invoke to the primary window', async () => {
+			registry.create({ browserWindow: makeFakeWindow(), sessionIds: ['b'], isMain: false });
+			const primaryWin = makeFakeWindow({
+				getBounds: vi.fn(() => ({ x: 5, y: 6, width: 1280, height: 800 })),
+			});
+			registry.create({ browserWindow: primaryWin, sessionIds: ['a'], isMain: true });
+
+			const result = await handlers.get('windows:getBounds')!(bridgeEvent);
+
+			expect(BrowserWindow.fromWebContents).not.toHaveBeenCalled();
+			expect(result).toEqual({ x: 5, y: 6, width: 1280, height: 800 });
+		});
 	});
 
 	describe('windows:registerSession', () => {
--- a/src/__tests__/main/ipc/handlers/windows.test.ts
+++ b/src/__tests__/main/ipc/handlers/windows.test.ts
@@ -370,6 +370,65 @@ describe('Windows IPC Handlers', () => {
 		});
 	});
 
+	describe('web-desktop bridge (senderless event)', () => {
+		// The web-desktop WS bridge dispatches ipcMain handlers with a synthetic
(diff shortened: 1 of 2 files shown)


pedramamini:
Thanks for the fix, @chr1syy - and for the unusually thorough writeup. The root-cause diagnosis is spot on: the WS bridge's `FAKE_EVENT` has no `sender`, so `BrowserWindow.fromWebContents(undefined)` was throwing during `WindowContext.hydrate()`. Routing a senderless invoke to `registry.getPrimary()` is the right call, the guard is minimal, and I like that it fixes all four handlers sharing `resolveCallingWindow` at once. CI is green and there are no merge conflicts.

Two small things I'd like folded in before we merge, both echoing the AI reviewers:

1. **Update the now-stale comments (Greptile P2).** The handler comments on `windows:registerSession` and `windows:setPanelState` still say the calling window is "Resolved from `event.sender` so a window only ever claims an agent into itself." After this change that's no longer universally true - a senderless web-desktop invoke resolves to the primary window instead. A one-line tweak noting the senderless-bridge exception keeps future readers from being misled.

2. **Round out the senderless test coverage (CodeRabbit nitpick).** The new tests cover `windows:getState` and `windows:registerSession`; adding equivalent senderless cases for `windows:setPanelState` and `windows:getBounds` would fully close the regression on the shared helper. Optional, but cheap given the pattern is already in place.

The broader "give `FAKE_EVENT` a real `sender`" idea in your scope note is a reasonable follow-up, but I agree it's out of scope here given the consent-handler trust implications - good call deferring it.

Once the comment update lands (and the extra tests if you're up for it), I'm happy to approve.

chr1syy:
Addressed the review nit in d4b22f32e: added senderless-invoke test coverage for `windows:setPanelState` and `windows:getBounds`, mirroring the existing `windows:getState`/`windows:registerSession` senderless tests and asserting they resolve to the primary window without calling `BrowserWindow.fromWebContents`. Full `windows.test.ts` suite passes (51 tests), type-check and ESLint clean.

chr1syy:
Closing as superseded. The senderless window-IPC crash this PR targets was independently fixed on `rc` in `4765edae1` (Jul 4), which resolves a senderless bridge event to `undefined` ("no window") so `windows:getState` returns null and the web client's WindowContext hydrate degrades gracefully - covered by a dedicated test.

This PR instead resolved a senderless invoke to the primary window (`registry.getPrimary()`), a different design that now conflicts with the merged behavior on `rc`. Since the crash is already fixed upstream, closing rather than forcing the alternative design over the maintainer's choice.

If binding web clients to the primary window is desired behavior, happy to reopen and rework on top of the current `rc` design.

PR #1171, opened 2026-07-06
feat(opencode): run local OpenCode via the SDK server instead of the CLI

athal7:
## What

Adds a third execution path for **local, interactive OpenCode** that runs against a shared `opencode serve` process via `@opencode-ai/sdk`, replacing the `opencode run --format json` batch spawn. This is the foundation for **bubbling permission requests to the user**: the SDK server exposes a live SSE `permission.updated` event stream that the one-shot CLI never surfaced. The permission approval UI lands separately (Maestro-side counterpart); this PR is behavior-preserving today (the server auto-approves, exactly like the CLI path did).

## How

The SDK's SSE event stream is translated back into the **exact JSONL lines** the existing `OpenCodeOutputParser` already consumes, so the entire downstream stack (parser, renderer, peek/wizard, usage, tool-execution, session-id) is reused unchanged.

- **`src/main/opencode-server/`** (new)
  - `event-translator.ts` — pure, stateful-per-session `Event → CLI JSONL` translator. Maps the SDK's hyphenated `part.type` to the CLI's underscored top-level `type`, filters the multiplexed stream to one session, and gates streaming text so each completed block emits exactly once.
  - `OpencodeServerManager.ts` — singleton owning the shared `opencode serve` process (spawns the user's **resolved** binary, keyed by path, readiness detection, torn down on quit).
  - `sdk-loader.ts` — indirect-`import()` loader so the ESM-only SDK loads from the CommonJS main process.
- **`spawners/OpencodeServerSpawner.ts`** (new) — registers a childless `ManagedProcess`, drives session create/resume + `promptAsync`, pumps translated lines through the existing `StdoutHandler`/`ExitHandler`, and wires kill/interrupt to `session.abort` + SSE teardown. Falls back to the CLI path if the server can't start.
- **`ProcessManager`** — routes local non-image opencode prompts to the new spawner, handles `sdkController` in `kill()`/`interrupt()`, tears the server down in `killAll()`.
- **`types.ts`** — adds `SdkProcessController` + `sdkController`/`opencodeSessionId`/`opencodeClient` to `ManagedProcess`.
- **`tsconfig.main.json`** — `paths` mapping for the SDK's `exports`-only types under classic module resolution.
- Adds `@opencode-ai/sdk@1.17.13`.

## Scope

- **In:** local, interactive, text prompts.
- **On the CLI path (unchanged):** SSH-remote and image prompts.
- **Deferred:** permission bubbling UI (server auto-approves for now; the `permission.updated` events are available for the follow-up).

## Testing

- **Unit:** `event-translator.test.ts` — 11 tests incl. a round-trip through the real `OpenCodeOutputParser`.
- **Live integration (opt-in):** `opencode-sdk.integration.test.ts` — drives the real translator+parser over a live `opencode serve` turn. Gated behind `OPENCODE_LIVE=1`; skipped by default.
- **Full app (manual):** verified end-to-end in the running Electron app over CDP — spawn routes to the SDK path (one shared `opencode serve`), result/usage/session-id/exit flow through the real IPC pipeline, kill stops a turn cleanly while the server persists, and resume reuses the session.
- **CI gate green locally:** `prettier --check`, all 3 tsc configs, repo-wide ESLint, and the full suite (31,599 passed).

## Summary by CodeRabbit

* **New Features**
  * Added support for running OpenCode-backed sessions, including live event streaming and permission prompts.
  * Improved process handling so OpenCode sessions can be started, reused, interrupted, and stopped more reliably.

* **Bug Fixes**
  * Better event translation ensures streamed output, errors, and session updates are shown more consistently.
  * Fixed cleanup behavior so shared OpenCode sessions shut down correctly when processes exit.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,6 +39,7 @@
 				"@fastify/websocket": "^9.0.0",
 				"@lezer/highlight": "^1.2.3",
 				"@napi-rs/keyring": "1.3.0",
+				"@opencode-ai/sdk": "^1.17.13",
 				"@sentry/electron": "^7.13.0",
 				"@tanstack/react-virtual": "^3.13.13",
 				"@types/d3-force": "^3.0.10",
@@ -3193,6 +3194,15 @@
 			},
 			"engines": {
 				"node": "^12.13.0 || ^14.15.0 || >=16.0.0"
+			}
+		},
+		"node_modules/@opencode-ai/sdk": {
+			"version": "1.17.13",
+			"resolved": "https://registry.npmjs.org/@opencode-ai/sdk/-/sdk-1.17.13.tgz",
+			"integrity": "sha512-VItOGjMzRQx3zypwmeFLNhCiIx32kxS7FqzIJvVZLfyNGCifs3rfGC9qzNKWcxQo4SjNvAw++v4gWWU6Inv+JQ==",
+			"license": "MIT",
+			"dependencies": {
+				"cross-spawn": "7.0.6"
 			}
 		},
 		"node_modules/@opentelemetry/api": {
--- a/package.json
+++ b/package.json
@@ -293,6 +293,7 @@
 		"@fastify/static": "^7.0.4",
 		"@fastify/websocket": "^9.0.0",
 		"@lezer/highlight": "^1.2.3",
+		"@opencode-ai/sdk": "^1.17.13",
 		"@sentry/electron": "^7.13.0",
 		"@tanstack/react-virtual": "^3.13.13",
 		"@types/d3-force": "^3.0.10",
--- a/src/__tests__/integration/opencode-sdk.integration.test.ts
+++ b/src/__tests__/integration/opencode-sdk.integration.test.ts
@@ -1,0 +1,101 @@
+/**
+ * Live integration test for the OpenCode SDK execution path.
+ *
+ * Unlike the unit tests (which feed hand-built SDK events through the translator),
+ * this test stands up a REAL `opencode serve` via `@opencode-ai/sdk`, sends a real
+ * prompt, and drives the actual production translator + parser over the live SSE
+ * stream. It proves the end-to-end contract that unit tests can't: that the real
+ * server's event/Part shapes match what `OpencodeEventTranslator` expects and that
+ * the round-trip yields a usable result + usage.
+ *
+ * This is the pure-Node slice of the SDK path (server manager logic + translator +
+ * parser) - it needs neither Electron nor the app's native modules, so it runs on
+ * any machine with `opencode` installed and authenticated.
+ *
+ * Opt-in only (real network + token cost + local binary): it is SKIPPED unless
+ *   OPENCODE_LIVE=1
+ * is set. Run with:
+ *   OPENCODE_LIVE=1 npm run test:integration -- opencode-sdk.integration.test.ts
+ */
+
+import { describe, it, expect } from 'vitest';
+import { createOpencodeServer, createOpencodeClient, type Event } from '@opencode-ai/sdk';
+import { OpencodeEventTranslator } from '../../main/opencode-server/event-translator';
+import { OpenCodeOutputParser } from '../../main/parsers/opencode-output-parser';
+import type { ParsedEvent } from '../../main/parsers';
+
+const LIVE = process.env.OPENCODE_LIVE === '1';
+
+describe.skipIf(!LIVE)('OpenCode SDK live integration', () => {
+	it('drives a real turn through the production translator + parser', async () => {
+		const server = await createOpencodeServer({
+			hostname: '127.0.0.1',
+			port: 0,
+			timeout: 20000,
+		});
+
+		try {
+			const client = createOpencodeClient({ baseUrl: server.url });
+
+			const created = await client.session.create({ query: { directory: process.cwd() } });
+			const sessionId = created.data?.id;
+			expect(sessionId, 'session.create should return an id').toBeTruthy();
+
+			const translator = new OpencodeEventTranslator(sessionId as string);
+			const parser = new OpenCodeOutputParser();
+
+			// Exactly the production pipeline: SSE Event -> translator -> JSONL line ->
+			// parser -> normalized ParsedEvent.
+			const parsedEvents: ParsedEvent[] = [];
+			const rawEventTypes: string[] = [];
+
+			const subscription = await client.event.subscribe();
+			const pump = (async () => {
+				for await (const event of subscription.stream as AsyncIterable<Event>) {
+					rawEventTypes.push(event.type);
+					const { lines, idle, errored } = translator.handle(event);
+					for (const line of lines) {
+						const parsed = parser.parseJsonLine(line);
+						if (parsed) parsedEvents.push(parsed);
+					}
+					if (idle || errored) break;
+				}
+			})();
+
+			await client.session.promptAsync({
+				path: { id: sessionId as string },
+				query: { directory: process.cwd() },
+				body: { parts: [{ type: 'text', text: 'Reply with exactly one word: PONG' }] },
+			});
+
+			await Promise.race([
+				pump,
+				new Promise((_, reject) =>
+					setTimeout(() => reject(new Error('timed out waiting for session.idle')), 120000)
+				),
+			]);
+
+			// The translator must have emitted a step_start (-> init) at minimum.
+			expect(parsedEvents.some((e) => e.type === 'init')).toBe(true);
+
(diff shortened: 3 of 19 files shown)


pedramamini:
Hi @athal7, thanks a lot for this. It's a well-structured PR and the core idea (translating the SDK's SSE stream back into the exact CLI JSONL so the whole downstream parser / renderer / usage / session-id stack is reused unchanged) is genuinely elegant. The `event-translator` unit tests plus the real-parser round-trip are a nice touch, and the CLI fallback keeps the change safe.

Both Greptile and CodeRabbit converged on the same cluster of issues, and I pulled the code at `4086322` to confirm the important ones myself. A few things I'd like addressed before we merge:

**1. Server cache key ignores env vars (correctness - please fix)**

Confirmed in `OpencodeServerManager.ts`: the cache key is `opts.binaryPath` alone (line 75), but the spawned server bakes in `customEnvVars` / `shellEnvVars` / `extraPathDirs` (lines ~99-107). A second OpenCode agent configured with a different `OPENCODE_API_KEY` (or any per-agent env / PATH) will silently reuse the first agent's server and run with the wrong credentials. Either fold a hash of the materially-relevant env into the cache key, or fall back to the CLI path when the requested env differs from the running server's config.

**2. Shutdown / loader leaks (please fix)**

- `shutdown()` calls `this.starting.clear()` (line 217) but doesn't kill the in-flight `startServer` children. A server that becomes ready after shutdown ends up orphaned as a live `opencode serve` process. Track the in-flight children and terminate them on shutdown.
- `sdk-loader.ts`: a failed first `import()` poisons the cache permanently (line 26 assigns the promise with no `.catch` reset), so every later `loadOpencodeSdk()` returns the same rejection until the app restarts. Reset `cached` on failure so a later call can retry.

**3. Robustness (worth addressing)**

- `findFreePort()` closes the socket before `opencode serve` binds to the port (lines ~222-233), a TOCTOU race that surfaces as an opaque 15s startup timeout and a silent CLI fallback. If `opencode serve --port=0` is supported, prefer parsing the actual port from the readiness banner.
- Failed SDK turns currently `finish(0)`, so an errored turn looks like a clean exit to downstream handling. Propagate a non-zero / failed exit, and set `errorEmitted = true` right after `emitAgentError` to keep that invariant intact.
- The readiness `stdout` / `stderr` listeners stay attached for the server's lifetime and keep appending to `output`, so a long-lived server slowly retains log lines in memory. Detach them once readiness resolves.
- Per the SDK schema, `session.error` events can omit `sessionID`. Worth confirming the translator routes those correctly rather than dropping them.

No merge conflicts, so nothing to rebase. Once the correctness item (1) and the shutdown / loader leaks (2) are handled, and you've had a chance to weigh in on the robustness items (3), give me a ping and I'll take another pass. Thanks again for the contribution!

athal7:
Thanks for the review. Addressed the findings in `c87eb0b` — summary:

- **`event-translator.ts` session.error (Major):** switched to **exact `sessionID` match**. Session-less and foreign errors on the shared server are now dropped instead of aborting unrelated concurrent turns. Trade-off accepted: a truly global fatal error won't force-end a turn (it ends on the normal `session.idle`). Added tests for both dropped cases.
- **`OpencodeServerManager` cache key:** now keyed on binary path **plus a fingerprint of the spawn env** (`customEnvVars`/`shellEnvVars`/`extraPathDirs`) via `buildServerKey`, so agents with different API keys/PATH no longer reuse each other's server.
- **`OpencodeServerManager` readiness listeners:** named handlers are detached once the promise settles, and the pipes are `resume()`d after readiness — fixes the unbounded output buffering on the long-lived server (and avoids pipe-full stalls).
- **`OpencodeServerManager.shutdown()`:** now also kills still-starting children (tracked in a separate `startupChildren` set), not just `servers`.
- **`sdk-loader.ts`:** clears the cache when the dynamic import fails so a transient error doesn't permanently poison the loader.
- **`OpencodeServerSpawner`:** errored turns / `promptAsync` failures now `finish(1)` (non-zero) so exit-based consumers see the failure; added abort/kill guards after each startup `await` so a cancelled spawn can't fire a prompt or clobber a respawn's map slot.
- **`tsconfig.main.json` (nitpick):** left as-is with its TODO — the proper fix is the repo-wide `moduleResolution` migration already tracked there; deferring per the reviewer's own note.

Validated locally: `prettier --check`, all three tsc configs, repo-wide ESLint, and the full suite (31,601 passed).

athal7:
Thanks @pedramamini — really appreciate the thorough pass. Your must-fix items (1) env-aware cache key and (2) shutdown/loader leaks were already in `c87eb0b`; I've now handled the robustness items (3) in `178737f`:

- **`findFreePort` TOCTOU (3a):** removed it. The server now binds with `--port=0` and I parse the actual URL from the readiness banner (verified `opencode serve --port=0` prints `opencode server listening on http://127.0.0.1:<port>`). This also resolves CodeRabbit's follow-up about partial-line parsing — I now buffer the trailing partial line and only scan complete newline-terminated lines, so a banner split across chunks can't resolve on a truncated URL.
- **Errored turns `finish(0)` (3b):** errored turns and `promptAsync` failures now `finish(1)`, and `emitAgentError` sets `errorEmitted` (idempotently) so ExitHandler's exit-based detection can't emit a duplicate generic error.
- **Readiness listener retention (3):** handlers are detached once readiness settles, and the pipes are `resume()`d to keep draining (done in `c87eb0b`).

On the last robustness point — **`session.error` without `sessionID`**: this is a deliberate call I'd like your read on. Because the shared server multiplexes one SSE stream across every concurrent turn (each turn has its own subscription + translator), treating a session-less error as "mine" meant a single global error aborted *every* active turn. I switched to **exact `sessionID` match**: per-turn provider/API errors carry the `sessionID` and are still surfaced normally; only genuinely unattributable global errors are dropped. The trade-off is that such an error won't force-end a turn — it ends on the normal `session.idle`. If you'd prefer we never drop them, the clean fix is a per-turn watchdog timeout as a safety net; happy to add that as a follow-up. Tests cover both the matched and dropped cases.

Validated locally: prettier, all three tsc configs, repo-wide ESLint, full suite (31,601 passed). Ready for another pass whenever you have a chance — thanks again!

athal7:
Addressed the permission-reply timeout finding in `5b0f2858`: the reply POST now uses `AbortSignal.timeout(5000)`, so an unresponsive `opencode serve` surfaces as a failed decision (returned `false` via the existing catch) instead of hanging `replyToPermission`. Full local gate green (prettier, tsc, ESLint, 31,610 tests).

jSydorowicz21:
Hey @athal7 same request for this one, can you retarget this at rc when you get a chance, since we're changing the whole spawn path I'd like a little cook time on this one too! Thanks again for all the contributions

athal7:
Done — retargeted this PR at `rc`. Makes sense to give the spawn-path change some cook time there. Ping me if you'd like anything rebased once it's had a soak.

athal7:
Follow-up note on coworking + the OpenCode SDK-serve transport (documenting a known limitation, tracked for later).

The `rc` coworking feature injects a per-session `MAESTRO_COWORKING_SESSION_ID` (and per-window `MAESTRO_COWORKING_SOCKET_OVERRIDE`) into the agent CLI env in `ProcessManager.spawn()`, so the coworking MCP subprocess can scope `list_terminals`/`read_terminal` to the calling session. That model assumes a per-spawn process.

The SDK-serve transport uses a **single, process-wide `opencode serve`** shared across sessions (keyed by an env fingerprint in `OpencodeServerManager.buildServerKey`). A per-session value baked into that shared server's env at start can't be correct for all sessions, and folding it into the fingerprint would spin up one server per session, defeating the shared-server design.

Resolution in `6818e6a` (`fix(opencode): skip coworking env injection on the SDK-serve path`): the serve decision now runs on the raw config **before** coworking injection and returns early, so no coworking env is threaded onto the serve path. The coworking bridge **fails closed** for the resulting env-less MCP subprocess (its parent is the shared `opencode serve` PID, not a tracked agent CLI → PID-tree resolution finds no owner → tools denied), so there's **no cross-session leak**. Added a regression test asserting the serve path forwards no coworking env.

Net: **not a regression** — coworking keeps working for PTY/child OpenCode, Claude Code, and Codex. Coworking *over the serve transport* is a new, not-yet-supported combination; making it work needs a **per-request/per-session coworking mechanism** on the serve path, which is a follow-up (the shared server can't carry a per-session value in its startup env).

jSydorowicz21:
@greptile re-review please

jSydorowicz21:
Merging into rc behind a plugin gate while we validate how this functions and how to serve the cowork mcp

PR #1172, opened 2026-07-06
feat(agents): Claude Code standard permission mode with live permission relay

athal7:
## What & why

Adds a **`standard` permission mode for Claude Code** backed by a live, interactive **allow/deny relay** - the missing piece that makes standard mode actually usable.

On Claude Code's API/print path, `standard` mode is non-functional without something to answer permission questions: Claude **aborts the whole run** on the first non-allowed tool call. This PR wires Claude's `--permission-prompt-tool` up to a Maestro UI prompt so the user can approve or deny each tool call interactively.

Supersedes #1155 (3-way permission mode), rebuilt cleanly on `main` and **scoped to Claude Code only** - no other-agent changes.

## Permission model (3-way)

- **Full Access** - bypass all prompts (`--dangerously-skip-permissions`)
- **Standard** - ask per tool call, via the relay (new)
- **Read-Only** - plan mode (`--permission-mode plan`)

`permissionMode` takes precedence over the legacy `yoloMode`/`readOnlyMode` booleans, which remain as a fallback for older configs. **Default behavior is unchanged** (unset resolves to full access).

## How the relay works

```
claude --print                         Maestro (main)
  │  --permission-prompt-tool mcp__maestro_permissions__approve
  │  --mcp-config {stdio: bridge.js}
  ▼
bridge.js (stdio MCP server) ──unix socket (0600)──▶ PermissionRelayServer
                                                          │
                                                     IPC  ▼
                                              PermissionPrompt modal
                                              [Allow]   [Deny]
```

- Claude spawns a tiny **stdio MCP bridge**; when it wants a non-allowed tool it calls the bridge's `approve` tool.
- The bridge dials back to a Maestro-owned **unix-domain socket** (`0600`, no TCP / no network surface; named pipe on Windows). A per-spawn token gates the connection.
- The main process surfaces the request to the renderer over the existing `responseChannel`-style IPC; a Zustand store + `PermissionPrompt` modal render it. The decision routes back through the socket to unblock Claude.
- ~300s auto-deny; **Escape denies** (fail-safe).

## Safety

- **claude-code API path only.** The interactive (maestro-p/TUI) path renders its own native prompts.
- **SSH + standard fails loud** - the local socket can't mediate a remote spawn, so it never silently downgrades.
- New **`supportsStandardPermissionMode`** capability gates the toggle so `standard` is hidden for agents without a working relay (only claude-code today) rather than exposing a non-functional option.

## Packaging

The bridge is bundled to a single self-contained file (`build-permission-relay-bridge.mjs`) and shipped **outside `app.asar`** via `extraResources`, because it runs as plain Node (`ELECTRON_RUN_AS_NODE`) and can't read inside the asar - same constraint as `maestro-p`.

## Verification

Verified end to end:
- Unit tests (relay core, arg building, capability gating)
- Transport round-trip (JSON-RPC over the socket)
- The **real `claude` 2.1.185 CLI** (Write triggers the relay; safe read-only Bash is auto-approved by Claude itself)
- The **live Electron UI** - Allow writes the file, Deny suppresses it, Claude reports each outcome
- A **packaged `.app`** - the bridge resolves outside asar under the real `resourcesPath`

Full suite: **31,615 passing**, lint + prettier clean.

## Summary by CodeRabbit

* **New Features**
  * Added a permission approval modal with **Allow/Deny** and Escape-to-deny.
  * Introduced **Full Access**, **Standard**, and **Read Only** permission modes with toolbar cycling, capability-aware visibility, and updated labels/tooltips.
  * Added a desktop permission-relay bridge so **Standard** tool access is handled consistently during spawning and replay.
* **Bug Fixes**
  * Made permission-mode behavior consistent across Auto Run, queued actions, remote spawns, and interactive replay, including correct coercion to **Read Only** when forced.
* **Chores**
  * Updated desktop packaging/build to include the new permission-relay bridge artifact.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,12 +26,13 @@
 		"dev:web-desktop": "vite --config vite.config.web-desktop.mts",
 		"build:web-desktop": "vite build --config vite.config.web-desktop.mts",
 		"dev:win": "powershell -NoProfile -ExecutionPolicy Bypass -File ./scripts/start-dev.ps1",
-		"build": "npm run build:main && npm run build:preload && npm run build:renderer && npm run build:web-desktop && npm run build:cli && npm run build:maestro-p",
+		"build": "npm run build:main && npm run build:preload && npm run build:renderer && npm run build:web-desktop && npm run build:cli && npm run build:maestro-p && npm run build:permission-relay-bridge",
 		"build:main": "tsc -p tsconfig.main.json",
 		"build:preload": "node scripts/build-preload.mjs",
 		"build:cli": "node scripts/build-cli.mjs",
 		"gen:cli-reference": "node scripts/gen-cli-reference.mjs",
 		"build:maestro-p": "node scripts/build-maestro-p.mjs",
+		"build:permission-relay-bridge": "node scripts/build-permission-relay-bridge.mjs",
 		"build:renderer": "vite build",
 		"package": "node scripts/set-version.mjs npm run build && node scripts/set-version.mjs electron-builder --mac --win --linux",
 		"package:mac": "node scripts/set-version.mjs npm run build && node scripts/set-version.mjs electron-builder --mac",
@@ -118,6 +119,10 @@
 					"to": "maestro-p.js"
 				},
 				{
+					"from": "dist/cli/permission-relay-bridge.js",
+					"to": "permission-relay-bridge.js"
+				},
+				{
 					"from": "src/prompts",
 					"to": "prompts/core",
 					"filter": [
@@ -165,6 +170,10 @@
 					"to": "maestro-p.js"
 				},
 				{
+					"from": "dist/cli/permission-relay-bridge.js",
+					"to": "permission-relay-bridge.js"
+				},
+				{
 					"from": "src/prompts",
 					"to": "prompts/core",
 					"filter": [
@@ -202,6 +211,10 @@
 				{
 					"from": "dist/cli/maestro-p.js",
 					"to": "maestro-p.js"
+				},
+				{
+					"from": "dist/cli/permission-relay-bridge.js",
+					"to": "permission-relay-bridge.js"
 				},
 				{
 					"from": "src/prompts",
--- a/scripts/build-permission-relay-bridge.mjs
+++ b/scripts/build-permission-relay-bridge.mjs
@@ -1,0 +1,55 @@
+#!/usr/bin/env node
+/**
+ * Build script for the permission-relay stdio MCP bridge using esbuild.
+ *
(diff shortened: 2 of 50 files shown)


pedramamini:
Thanks for this, @athal7 - genuinely clean piece of work, and much appreciated the thorough writeup plus the end-to-end verification notes (real CLI, live UI, and a packaged `.app`).

I reviewed the PR alongside the Greptile and CodeRabbit passes, and every substantive finding they raised has already been resolved in your follow-up commits (`4e3bdf0`, `93b8ac4`):

- **Socket permission race** (`listen()` vs `chmod` window): closed by constraining the umask to `0o177` around `listen()`, with the `chmod` kept as a backstop.
- **Relay token in argv + Windows shell mangling of inline `--mcp-config`**: resolved by writing the MCP config to a `0600` temp file and passing it by path, with deletion wired into session-exit cleanup.
- **`clearSession` dropping pending requests**: now sends an explicit deny so the awaiting bridge call unblocks immediately instead of waiting on the ~300s registry timeout.
- **`respondPermission` IPC rejection**: now caught and logged via `sendDecision` rather than an unguarded fire-and-forget.
- **Fail-safe focus**: `autoFocus` moved to Deny so a stray Enter/Space can't approve a tool call.
- **Windows named-pipe ACL claim**: comment corrected to describe the actual guarantee.

No merge conflicts against `main` and CI is green.

Only nitpicks remain, none blocking, entirely at your discretion:
- `integration.ts` still calls `win.webContents.send` without a `try/catch` (the destroyed-window guard covers the realistic case; CodeRabbit's note was defense-in-depth).
- Cosmetic ordering of `PERMISSION_PROMPT` in `modalPriorities.ts`.
- Optional extra test coverage for standard-mode propagation and cross-spawn token isolation.

Approving. Nice job scoping this cleanly to claude-code and failing loud on the SSH path.

jSydorowicz21:
@athal7 could you retarget this to rc instead of main please, this is definitely one we'd want a little bit of soak time on since it changes permissions pathing. Thanks again for revising this pr and adding this feature, I know a lot of people are going to appreciate it

jSydorowicz21:
@greptile please re-review

pedramamini:
Hey @athal7 - this now conflicts with `rc` (the Claude Code agent + permission plumbing has moved underneath it). Could you rebase onto the latest `rc` and resolve the conflicts, then re-push? Keen to get the standard permission mode + live relay in once it is green.

jSydorowicz21:
I rebased, merging to rc

PR #1179, opened 2026-07-08
Refactoring

reachrazamair:
## Summary

Decomposes two large main-process files into focused, single-purpose modules. No behavior changes; this is a pure structural refactor.

- `src/main/app-lifecycle/window-manager.ts` (655 lines removed) is split into six modules under `src/main/app-lifecycle/`: guest webview security hardening, main-window navigation guards, spell-check context menu, window crash handlers, window positioning, and dev-mode auto-updater stubs. `window-manager.ts` keeps the shared `createBrowserWindow` factory, multi-window registry wiring, and plugin panel hardening, and now delegates to the extracted `attach*` helpers.
- `src/main/ipc/handlers/process.ts` (1,332 lines removed) has its `process:spawn` orchestration moved into `src/main/ipc/handlers/process/`: spawn types, Claude transcript sanitization, Claude spawn-context resolution, local interactive (`maestro-p`) spawn handling, session-mode persistence, SSH spawn wrapping, and the top-level `handle-spawn.ts` orchestrator. `registerProcessHandlers` remains the IPC entry point in `process.ts`; other handlers (`write`, `kill`, `runCommand`, etc.) are untouched.
- Local `maestro-p` spawns are routed through `applyClaudeSpawnDecision`, aligning desktop spawn behavior with Cue, group chat, and tab naming.
- Fixes a broken import in `window-manager.test.ts` left over from the module split.

## Behavior Preservation

Intended to be behavior-preserving only.

- Security rationale comments, Sentry noise-filtering comments, and off-screen window placement comments are preserved verbatim in their new locations.
- All existing spawn-path comments are preserved.
- Existing `process.test.ts` coverage (98 tests) is unchanged and still passes against the new module boundaries.

## Test Coverage

- `src/__tests__/main/app-lifecycle/window-manager.test.ts` updated for the new module layout (import fixes only, no new behavior).

## Verification

- [x] `npx tsc --noEmit --pretty` — clean
- [x] `npx vitest run src/__tests__/main/app-lifecycle src/__tests__/main/ipc/handlers/process.test.ts` — 5 files, 220 tests passed

## Manual Test Plan

- Launch the app and confirm the primary window opens, positions correctly, and DevTools/context-menu spell-check suggestions still work.
- Open a secondary window and verify navigation guards and webview security behavior are unchanged.
- Force a renderer crash (or simulate) and confirm crash handling/Sentry reporting still fires.
- Spawn a local Claude Code (`maestro-p`) agent and confirm tab naming, transcript resume, and interactive-mode persistence behave as before.
- Spawn an agent over an SSH remote and confirm the SSH-wrapped command still launches correctly.
- Spawn Codex/OpenCode/other non-Claude agents and confirm `process:spawn` still routes correctly through `handle-spawn.ts`.

## Summary by CodeRabbit

* **New Features**
  * Added spell-check suggestions with “Add to Dictionary” for editable fields.
  * Strengthened embedded webview and main-window navigation/permission protections.
  * Added renderer crash diagnostics with automatic recovery.
  * Added development-mode auto-updater IPC stubs with explicit “disabled” responses.

* **Bug Fixes**
  * Restores windows to visible screen work areas after monitor changes or invalid saved bounds.
  * Improved reliability for SSH/interactive spawning flows and transcript resume handling.

* **Tests**
  * Updated lifecycle tests to use the shared visible-window positioning utility.


pedramamini:
Thanks for this, @reachrazamair - genuinely clean decomposition. I read through both the Greptile and CodeRabbit reviews and diffed every extracted module against the old `process.ts` / `window-manager.ts`, and I can confirm the split is behavior-preserving: every finding the bots raised maps to code that already lived on `rc` and was relocated verbatim (the SSH-local fallback, the API-replay MCP args, the interactive-path `sessionCustomPath` precedence, the crash-handler dedup, and the console-level comment are all identical to their originals). Nice job keeping the comments and logic intact across the move, and thanks for fixing the leftover test import.

Two small things I'd like folded in before merge. Both are zero-risk (comment/style only) and both live in blocks you're relocating anyway, so this is the cleanest moment to touch them:

1. **`window-crash-handlers.ts` (~L132)** - the comment reads `// Level 2 = error` but the guard checks `level === 3` (and the parenthetical correctly lists `3=error`). Since this line is moving into a brand-new file, worth correcting the heading to `Level 3 = error` so a future editor doesn't "fix" the check down to `2`.
2. **`window-crash-handlers.ts` (~L67)** - you already compute `intentionalTermination` just above, so the auto-reload guard can reuse `if (!intentionalTermination)` instead of re-listing `details.reason !== 'killed' && details.reason !== 'clean-exit'`. Keeps the two branches from drifting apart later.

On Greptile's three P1s (SSH remote silently falling back to local, replayed API turns dropping MCP/plugin args, and `sessionCustomPath` losing precedence to the detected binary): those are all real, but they're pre-existing on `rc` and predate this PR, so I'd rather not pull them into a "no behavior changes" refactor. They're worth tracking as separate issues (the SSH-local one especially, given our own CLAUDE.md "fail loudly" guidance for the CLI path), but no need to touch them here.

No merge conflicts against `rc`, so nothing to rebase. Once those two comment/style tweaks land I'm happy to approve. Thanks again for chipping away at these big files!

PR #1188, opened 2026-07-09
refactor: extract AppShell and peel App.tsx callbacks into domain hooks

reachrazamair:
Pull MaestroConsoleInner layout chrome into AppShell.tsx (title bar, loading/ empty states, sidebars, center workspace, plugin docks, toasts, overlays). App.tsx still wires AppModals, AppStandaloneModals, log viewer, and group chat as slots.

Move the remaining inline callbacks into their domain hooks:
- handleCloseCreateGroupModal → useGroupManagement
- handlePRCreated → useWorktreeHandlers (optional rightPanelRef dep)
- handleSaveBatchPrompt → useBatchHandlers
- handleFileSearchSelect → useSessionSwitchCallbacks

App.tsx drops ~230 lines. No behavior change; modal prop assembly stays in App.tsx for a follow-up useAppModalsProps pass.

## Summary by CodeRabbit

* **New Features**
  * Introduced a refreshed app shell with improved top bar labeling, panel layout, and mobile swipe/drawer interactions.
  * File search results now allow opening files directly.
  * Added the ability to save batch prompts from the interface.
* **Bug Fixes**
  * Improved PR creation flow with clearer success feedback and more reliable right-side panel updates/history.
  * Group creation modal workflow now closes more consistently and integrates cleanly with the updated UI flow.

### DIFF
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -180,7 +180,6 @@ import type { RightPanelTab, Session, QueuedItem, CustomAICommand, ThinkingItem 
 import { THEMES } from './constants/themes';
 import { usePluginContributions } from './hooks/usePluginContributions';
 import { resolvePluginTheme } from './utils/pluginThemes';
-import { generateId } from './utils/ids';
 import { getActiveOutputSearchKey } from './utils/outputSearch';
 import { reorderQueueItem } from './utils/executionQueue';
 import { getContextColor } from './utils/theme';
@@ -320,7 +319,6 @@ function MaestroConsoleInner() {
 		// Worktree Modals
 		createWorktreeSession,
 		createPRSession,
-		setCreatePRSession,
 		deleteWorktreeSession,
 		// Tab Switcher Modal
 		setTabSwitcherOpen,
@@ -483,7 +481,7 @@ function MaestroConsoleInner() {
 		encoreFeatures,
 	} = settings;
 
-	// Reset modal-open flags when their Encore Feature toggle is disabledP
+	// Reset modal-open flags when their Encore Feature toggle is disabled
 	useEffect(() => {
 		if (!encoreFeatures.symphony) setSymphonyModalOpen(false);
 	}, [encoreFeatures.symphony, setSymphonyModalOpen]);
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -9,27 +9,16 @@ import React, {
 	type ReactNode,
 } from 'react';
 import { useFocusAfterRender } from './hooks/utils/useFocusAfterRender';
-import { withMonoFallback } from '../shared/fontStack';
 import { isWebDesktop } from './utils/runtimeContext';
 import { isCoarsePointer } from './utils/touch';
-// SettingsModal is now lazy-loaded inside AppStandaloneModals
-import { SessionList } from './components/SessionList';
-import { RightPanel, RightPanelHandle } from './components/RightPanel';
 import { slashCommands } from './slashCommands';
-import { AppModals, type PRDetails, type FlatFileItem } from './components/AppModals';
+import { AppModals } from './components/AppModals';
 import { AppStandaloneModals } from './components/AppStandaloneModals';
+import { AppShell } from './components/AppShell';
 import { initializeRendererPrompts } from './services/promptInit';
-import { ErrorBoundary } from './components/ErrorBoundary';
-import { MainPanel, type MainPanelHandle } from './components/MainPanel';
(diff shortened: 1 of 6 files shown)


pedramamini:
Thanks for taking this on, @reachrazamair - `App.tsx` has needed this for a while, and peeling the callbacks into their domain hooks (rather than just relocating them) is exactly the right shape. A few things to clean up before this goes in.

I checked the refactor's core claim of "no behavior change" and it holds up:

- The `AppShell` child order and every render condition match the original tree verbatim (title bar, modals, standalone modals, loading, empty state, `SessionList`, plugin left dock, mobile backdrop, edge-swipe zones, log viewer, group chat, `MainPanel`, plugin main dock, `RightPanel`, plugin right dock, toasts, center flash, thought stream). The old `<>` fragment only ever wrapped the shell `div`, so nothing was dropped.
- No temporal dead zone problems from the new hook deps: `rightPanelRef` is declared at line 869 and consumed at 1196, `handleFileClick` at 1208 and consumed at 1548.
- The switch to point-in-time store reads is equivalent or better. `selectActiveSession(useSessionStore.getState())` is exactly what `useActiveSession()` subscribes to, and `getData('createPR')?.session ?? null` matches how `useModalHandlers` derived `createPRSession`.
- `handleCloseCreateGroupModal` with `[]` deps is safe: `setCreateGroupModalOpen` is a local `useState` setter, so it is stable. Matches the existing `createNewGroup` pattern right above it.
- All four extracted callbacks are still wired to the same props (`onPRCreated`, `onCloseCreateGroupModal`, `onSaveBatchPrompt`, `onFileSearchSelect`).

No merge conflicts, and both CI matrix legs are green, so no rebase needed.

## Please address

**1. Dead code left behind by the extraction (`src/renderer/App.tsx`)**

Two symbols are now unreferenced now that `handlePRCreated` moved out:

- Line 183: `import { generateId } from './utils/ids';`
- Line 323: `setCreatePRSession,` destructured from `useModalHandlers`

CI does not catch these: `npm run lint` runs `tsconfig.lint.json`, which sets `noUnusedLocals: false`, and ESLint reports them as warnings rather than errors. But under the repo's strict `tsconfig.json` they are hard errors:

```
src/renderer/App.tsx(183,1): error TS6133: 'generateId' is declared but its value is never read.
src/renderer/App.tsx(323,3): error TS6133: 'setCreatePRSession' is declared but its value is never read.
```

I confirmed the parent commit type-checks clean under the same config, so these two are introduced here. `createPRSession` (line 322) is still used and should stay.

**2. Typo in a comment (`src/renderer/App.tsx:483`)**

A stray `P` rode along during the cut: `// Reset modal-open flags when their Encore Feature toggle is disabledP`. Greptile flagged this one too.

**3. Em-dash in `src/renderer/components/AppShell.tsx:2`**

`* AppShell — spatial layout for MaestroConsoleInner.` uses a U+2014 em-dash. `CLAUDE.md` prohibits em-dashes and en-dashes everywhere, including code comments. A spaced hyphen works here. This is the only one in the diff.

**4. Hoist the `rightPanelRef` fallback (`src/renderer/hooks/worktree/useWorktreeHandlers.ts:143`)**

Both bots caught this and they are right. `const { rightPanelRef = { current: null } } = deps;` mints a fresh object every render, and it sits in `handlePRCreated`'s `useCallback` dep array, so any caller that omits the dep gets the callback rebuilt on every render.

To be clear about the blast radius: `App.tsx` is the only production caller and it always passes the ref, so nothing is broken today. The bare `useWorktreeHandlers()` calls in `src/__tests__/renderer/hooks/useWorktreeHandlers.test.ts` are what exercise the default path. Still worth a module-level `const EMPTY_RIGHT_PANEL_REF` so it cannot bite a future caller.

## Optional

**Derive the shell padding from `showTitleBar` (`AppShell.tsx:106-124`).** You compute `showTitleBar`, then re-spell its De Morgan inverse three lines later for the `pt-0` / `pt-10` choice. They agree today, but they can silently drift. CodeRabbit's suggested `showTitleBar ? 'pt-10' : 'pt-0'` is a clean fix.

## One you can skip

CodeRabbit flagged `setCreatePRSession(null)` not being in a `finally` block in `handlePRCreated` and rated it Major. That behavior is unchanged from what was in `App.tsx` before this PR: the same `await window.maestro.history.add(...)` preceded the same unguarded clear. It is a pre-existing issue, not a regression, and wrapping it in `try/catch/finally` would be the one real behavior change in an otherwise behavior-preserving PR. I would leave it for a follow-up unless you feel strongly. Your call.

Once 1 through 4 are in, this is good to merge. Nice work.

PR #1191, opened 2026-07-10
feat(settings): dropdown of known auth accounts for CLAUDE_CONFIG_DIR / CODEX_HOME

jSydorowicz21:
## What & why

Community request ([Discord](https://discord.com)): people running multiple Claude/Codex accounts currently have to remember and hand-type `CLAUDE_CONFIG_DIR` / `CODEX_HOME` paths in agent settings. This turns those two env values into a **dropdown of already-known account paths** with a `Custom…` free-text escape hatch.

## How

- New `agents:getKnownAuthDirs` IPC handler enumerates known paths **only from persisted agent/session `customEnvVars`**, canonicalized/deduped via the existing `resolveConfigDirKey`/`resolveCodexHomeKey` resolvers. **No disk scanning** — preserves the deliberate invariant that avoids triggering OAuth popups.
- SSH/URI-backed sessions are excluded (remote paths live on the remote host); SSH agent configs keep plain manual input.
- Shared `AuthPathValueInput` select+custom component used in both global Environment settings and agent config panels. Known value preselected; empty known-list falls back to exactly today's text input.
- Extended the existing absolute-path validation to `CODEX_HOME` (previously `CLAUDE_CONFIG_DIR` only).

## Scope

Local paths only. SSH-side enumeration and sessions-list override badges are follow-ups.

## Testing

- 4 test files / 167 tests: IPC handler shape, preload exposure, EnvVarsEditor dropdown behavior (magic keys only, custom reveal, empty-list fallback), AgentConfigPanel integration.

## Summary by CodeRabbit

* **New Features**
  * Added known authentication path suggestions for Claude Code and Codex environment variables.
  * Users can select a known path or enter a custom path in environment settings.
  * Authentication paths are displayed using a shorter home-directory format when applicable.
* **Bug Fixes**
  * Added validation to ensure custom Codex home paths are absolute.
  * Excluded remote or non-local sessions from local path suggestions.
* **Tests**
  * Expanded coverage for path discovery, selection, validation, and environment variable updates.

### DIFF
--- a/src/__tests__/main/ipc/handlers/agents.test.ts
+++ b/src/__tests__/main/ipc/handlers/agents.test.ts
@@ -5,6 +5,7 @@
  */
 
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import path from 'path';
 import { ipcMain } from 'electron';
 import {
 	registerAgentsHandlers,
@@ -186,6 +187,7 @@ describe('agents IPC handlers', () => {
 				'agents:getRemoteMaestroPAvailable',
 				'agents:getClaudeUsageSnapshots',
 				'agents:getClaudeUsageAccountKeys',
+				'agents:getKnownAuthDirs',
 				'agents:getLimitResetAt',
 				'claude:usage:refresh-all',
 				'agents:getCodexUsageSnapshots',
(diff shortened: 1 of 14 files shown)


pedramamini:
@jSydorowicz21 Thanks for the contribution! This is a well-scoped feature and the write-up is excellent. The "no disk scanning, enumerate only from persisted `customEnvVars`" invariant is exactly the right call, and calling that out explicitly in a code comment next to the handler is appreciated. Tests are thorough and both CI legs are green.

A few things I would like addressed before merge.

### 1. `ABSOLUTE_PATH_KEYS` regression (blocking)

`src/renderer/components/Settings/EnvVarsEditor.tsx`

The `Set` was swapped for a plain object literal, and the lookup went from `.has(entry.key)` to `ABSOLUTE_PATH_KEYS[entry.key]`. A bare object literal inherits from `Object.prototype`, so that indexed lookup now returns truthy for inherited members:

```js
ABSOLUTE_PATH_KEYS['constructor']    // -> true
ABSOLUTE_PATH_KEYS['toString']       // -> true
ABSOLUTE_PATH_KEYS['valueOf']        // -> true
ABSOLUTE_PATH_KEYS['hasOwnProperty'] // -> true
ABSOLUTE_PATH_KEYS['__proto__']      // -> truthy (Object.prototype)
```

`entry.key` is free text from the key input, so an env var named any of those now falsely trips `"<key> must be an absolute path (starting with /)"` and blocks the save. The `Set` version was immune to this.

The refactor also is not needed for this PR: the only change required was adding `CODEX_HOME` to the existing collection. Please revert to:

```ts
const ABSOLUTE_PATH_KEYS = new Set<string>(['CLAUDE_CONFIG_DIR', 'CODEX_HOME']);
```

and restore the `.has(entry.key)` call. Smaller diff and no behavior change beyond the one you intended.

### 2. Stale suggestions after the SSH toggle flips (blocking)

`src/renderer/hooks/agent/useKnownAuthDirs.ts`

Greptile flagged this and I confirmed it. `AgentConfigPanel` takes `isSshEnabled` as a prop and passes `useKnownAuthDirs(!isSshEnabled)`. Toggling an agent to SSH flips the prop on the existing component instance rather than remounting it, so the effect hits `if (!enabled) return;` and leaves the previously fetched local paths in state. The dropdown keeps rendering local `CLAUDE_CONFIG_DIR` / `CODEX_HOME` options for what is now a remote agent, and the user can save a local-only path onto an SSH config.

This defeats the "SSH agent configs keep plain manual input" invariant from your PR description. Greptile's suggested fix is right:

```ts
useEffect(() => {
    if (!enabled) {
        setKnownAuthDirs(EMPTY_KNOWN_AUTH_DIRS);
        return;
    }
    // ...
```

A regression test that flips `isSshEnabled` on a mounted `AgentConfigPanel` and asserts the plain text input comes back would be worth adding, since the current tests only cover the initial-mount case.

### 3. Custom input collapses mid-typing (non-blocking, please fix)

`src/renderer/components/shared/AuthPathValueInput.tsx`

```ts
useEffect(() => {
    setShowCustomInput(!hasKnownValue);
}, [hasKnownValue]);
```

`hasKnownValue` is derived from `value`, so this effect force-collapses the custom text input the instant the typed value happens to equal a known path. Concretely: a user with two accounts picks `Custom…` and hand-types the second account's path. On the keystroke that completes the exact match, the input unmounts under the cursor and the select jumps. Gating on whether the user explicitly chose `Custom…` (rather than re-deriving it from the value on every change) avoids this.

### 4. SSH-enabled check diverges from codebase convention (nit)

`src/main/ipc/handlers/agents.ts:84`

`sshRemoteConfig.enabled === true` is stricter than everywhere else in the codebase, which uses a truthy check:

- `src/renderer/stores/tabStore.ts:630` -> `!!(session.sessionSshRemoteConfig?.enabled || session.sshRemoteId)`
- `src/renderer/stores/agentStore.ts:413` -> `!!session.sshRemoteId || !!session.sessionSshRemoteConfig?.enabled`
- `src/renderer/utils/spawnHelpers.ts:49` -> `opts.session.sshRemoteId || opts.session.sessionSshRemoteConfig?.enabled`

I do not think Greptile's "remote paths leak" scenario is reachable without a corrupted store, since `enabled` is typed `boolean`. But the inconsistency means a malformed persisted value would be treated as remote everywhere else and local here, which is the wrong direction to fail. Matching the truthy convention costs nothing.

### 5. Reference the shared type in `global.d.ts` (nit)

CodeRabbit's point is valid. `src/renderer/global.d.ts:1286` re-declares the `KnownAuthDirs` shape inline while neighboring entries use `import('../shared/...')`. Since you already added `src/shared/authPaths.ts`, use it:

```ts
getKnownAuthDirs: () => Promise<import('../shared/authPaths').KnownAuthDirs>;
```

### One question, not a change request

`collectKnownAuthPaths` only records explicitly-set env var values, so the default account (`~/.claude` / `~/.codex`) never appears in the dropdown. A user with one custom account plus the default will see only the custom one listed and has to go through `Custom…` to get back to the default. I understand this follows from the no-disk-scan rule, and I am not asking you to scan. Would seeding the list with the two resolver defaults (which `resolveConfigDirKey` / `resolveCodexHomeKey` already return for an unset env, no filesystem access needed) be reasonable here, or do you see a downside?

No merge conflicts, so nothing to rebase. Once 1 and 2 are handled I am happy to take another look.

PR #1192, opened 2026-07-10
feat(groups): standard icons + label colors for agent-panel folders

jSydorowicz21:
## What & why

Community request ([Discord](https://discord.com)): group folders in the agents left panel are emoji-only. This adds **standard icons with label colors** as an alternative — pick an emoji OR a Lucide icon from a curated 16-icon set, plus an optional accent color from an 8-color palette.

## Model change (groundwork for nested groups)

- `Group` gains optional `kind: 'user' | 'worktree'`, `icon`, `color`; `emoji` retained.
- **Kills the emoji-as-sentinel hack**: worktree groups were detected by the literal `group.emoji === '🌳'` scattered across SessionList (including delete-cascade logic). All sites now go through a single `isWorktreeGroup()` helper (`kind === 'worktree'` with legacy-emoji fallback), so old persisted groups behave identically. Repo grep confirms the literal survives only inside the helper.
- User-created groups persist `kind: 'user'` from both desktop and remote/web creation paths.

This is deliberately scoped as the model groundwork PR — nested groups (Company > Project > Agents) builds on it next.

## UI

- Create/Rename group modals get an appearance picker: emoji (existing picker) or icon + color; the two are mutually exclusive.
- Group headers render the icon tinted with the label color when set; color accents label/icon only — rows are untouched.
- Rename plumbing extended through the existing modalStore/AppModals/QuickActions threading (no parallel channel).

## Testing

- 6 test files / 312 tests: `isWorktreeGroup` matrix (kind/legacy/both/neither), picker mutual-exclusion transitions, modal round-trips preserving unrelated persisted fields, SessionList render + delete-cascade parity.

## Summary by CodeRabbit

* **New Features**
  * Added full group appearance customization: standard icons and label colors alongside emoji.
  * Expanded group create/rename flows (including Quick Actions) to save, display, and clear icon/color.
  * Session list now renders group icons and applies the selected label color.
* **Bug Fixes**
  * Improved worktree group detection to handle both modern flags and legacy emoji data, fixing related context-menu and deletion behavior.
* **Tests**
  * Expanded coverage for icon/color selection, rename/create modal state, Quick Actions wiring, and worktree detection.


pedramamini:
Hey @jSydorowicz21 - thanks a lot for this one. The emoji-as-sentinel cleanup is genuinely welcome, the test coverage is thorough (312 tests across 6 files), and threading rename state through the existing modalStore/AppModals/QuickActions channel instead of opening a parallel one was the right call.

There is one blocking issue plus a few smaller things before we merge.

### 1. Blocking: `isWorktreeGroup` ignores an explicit `kind` (`src/shared/types.ts:185`)

Both Greptile and CodeRabbit flagged this, and I confirmed it locally:

```ts
return group.kind === 'worktree' || group.emoji === '🌳';
```

`{ kind: 'user', emoji: '🌳' }` returns `true`. Since `CreateGroupModal` now stamps `kind: 'user'` on every new group, and the emoji picker exposes the full emoji-mart set, a user who simply picks the tree emoji for an ordinary folder gets worktree treatment: the context menu offers "Remove Group and Agents", and delete routes through `deleteWorktreeGroup`, which kills every agent's PTYs in that group. The remote/web creation path has the same exposure.

The fix both bots suggest is the right one:

```ts
return group.kind === 'worktree' || (group.kind === undefined && group.emoji === '🌳');
```

I checked out your branch to verify rather than take it on faith: the bug reproduces on `6f80d79`, the one-line change fixes it, and all 312 tests in the six affected files still pass afterward. Legacy persisted groups carry no `kind`, so they keep their worktree behavior.

### 2. Context worth having: nothing constructs `kind: 'worktree'`

This changes how the helper should be read. `git log -S` shows the only producer of a tree-emoji group lived in `App.tsx` and was removed in `5d91dd163`. Repo-wide, `kind: 'worktree'` now appears only in your new tests.

So the legacy emoji fallback is currently the *only* live detection path, and `'worktree'` is an unconstructed variant. That is fine as groundwork for nested groups, but please add a comment on the helper saying so. Otherwise the next reader takes `kind === 'worktree'` for the primary path and assumes the emoji check is vestigial, when it is actually the load-bearing one.

### 3. Selecting an icon blanks the emoji, which is lossy

In `GroupAppearancePicker`, choosing an icon calls `onEmojiChange('')`, and `RenameGroupModal` persists that empty string into the model.

Two consequences:

- Renaming a legacy worktree group (`kind` undefined, `emoji: '🌳'`) and picking an icon silently clears the sentinel, so the group quietly stops being a worktree group.
- `SessionList` resolves the icon via `GROUP_ICON_OPTIONS.find((o) => o.id === group.icon)?.Icon`. If that id is ever unknown (an icon renamed or dropped from the curated 16), it falls back to `<span>{group.emoji}</span>` and renders nothing at all.

You do not need to blank it. `SessionList` already prefers the icon whenever one is set, so leaving `emoji` intact gives you render-time mutual exclusion for free, preserves the sentinel, lets "clear icon" restore the previous emoji, and provides a real fallback glyph. Keep the mutual exclusion in the picker UI, just do not destroy the value in the model.

### 4. Nit: extend the negative test

CodeRabbit's nitpick is a good one. The current negative case uses `📁`, so it passes even with the bug present. Add the case that would actually have caught #1:

```ts
expect(isWorktreeGroup(createGroup({ kind: 'user', emoji: LEGACY_WORKTREE_EMOJI }))).toBe(false);
```

### 5. Nit: hardcoded palette

`GROUP_LABEL_COLORS` is eight fixed hex values applied directly as `style={{ color }}`, independent of the active theme. On lighter themes some of these (yellow `#EAB308` especially) will be low contrast against the panel background. Not a blocker, but worth a look.

---

No merge conflicts, the branch is clean against `rc`, so no rebase needed. `lint-and-format` is green; both `test (ubuntu-latest)` and `test (windows-latest)` were still running when I looked, and per our contribution rules both legs need to pass before merge.

Ping me once #1 and #3 are addressed and I will take another pass. Thanks again for the contribution.

jSydorowicz21:
Folded into #1193 — the branch is stacked on this one, so #1193 retargeted to rc now carries both the icons/colors groundwork and the nesting feature as one PR.

PR #1194, opened 2026-07-10
Fix(group-chat): debounce draft persistence to stop per-keystroke app re-render

jeonghyeonlee-dev:
## Observed behavior
I set up a group chat and had two agents, `claude_fable` and `codex 5.6 sol`, take turns discussing a topic for three rounds each. Even though the conversation was only a handful of messages long, the whole UI became sluggish: typing in the message box lagged noticeably.

## Root cause
`GroupChatInput` calls `onDraftChange` on every keystroke, which rewrites the global `groupChats` array in the store. `App.tsx` subscribes to `groupChats`, so every character re-renders the entire app tree. With a large message mounted, reconciling that tree on each keystroke is what makes typing feel slow, regardless of how short the conversation is.

## Fix
- Keep the textarea responsive through local state and debounce (250ms) the store write, reusing the existing `useDebouncedCallback` hook.
- `flush()` on blur so the draft is persisted before leaving the field.
- `cancel()` on send and on chat switch so a stale pending write cannot repopulate the box or land on the wrong chat.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved message draft saving by reducing unnecessary draft updates while typing via debounced persistence.
  * Ensured drafts are flushed to storage when leaving the message field and before switching chats.
  * Prevented stale/incorrect drafts from being written to the wrong chat after chat switches.
  * Cancelled pending draft writes and cleared saved drafts after sending a message.
* **Tests**
  * Added coverage for debounced draft persistence, flushing, cancellation, and unmount behavior.

### DIFF
--- a/src/renderer/components/GroupChatInput.tsx
+++ b/src/renderer/components/GroupChatInput.tsx
@@ -643,6 +643,7 @@ export const GroupChatInput = React.memo(function GroupChatInput({
 							ref={inputRef}
 							value={message}
 							onChange={handleChange}
+							onBlur={flushDraft}
 							onKeyDown={handleKeyDown}
 							onPaste={handlePasteWrapped}
 							onDrop={handleDropWrapped}


pedramamini:
@jeonghyeonlee-dev Thanks for this - the diagnosis is spot on and well written up. You correctly traced the jank to `onDraftChange` rewriting the global `groupChats` array on every keystroke, and `App.tsx` subscribing to it. Debouncing the store write while keeping the textarea on local state is exactly the right shape for the fix, and reusing the existing `useDebouncedCallback` hook rather than rolling a new one is appreciated.

Both Greptile and CodeRabbit flagged the chat-switch path, and I went and verified their claims against the code rather than taking them at face value. They're both right, and they're describing two symptoms of the same root cause. I'd like to see this addressed before merge.

### Root cause: the debounced write isn't bound to the chat that produced it

`handleGroupChatDraftChange` (`src/renderer/hooks/groupChat/useGroupChatHandlers.ts:760-766`) resolves the target chat from the store at **call** time:

```ts
const { activeGroupChatId, setGroupChats } = useGroupChatStore.getState();
```

That was harmless when the write was synchronous with the keystroke. Now that it fires up to 250ms later, `activeGroupChatId` may have moved on. Your comment in the switch effect shows you spotted this, and `cancelDraft()` is a reasonable defensive move given the current handler signature - but it trades a correctness bug for a data-loss bug.

### Consequence 1: pending draft is silently dropped on chat switch

In `GroupChatInput.tsx` (~line 233), `cancelDraft()` on switch discards the only copy of the last <=250ms of typing.

The `onBlur={flushDraft}` guard covers mouse-driven switches (mousedown blurs the textarea, flush runs while the old chat is still active). It does **not** cover keyboard switches. `handleKeyDown` deliberately lets meta/ctrl combos bubble:

```ts
// Let global shortcuts bubble up (Cmd+K, Cmd+,, Cmd+/, etc.)
// Don't stop propagation for meta/ctrl key combinations not handled above
return;
```

and `cyclePrev` / `cycleNext` are registered globally (`useMainKeyboardHandler.ts:471-479`), with `useCycleSession` cycling through group chats as well as agents. So:

**Repro:** focus the group chat input, type `hello`, and within 250ms hit the cycle shortcut. Focus never leaves the textarea, so no blur, so no flush. `cancelDraft()` runs, `hello` never reaches the store, and switching back shows the stale draft.

### Consequence 2: same loss on unmount

`GroupChatPanel` is conditionally rendered (`App.tsx:3326`, `groupChats.find(...) && <GroupChatPanel .../>`). Cycling from a group chat to an agent sets `activeGroupChatId` to `null` and unmounts the input. `useDebouncedCallback`'s unmount cleanup only calls `clearTimeout` - it never flushes - so the pending draft is dropped there too. This one isn't fixed by anything in the current diff.

One trap if you go after this: React runs effect cleanups in **declaration order**, and the hook's own cleanup is registered inside `useDebouncedCallback` at the top of the component. A naive `useEffect(() => () => flushDraft(), [flushDraft])` declared further down will find the timer already cleared and no-op. You'll likely need to either flush from a ref-tracked latest value, or add a `flushOnUnmount` option to the hook.

### Suggested direction

Thread the originating `groupChatId` through the debounced write so the store update targets the chat the text was typed into, not whatever is active when the timer fires. Once the write is bound to its chat:

- `flushDraft()` becomes safe on chat switch, and you can use it **instead of** `cancelDraft()` - no lost keystrokes, no cross-chat contamination.
- The switch/timer race CodeRabbit describes closes on its own.
- `cancelDraft()` on send stays exactly as you have it, which is correct.

Note that flushing on switch is only safe *after* the binding change. Flushing today would write the old chat's text onto the new chat, since by the time the effect runs `activeGroupChatId` has already advanced. That is precisely why `cancel()` was the only option available to you.

A test with fake timers covering "type, switch chat before the debounce fires, switch back" would lock this down nicely, plus one for the unmount case.

Everything else looks good to me: `cancel()` on send is right, the dependency arrays are updated correctly, the hook's `flush`/`cancel` identities are stable so the `useCallback` deps don't churn, and there are no merge conflicts against `main`. Ping me when you've pushed and I'll take another look.

jeonghyeonlee-dev:
Addressed the chat-switch and unmount data-loss paths: the debounced write now carries the originating groupChatId, flush() replaces cancel() on switch, useDebouncedCallback gained a flushOnUnmount option, plus fake-timer tests for switch-before-debounce and unmount. Ready for another look.

pedramamini:
Thanks! This targets `main` and currently conflicts. Could you rebase on the latest `main`? I will review and merge once it is clean and CI is green.

jSydorowicz21:
A pending local save can overwrite a newer external draft. Type locally, then let Prompt Composer update the same chat before the 250 ms debounce expires: the effect at `GroupChatInput.tsx:251` adopts the external value into the textarea but leaves the queued local callback intact. When it fires, the store reverts to the old local value.

I reproduced this in Electron. The textarea showed `external update v2` while the store had been overwritten with `local pending v2`. Please cancel or version the queued local write when accepting an external update, and add a fake-timer regression covering local type, external update, then debounce completion.

The original cross-chat timeout concern is fixed on the current head, so I did not repeat it.

jeonghyeonlee-dev:
Rebased this PR on the latest `main` and force-pushed the branch.

While rebasing, the newer external-draft race fix and its regression coverage were already present on `main`, so I skipped the now-duplicate commits. The remaining PR delta is the `onBlur={flushDraft}` safeguard in `GroupChatInput`, which keeps pending draft persistence flushed when the input loses focus.

PR #1195, opened 2026-07-10
feat(plugins): hostViews contribution type — plugin-driven Concerto views

jSydorowicz21:
## What & why

Implements the `hostViews` contribution type that #1178 scoped and deliberately deferred: plugins can now render **native, host-styled views** on the Concerto surfaces (Movement panels, Cadenza HUD cards) by contributing **block data** that the host's BlockView engine renders. Plugin code never executes in the renderer — the host owns rendering, theming, and placement.

## Design

**Contribution (tier 0, data-only):** `hostViews: [{ id, surface: 'movement' | 'cadenza', title, description?, blocks? }]` — a static `blocks` payload lets a zero-code tier-0 plugin ship a live-looking panel. Ids are namespaced `<pluginId>/<localId>`; standard collect/aggregate rules apply (invalid dropped with recorded error, first-wins, built-in wins).

**Runtime (tier 1, brokered):** new `ui:hostView` capability gates `maestro.ui.hostView.update(localId, blocks)` / `.remove(localId)` through the standard table-driven broker path. The handler enforces:
- closed param schema; owner-local **declared** view ids only (undeclared → error)
- host resolves surface/title from the declaration — callers cannot pick surface, title, session, or options
- `decision` payloads rejected (no plugin reply channel in v1)
- `plugins` + `concerto` Encore flags gate at the main boundary; flag off → clean error, nothing buffered

**Lifecycle:** a main-owned `PluginHostViewRegistry` records forwarded views; purges on disable/uninstall/sandbox stop/crash/flag-off (extends `purgePluginData`); replays after renderer load; refreshes tier-0 static blocks on hot reload. Cadenza remove paths never spawn an empty always-on-top HUD.

**Provenance:** plugin views stamp the manifest name; Movement headers and Cadenza cards show `from <plugin name>`.

**Versioning:** `HOST_API_VERSION` 1.8.0 → 1.9.0 (MINOR, additive); `@maestro/plugin-sdk` vendored contracts bumped in lockstep; drift-guard green.

## Testing

- Focused: 9 files / 142 tests (contract validation, broker gating, handler rejection paths, registry lifecycle, renderer provenance).
- SDK: 28 tests incl. type-level drift guard.
- Regression: full existing plugin + pianola + Movement/Cadenza store suites — 48 files / 699 tests.

## Summary by CodeRabbit

* **New Features**
  * Plugins can now declare host-rendered views for Movement and Cadenza surfaces.
  * Authorized plugins can update or remove their declared views at runtime.
  * Plugin-contributed views display their source plugin for clearer attribution.
  * Added an Insights plugin category.
* **Documentation**
  * Updated plugin development guidance with host-view declarations, permissions, size limits, and API requirements.
* **Compatibility**
  * Updated the host API contract to version 1.9.0 and released the plugin SDK as version 0.4.0.


pedramamini:
Hi @jSydorowicz21, thanks so much for this contribution! The `hostViews` design is genuinely clean: keeping plugin code out of the renderer and having the host own rendering, theming, and placement is exactly the right call, and the tier-0/tier-1 split plus the lifecycle purge coverage is thorough. Nice work.

Before we merge, I went through the Greptile and CodeRabbit reviews and verified two issues against the code that I'd like you to address:

**1. Orphaned runtime views survive a declaration removal** (`src/main/plugins/plugin-host-view-registry.ts`, `sync()`) - flagged by both Greptile (P1) and CodeRabbit. The pruning loop does `if (live.source !== 'static') continue;`, so a view that was updated via the tier-1 `update()` path is never pruned once its static declaration disappears entirely (e.g. a manifest hot-reload that drops the `hostViews` entry without disabling or uninstalling the plugin). The stale view stays rendered, `replay()` can re-forward it, and `remove()` can no longer target it because `getDeclared()` now returns `null`. CodeRabbit's proposed fix (prune any live entry when its declaration is fully gone, and keep the `blocks === undefined` check only for `static` entries) looks correct to me.

**2. Cadenza close can miss the main renderer** (`src/main/index.ts`, `forwardPluginHostView`) - Greptile P1. The open path (`deliverCadenza`) can fall back to the main renderer, but the remove path short-circuits on `deliverCadenzaToExistingHud(payload)` and returns as soon as a HUD exists, even if that HUD never rendered the card. So if a card was ever delivered to the main renderer and a HUD is later created (for a different card), the close only reaches the HUD and the in-app card leaks through purge/disable/uninstall/runtime-remove. Granted the plugin open path usually creates the HUD lazily, so the precondition is narrow, but the open/remove asymmetry is real. Please either deliver the close to both surfaces or otherwise guarantee it reaches wherever the card actually lives (or let me know if I've missed a guard that already covers this).

**3. (nice-to-have)** CodeRabbit's nitpick to add a `sync()` regression test for the hot-reload orphan case would be a good guard once #1 lands.

No merge conflicts against `rc`, so no rebase needed. One heads-up: `test (windows-latest)` was still running when I looked, so worth a glance once it settles.

Thanks again for the solid work here. Ping me once these are addressed and I'll take another pass.

PR #1200, opened 2026-07-10
perf(renderer): cut MaestroConsoleInner re-renders from cycling and queue recovery

reachrazamair:
`MaestroConsoleInner` still re-renders on many store updates because handler hooks subscribe to broad Zustand slices. This PR is an early slice of peeling that work toward event-time `getState()` reads and narrow selectors, so App only re-renders when the shell actually needs to paint.

## Summary
- **Test infra:** shared `resetAllStores` / `resetStore` / `resetStores` helpers (Zustand v5 `getInitialState` + Set/Map cloning) and TEST-PATTERNS guidance; adopt in a few suites so resets stop drifting from hand-rolled snapshots.
- **Cycling:** extract plain `cycleSession(dir, deps)` that reads stores at keypress time; keep a thin `useCycleSession` wrapper with a stable callback for the keyboard handler (no reactive store subscriptions).
- **Queue recovery:** stop selecting the full `sessions` array in `useQueueProcessing`; subscribe only to `sessionsLoaded` + an idle-queue signature, and read session objects via `getState()` inside effects.
- Drop the unused `UseCycleSessionDeps` alias in favor of `CycleSessionDeps` only.

## Future Work
Keep cutting App’s session-store subscriptions so streaming and sidebar updates don’t re-render the whole console. Handlers and persistence should read via getState() at event time; only components that paint data should subscribe. Same pattern for useActiveSession and the remaining App-mounted hooks, plus broader resetAllStores adoption in tests.

## Summary by CodeRabbit

* **Improvements**
  * Refined session cycling so navigation stays consistent across starred items, bookmarks, group chats, active filters, and multi-window contexts.
  * Reduced unnecessary UI re-renders during queued agent processing for smoother streaming/console interaction.
  * Strengthened queued-task recovery when sessions move from busy to idle.
* **Documentation**
  * Updated testing guidance with safer, clearer instructions for resetting Zustand store state between test runs.
* **Tests**
  * Added shared store reset helpers and expanded coverage to confirm resets fully isolate state per test.

### DIFF
--- a/docs/agent-guides/TEST-PATTERNS.md
+++ b/docs/agent-guides/TEST-PATTERNS.md
@@ -270,7 +270,7 @@ beforeEach(() => {
 });
 ```
 
-`resetAllStores()` / `resetStore()` / `resetStores(...)` live in `src/__tests__/helpers/resetStores.ts` (re-exported from `src/__tests__/helpers`). They use Zustand v5 `getInitialState()` with replace mode and clone top-level `Set`/`Map` values so in-place mutations cannot poison later tests.
+`resetAllStores()` / `resetStore()` / `resetStores(...)` live in `src/__tests__/helpers/resetStores.ts` (re-exported from `src/__tests__/helpers`). They use Zustand v5 `getInitialState()` with replace mode and deep-clone `Set`/`Map`/array/plain-object defaults so in-place mutations cannot poison later tests. Store actions keep their original function references.
 
 For a single-store unit suite, `resetStore(useSessionStore)` (or `resetStores(useSessionStore, useUIStore)`) is enough.
 
--- a/docs/agent-guides/TEST-PATTERNS.md
+++ b/docs/agent-guides/TEST-PATTERNS.md
@@ -253,24 +253,28 @@ Never hardcode `/`, `\`, or `process.env.HOME` in path assertions or path fixtur
 
 ### Mocking Zustand Stores
 
-Zustand stores are singletons. Two approaches:
+Zustand stores are singletons. Prefer seeding the **real** store with `setState` after a full reset. Do not hand-roll partial default snapshots that drift from each store's create-time initial state.
 
-**Approach 1: Direct setState (preferred for store tests)**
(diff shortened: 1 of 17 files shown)


pedramamini:
Hi @reachrazamair, thanks for the contribution! This is a clean slice of the re-render peeling work, and the event-time `getState()` approach reads well.

A few notes after going through the PR and the Greptile / CodeRabbit reviews:

**Already handled (nice):**
- Greptile's "nested defaults stay shared" and CodeRabbit's "add `useImageAnnotatorStore`" on `resetStores.ts` both look addressed in `e89ec1d` (deep-clone via `cloneCollections` plus the store added to the reset list). Those look good.

**One item worth addressing before merge - CodeRabbit's "Major" on `useQueueProcessing.ts` (runtime recovery re-run):**

I traced this against the current code and it holds. The runtime-recovery effect now triggers on `idleQueuedSignature` instead of the full `sessions` array:

```ts
useEffect(() => {
	if (!sessionsLoaded || !startupRecoveryComplete.current) return;
	...
}, [sessionsLoaded, idleQueuedSignature, dispatchQueuedItem]);
```

`startupRecoveryComplete.current` is a ref, so flipping it to `true` inside the 500ms startup timer does not re-run this effect. If a session becomes idle with a runnable item *during* that startup window, the signature-change edge gets consumed by the early `return` (flag still false), and once the flag flips there's no dependency change to re-fire, so that item can sit until some later unrelated queue/idle event moves the signature again. The old full-`sessions` subscription masked this because it re-ran on constant session churn; the narrower signature selector (which is the whole point of the PR) removes that safety net.

It's a narrow timing window and it self-heals on the next queue event, so not a showstopper, but it's a real behavior change introduced by the optimization. Simplest fix is to lift the completion flag into state (so the effect re-runs when it flips), or re-scan idle+runnable sessions when the 500ms timer fires.

No merge conflicts and both CI legs are green, so no rebase needed. Once that last one is handled I'm happy to approve. Thanks again for pushing this forward!

PR #1201, opened 2026-07-10
feat(plugins): broad mediated UI surfaces + trusted-chrome guard

jSydorowicz21:
## What & why

Extends the plugin interface so plugins can contribute UI into a broad set of **host-mediated** surfaces, while a hardcoded **trusted-chrome** set stays permanently off-limits to any plugin. Host owns every frame and theme; plugin code never runs in the main renderer. This is the "widen mediated mount points" model (chosen over arbitrary UI mutation) — the safe way to let plugins customize most of the app.

## New contributable surfaces

Extends the existing `ui:contribute`/`uiItems` mechanism (declarative label/icon/command/tooltip only) to: `toolbar`, `tabBar`, `sessionRowBadge`, `groupHeaderBadge`, `settingsSection`, `rightPanelTab`, `contextMenuItem`, `emptyState`. The surface union derives from one `UI_SURFACES` catalog — adding a future safe surface is a one-line entry. Items are host-rendered, themed, provenance-tagged (`from <plugin>`), and can only dispatch the plugin's own namespaced command.

## Trusted-chrome guard (the point of this PR)

Four zones are **permanently unavailable** to every plugin surface — declarative *and* the high-trust `ui:render-unsafe` tier:
1. **plugin-management** + enable/disable
2. **permission/consent dialogs**
3. **uninstall / grant-revoke flows**
4. **security indicators** (SSH status, permission-mode, agent identity)

Enforcement is layered:
- Protected zones are **not members** of the contributable `PluginUiSurface` allowlist to begin with.
- `validatePluginUiMount` rejects protected/unknown targets, called at BOTH `parseUiItem` and `aggregateContributions` (parse-time + pre-insertion revalidation), recording a "protected chrome" error.
- **Render-boundary backstop**: `PluginUiItemsSlot` re-checks the positive allowlist and requires canonical `<pluginId>/<localId>` provenance, so a forged renderer object can't address a protected surface.
- `ui:render-unsafe` cannot unlock `ui:contribute`/`ui:panel` (regression-tested).
- **Cutover**: the plugin-management screen previously hosted a sandboxed `PluginPanelSlot`; moved it to neutral Display settings, with a test asserting plugin panels are absent from management controls.

## Testing

- `it.each(PROTECTED_UI_SURFACES)`: every zone rejects declarative + render-unsafe attempts (the security regression test).
- Surface enum/schema validation, aggregation rules (namespacing, first-wins, built-in-wins), render-slot smoke, lifecycle purge (disable → items vanish), flag-off → no UI, SDK drift.
- `HOST_API_VERSION` 1.13.0 (additive/MINOR), `@maestro/plugin-sdk` 0.8.0 in lockstep. `bun run lint` + prettier + eslint clean.

Rebased onto current `rc`; the additive host API and SDK versions are linearized at 1.13.0 and 0.8.0 respectively.

## Summary by CodeRabbit

* **New Features**
  * Added plugin UI extension points across empty states, tabs, session lists, context menus, settings, and panels.
  * Upgraded plugin UI actions with **inline/menu** presentation, per-item icons, and tooltips; added an additional UI item slot in the right-side tab header.

* **Improvements**
  * Advanced the plugin/host UI contract and **tightened “trusted chrome” protections** with an explicit protected-surface denylist.
  * Clarified that **`ui:render-unsafe` remains restricted** to host-approved, non-protected regions.
  * Updated special `placement: "settings"` behavior to render only in neutral display settings.

* **Documentation**
  * Refreshed plugin development guidance, including `uiItems` shape and trusted-chrome constraints.

* **Tests**
  * Expanded contract drift and UI mount-policy coverage, plus renderer hardening tests.

### DIFF
--- a/CLAUDE-PLUGINS.md
+++ b/CLAUDE-PLUGINS.md
@@ -11,7 +11,7 @@ A plugin is one folder under `<userData>/plugins/` containing a `plugin.json` ma
 - Entire system is gated on `encoreFeatures.plugins === true` (off by default), re-read per call.
 - Every `plugins:*` IPC channel throws the sentinel `'PluginsDisabled'` when the flag is off, so the renderer can distinguish "feature off" from "no plugins installed". The gate runs OUTSIDE `withIpcErrorLogging` so the sentinel is not logged as a real failure.
 - `PluginManager.getActiveRecords()`, `getContributions()`, and `getAgentRegistry()` all return empty when the flag is off, regardless of what is on disk.
-- `HOST_API_VERSION = '1.12.0'` (`src/shared/plugins/host-api.ts`) is the single source of truth for the host surface version.
+- `HOST_API_VERSION = '1.13.0'` (`src/shared/plugins/host-api.ts`) is the single source of truth for the host surface version.
 
 ## File map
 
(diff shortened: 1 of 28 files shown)


pedramamini:
Thanks for this, @jSydorowicz21 - really nice work, and thank you for the contribution. The layered trusted-chrome guard is well thought through: keeping `PROTECTED_UI_SURFACES` out of the positive `UI_SURFACES` allowlist, failing closed in `validatePluginUiMount` at both parse-time and aggregation, and the render-boundary backstop in `PluginUiItemsSlot` (positive allowlist plus canonical `<pluginId>/<localId>` provenance) together give genuine defense in depth. The `it.each(PROTECTED_UI_SURFACES)` regression test is exactly the right way to lock that down.

I reviewed the PR alongside the Greptile and CodeRabbit feedback, and I'd like a couple of small changes before we merge. Both are real, and both trace back to `PluginUiItemsSlot`:

1. **Group-header plugin buttons can't be keyboard-activated** (`SessionList.tsx:1540`, CodeRabbit "Major"). The `groupHeaderBadge` slot is nested inside the `LongPressable`, whose `onKeyDown` intercepts Enter/Space, calls `e.preventDefault()`, and toggles the group. Mouse clicks work only because the slot's `onClick` stops propagation, but there's no equivalent for keyboard, so the contributed button ends up unreachable by keyboard. CodeRabbit's suggested guard (`if (e.target !== e.currentTarget) return;` at the top of the header's `onKeyDown`) looks like the right fix.

2. **Context-menu plugin items don't dismiss the menu** (`SessionContextMenu.tsx:638`, Greptile P2 / CodeRabbit "Minor"). Every built-in item calls `onDismiss()` on activation, but the plugin slot only invokes the command, so the menu stays open until the user closes it manually. Threading an activation callback (e.g. `onActivate`) through `PluginUiItemsSlot` and passing `onDismiss` here would bring it in line with the native items.

Good news on the merge front: no conflicts, so nothing to rebase.

Once those two are addressed I think this is ready to go. Thanks again for the solid work on the security model here.

jSydorowicz21:
There is an out-of-order read race in `usePluginContributions`. If the initial `plugins.contributions()` request starts while a plugin is active, then disable or revoke triggers a newer empty read, the older request can still resolve last and restore the revoked UI item. The hook only guards unmount, not superseded loads.

I reproduced this with a delayed first response: the empty change response applied, then the older active response rendered the revoked control. Please version the loads or otherwise ignore responses that are no longer the newest registry snapshot, with an out-of-order regression test.

I did not repeat the existing group-header keyboard and context-menu dismissal threads; both remain valid on this head.

jSydorowicz21:
@greptile re-review please

PR #1204, opened 2026-07-10
fix: eliminate group chat typing lag under load

felipeggv:
## Summary
- Keep the Group Chat draft local while typing and debounce global persistence.
- Memoize the message history with stable callbacks so draft updates do not rebuild it.
- Virtualize long histories while preserving auto-scroll, message jumps, and keyboard navigation.

## Measured result
Target: Group Chat bcb41d92-c7e6-4c80-a851-90899ae01422 with 423 messages.

- Input p50: 246.72 ms to 2.10 ms.
- Input p95: 381.81 ms to 2.44 ms.
- DOM nodes: about 130,340 to 1,150.
- Rendered message rows: 423 to 6.
- Typing-related long tasks over 50 ms: 73 to 0.

## Validation
- 33,657 tests passed, 108 skipped, 0 failed.
- TypeScript lint passed.
- ESLint passed.
- Prettier passed.
- Full production build passed.
- Live top, bottom, Shift+Arrow navigation, draft restoration, and no-send probe passed.

ClickUp: https://app.clickup.com/t/86ajfwjh8

## Summary by CodeRabbit

* **Improvements**
  * Group chat drafts are now saved with debouncing and flush correctly when switching chats, sending, closing/unmounting, and when opening the Prompt Composer (including the Cmd+P shortcut).
  * Improved draft reliability when handling drop actions and external draft updates.
  * Enhanced group chat actions: quick mode toggling (markdown/raw), success flash feedback, and publishing group chat text as a gist.
* **Performance**
  * Group chat messages are fully virtualized for faster rendering, smoother scrolling/navigation, and visible scroll-to-message highlighting, including the typing indicator.
* **Tests**
  * Added/expanded coverage for draft persistence/flush behavior and virtualization-based scrolling/auto-scroll.

### DIFF
--- a/src/__tests__/renderer/components/GroupChatInput.test.tsx
+++ b/src/__tests__/renderer/components/GroupChatInput.test.tsx
@@ -140,6 +140,26 @@ describe('GroupChatInput', () => {
 			expect(onDraftChange).toHaveBeenCalledWith('keep this draft', 'test-group-chat');
 		});
 
+		it('cancels a pending local draft when an external draft arrives', () => {
+			vi.useFakeTimers();
+			const onDraftChange = vi.fn();
+			const { rerender } = render(
+				<GroupChatInput {...createDefaultProps({ draftMessage: '', onDraftChange })} />
+			);
+
+			const textarea = screen.getByPlaceholderText(/Type a message/i) as HTMLTextAreaElement;
+			typeInTextarea(textarea, 'stale local draft');
+			rerender(
+				<GroupChatInput
+					{...createDefaultProps({ draftMessage: 'new external draft', onDraftChange })}
+				/>
+			);
+			vi.advanceTimersByTime(300);
+
+			expect(textarea.value).toBe('new external draft');
+			expect(onDraftChange).not.toHaveBeenCalled();
+		});
+
 		it('publishes fresh text before opening Prompt Composer', () => {
 			vi.useFakeTimers();
 			const onDraftChange = vi.fn();
--- a/src/__tests__/renderer/components/GroupChatInput.test.tsx
+++ b/src/__tests__/renderer/components/GroupChatInput.test.tsx
@@ -10,7 +10,7 @@
  * Regression test for: Group chat @mention tab completion
  */
 
-import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { render, screen, fireEvent } from '@testing-library/react';
 import { GroupChatInput } from '../../../renderer/components/GroupChatInput';
 import type { Session, Group, GroupChatParticipant } from '../../../renderer/types';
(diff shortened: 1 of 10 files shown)


jSydorowicz21:
Reviewed. Two functional regressions worth fixing before merge; otherwise the debounce/virtualization work looks sound and the added tests cover coalescing, flush-on-switch/unmount, and the small-window case.

**1. (P1) Stale debounced inline draft can overwrite a newer Prompt Composer draft** — `src/renderer/components/GroupChatInput.tsx` ~L242-249. Repro path: type in the inline group-chat textarea, then within the 300 ms debounce window send/save a different value from Prompt Composer (`handlePromptComposerSubmit`/`Send` writes `draftMessage` immediately). The external-draft effect accepts the new value, but the still-armed inline timer later fires `onDraftChange` and resurrects the stale inline text over the saved/cleared draft. Fix: cancel the pending debounce before applying any external draft that doesn't match `lastPersistedDraftRef` (and add `cancelDraft` to that effect's deps); keep the self-persisted guard as-is. Test gap: external `draftMessage` update while an inline debounce is pending.

**2. (P2) Typing indicator can be scrolled out of view in virtualized mode** — `src/renderer/components/GroupChatMessages.tsx` ~L193-194. The auto-scroll effect scrolls to `scrollHeight`, but the new `virtualizer.scrollToIndex(messages.length - 1, { align: 'end' })` realigns the *last message* to the viewport bottom. When `state !== 'idle'` the typing indicator renders after the virtual spacer (not as a virtual item), so a message append mid-turn hides the 'Moderator is thinking…' row below the fold — pre-patch bottom-scroll included it. Fix: model the indicator as a virtual footer row, or include the footer height in the final bottom-scroll target. Test gap: non-idle auto-scroll with a trailing typing row.

**Manual checks suggested before merge:** inline type → immediate composer send → wait >300 ms (old draft must not return); 400+ message chat with non-idle state after append (indicator stays visible); auto-scroll on/off at top/bottom; Shift+Arrow navigation; `scrollToMessage` highlight for offscreen targets; image-heavy variable-height rows after load; markdown/raw toggle + copy/gist on virtualized rows.

**Conflict note:** no plugin-contract/semver impact (doesn't touch `src/shared/plugins` or the SDK). Main conflict risk is `App.tsx`/app-shell wiring against the Groups+ stack (#1193/#1199/#1202).

felipeggv:
Implemented the final review fixes in `fbf884660`.

- Pending inline draft persistence is cancelled before a true external draft is applied, with a regression test.
- The busy-state indicator is now a virtual footer row and remains visible at the bottom after message appends.
- Offscreen message highlighting uses a bounded 12-frame retry.
- Prompt Composer button, hotkey, and drop paths flush the current draft first.

Validation on the final head:

- TypeScript lint: passed
- ESLint: passed
- Production build: passed
- Vitest: 33,666 passed, 108 skipped, 0 failed
- Real 423-message busy chat: indicator visible at the bottom
- Chromium typing trace: p50 1.3 ms, p95 1.4 ms, 0 long tasks, 97.58% DOM reduction

@coderabbitai review

felipeggv:
Final verification was repeated in the actual Electron Group Chat window on head `fbf884660`, not in a standalone demo page.

- Loaded the real long-history chat from the corrected worktree
- Typed 65 characters through Chromium input events
- Input round trip: p50 2.02 ms, p95 2.50 ms, max 2.65 ms
- Next frame: p50 16.26 ms
- Rendered messages: 10
- Total DOM nodes: 1,662
- Main-thread tasks >= 50 ms: 0
- Existing user draft restored exactly after capture

All CI checks are green on Ubuntu, Windows, lint, and CodeRabbit. I attempted the squash merge, but the authenticated contributor account does not have `MergePullRequest` permission on `RunMaestro/Maestro`. This PR is clean and ready for maintainer merge.

pedramamini:
Thanks for the great work here, @felipeggv - this is a thorough fix with strong before/after numbers and solid test coverage.

I reviewed the PR alongside the Greptile, CodeRabbit, and human review feedback, and confirmed all flagged items are resolved in `fbf884660`:

- **Stale debounced draft race** (Greptile P1 / CodeRabbit Major): `cancelDraft()` now fires before an external draft is applied, with `cancelDraft` added to the effect deps. Verified, plus a regression test.
- **Typing indicator scrolled out of view** (Greptile P2): the indicator is now a true virtual footer row counted in `virtualItemCount`, and auto-scroll targets `virtualItemCount - 1`, with a clean inline fallback for empty chats.
- **Offscreen highlight miss in `scrollToMessage`** (CodeRabbit Minor): replaced the two fixed rAFs with a bounded 12-frame retry.

Also confirmed:
- CI green on both `test (ubuntu-latest)` and `test (windows-latest)`, plus lint-and-format and CodeRabbit.
- No merge conflicts against `rc` (mergeable, clean).

Approving. Since your account does not have merge permission, a maintainer will handle the squash merge.

felipeggv:
@jSydorowicz21 could you review the current head `fbf884660` when available? Your two previous findings were fixed with regression coverage, and the final CI is green on Ubuntu, Windows, lint, and CodeRabbit. The real Electron Group Chat trace now shows p50 2.02 ms, p95 2.50 ms, and zero main-thread tasks above 50 ms.

jSydorowicz21:
Re-reviewed at head fbf884660 — both prior findings are properly fixed. LGTM.

1. **Stale debounced draft**: the external-draft sync effect now calls `cancelDraft()` before applying an externally-changed `draftMessage` (GroupChatInput.tsx:253-261), with flush-before-switch, clear-on-send, and `lastPersistedDraftRef` guarding the rest. The cross-chat plumbing (`groupChatId ?? activeGroupChatId` in useGroupChatHandlers:795-800) also means a flush during a chat switch writes to the *original* chat — nice touch.
2. **Typing indicator**: now modeled as a real virtual row (`virtualItemCount` includes it; `getItemKey` = `typing-indicator`), and bottom-scroll targets `virtualItemCount - 1`, so the indicator is the scroll anchor when present.

Both previously-flagged test gaps are covered (external draft during pending debounce: GroupChatInput.test.tsx:138-158; non-idle append keeps indicator visible: GroupChatMessages.test.tsx:202-229). No new regressions found in the touched draft/flush/keyboard/virtualization paths.

jSydorowicz21:
Merging to RC

PR #1205, opened 2026-07-10
fix: address Greptile findings from merged plugin-surface PRs

jSydorowicz21:
Follow-up fixes for review findings on already-merged #1191, #1193, #1195:

- **#1191** — `useKnownAuthDirs` now clears suggestions when the panel switches to SSH (stale local paths could be saved into a remote config), and `agents:getKnownAuthDirs` excludes SSH sessions on any truthy `enabled` (was `=== true`, letting migrated truthy values leak remote paths into the local dropdown).
- **#1195** — `PluginHostViewRegistry.sync()` now purges runtime-sourced views whose declaration disappeared on hot-reload (previously stuck: replayed forever, un-removable since `remove()` rejects undeclared ids), and cadenza host-view close is delivered to BOTH the HUD and the main renderer (an in-app fallback card opened pre-HUD survived purge/disable).
- **#1193** — the reported persisted-cycle normalization gap was **refuted**: rc's normalization already promotes cycle members to root (only links to an existing root parent survive). Added explicit A↔B and over-depth-chain regression tests to pin that behavior.

Verification: 4 files / 98 focused tests, typecheck ×3, eslint, prettier clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session “local vs non-local” detection by treating SSH enablement truthy values (e.g., `"true"`, `"TRUE"`, `"1"`, `1`) as enabled, not just strict booleans.
  * Fixed plugin host view forwarding/cleanup to properly remove outdated live Cadenza/HUD views and correctly propagate Cadenza close/open behavior.
  * Ensured known authentication directories are cleared when the feature is disabled.
  * Strengthened group hierarchy normalization for persisted parent cycles and invalid links.

* **Tests**
  * Expanded coverage for SSH enablement parsing, known auth dirs hook reset behavior, plugin host view lifecycle/forwarding (open/close, remove, purge), and group hierarchy normalization edge cases.

### DIFF
--- a/src/__tests__/main/ipc/handlers/agents.test.ts
+++ b/src/__tests__/main/ipc/handlers/agents.test.ts
@@ -101,7 +101,7 @@ import { buildSshCommand } from '../../../../main/utils/ssh-command-builder';
 import * as fs from 'fs';
 
 describe('agents IPC handlers', () => {
-	let handlers: Map<string, Function>;
+	let handlers: Map<string, (...args: never[]) => unknown>;
 	let mockAgentDetector: {
 		detectAgents: ReturnType<typeof vi.fn>;
 		getAgent: ReturnType<typeof vi.fn>;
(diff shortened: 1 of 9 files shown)


pedramamini:
Thanks for the follow-up work here, @jSydorowicz21 - closing the loop on the earlier Greptile findings with focused regression tests is exactly the right approach, and the extraction of the host-view forwarder reads cleanly. 🙏

CodeRabbit came back clean and there are no merge conflicts. There is one new finding from this round's automated review that I'd like to see addressed before merge, plus a second I'd ask you to sanity-check:

**1. (please fix) `Boolean(sshRemoteConfig.enabled)` treats the string `"false"` as enabled** - `src/main/ipc/handlers/agents.ts` (`isLocalSession`)

This is the mirror of the bug you just fixed. Moving from `=== true` to `Boolean(...)` correctly catches migrated truthy values like `"true"` / `1`, but `Boolean("false") === true` in JS, so a persisted `enabled: "false"` (or `"0"`) now flips a *local* session to remote and drops its valid `CLAUDE_CONFIG_DIR` / `CODEX_HOME` from the dropdown - the exact symptom this handler is meant to avoid. If migrated data can hold stringified booleans, it can hold `"false"` too. Suggest normalizing explicitly, e.g. treat it as enabled only when `enabled !== false && enabled !== 'false' && Boolean(enabled)`, or coerce to a real boolean up front.

**2. (please confirm) `sync()` purges runtime host views when a declaration disappears** - `src/main/plugins/plugin-host-view-registry.ts`

I see the purge-on-disappearance is intentional (it is what unsticks the replay-forever case), so no objection to the design. The thing worth double-checking: if `getHostViews()` can return empty *transiently* during a hot-reload or manifest re-read, `sync()` will `removeLive()` runtime-sourced views, and unlike static views those do not self-restore (the plugin has to push another update). If `sync()` only ever runs once declarations have settled, this is a non-issue; if it can fire mid-reload, a small guard (skip the purge when declarations are empty, or only drop runtime views on explicit teardown) would be safer.

Once #1 is handled and you have had a look at #2, this should be good to go. Thanks again for tightening these up!

jSydorowicz21:
@greptileai review — both findings addressed in 54f13a6: SSH coercion now treats 'false'/'0' as local (only true/'true'/1/'1' count as SSH-enabled), and host-view reconciliation purges undeclared runtime views only when the plugin record's loadStatus is 'ok', retaining them through transient unavailability.

jSydorowicz21:
@greptileai Addressed the failed-load finding: the reconciliation gate is now record-PRESENCE (`isPluginRecordPresent`, any loadStatus) instead of `loadStatus === 'ok'`. A reload that lands in `invalid`/`incompatible` keeps the record present, so its undeclared runtime views purge; retention now applies only while the record is genuinely absent (transient reload window). New regression: record persists in failed load state → runtime view removed.

PR #1211, opened 2026-07-12
fix: keep composer mention highlights aligned while editing

felipeggv:
## Summary

- Keep native textarea glyphs and caret visible while the mention overlay paints decoration only.
- Hide the decorative overlay during native text selection to prevent doubled selection rendering.
- Synchronize the overlay scroll position after mount and text updates so mention highlights do not shift vertically.

## Validation

- Prettier check passed for both changed files.
- TypeScript lint passed.
- ESLint passed.
- Focused InputArea suite: 129 passed.
- Full Vitest suite: 33,907 passed, 108 skipped.
- Renderer production build passed.
- Manually verified cursor positioning, text selection, multiline scrolling, and agent mention alignment in Electron.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mention highlighting in the input area.
  * Prevented overlay text from interfering with normal typing and text selection.
  * Temporarily hides mention decorations while selecting text and restores them afterward.
  * Improved overlay alignment and scrolling during input and initial display.
  * Ensured overlays appear only when recognized mentions are present.

### DIFF
--- a/src/__tests__/renderer/components/InputArea.test.tsx
+++ b/src/__tests__/renderer/components/InputArea.test.tsx
@@ -2,6 +2,7 @@ import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { render, screen, fireEvent, act, within, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { InputArea } from '../../../renderer/components/InputArea';
+import { InputTextarea } from '../../../renderer/components/InputArea/components/InputTextarea';
 import { useComposerInputStore } from '../../../renderer/stores/composerInputStore';
 import { useSessionStore } from '../../../renderer/stores/sessionStore';
 import { formatEnterToSend } from '../../../renderer/utils/shortcutFormatter';
@@ -278,6 +279,44 @@ describe('InputArea', () => {
 			expect(overlay).toHaveStyle({ visibility: 'visible' });
 			expect(textarea).toHaveStyle({ color: mockTheme.colors.textMain });
 		});
+
+		it('aligns a newly mounted mention overlay with the textarea scroll position', () => {
+			const frames: FrameRequestCallback[] = [];
+			vi.spyOn(window, 'requestAnimationFrame').mockImplementation((callback) => {
+				frames.push(callback);
+				return frames.length;
+			});
+			vi.spyOn(window, 'cancelAnimationFrame').mockImplementation(() => undefined);
+			const inputRef = { current: null } as React.RefObject<HTMLTextAreaElement>;
+
+			const { container } = render(
+				<InputTextarea
+					session={createMockSession({ inputMode: 'ai' })}
+					theme={mockTheme}
+					isTerminalMode={false}
+					inputValue="check @src/index.ts now"
+					spellCheckEnabled={false}
+					inputRef={inputRef}
+					onInputFocus={vi.fn()}
+					onChange={vi.fn()}
+					handleInputKeyDown={vi.fn()}
+					handlePaste={vi.fn()}
+					handleDrop={vi.fn()}
+				/>
+			);
+			const textarea = screen.getByRole('textbox') as HTMLTextAreaElement;
+			const overlay = container.querySelector('.maestro-input-text-overlay') as HTMLDivElement;
+			textarea.scrollTop = 20;
+			textarea.scrollLeft = 7;
+
+			act(() => {
+				frames.splice(0).forEach((callback) => callback(0));
+			});
+
+			expect(overlay.scrollTop).toBe(20);
+			expect(overlay.scrollLeft).toBe(7);
+		});
+
 		it('renders the notification settings button', () => {
 			const props = createDefaultProps();
 			render(<InputArea {...props} />);
--- a/src/__tests__/renderer/components/InputArea.test.tsx
+++ b/src/__tests__/renderer/components/InputArea.test.tsx
@@ -3,6 +3,7 @@ import { render, screen, fireEvent, act, within, waitFor } from '@testing-librar
 import userEvent from '@testing-library/user-event';
 import { InputArea } from '../../../renderer/components/InputArea';
 import { useComposerInputStore } from '../../../renderer/stores/composerInputStore';
+import { useSessionStore } from '../../../renderer/stores/sessionStore';
 import { formatEnterToSend } from '../../../renderer/utils/shortcutFormatter';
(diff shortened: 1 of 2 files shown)


felipeggv:
@jSydorowicz21 could you review this PR when available? It fixes the prompt editor cursor, selection, and mention-highlight alignment issues, includes regression coverage for the scrolled overlay case, and is rebased on the current rc. Local validation is green with 33,907 tests passed.

pedramamini:
@felipeggv Thanks for taking this on, and thanks for the unusually thorough validation write-up. Inverting the layering so the native textarea owns the glyphs, caret, and selection while the overlay paints decoration only is the right call. It kills a whole class of caret-drift bugs instead of papering over them, and the zero-inline-advance chip trick is preserved correctly. The four new focused tests are well targeted, too.

I found one issue I'd like fixed before we merge.

### `hasSelection` is sticky across blur

`hasSelection` is only ever cleared by events on a *focused* textarea: `onChange`, `onSelect`, and the `selectionchange` listener, which early-returns via `document.activeElement !== textarea`. Nothing resets it on blur, and browsers preserve a textarea's selection range after it loses focus.

So this sequence leaves the decoration stuck off:

1. Composer contains `check @src/index.ts now`, so the chip renders.
2. Drag-select or double-click any word. `hasSelection` becomes `true` and the overlay goes `visibility: hidden` (intended).
3. Click away, for example onto the Left Bar or any button. The textarea blurs but its selection range survives.
4. `hasSelection` stays `true`, so the overlay stays hidden. The mention chip background and border are gone from an unfocused composer until the user clicks back in and collapses the caret.

I confirmed this against your branch with a throwaway test rather than by eyeballing it:

```tsx
textarea.focus();
textarea.setSelectionRange(0, 5);
fireEvent(document, new Event('selectionchange'));
expect(overlay).toHaveStyle({ visibility: 'hidden' }); // passes

fireEvent.blur(textarea);
expect(overlay).toHaveStyle({ visibility: 'visible' }); // FAILS: still hidden
```

This is the same root cause as Greptile's P2 note about the stale flag surviving an overlay remount. Both come from `hasSelection` outliving the selection's relevance, so one fix covers both.

### Suggested fix

Clear the flag on blur. Note `onInputBlur` is typed `() => void`, so it takes no argument:

```tsx
onBlur={() => {
	setHasSelection(false);
	onInputBlur?.();
}}
```

I applied exactly this in a scratch worktree and ran the full `InputArea.test.tsx` suite: 130/130 pass, including all four of your new tests plus the repro above. A regression test for the blur path would be a welcome addition alongside it.

### Non-blocking notes

- **CI:** the `test (windows-latest)` leg is red, but it is **not your fault**. The only failure is `AutoRunBlurSaveTiming.test.tsx > very long content can be saved` timing out at 10s on the Windows runner, which this PR does not touch. We'll get it re-run. Both matrix legs do need to be green before this can land.
- **Chip text color:** `mentionChipStyle` now hardcodes `color: 'transparent'`, so the live composer chip always shows `textMain` glyphs from the textarea. No shipped theme sets `mentionChipText` today, so there is zero visible change, but a custom theme that sets it would see the composer chip and the sent-transcript pill diverge. Fine to leave as is; just flagging it.

No merge conflicts, and the branch is otherwise clean. Push the blur fix and I'll take another look. Nice work.

pedramamini:
Thanks for this! Reviewed as part of a PR sweep. It currently conflicts with `rc` and the windows-latest test leg is red, so I could not merge it. Could you rebase on the latest `rc` and get both CI legs green? Happy to review and merge once it is clean.

pedramamini:
@felipeggv Checking back in on this one. Thanks again for the work here, the layering inversion is still the approach we want and the new tests are good.

Nothing has landed on the branch since my last note on July 15, so the same three items are still open:

1. **Merge conflicts.** The branch now conflicts with `rc`. Could you rebase onto the latest `rc` and resolve them so we get a clean merge?
2. **`hasSelection` sticky across blur.** Still unaddressed on the current head: `onBlur={onInputBlur}` is passed through untouched, so nothing clears the flag when the textarea loses focus while a selection range survives. The mention chip decoration stays hidden on an unfocused composer. The one-line fix and repro are in my [earlier comment](https://github.com/RunMaestro/Maestro/pull/1211#issuecomment-4950728926). A regression test for the blur path alongside it would be great.
3. **CI.** `test (windows-latest)` is red. The failure itself (`AutoRunBlurSaveTiming.test.tsx`) is not caused by this PR, but a rebase plus a re-run should clear it. Both matrix legs need to be green before this can land.

No rush, and no hard deadline on our side. Just let me know if you would rather hand it off and we can pick up the rebase and the blur fix ourselves. Happy to re-review as soon as it is pushed.

pedramamini:
@felipeggv Thank you for this contribution, and for sticking with the composer alignment problem until it was actually solved.

Good news: I think this PR has already landed in a different form, so there is nothing left to rebase.

Your own #1329 (merged into `rc` on July 29) carries the same approach as this branch: the native textarea owns the glyphs, caret, and selection, the overlay paints transparent decoration only, it renders only when a recognized mention exists (`overlayRendered`), and `selectionchange` hides it during a drag. #1330 from @chr1syy followed up with the two open items from my earlier review: `hasSelection` is cleared in `onBlur`, and an effect resets it whenever the overlay unmounts (terminal mode flip, last mention deleted), which also covers Greptile's stale-selection P2.

That is why this branch now conflicts with `rc` in both files. The conflicts are this older copy of the change colliding with the merged one, not new work that needs reconciling. I compared the two:

- **Layering and transparent chip glyphs:** identical to what is on `rc` today.
- **Scroll sync on mount:** `rc` does it in a synchronous `useLayoutEffect` keyed on the rendered segments rather than a `requestAnimationFrame`, so the overlay already has its full `scrollHeight` when the copy runs (no one-row clamp, no deferred frame).
- **Blur / remount stale selection:** fixed on `rc` by #1330.
- **Tests:** the native-text, recognized-mention, and selection tests from this PR are already in `InputArea.test.tsx` on `rc`, and `InputArea/components/InputTextarea.test.tsx` covers overlay scroll sync plus the blur and unmount regressions.

So unless you know of a behavior on this branch that `rc` is still missing, I would suggest closing this one as superseded by #1329 and #1330. If there is something `rc` still gets wrong, a fresh small PR off the latest `rc` with a repro would be the cleanest path, and I am happy to review it.

Thanks again for the fix. It is a real improvement for everyone typing mentions in the composer.

PR #1216, opened 2026-07-13
feat: Virtuosos - multi-provider account multiplexing with automatic switching (Encore-gated)

openasocket:
## Intent

Claude Code subscriptions (Pro/Max) enforce rolling per-account usage windows. A single heavy Maestro user hits those limits mid-session, which kills long Auto Run batches and blocks interactive work until the window resets. This PR lets one Maestro install drive multiple provider accounts ("Virtuosos"): isolate their credentials, route every spawn to an account, track per-account usage, and automatically switch/recover when an account is rate-limited or its auth expires.

Built for the RC channel: **the entire feature is gated behind an Encore flag** (Settings -> Plugins -> Virtuosos, default off). With the flag off, Maestro behaves as if the feature does not exist - no UI surfaces, no env injection, no account event traffic - so RC users who do not opt in are completely unaffected.

## Overview

User-facing name is **Virtuosos**; all internal identifiers say `account` (see `docs/agent-guides/ACCOUNT-MULTIPLEXING.md`, the checked-in design guide, for the full architecture, event contracts, and validation checklists).

- Each account lives in its own config dir (e.g. `~/.claude-work`) so OAuth credentials never mix; shared resources (`projects/`, `settings.json`, ...) are symlinked so `--resume` works across accounts.
- A single spawn-time hook (`account-env-injector.ts`) resolves an account (explicit > assignment > default > capacity-aware selection) and sets the provider env var on every spawn path: desktop AI turns, group chat, context grooming, and CLI playbooks.
- Usage is aggregated per account into midnight-aligned windows (stats.db migration v9) with warning/limit thresholds and P90 predictions in the Usage Dashboard.
- On `rate_limited`: record throttle, mark account, pick the next same-provider account, prompt or auto-switch, kill + respawn with `--resume`, re-send the interrupted prompt exactly once (coordinated with the Agent Resilience retry engine via `cancelRetry`). On `auth_expired`: automatic provider login with credential-sync fallback. All accounts exhausted: Auto Runs pause and auto-resume when a window resets (recovery poller).

## Features

- **Multi-provider parity.** Full multiplexing for Claude Code (`CLAUDE_CONFIG_DIR`), OpenAI Codex (`CODEX_HOME`), and OpenCode (`XDG_DATA_HOME`), driven by a single metadata module (`src/shared/accountProviderMeta.ts`). Gemini CLI and Factory Droid are import/observe only (their CLIs have no config-dir override; dropdown disables them with an explanation). Selection, defaults, throttle targets, and the toolbar selector are always provider-scoped, so a Codex account can never be routed into a Claude session.
- **Virtuosos modal** (hamburger menu): discover existing account dirs across providers, create new isolated dirs with per-provider login commands, symlink validate/repair, per-account limits, switch config.
- **Per-agent AccountSelector pill** (AI toolbar) with live usage; manual switches intentionally do NOT respawn (the last turn already completed; only automatic switches resume an interrupted turn).
- **Provider switching (vertical swapping)**, layered on top: consecutive-error failover suggestions, groomed-context migration to a new provider session with archive/merge-back provenance, provider health dashboard.
- **CLI parity:** `maestro-cli playbook --account <name>` / `--account-rotation`, provider-scoped.
- **Encore gating:** main process serves all account machinery through gated getters that re-read the flag live per call; renderer hides every surface; CLI warns and ignores account options; recovery poller starts/stops on flag toggle. Toggling requires no restart, and account data is preserved while off.

## Test coverage

All automated checks pass locally: `npm run lint`, `npm run lint:eslint`, full suite (34,200 passed / 108 skipped across 1,383 files) via the repo pre-push gate.

Unit coverage (mapped in detail in the guide, section 5.1): registry (CRUD, normalization, capacity scoring, provider scoping), env injector (guards, resolution order, per-provider env vars, cross-provider isolation), setup (discovery, per-provider create/login/sync, base-dir protection), switcher + IPC wiring, throttle handler, auth recovery (login success/timeout/fallback, concurrency guard), recovery poller, window math, error-listener routing, usage listener thresholds, stats queries, CLI account reader, renderer selector/switch-modal/usage hooks, dashboard tab gating.

**Known intentional gaps** (UI-heavy, exercised via the manual checklist in the guide, section 5.2): `AccountsPanel.tsx`, `VirtuososModal.tsx`, `AccountUsageHistory.tsx`, `AccountUsageDashboard.tsx` internals, `src/cli/commands/accounts.ts`.

## What to test more rigorously

In rough priority order for RC soak:

1. **Switch/resume correctness under real rate limits** - the automatic switch must re-send the interrupted prompt exactly once (watch for double-send with the Agent Resilience retry card, and for prompt loss on batch vs PTY paths).
2. **Flag-off invariance** - with Virtuosos disabled, absolutely nothing should change: no env injection (verify `ps eww <pid>`), no badges, no toasts, error handling identical to current RC behavior. This is the RC safety property.
3. **Codex/OpenCode isolation in practice** - `CODEX_HOME` and `XDG_DATA_HOME` relocation against current CLI versions; note OpenCode's account dir becomes the XDG data root for child tools too (documented limitation).
4. **Credential lifecycle** - auth expiry mid-turn, login timeout fallback to credential sync, concurrent recovery attempts.
5. **Assignment persistence across restart** and reconciliation when agents are deleted.
6. Windows CI leg - path handling for account dirs is exercised in tests, but the feature has only been manually validated on macOS.

## What's left / follow-ups

- No automatic remote provisioning for SSH remotes (account dirs must exist on the remote host; `accounts:validate-remote-dir` exists for pre-flight checks).
- Usage windows are midnight-aligned advisories, not mirrors of Anthropic's rolling quotas.
- The CLI reads `maestro-accounts.json` from default userData paths only; custom desktop sync paths are invisible to it.
- Gemini CLI / Factory Droid become multiplexable only if upstream ships a config-dir override (tracked in the guide; gemini-cli#2815).
- E2E coverage for the switch/resume flow (currently unit + manual).

## How to test

1. `npm run dev`, enable Settings -> Plugins -> Virtuosos.
2. Hamburger -> Virtuosos: Discover or Create accounts (copyable login command per provider).
3. Set a tiny `tokenLimitPerWindow` (e.g. 1000) and enable auto-switching to exercise warning -> limit -> switch -> resume without waiting for a real rate limit.
4. Full manual checklist: `docs/agent-guides/ACCOUNT-MULTIPLEXING.md` section 5.2.

No screenshots included: this machine is mid-network-outage; happy to add them in a follow-up comment.

---

## Summary by CodeRabbit

* **New Features**
  * Introduced Virtuosos (account multiplexing) with per-session account routing, rotation-based selection, auth recovery, throttling, and automatic resume workflows, plus account/provider switching, usage dashboards, provider failover suggestions, and unarchive conflict handling.
  * Extended the CLI with a new `accounts` command and `playbook` options `--account` / `--account-rotation` (feature-flagged).
* **Documentation**
  * Added the Virtuosos end-to-end guide and updated the reference checklist.
* **Chores**
  * Updated ignore rules to exclude local Virtuosos cache artifacts.
* **Tests**
  * Added extensive new test coverage and updated assertions for the new Virtuosos and stats functionality.


pedramamini:
@openasocket First off, thank you for this. The scope here is genuinely impressive: multi-provider parity, a coherent event contract, a checked-in design guide, and ~34k passing tests. The fact that you called out switch/resume double-send as the thing to stress-test first tells me you know where the bodies are buried, and you were right to. I dug into exactly that area and found some things worth fixing before this lands on `rc`.

Good news up front: no merge conflicts, the branch is clean against `rc`.

I verified everything below directly against the code on your branch (not just relaying bot output).

## Blockers

**1. Windows CI is red.** `test (windows-latest)` fails on two tests this PR adds, both in `src/__tests__/cli/services/account-reader.test.ts`:
- `readAccountsFromStore > falls back to filesystem discovery when store file missing`
- `readAccountsFromStore > handles macOS store path`

Both hardcode POSIX separators, so `path.join` returns `\home\testuser\.claude-personal` on Windows and the assertion fails. Product code looks correct; it is the assertions that need to build expected paths with `path.join`. Per CONTRIBUTING, both matrix legs have to be green before merge.

**2. The Encore gate does not fully hold, which undercuts the PR's headline safety claim.** The main-process gating is genuinely careful (I traced every spawn path: desktop, group chat, grooming, CLI, and confirmed no env injection with the flag off; the recovery poller correctly does not tick). But two renderer surfaces run with the flag OFF:

- `App.tsx:4169` mounts `<VirtuososModal>` unconditionally. Only the `isOpen` *prop* is gated. `VirtuososModal.tsx:44` calls `useProviderHealth(sessions)` *above* the `if (!isOpen) return null` on line 69, so the hook always runs: a 10s `setInterval` (`useProviderHealth.ts:251`) firing `agents.detect()`, `providers.getAllErrorStats()`, `settings.get()`, and `stats.getStats()` on every tick, plus `onFailoverSuggest`/`onStatsUpdate` subscriptions. Since `refresh` depends on `[sessions]`, the effect also re-fires on every sessions identity change, i.e. during streaming.
- `SessionList.tsx:167` calls `useAccountUsage()` with no flag check: `accounts.list()`, `accounts.getAllUsage()`, a 30s interval, and an IPC listener, all on an install that never enabled the feature.

Related: `src/shared/plugins/first-party.ts:604` says "Account multiplexing itself is always on ... the flag only gates the renderer surfaces," which contradicts the PR description. Worth reconciling so the intended contract is unambiguous.

**3. Automatic respawn targets the ACTIVE tab, not the throttled one.** In the `onSwitchRespawn` handler, the incoming `sessionId` is the decorated process ID (`${base}-ai-${tabId}`), but the handler resolves the tab with `getActiveTab(session)` and rebuilds `targetSessionId` from that (`App.tsx:1208-1214`). The originating tab ID is discarded. If the throttled turn was on a tab that is not currently active, the interrupted prompt gets replayed into a different tab's conversation and resumed with *that* tab's `agentSessionId` (so `--resume` points at the wrong provider session), while the throttled tab never resumes.

**4. Missing window-ownership gate on the account listeners, which is a real double-send vector.** `onSwitchRespawn` and `onSwitchExecute` in `App.tsx` do not use `useOwnedSessionGate()`. Every other process listener does (`useAgentErrorListener`, `useAgentExitListener`, etc.), because `safeSend` broadcasts to all windows and web-desktop clients, and every renderer holds all agents. With two windows open (or one window plus a web client), a single `rate_limited` event runs the switch in each renderer, so the same prompt gets spawned N times against the same `targetSessionId`. Each earlier spawn is killed by the next, but only *after* it may already have run tools and written files. This is the double-send you flagged, and it is orthogonal to the `cancelRetry` coordination (which does look correct for the 30s-backoff path).

**5. The respawn spawn config diverges from the canonical dispatch** in `agentStore.ts`. It omits `appendSystemPrompt` (per `spawnHelpers.ts:20-21`, every spawn site resuming a session must pass it, so the resumed turn silently loses the Maestro system prompt), `sendPromptViaStdin`/`sendPromptViaStdinRaw` (on Windows the normal path sends prompts via stdin; here the prompt goes on argv, so long/multiline/quoted prompts can corrupt), `images` (an image-bearing turn resumes as text-only), and `sessionCustomEffort`.

## Security

**6. Shell injection in `validateRemoteAccountDir`** (`account-setup.ts:565-588`). `configDir` is interpolated raw into the remote command string. Double quotes do not stop `$(...)` or backtick expansion in `sh`, so a config dir containing either executes on the remote host. Please use `shellEscapeForDoubleQuotes()` from `src/main/utils/shell-escape.ts` (CLAUDE.md lists it as the canonical helper). CodeRabbit flagged this too and it is a quick fix.

## Project convention

**7. 93 em/en-dashes across ~30 of the added files.** CLAUDE.md marks this one as non-negotiable. Worst offenders: `account-setup.ts` (13), `useProviderDetail.ts` (6), `VirtuosoUsageView.tsx` (6), `ProviderDetailView.tsx` (5), `account-registry.ts` (5), `App.tsx` (5). At least one is user-facing copy: the respawn toast reads `Switched to account ${toAccountName} (${reason}) — resuming interrupted turn`. A sweep should knock these out quickly.

## Worth picking up from CodeRabbit

These I spot-checked and agree with:
- `account-env-injector.ts:71-98`: the explicit `accountId` path skips the `status === 'active'` check that the assignment and default paths enforce, so a caller can inject a known-throttled account.
- `account-registry.ts:252-285`: `rotationIndex` is global but `selectNextAccount()` filters by `agentType` first, so alternating providers makes round-robin skip and repeat. Scope the cursor per provider.
- `provider-error-tracker.ts:88-98`: `toolType` is never refreshed on an existing session state, so post-switch errors get attributed to the old provider and corrupt the health dashboard.
- `account-switcher.ts:85-98`: the fixed 1s `SWITCH_DELAY_MS` is shorter than `PTY_KILL_ESCALATION_MS` (2s), so respawn can race PTY teardown. Waiting on the `exit` event with a timeout fallback would be sturdier.
- Several catch-alls swallow unexpected errors without `captureException` (`account-switcher.ts:139-156`, `account-throttle-handler.ts:136-143`, `account-env-injector.ts:119`), which the repo's Sentry guidance asks you to let bubble.
- A few `.toFixed(2)` cost formats should use the canonical `formatCost()`.

## Note on review coverage

Heads up that Greptile never actually reviewed this: it bailed with "Too many files changed (140 files, 100 file limit)." I am tagging it separately so you get that second pass. CodeRabbit's `Docstring Coverage` pre-merge check is also failing at 66% against an 80% threshold.

---

None of this is a knock on the design, which I think is sound. Items 2, 3, and 4 are the ones I would want fixed before this goes into RC soak, since they are precisely the failure modes that are hardest to diagnose in the field. Happy to re-review as soon as you have pushed. Thanks again for the very thorough writeup and the design guide, they made this much easier to review.

pedramamini:
@greptile-apps review

(Bypassing the 100-file limit so this PR gets a second AI review pass alongside CodeRabbit.)

jSydorowicz21:
Did a second pass on this at 00f157f since nothing new has been pushed since @pedramamini's review. I checked his blockers against the code myself rather than taking them on faith, and unfortunately they all hold up. Not ready to merge yet. The short version of what I'd want fixed, in the order I'd fix it:

**Windows CI is still failing.** Both broken tests are in `src/__tests__/cli/services/account-reader.test.ts`: the assertions hardcode strings like `/home/testuser/.claude-personal` while the code under test builds paths with `path.join`, which produces backslashes on Windows. The product code is fine; just build the expected values with `path.join` in the tests.

**The flag-off story doesn't match the PR description, and the code agrees with neither.** `src/shared/plugins/first-party.ts:602` literally says "Account multiplexing itself is always on" and the comment at 610-613 says the flag only gates renderer surfaces. Meanwhile two renderer surfaces run even with the flag off:
- `App.tsx:4168` mounts `VirtuososModal` unconditionally (only `isOpen` is gated), and `VirtuososModal.tsx:44` calls `useProviderHealth(sessions)` above the `if (!isOpen) return null` on line 69. That hook sets up a 10s `setInterval` doing `agents.detect()` + stats + settings IPC on every tick, for users who never opted in.
- `SessionList.tsx:167` calls `useAccountUsage()` with no flag check: `accounts.list()`, `getAllUsage()`, a 30s interval, and an IPC listener.

Either the description or the code needs to change so the contract is unambiguous, and the two hooks need gating regardless.

**Respawn targets the wrong tab.** In the `onSwitchRespawn` handler (`App.tsx:1207-1214`), the incoming `sessionId` carries the tab suffix (`${base}-ai-${tabId}`), but the handler throws it away and resolves via `getActiveTab(session)`. If the throttled turn was on a background tab, its prompt gets replayed into whatever tab the user happens to be looking at, resumed against that tab's `agentSessionId`.

**Multi-window double-send.** `onSwitchRespawn`/`onSwitchExecute` don't use `useOwnedSessionGate()` the way every other broadcast process listener does (compare `useAgentStderrListener.ts:21-25`). `safeSend` broadcasts to all windows, so with two windows open one `rate_limited` event runs the whole switch-and-respawn flow in each of them. Given you called out double-send as the top thing to stress test, this is probably the mechanism you'd hit first.

**Two security issues in `account-setup.ts`:**
- `validateRemoteAccountDir` (lines 566-584) interpolates `configDir` raw into SSH command strings inside double quotes, so `$(...)` and backticks execute on the remote host. `shellEscapeForDoubleQuotes()` in `src/main/utils/shell-escape.ts` exists for exactly this and isn't imported here.
- The account removal safety check (lines 522-535) validates only `path.basename(configDir)` against managed prefixes before `fs.rm(configDir, { recursive: true, force: true })`. `/tmp/.claude-work` passes. It needs a resolved-path containment check against the home dir, not just a basename prefix.

**Smaller but real:** the explicit-`accountId` path in `account-env-injector.ts:97-98` skips the `status === 'active'` check the other resolution branches enforce, so callers can inject a known-throttled account. And there are still 23 en/em dashes across `src/main/accounts/*.ts` alone (13 in `account-setup.ts`), which CLAUDE.md treats as a hard rule.

The design itself reads well and the test coverage on the main-process side is genuinely thorough. The tab routing and window ownership issues are the two I'd treat as hard blockers, since they corrupt sessions in ways users can't easily diagnose. Happy to take another look once these are pushed.

openasocket:
Do not have windows env to work on, so someone else will need to test/validate/dev that side of things because i can not, I can take a shot at it blind but someone will need to test.

jSydorowicz21:
Pushed 4d6ade7 addressing everything raised above. Rundown of what changed and how it was verified:

**Windows CI.** The two `account-reader.test.ts` failures were assertions hardcoding POSIX separators; expectations are now built with `path.join`. Verified on an actual Windows machine: 13/13 pass. While in there, the CLI filesystem fallback discovery no longer hardcodes `claude-code`; it infers the provider from the directory prefix via `ACCOUNT_PROVIDER_META`, with tests for codex and opencode prefixes (greptile's P2).

**Flag-off invariance.** `VirtuososModal` now only mounts when the Encore flag is on. `useProviderHealth` and `useAccountUsage` both take an `enabled` option and are fully inert when disabled: no IPC, no intervals, no subscriptions. `SessionList`, `AccountSelector`, `AccountsPanel`, and the usage dashboard pass the flag through. The `first-party.ts` copy that claimed "account multiplexing itself is always on" is rewritten to match the actual contract: the flag gates the entire feature and account data is preserved while off.

**Switch/resume correctness.** The respawn handler now parses the tab id out of the decorated session id and resumes that tab's `agentSessionId`, falling back to the active tab only when the suffix doesn't resolve. Both `onSwitchRespawn` and `onSwitchExecute` are scoped with `useOwnedSessionGate()`, so multi-window installs execute a switch exactly once instead of once per renderer. The respawn spawn config now matches the canonical dispatch: `appendSystemPrompt`, stdin prompt flags (so Windows prompts don't ride argv), images, and per-tab model/effort. `useSessionLifecycle` captures the previous account id before `setSessions`, so manual switches between two accounts actually reach `executeSwitch` instead of always falling into `assign`.

**Security.** `validateRemoteAccountDir` escapes `configDir` with `shellEscapeForDoubleQuotes()` in all three remote commands. Destructive account removal and symlink repair now require the resolved path to be a managed-prefix directory sitting directly under the home directory, so `/tmp/.claude-work` no longer passes.

**The CodeRabbit/greptile batch.** Explicit `accountId` injection enforces active status; round-robin cursors are per provider (with legacy numeric cursor migration); the provider error tracker re-attributes `toolType` after a switch; the error listener resolves the tracker getter per event so enabling the flag mid-session connects it; the groomer no longer persists throwaway assignments; auth recovery kills suffixed `-ai-` process ids; throttle handling proceeds without the stats DB instead of no-oping; the switcher waits on process exit (3s fallback) instead of sleeping 1s under the 2s PTY escalation; unexpected errors in the switch/throttle/credential-sync paths report to Sentry; `UnarchiveConflictModal` drops the redundant `onKeyDown` handlers that double-fired destructive actions; usage views use `formatCost()` and `AccountUsageHistory` guards against stale async responses; `ProviderDetailView` uses `useEventListener`; session restoration handles provider-specific env vars and reconciliation runs off the startup path.

**Punctuation sweep.** Zero em/en dashes remain on any line this PR adds.

Verification: full pre-push gate on Windows passed end to end: prettier check, all three tsc configs, eslint, and the complete suite (1,383 files, 34,210 passed / 109 skipped). LSP diagnostics clean on all 55 changed files.

Not addressed here: the docstring-coverage bot warning, screenshots, and E2E for switch/resume, which were already flagged as follow-ups in the PR description.

openasocket:
## Windows parity for account multiplexing (Virtuosos)

Summary of the Windows work now on this branch (commit `319dcb4f0`), for reviewers.

### 1. The two failing Windows CI tests are fixed

Both failures were test-side only: the account test assertions hardcoded POSIX `/` separators in expected paths, so they could never pass on the windows-latest leg. Expected paths are now built with `path.join` (and the shared `pathExpect` helpers). The product code was already separator-correct; no behavior changed to make CI green.

### 2. Four additional Windows blockers found by independent audit (not flagged in review)

While fixing CI, an audit of the Windows code paths surfaced four real blockers. What each would have done to a Windows user:

1. **Symlink privilege broke account creation.** Creating a symlink on Windows requires `SeCreateSymbolicLinkPrivilege` (admin elevation or Developer Mode), so on a stock install every "Create New Virtuoso" failed with EPERM and left a half-provisioned directory. Account dirs are now built with privilege-free NTFS directory junctions for shared directories and plain file copies for shared files, and validate/repair treats copied files as intact.
2. **The `.cmd` shim broke automatic auth recovery.** Provider CLIs on Windows are `.cmd`/`.ps1` npm shims that Node cannot exec directly, so the auth-recovery `claude login` spawn silently failed with ENOENT and recovery always fell through to "run it manually". Login spawns now go through the preferred Windows shell (PowerShell), mirroring `handle-spawn.ts`.
3. **Login command strings were POSIX-only.** The copyable command was `CLAUDE_CONFIG_DIR="..." claude login`, which is a syntax error in both cmd.exe and PowerShell, so manual re-auth of an isolated account was impossible as instructed. On Windows all login commands now render in PowerShell form: `$env:CLAUDE_CONFIG_DIR='C:\Users\...'; claude login` (same for `CODEX_HOME` / `XDG_DATA_HOME`).
4. **OpenCode base-dir / XDG resolution.** Credential sync looked for OpenCode's base auth file where it would never be on Windows. Verified against sst/opencode (`packages/core/src/global.ts`) and sindresorhus/xdg-basedir: opencode resolves its data root with NO platform branch, `$XDG_DATA_HOME` else `<home>/.local/share` on every OS, never `%APPDATA%`. The sync now mirrors xdg-basedir exactly, which also means OpenCode multiplexing IS fully supported on Windows rather than gated off.

Also in this pass: managed account-dir prefix checks are case-insensitive on Windows (NTFS), including the removal/repair safety helper added in the review-findings commit, and the CLI probes the `%APPDATA%\Maestro` store location.

### 3. Respawn stdin-flag fix (matters most on Windows)

The account-switch respawn now builds its spawn config through the shared `buildAccountSwitchRespawnConfig` -> `buildSpawnConfigForAgent` path instead of a hand-rolled field list, so a resumed turn is equivalent to a normal turn: `appendSystemPrompt`, recorded images (`lastImages` now rides along with the recorded prompt), per-tab model/effort, and crucially the `sendPromptViaStdin`/`sendPromptViaStdinRaw` transport flags. Normal turns deliver prompts via stdin on Windows because cmd.exe caps the command line around 8191 chars; the respawn path replays exactly the kind of long, multiline interrupted prompt that hits that cap, so putting it on argv could truncate or corrupt it. This composes with the review-findings commit: the throttled-tab resolution and multi-window ownership gate from `4d6ade7` are kept, with the spawn-config build routed through the shared helper so the two paths cannot drift again.

### 4. Manual Windows validation still needed

The author has no Windows environment; all of the above is implemented against documented Windows behavior and locked in with `process.platform = 'win32'` unit tests, but CI cannot exercise the interactive flows. Before this ships, someone with a real Windows machine should run the checklist now in `docs/agent-guides/ACCOUNT-MULTIPLEXING.md`, section 5.5 "Needs manual Windows validation": account creation (junctions + spawn env), PowerShell login command paste-and-authenticate, automatic auth recovery with turn resume, Codex `CODEX_HOME` isolation, OpenCode `XDG_DATA_HOME` isolation and credential sync, and symlink repair on a broken dir. Section 5.4 of the same guide documents the implementation rationale for each Windows behavior.

openasocket:
## Windows parity follow-up (commit 1b54d1d)

I re-audited the previous Windows-parity commit (319dcb4) against the canonical spawn paths and found four gaps that the `process.platform`-mocked tests did not catch. All four are now fixed, with regression tests. Pushed on top of this branch.

### 1. PowerShell auto-login was broken for spaced binary paths (the important one)
`runProviderLogin` (auth recovery) pre-joins the provider CLI into a single shell command string. When the resolved binary path contains a space - the common case on Windows, e.g. `C:\Users\First Last\AppData\Roaming\npm\claude.cmd` - the quoting turns it into `'C:\Users\First Last\...\claude.cmd' login`. PowerShell parses a **leading quoted token as a string literal, not a command**, so it errors on the trailing `login` and the login browser never opens. Automatic auth recovery would silently fail to re-authenticate on a large slice of Windows installs.

Fix: prepend the `&` call operator when the shell is PowerShell (`isPowerShellShell`); cmd.exe runs a quoted leading token as a command and gets no prefix. This is the same shape the proven `ChildProcessSpawner` path relies on.

### 2. Auto-login now honors the user's custom shell
`runProviderLogin` called `getWindowsShellForAgentExecution()` with no args, always defaulting to PowerShell. It now threads `customShellPath` from settings, matching `handle-spawn.ts`.

### 3. Login-timeout kill now terminates the process tree on Windows
On timeout the child is the shell wrapper, so `child.kill('SIGTERM')` orphaned the login grandchild. Windows now uses `taskkill /pid <pid> /t /f`.

### 4. Respawn env var is provider-aware
The account-switch respawn hardcoded `CLAUDE_CONFIG_DIR`. A Codex/OpenCode switch stamped a stray Claude var (harmless only because main's `injectAccountEnv` sets the real var). It now derives the env var from `getAccountProviderMeta(session.toolType).envVar` (`CODEX_HOME` / `XDG_DATA_HOME` / `CLAUDE_CONFIG_DIR`).

### 5. Remote (SSH) account validation is provider-aware
`validateRemoteAccountDir` hardcoded `.claude.json` / `projects`, so Codex/OpenCode accounts on an SSH remote always reported `hasAuth: false`. It now derives the auth files and the resume-critical shared resource from provider meta (OpenCode shares nothing, so it reports symlinks valid without a probe). Still POSIX `test` - runs on the remote host, never on Windows.

### Tests / validation
Added: the `&` prefix asserted on all three PowerShell login paths, a cmd.exe no-prefix case, a customShellPath-honored case, respawn provider-env assertions (codex -> `CODEX_HOME`, opencode -> `XDG_DATA_HOME`), and codex/opencode remote-validation cases. Full suite green on the pre-push hook: **34,360 passed / 108 skipped, 0 failures**. Docs updated in `ACCOUNT-MULTIPLEXING.md` section 5.4 (items 2, 6) plus new item 8.

**Still needs a Windows tester.** These paths remain covered only by `process.platform`-mocked unit tests - which is exactly how bug #1 slipped through the first pass. The section 5.5 manual checklist is the real gate; the highest-value item to exercise is automatic auth recovery actually opening a login window (bug #1's failure mode), plus Codex/OpenCode account isolation.

pedramamini:
Reviewed this in a batch pass (security + quality). The security audit came back clean: account credentials are never stored raw (config-dir pointer model + local file copies), no new external network endpoints in source, the plugin egress guard is default-deny and blocks loopback/RFC1918/link-local/metadata, and the Encore gating holds. Nice work on the defense-in-depth.

Holding it out of this batch merge for two non-security reasons:

1. Size/scope: ~24.9k lines across 145 files, and it bundles several subsystems (Virtuosos + plugins + pianola + coworking + browser-sessions) into one PR. That is a lot to land atomically on rc. If you can split it (Virtuosos core vs the bundled subsystems), each piece gets a faster, safer review + merge.
2. Red CI: the Windows leg is failing. It looks test-only (account tests hardcode POSIX '/' separators in expected paths while the product code is separator-correct) - building the expectations via path.join should green it up.

Once CI is green (and ideally split), I am happy to merge. Flagging rather than force-landing given the surface area.

jSydorowicz21:
I found three paths that should be fixed before this merges:

1. `account-setup.ts:371` builds `configDir` directly from the UI-supplied account name. On Windows, a name such as `work\\..\\..\\outside` resolves outside the managed account prefix, after which this function creates the directory and links or copies provider resources into it. Please reject path separators and validate the resolved path with `validateManagedAccountDirectory` before any filesystem call.

2. The setup handlers in `ipc/handlers/accounts.ts` are callable while Virtuosos is disabled. The preload exposes `createDirectory` unconditionally, and `accounts:create-directory` reaches the filesystem without checking the live Encore flag. The same applies to discovery, repair, and sync operations. Please gate the whole account IPC surface, not only the registry getters and renderer UI.

3. `account-switcher.ts:77` only checks that the target exists. The selector can still send throttled, expired, or disabled accounts, so a stale or multi-window selection kills the current process and respawns against a known-bad account. Please reject non-active targets in the switcher and filter them from the selector.

I did not repeat the existing SSH `configDir` thread; that issue is still present on this head.

pedramamini:
@openasocket First, thank you for this. Virtuosos is a big, careful piece of work, and the iteration since the first round has been genuinely responsive: the Windows parity passes, the provider-meta abstraction, and the checked-in design guide all raise the quality bar. I re-reviewed the current head (`1b54d1db1`) against the CodeRabbit and Greptile passes and the two human review rounds, and I verified each item in the code myself rather than relaying bot output.

## Confirmed fixed since the last round

Credit where it is due, these all check out on this head:

- **Punctuation sweep.** Zero em/en dashes on any line this PR adds. Verified with a diff scan against the merge base.
- **Flag-off invariance in the renderer.** `App.tsx:4153` now gates the `<VirtuososModal>` mount itself, not just `isOpen`, and `SessionList.tsx:168` passes `{ enabled: virtuososEnabled }` into `useAccountUsage`. Both of the earlier always-on interval/IPC leaks are closed.
- **SSH shell injection in `validateRemoteAccountDir`.** `account-setup.ts:645` escapes `configDir` with `shellEscapeForDoubleQuotes()` and applies it to all three remote commands, including the per-file auth probe. For the record, the note in the most recent review that this issue "is still present on this head" does not hold; it is fixed.

## Blocking before merge

**1. Merge conflicts.** The branch no longer merges cleanly into `rc`. 24 files conflict, including several load-bearing ones: `src/renderer/App.tsx`, `src/main/index.ts`, `src/main/ipc/handlers/index.ts`, `src/cli/index.ts`, `src/renderer/global.d.ts`, `src/renderer/hooks/session/useSessionLifecycle.ts`, `src/shared/settingsMetadata.ts`, `src/renderer/components/SessionList/SessionList.tsx`, plus `CLAUDE.md` and `.gitignore`. Could you rebase onto current `rc` and resolve these? Given the spawn-path and IPC-registration files in that list, I would rather you resolve them than have a reviewer guess at intent.

**2. Path traversal in `createAccountDirectory` (`account-setup.ts:371`).** `configDir` is built as `path.join(homeDir, meta.dirPrefix + accountName)` straight from the UI-supplied name, with no validation. `path.join` normalizes, so a name like `../../../tmp/evil` (or the backslash form on Windows) resolves outside the managed prefix, and the function then `mkdir`s it and links or copies provider resources into it. The guard already exists: `validateManagedAccountDirectory()` is called on the removal path (line 490) and the repair path (line 610), but not here. Please reject path separators in the name and run the resolved path through that same helper before any filesystem call.

**3. The account setup IPC surface ignores the Encore flag.** The registry-backed handlers are safely gated by `getAccountRegistry()` returning null, but the setup handlers do not go through the registry at all and never consult the flag: `accounts:validate-base-dir`, `discover-existing`, `create-directory`, `validate-symlinks`, `repair-symlinks`, `read-email`, `get-login-command`, `remove-directory`, `validate-remote-dir`, and `sync-credentials` (`ipc/handlers/accounts.ts:336-436`). With Virtuosos off, those still enumerate the home directory, create and delete directories, copy credentials, and open SSH connections. That is at odds with the PR's headline "behaves as if the feature does not exist" claim. Please gate the whole namespace, not only the registry getters and the renderer surfaces.

**4. `executeSwitch` accepts non-active targets (`account-switcher.ts:77-81`).** The only check on `toAccountId` is that the account exists. A stale selector state, a multi-window race, or a disabled/throttled/expired account all get through, and the very next step kills the running process and respawns against that known-bad account, which is the worst possible ordering. Please reject targets whose `status` is not `active` (and confirm the provider matches the session's `toolType`) before the kill, and filter them out of the selector as well.

**5. CI has not run on this head.** There are no check runs at all for `1b54d1db1`, so neither `test (ubuntu-latest)` nor `test (windows-latest)` has been exercised on the current code. Per CONTRIBUTING both legs need to be green before merge, and the local pre-push gate on a single OS is not a substitute. The rebase should retrigger them; please confirm both are green afterward.

## Still worth considering

The size and scope concern from the earlier review has not gone away: ~25k added lines across 145 files, bundling Virtuosos with several unrelated subsystems. Since a rebase is needed anyway, this would be a natural moment to split Virtuosos core from the bundled subsystems. Each piece would get a faster and safer review, and the blast radius on `rc` would be a lot smaller. Entirely your call, but I wanted to flag it before the conflict resolution work goes in.

Also still open and not blocking: the docstring coverage warning from CodeRabbit, and the section 5.5 manual Windows checklist, which really does need a human on a Windows box given that the mocked `process.platform` tests already missed one real bug.

Happy to take another pass as soon as the rebase and the four code items land. Thanks again for sticking with this one.

pedramamini:
@openasocket Thanks again for all the work on Virtuosos. The design guide and the Windows parity passes made this a lot easier to review, and we still want to see it land.

Checking in because it has been about a month since the last review and nothing new has been pushed. The head is still `1b54d1db1`, so the items from the [Aug 13 review](https://github.com/RunMaestro/Maestro/pull/1216#issuecomment-5275113603) are all still open. The conflicts have also grown while the branch sat.

## Please rebase onto current `rc`

The branch now conflicts with `rc` in **41 files** (it was 24 in August). The ones that matter most are the spawn and IPC wiring: `src/renderer/App.tsx`, `src/main/index.ts`, `src/main/ipc/handlers/index.ts`, `src/cli/services/agent-spawner.ts`, `src/cli/services/batch-processor.ts`, `src/main/utils/context-groomer.ts`, `src/main/process-listeners/index.ts`, `src/renderer/hooks/session/useSessionLifecycle.ts`, `src/renderer/hooks/session/useSessionRestoration.ts`, `src/renderer/global.d.ts`, and `src/shared/settingsMetadata.ts`. Could you rebase and resolve them yourself? A reviewer guessing at intent in those files is how resume and routing bugs get in.

**One conflict to watch, since it won't show up as a merge error:** `src/main/stats/migrations.ts`. This PR adds its account-usage migration as `version: 9`, but `rc` has since added its own migrations at versions 9, 10, and 11. After the rebase, the account migration needs to become the next free number (currently 12). If it keeps a number that is already taken, any database that has already reached v11 will skip it, and the account usage tables will never be created for existing `rc` users.

## Still blocking from the Aug 13 review (unchanged on this head)

1. **Path traversal in `createAccountDirectory`** (`account-setup.ts:371`): reject path separators in the account name and run the resolved path through `validateManagedAccountDirectory()` before any filesystem call, as the removal and repair paths already do.
2. **The account setup IPC surface ignores the Encore flag** (`ipc/handlers/accounts.ts`: `create-directory`, `remove-directory`, `discover-existing`, `sync-credentials`, `validate-remote-dir`, and the others). With Virtuosos off, these can still create and delete directories, copy credentials, and open SSH connections. Please gate the whole namespace.
3. **`executeSwitch` accepts non-active targets** (`account-switcher.ts:77-81`): before the kill, reject accounts that aren't `active` or don't match the session's provider, and filter them out of the selector.
4. **CI has never run on this head.** There are no check runs for `1b54d1db1`. The rebase push should trigger them, and both `test (ubuntu-latest)` and `test (windows-latest)` need to pass.

Since you'll be touching most of these files during the rebase anyway, now is also a good time to think about splitting Virtuosos core from the bundled subsystems. That's optional, but it would make the rebase and the next review much smaller.

Happy to take another look as soon as the rebased branch is up. Thanks for sticking with this one.

PR #1217, opened 2026-07-13
perf(session): stop App full sessions subscription during streaming

reachrazamair:
Peel MaestroConsoleInner off useSessionStore(s => s.sessions) so log/token flushes do not re-render the console shell. Persistence, fork, nav, and cue read at event time; leaves self-source with narrow selectors. Sidebar equality and useActiveSession remain as follow-ups.

## Summary by CodeRabbit

* **Performance**
  * Reduced re-render churn during streaming by switching session-driven UI to use targeted store subscriptions and stable equality checks.
  * Improved sidebar/session ordering and initialization of debounced session persistence.

* **Bug Fixes**
  * More reliable session navigation, conversation forking, cue auto-discovery, and session persistence (including flush and failure handling).
  * Stabilized timing for event payload validation and ensured mention autocomplete and thinking indicators stay in sync with current session state.

* **Tests**
  * Refreshed test setup to reset and seed shared session state for better isolation and correctness.

### DIFF
--- a/src/__tests__/main/cue/cue-reconciler.test.ts
+++ b/src/__tests__/main/cue/cue-reconciler.test.ts
@@ -273,9 +273,10 @@ describe('reconcileMissedTimeEvents', () => {
 		});
 
 		const sleepDuration = 60 * 60 * 1000; // 1 hour
+		const wakeTimeMs = Date.now();
 		const config = makeConfig({
-			sleepStartMs: Date.now() - sleepDuration,
-			wakeTimeMs: Date.now(),
+			sleepStartMs: wakeTimeMs - sleepDuration,
+			wakeTimeMs,
 			sessions,
 		});
 
--- a/src/__tests__/renderer/components/GroupChatInput.test.tsx
+++ b/src/__tests__/renderer/components/GroupChatInput.test.tsx
@@ -13,6 +13,7 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { render, screen, fireEvent } from '@testing-library/react';
 import { GroupChatInput } from '../../../renderer/components/GroupChatInput';
+import { useSessionStore } from '../../../renderer/stores/sessionStore';
 import type { Session, Group, GroupChatParticipant } from '../../../renderer/types';
 import { createMockSession as baseCreateMockSession } from '../../helpers/mockSession';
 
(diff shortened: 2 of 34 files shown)


pedramamini:
@reachrazamair Thanks for this - peeling `MaestroConsoleInner` off the full `sessions` subscription is exactly the right move, and the approach here is careful. The narrow equality helpers in `sessionEquality.ts` are well-scoped and clearly commented, self-sourcing the leaves rather than threading props is the correct direction, and the test migration to seeding the store is thorough. CI is green on both matrix legs and the branch merges cleanly, so no rebase needed.

I went through the bot findings and verified each against the code rather than taking them at face value. One is a genuine regression from this PR and I'd like it addressed before merge. The rest are either pre-existing or optional.

## Needs a fix

**`src/renderer/App.tsx` - `participantSessionPaths` can serve a stale `projectRoot`**

This is the one real regression, and CodeRabbit is right about it. The new memo derives from `sessionsForSidebar`:

```ts
return new Map(
    sessionsForSidebar.filter((s) => ids.has(s.id)).map((s) => [s.id, s.projectRoot])
);
```

`sessionsForSidebar` is gated by `sidebarSessionEquality`, which compares `id`, `name`, `state`, `toolType`, `bookmarked`, `groupId`, `parentSessionId`, `worktreesExpanded`, `worktreeBranch`, `isGitRepo`, `isLive`, `inputMode`, `activeTabId`, and `aiTabs` - but **not** `projectRoot`. So if an agent's project root moves and no other sidebar-tracked field changes, the array reference never flips, the memo never recomputes, and `GroupChatRightPanel` keeps rendering the old path. The previous code read the live `sessions` array, so this staleness is new.

A dedicated selector over `{ id, projectRoot }` with matching equality is the cleanest fix and keeps the streaming isolation you're after.

## Nits (your call)

**`selectCueDiscoverySignature` includes `projectRoot`, but nothing acts on it.** The signature is keyed on `id + projectRoot`, so a root change does re-fire the effect - but the effect body reconciles by id only (added/removed sets), so it no-ops. The underlying gap predates this PR, but the signature now implies it's handled. Either act on the change (remove the old root, refresh the new one) or drop `projectRoot` from the signature so it isn't a dead input.

**`hasSessions` and `hasNoAgents` are exact inverses.** Two subscriptions computing `length > 0` and `length === 0`. Worth collapsing to one.

## Pre-existing, not regressions - please leave them

I checked both of these against `origin/rc` and the behavior is unchanged by your diff, so per our scope-discipline guidance please don't expand the PR to cover them:

- **`useForkConversation` forcing `state: 'idle'` on spawn failure** (CodeRabbit, Major). The session-level `idle` write in the failure path is already on `rc` verbatim. The bot flagged it only because the file was largely rewritten. Real issue, separate PR.
- **`useDebouncedPersistence` dropping a flush when the debounce timer fires during an in-flight write** (CodeRabbit, Major). The `flushingRef` early-return with no reschedule already exists on `rc`. It's a real (if narrow - it needs an IPC write slower than the 2s debounce) bug and a good follow-up, but it isn't introduced here.

Greptile's P2 on `useActiveSession` still subscribing to the full active-session object is fair, and you already call it out as a follow-up in the description. Fine to defer.

Fix the `participantSessionPaths` staleness and I'm happy to approve.

PR #1218, opened 2026-07-14
fix(omp): recoverable no-output error + OmpSessionStorage History parity

chr1syy:
## Summary

Two related omp reliability fixes surfaced while diagnosing the recurring
"OMP session started from web-desktop, never went busy, appeared done, no
answer" report.

### 1. `fix(omp)`: recoverable error when a turn exits with no output

An omp turn can exit cleanly (code 0) right after startup / TTSR-rule
registration having emitted **no `agent_end`, no result, and no streamed
text**. Every existing exit branch then no-ops:

- `detectErrorFromExit` returns `null` on code 0, and
- the streamed-text exit fallback has nothing to flush,

so the tab clears its busy pill to an empty "done" state - **no answer and no
error**, indistinguishable from success. That is the reported turn.

`ExitHandler` now emits a recoverable, **non-auto-retrying** `agent_crashed`
error in that exact case, so the turn visibly fails and the user can resend.

- Scoped to `omp` to avoid tripping legitimate empty helper turns of other agents.
- User stops are excluded: `kill()` removes the process before `close`, and
  `interrupt()` now sets a new `ManagedProcess.interrupted` flag the guard
  checks (SIGINT `close` codes are coerced to `0`, so without the flag an
  instant Stop would masquerade as a crash).
- `agent_crashed` is in `NON_RETRYABLE_TYPES`, so this shows a clear error with
  manual resend rather than an auto-retry loop.

### 2. `feat(omp)`: register `OmpSessionStorage` for History/resume parity

omp had **no registered session storage**, so omp `.jsonl` transcripts under
`~/.omp/agent/sessions` were invisible to History/resume and the peer-frontend
transcript reconstruction path (`agentSessions.read('omp', ...)`) returned
empty, unlike every other storage-backed agent. That is why omp AI-tab
transcripts did not self-heal / sync across frontends (e.g. web-desktop after a
reload).

`OmpSessionStorage` (extends `BaseSessionStorage`) is added, registered, and
omp's `supportsSessionStorage` capability is flipped on. Listing scans every
session directory and filters by the **authoritative `cwd`** carried in each
transcript's `session` event, so it never depends on reproducing omp's internal
directory-slug scheme (a stale or unexpected slug dir can't hide a match).
Usage, cost, per-model split, title, and a system-context-skipping first-message
preview are recovered from the JSONL.

**Known v1 limitation:** local transcripts only; SSH-remote omp History degrades
to empty (no throw). Remote list/read (as Factory Droid does) is a documented
follow-up.

## Testing

- `ExitHandler.test.ts`: +6 cases (fire / interrupted / streamed-text present /
  non-omp / synopsis+tab-naming / no-double-report). Full ExitHandler suite: 24 passed.
- `omp-session-storage.test.ts`: +6 cases over **real temp fixtures** (no reliance
  on a developer's `~/.omp`): listing + metadata/usage, cwd filtering,
  cross-directory scan, message extraction (skips thinking/toolResult),
  delete-message-pair, SSH degradation.
- Full `process-manager` suite: 283 passed. ESLint clean on all touched files.
- `tsc -p tsconfig.main.json`: my changes add **zero** new type errors (verified
  by stash comparison against clean `rc`). The 5 pre-existing
  `@opencode-ai/sdk` moduleResolution errors in `src/main/opencode-server/*`
  are present on unmodified `rc` in a bare local worktree and are unrelated to
  this PR (resolved under CI's dependency install).

## Not included (scoping)

- **TTSR rule interrupts misclassified as session errors** - already merged via
  #1210; nothing to do here.
- **Agent-agnostic live session/transcript sync + stale-client `sessions:setMany`
  clobber** - a separate, load-bearing cross-renderer change whose root cause
  was not fully confirmed (needs the web-desktop console diagnostics from the
  prior investigation). Deliberately excluded to keep this PR focused and
  reviewable; happy to open a dedicated PR once the source is confirmed.

## Summary by CodeRabbit

* **New Features**
  * Added local Oh My Pi (omp) session history support, including listing sessions, reading messages, searching text, and deleting message pairs from transcripts.
  * Enabled omp sessions storage support for session history.
* **Bug Fixes**
  * Hardened omp “silent exit” handling to surface a recoverable crash error when no output/result is produced.
  * Properly treats user interruptions as non-crash stops and avoids duplicate/false error reporting.
* **Tests**
  * Extended exit-handling and omp session storage coverage for scoping, cwd matching, and remote/ssh scenarios.

### DIFF
--- a/src/__tests__/main/process-manager/handlers/ExitHandler.test.ts
+++ b/src/__tests__/main/process-manager/handlers/ExitHandler.test.ts
@@ -595,4 +595,109 @@ describe('ExitHandler', () => {
 			expect(elapsed).toBeLessThan(200);
 		});
 	});
+
+	describe('omp silent-exit hardening', () => {
+		function ompProc(overrides: Partial<ManagedProcess> = {}): ManagedProcess {
+			return createMockProcess({
+				toolType: 'omp',
+				isStreamJsonMode: true,
+				outputParser: createMockOutputParser({ agentId: 'omp' }),
+				resultEmitted: false,
+				errorEmitted: false,
+				streamedText: '',
+				...overrides,
+			});
+		}
+
+		it('surfaces a recoverable agent_crashed error when omp exits clean with no result, error, or output', async () => {
+			processes.set('sess-ai-tab1', ompProc());
+
+			const errors: Array<{ type: string; recoverable: boolean; message: string }> = [];
+			const exitEvents: number[] = [];
+			emitter.on('agent-error', (_sid: string, err) => errors.push(err));
+			emitter.on('exit', (_sid: string, code: number) => exitEvents.push(code));
+
+			await exitHandler.handleExit('sess-ai-tab1', 0);
+
+			expect(errors).toHaveLength(1);
+			expect(errors[0].type).toBe('agent_crashed');
+			expect(errors[0].recoverable).toBe(true);
+			expect(errors[0].message).toContain('without producing a response');
+			// The exit event still fires so the tab leaves the busy state.
+			expect(exitEvents).toEqual([0]);
+		});
+
+		it('does not fire when the user interrupted the turn (null signal coerced to code 0)', async () => {
+			processes.set('sess-ai-tab1', ompProc({ interrupted: true }));
+
+			const errors: unknown[] = [];
+			emitter.on('agent-error', (_sid: string, err) => errors.push(err));
+
+			await exitHandler.handleExit('sess-ai-tab1', 0);
+
+			expect(errors).toHaveLength(0);
+		});
+
+		it('does not fire when omp streamed text that the exit fallback flushes as the result', async () => {
+			processes.set('sess-ai-tab1', ompProc({ streamedText: 'here is my answer' }));
+
+			const errors: unknown[] = [];
+			const dataEvents: string[] = [];
+			emitter.on('agent-error', (_sid: string, err) => errors.push(err));
+			emitter.on('data', (_sid: string, data: string) => dataEvents.push(data));
+
+			await exitHandler.handleExit('sess-ai-tab1', 0);
+
+			expect(errors).toHaveLength(0);
+			expect(dataEvents).toContain('here is my answer');
+		});
+
+		it('does not fire for non-omp agents that exit clean with no output', async () => {
+			processes.set('sess-ai-tab1', ompProc({ toolType: 'claude-code' }));
+
+			const errors: unknown[] = [];
+			emitter.on('agent-error', (_sid: string, err) => errors.push(err));
+
+			await exitHandler.handleExit('sess-ai-tab1', 0);
+
+			expect(errors).toHaveLength(0);
+		});
+
+		it('does not fire for background omp synopsis or tab-naming sessions', async () => {
+			processes.set('sess-synopsis-123', ompProc());
+			processes.set('tab-naming-abc', ompProc());
+
+			const errors: unknown[] = [];
+			emitter.on('agent-error', (_sid: string, err) => errors.push(err));
+
+			await exitHandler.handleExit('sess-synopsis-123', 0);
+			await exitHandler.handleExit('tab-naming-abc', 0);
+
+			expect(errors).toHaveLength(0);
+		});
+
+		it('does not double-report when a non-zero exit already emitted an error', async () => {
+			// detectErrorFromExit fires for non-zero codes; the guard must see
+			// errorEmitted and stay quiet so only one error surfaces.
+			const parser = createMockOutputParser({
+				agentId: 'omp',
+				detectErrorFromExit: vi.fn(() => ({
+					type: 'agent_crashed',
+					message: 'Oh My Pi exited with code 1',
+					recoverable: true,
+					agentId: 'omp',
+					timestamp: Date.now(),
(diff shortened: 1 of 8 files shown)


pedramamini:
@chr1syy thanks for the contribution! This is a genuinely well-diagnosed pair of fixes. The writeup connecting the "started, never went busy, appeared done, no answer" report to the `detectErrorFromExit` code-0 no-op plus the empty streamed-text fallback is exactly the kind of root-cause tracing that makes a PR easy to review. Filtering `listSessions()` on the authoritative `cwd` from the `session` event rather than trying to reproduce omp's directory-slug scheme is the right call, and the `interrupted` flag to keep an instant Stop from masquerading as a crash is a nice catch that would have bitten in the field.

A few things to address before we merge.

### 1. Blocking: `findTranscriptPath()` ignores `projectPath` (also the red CI)

`lint-and-format` is failing:

```
src/main/storage/omp-session-storage.ts(336,35): error TS6133: 'projectPath' is declared but its value is never read.
```

That is the same root cause as Greptile's P1. `findTranscriptPath(projectPath, sessionId)` takes `projectPath` and then matches purely on `sessionId` across every transcript directory, never checking the transcript's `cwd`. `listSessions()` correctly scopes by `cwd`, so the two paths disagree: a stale History entry or a direct `readSessionMessages()` / `deleteMessagePair()` for project B can resolve and, in the delete case, rewrite a transcript belonging to project A.

Parsing the transcript and confirming `path.resolve(transcript.cwd)` matches the resolved `projectPath` before returning fixes the scope bug and the unused-parameter lint error in one go. Worth a test that a matching `sessionId` under a different `cwd` is not returned.

### 2. `sshRemoteId` missing on the new silent-exit error

Every other error construction in `ExitHandler.ts` sets `sshRemoteId` from `managedProcess` (lines 154, 198, 222, 472), but the new `agent_crashed` at ~line 254 omits it. For an omp session running over an SSH remote, listeners lose the association needed to keep per-remote status pills independent of the local snapshot. One-line fix.

### 3. Case-sensitive `cwd` comparison in `listSessions()`

`path.resolve(transcript.cwd) !== target` is a case-sensitive string compare. On macOS and Windows (case-insensitive by default) a `cwd` that differs only in casing, drive-letter casing being the classic one, gets silently excluded. That is the exact failure mode the doc comment says the authoritative-cwd filter avoids, so it is worth normalizing casing on those platforms.

### 4. Question, not a blocker: `getSessionPath()` returning `null`

Your rationale in the comment is reasonable, but with `supportsSessionStorage` now on, Greptile flags that the starred-transcript mirror has no source file to snapshot from, so a starred omp conversation cannot be rehydrated once the underlying `~/.omp` transcript is gone. Is that acceptable for v1 alongside the documented SSH-remote limitation, or is it worth resolving? Happy either way as long as it is a deliberate call and noted as such.

Nothing needed on the merge front, the branch is conflict-free against `rc`. The SSH-remote-History-degrades-to-empty scoping and the deliberate exclusion of the cross-renderer sync work both read as the right decisions to me. Ping me once the above is in and I will take another pass.

chr1syy:
Thanks @pedramamini - all four addressed in b6235313b.

**1. `findTranscriptPath()` project scope + red CI (blocking, = Greptile P1).** It now parses each candidate and requires BOTH the session id AND the authoritative `cwd` (resolved) to match `projectPath` before returning, so a stale History entry or a direct `readSessionMessages`/`deleteMessagePair` for project B can no longer resolve or rewrite a project-A transcript. Using `projectPath` also clears the `TS6133` unused-parameter error that was failing `lint-and-format`. Added a test: a same-`sessionId` transcript under a different `cwd` is neither read nor deleted.

**2. `sshRemoteId` on the silent-exit error.** Added `sshRemoteId: managedProcess.sshRemoteId` to the `agent_crashed` construction, matching the other four error sites in `ExitHandler.ts`.

**3. Case-sensitive `cwd` compare.** `listSessions()` and `findTranscriptPath()` now compare through a shared `samePath()` helper that lowercases on Windows/macOS (via `isWindows`/`isMacOS` from `platformDetection`) and stays case-sensitive elsewhere. Added a test that overrides `process.platform` to `darwin` and asserts a casing-only-different `cwd` still lists.

**4. `getSessionPath()` returning `null` (your question, = Greptile P2).** Kept as a deliberate call and documented as such in the code. An omp transcript lives at a timestamped filename under a cwd-slug dir, so it isn't derivable synchronously from a `sessionId` - this matches `CodexSessionStorage.getSessionPath()`, which returns `null` for the same reason and resolves via an async search. A synchronous directory scan here would be an O(n^2) blocking read in the `agentSessions.ts` loop callers (~lines 662/701). Consequence, same as Codex: the starred-transcript mirror can't snapshot an omp session and the `getSessionPath`-based name backfill skips omp; omp History itself is unaffected (served by `listSessions`/`readSessionMessages`). If you'd rather have starred omp snapshots land in this PR, I can add a small list/read-warmed path cache so the sync accessor resolves for any session the user has browsed - happy to, just say the word.

Also picked up the CodeRabbit nit: the SSH-degradation test now covers `deleteMessagePair` too.

Green on my end: 32 storage + exit tests pass, eslint and prettier clean, and `tsc -p tsconfig.main.json` adds zero new errors vs `rc`. Ready for another pass.

PR #1219, opened 2026-07-14
perf(session): peel App session subscriptions so streaming skips the console shell

reachrazamair:
## Summary
- Peel App-mounted input, interrupt, and batch hooks off full `selectActiveSession` / `sessions` subscriptions (event-time `getState()` or narrow field selectors) so streaming log/token flushes do not re-render `MaestroConsoleInner` through those paths.
- Hold App's active agent with `activeSessionChromeEquality`; MainPanel self-sources the full session for live chat; `useTabDerivedState` uses the same chrome equality.
- Stabilize empty `stagedImages` on a module-level empty array so stream rebuilds of `[]` do not wake App.
- Update MainPanel and GoalRunner tests to seed `useSessionStore` now that paint/batch no longer rely on injected full session arrays.

## Follow-ups (not in this PR)
- Remaining App-mounted `selectActiveSession` hooks (`useModalHandlers`, lifecycle, etc.)
- Relocate hook mounts out of App into leaf hosts; shrink chrome equality over time

## Summary by CodeRabbit

* **Performance Improvements**
  * Reduced unnecessary UI updates while sessions stream logs and responses, especially across the main panel, tab strip, and input handling.
  * Improved responsiveness for interruptions, submissions, and tab completion by using the latest session state at action time.

* **Bug Fixes**
  * Improved session consistency during terminal persistence, multi-window interactions, and batch workflows.
  * Fixed focus/selection behavior related to composer input so the correct target is preserved.

* **Tests**
  * Updated and expanded automated coverage to match the streamlined session-state handling and rendering behavior.

### DIFF
--- a/src/__tests__/renderer/components/MainPanel.test.tsx
+++ b/src/__tests__/renderer/components/MainPanel.test.tsx
@@ -494,6 +494,26 @@ describe('MainPanel', () => {
 		onNewTab: vi.fn(),
 	};
 
+	function seedSessionStore(session: Session | null | undefined) {
+		useSessionStore.setState({
+			sessions: session ? [session] : [],
+			activeSessionId: session?.id ?? '',
+		} as never);
+	}
+
+	function renderMainPanel(overrides: Partial<typeof defaultProps> = {}) {
+		const props = { ...defaultProps, ...overrides };
+		const session = props.activeSession as Session | null | undefined;
+		seedSessionStore(session);
+		const result = render(<MainPanel {...(props as any)} />);
+		const rerenderMainPanel = (nextOverrides: Partial<typeof defaultProps> = {}) => {
+			const nextProps = { ...defaultProps, ...nextOverrides };
+			seedSessionStore(nextProps.activeSession as Session | null | undefined);
+			result.rerender(<MainPanel {...(nextProps as any)} />);
+		};
+		return { ...result, rerender: rerenderMainPanel };
+	}
+
 	beforeEach(() => {
 		vi.clearAllMocks();
 		vi.useFakeTimers({ shouldAdvanceTime: true });
@@ -569,7 +589,7 @@ describe('MainPanel', () => {
 
 	describe('Render conditions', () => {
 		it('should render LogViewer when logViewerOpen is true', () => {
-			render(<MainPanel {...defaultProps} logViewerOpen={true} />);
+			renderMainPanel({ logViewerOpen: true });
 
 			expect(screen.getByTestId('log-viewer')).toBeInTheDocument();
 			expect(screen.queryByTestId('terminal-output')).not.toBeInTheDocument();
@@ -577,9 +597,7 @@ describe('MainPanel', () => {
 
 		it('should close LogViewer and call setLogViewerOpen when close button is clicked', () => {
 			const setLogViewerOpen = vi.fn();
-			render(
-				<MainPanel {...defaultProps} logViewerOpen={true} setLogViewerOpen={setLogViewerOpen} />
-			);
+			renderMainPanel({ logViewerOpen: true, setLogViewerOpen: setLogViewerOpen });
 
 			fireEvent.click(screen.getByTestId('log-viewer-close'));
 
@@ -587,7 +605,7 @@ describe('MainPanel', () => {
 		});
 
 		it('should render AgentSessionsBrowser when agentSessionsOpen is true', () => {
-			render(<MainPanel {...defaultProps} agentSessionsOpen={true} />);
+			renderMainPanel({ agentSessionsOpen: true });
 
 			expect(screen.getByTestId('agent-sessions-browser')).toBeInTheDocument();
 			expect(screen.queryByTestId('terminal-output')).not.toBeInTheDocument();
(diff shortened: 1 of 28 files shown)


pedramamini:
@reachrazamair Thanks for taking this on! Peeling the App-level session subscriptions is exactly the right lever for the streaming-repaint problem, and the structure here is good: `getState()` at event time for handlers, narrow primitive selectors for reactive reads, and letting `MainPanel` self-source the full `Session` for paint. The `EMPTY_STAGED_IMAGES` sentinel and the `depsRef` pattern in `useInterruptHandler` are both clean. CI is green on both matrix legs and there are no merge conflicts.

There is one systemic issue I'd like resolved before merge, plus a few smaller items.

## Blocking: `activeSessionChromeEquality` is an allowlist, and it is missing fields its consumers actually read

The comparator is opt-in per field, so any field it omits becomes silently stale for **every** consumer holding the chrome-gated `activeSession`. Right now `MaestroConsoleInner` still hands that slice to several hooks and components, and at least three of them read omitted fields.

**1. `fileTree` (most severe, I don't think either bot flagged this).**

`App.tsx:2218` passes the chrome-gated `activeSession` into `useFileTreeManagement`, whose `filteredFileTree` memo reads it directly:

```ts
// src/renderer/hooks/git/useFileTreeManagement.ts:1029
const filteredFileTree = useMemo(() => {
    if (!activeSession || !fileTreeFilter || !activeSession.fileTree) {
        return activeSession?.fileTree || [];
    }
    ...
    return filterTree(activeSession.fileTree);
}, [activeSession, fileTreeFilter]);
```

That value is passed straight to the Files panel (`App.tsx:3376`). `useFileTreeManagement` does not self-subscribe to the active session (only to `sessionsLoaded`), so it depends entirely on the prop. `fileTree` is not compared by `activeSessionChromeEquality`, and `refreshFileTree` writes `fileTree:` on its own (lines 466, 612, 768, 801, 866, 917) without touching any compared field.

Failure scenario: open an agent, let the tree load, then create/rename/delete a file on disk (or hit refresh, or switch branches). The store gets a new `fileTree`, chrome equality returns `true`, App does not re-render, and the Right Bar Files panel keeps showing the old tree until something unrelated (agent state flip, tab switch) happens to bust equality. This is a user-visible regression and worth a test.

**2. `contextUsage` + tab `logs`.** You correctly converted the `canSummarizeActiveTab` getter at `App.tsx:2517` to `selectActiveSession(useSessionStore.getState())`, but the identical computation in JSX at `App.tsx:3310` still reads the chrome-gated value:

```ts
canSummarizeActiveTab={
    activeSession
        ? canSummarize(
                activeSession.contextUsage,
                activeSession.aiTabs.find((t) => t.id === activeSession.activeTabId)?.logs
            )
        : false
}
```

Both `contextUsage` and `logs` are deliberately excluded from the comparator, so this prop goes stale mid-stream. CodeRabbit flagged this one too. The getter you already wrote is the fix.

**3. Tab strip fields.** CodeRabbit's list is right: `AITab.isGeneratingName`, `AITab.agentError`, `AITab.pendingMergedContext`, and `BrowserTab.customTitle` are rendered but not compared, so auto-naming and error badges can lag.

### Suggested direction

Rather than only growing the allowlist (which puts a standing burden on every future field addition), I'd prefer the fix that matches the philosophy already in this PR: **have each App-mounted consumer self-source the fields it needs**, the way `MainPanel` now does. Concretely, `useFileTreeManagement` should read `fileTree` from the store itself instead of receiving it via the chrome slice. Adding `fileTree` to the comparator by reference would also work (tree updates are not streaming-frequency), but self-sourcing scales better.

Whichever you pick, could you add a short comment on `activeSessionChromeEquality` stating the invariant, something like "any consumer of this slice must not read fields absent from this comparator", plus the tab fields CodeRabbit listed? A regression test covering "file tree refresh repaints the Files panel" would give real protection here.

## Secondary

- **The `storeActiveId` guard in the completion hooks** (CodeRabbit, `useAtMentionCompletion` / `useTabCompletion`): `selectActiveSession()` already falls back to `sessions[0]`, and restoration sets `sessions` before `hydrateActiveSessionId()`, so gating on a truthy `activeSessionId` can blank `fileTree`/`cwd`/`autoRunFolderPath` during startup. Worth dropping the guard.

- **Greptile's two P1s** (deferred replay target, blur sync target): I looked at both and I think they are lower severity than labelled. `processInputRef.current` is refreshed every render, so the `setTimeout(0)` in `handleReplayMessage` already resolved to the freshest closure before this PR; the send target behavior is essentially unchanged. I also confirmed the `syncAiInputToSession('')` clears in `useInputProcessing` (476, 557, 665, 1107) run synchronously inside `processInput` (the nearby `await`s are inside nested `(async () => {})()` IIFEs), so there is no widened async window there. The only real exposure is a 0ms macrotask in `handleReplayMessage`. Not a blocker, but since you are already resolving `activeSession` at the top of that callback, capturing `activeSessionId`/`activeTabId` there and threading them through would close it cheaply.

## Nits

- `MainPanel` destructures `activeSession: _activeSessionFromApp` and ignores it while App still computes and passes it. If it is genuinely unused for paint, consider dropping it from `MainPanelProps` rather than leaving an intentionally-ignored prop for the next reader to puzzle over.
- `useTabCompletion`'s no-arg path builds `fileTree: storeFileTree ?? []`, allocating a fresh `[]` every render when the session has no tree. That busts the `fileNames` memo (dep is `fields?.fileTree`), which in turn changes `getSuggestions`' identity every render, defeating the memoization the hook's own comments promise. A module-level `EMPTY_FILE_TREE` constant fixes it, and there is already precedent for exactly that constant in `useFileTreeManagement`.
- The `useBatchProcessor` getter-backed fake `MutableRefObject` with a no-op setter works, but it silently swallows `sessionsRef.current = ...` writes. A `getSessions()` callback or a `{ readonly current: Session[] }` type would fail loudly instead of quietly doing nothing.

Happy to re-review as soon as the file-tree staleness and the `canSummarizeActiveTab` call site are addressed. Thanks again, the perf direction here is solid and worth landing.

PR #1220, opened 2026-07-14
fix(file-preview): stop stats bar hide/show jitter on barely overflowing content

reachrazamair:
Hiding the Size/Lines/Modified chrome on scroll grew the viewport enough that tiny overflow collapsed, scrollTop snapped to 0, and the bar reappeared in a loop. Only collapse that chrome when leftover overflow is larger than the header chrome.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the file preview “Lines” stats bar visibility during scrolling, including cases with minimal overflow.
  * Prevented the stats bar from flickering/toggling when content height is near the viewport and during layout changes.
  * Correctly initializes and restores the stats bar visibility when switching files and returning to the top.

* **Tests**
  * Added coverage for stats bar visibility across scroll edge-cases and file rerenders.

### DIFF
--- a/src/__tests__/renderer/components/FilePreview.test.tsx
+++ b/src/__tests__/renderer/components/FilePreview.test.tsx
@@ -1667,7 +1667,7 @@ print("world")
 			const container = document.querySelector('.overflow-y-auto') as HTMLElement;
 			expect(container).not.toBeNull();
 
-			// Barely overflowing — hiding the stats chrome would clamp scrollTop to 0
+			// Barely overflowing - hiding the stats chrome would clamp scrollTop to 0
 			// and re-show the bar (the jitter loop). Stay visible instead.
 			Object.defineProperty(container, 'scrollHeight', { configurable: true, value: 520 });
 			Object.defineProperty(container, 'clientHeight', { configurable: true, value: 500 });
@@ -1681,7 +1681,7 @@ print("world")
 			expect(screen.getByText('Lines:')).toBeInTheDocument();
 		});
 
-		it('hides the stats bar when scrolled with enough leftover overflow', () => {
+		it('hides the stats bar when scrolled past the chrome height with enough overflow', () => {
 			render(
 				<FilePreview
 					{...defaultProps}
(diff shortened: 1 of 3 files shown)


pedramamini:
@reachrazamair Thanks for the contribution, and thanks for digging into the root cause here rather than papering over it. The core insight is right: collapsing the chrome grows the viewport, which can clamp `scrollTop` back into the "show" band, and the hysteresis + `overflow > STATS_CHROME_PX` guard is a sound way to break that loop. CI is green on both matrix legs and the branch merges cleanly.

Before we merge, there is one behavioral regression I was able to reproduce, plus the CodeRabbit finding I think is worth taking.

### 1. Stats bar can get stuck hidden when a saved scroll position is restored (blocking)

The new initial/reset value uses only `initialScrollTop`:

```ts
const [showStatsBar, setShowStatsBar] = useState(
    () => initialScrollTop === undefined || initialScrollTop <= 10
);
```

The only thing that can ever flip it back to `true` is a `scroll` event, and there is no resize/ResizeObserver path. So any time a file opens with a saved `initialScrollTop > 10` and no scroll event follows, the bar stays hidden with no way for the user to get it back. Two concrete cases:

- **Restore clamps to 0.** A tab has a saved `scrollTop` of 300, but the content no longer overflows (window/pane got taller, right bar collapsed, file shrank). `contentEl.scrollTop = 300` clamps to 0, so no scroll event fires. `showStatsBar` was initialized to `false` and stays `false`. The user cannot scroll to fix it because there is no overflow. On `rc` today this initializes to `true`, so this is new behavior.
- **Restore lands inside the chrome band.** Saved `scrollTop` of 40 with plenty of overflow. The restore fires a scroll event, `handleScroll` sees `40 <= STATS_CHROME_PX + AT_TOP_PX` and returns `prev` (`false`). But scrolling manually to 40 with the same overflow keeps the bar visible, which is exactly what your own "does not hide the stats bar on a shallow scroll" test asserts. The restore path and the scroll path disagree.

Both of these fail against the PR head:

```tsx
it('keeps the stats bar visible when a saved scroll position cannot be restored', () => {
    render(
        <FilePreview
            {...defaultProps}
            file={{ name: 'short.md', content: 'Short\n', path: '/test/short.md' }}
            initialScrollTop={300}
        />
    );
    expect(screen.getByText('Lines:')).toBeInTheDocument(); // currently fails
});

it('keeps the stats bar visible when a saved scroll position lands in the chrome band', () => {
    render(
        <FilePreview
            {...defaultProps}
            file={{ name: 'tall.md', content: 'Some content\n'.repeat(20), path: '/test/tall.md' }}
            initialScrollTop={40}
        />
    );
    expect(screen.getByText('Lines:')).toBeInTheDocument(); // currently fails
});
```

The root cause is that the initial/reset value guesses from `initialScrollTop <= 10` while the hide decision now depends on real geometry (`overflow` and the chrome height). Suggested direction: pull the visibility rule out of `handleScroll` into a small `evaluateStatsBar(contentEl)` callback, and call it from the `useLayoutEffect` right after you set `scrollTop` (both the restore branch and the reset-to-0 branch). That way the restored state is derived from what actually happened to the DOM instead of from the requested position, and the two paths can't drift apart. Please add the two cases above as tests.

### 2. Ref mutation during render (agreeing with CodeRabbit)

I know it looks like a style nit, but it has teeth here. If React discards or replays that render (concurrent rendering, Suspense, StrictMode), `showStatsBarPathRef.current` keeps the new path while the `setShowStatsBar` never commits. On the next render the `file?.path !== ref.current` check is already satisfied, so the reset is skipped and the bar stays hidden. That is precisely the stuck-hidden bug this reset exists to prevent. CodeRabbit's prev-state version is the documented React pattern for this and is a two-line change, so let's take it.

### 3. Nit, non-blocking: `STATS_CHROME_PX = 64`

The constant is a hand-measured estimate of markup that lives in a different file (`FilePreviewHeader.tsx`), with nothing tying the two together. If that chrome ever grows past 64px the guard silently stops guarding. Not worth blocking on, and the failure mode degrades to a single snap-back rather than the old freeze, but if `evaluateStatsBar` is being extracted anyway, measuring the header element would make this self-maintaining. At minimum, a short comment in `FilePreviewHeader.tsx` noting that its height feeds `STATS_CHROME_PX` would help the next person.

### Already handled

Greptile's P1 about the bar staying hidden across a file switch was fixed by 2953ca6 (the render-time reset); it reviewed the earlier commit only, so no action needed there.

Nice work on the diagnosis. Ping me when the above is addressed and I'll take another pass.

PR #1221, opened 2026-07-15
fix(context): anchor Context Timeline to the clicked gauge and read omp's real context window

chr1syy:
## Summary

Two related context-window fixes:

### 1. Context Timeline opens on the context gauge you click
The Context Timeline previously always docked at the bottom-left, regardless of where it was opened from. It now anchors to the header context gauge that opened it (mouse click or keyboard activation).

- `contextTimelineStore.openPanel(sessionId, anchorRect?)` stores a plain, clone-safe rect of the trigger; `closePanel` clears it.
- `MainPanelHeader` passes the widget's `getBoundingClientRect()` on both click and Enter/Space.
- `ContextTimelinePanel` computes a viewport-clamped position from that rect (right-aligned under the trigger, flips above it when it would overflow the bottom, width and height clamped to the viewport) and falls back to the bottom-left dock when opened without an anchor.

### 2. Oh My Pi reads its actual (model-dependent) context window
For omp the gauge showed a static 200k even when running a 1M model (e.g. opus). Root cause was twofold: the parser never surfaced a window, and the renderer prioritized omp's `contextWindow` config default (200k) over any reported value.

omp streams the per-turn model (for example `claude-opus-4-8`), and its own catalog (`omp models --json`) maps that model to its real window. This PR wires that through:

- New synchronous `omp-model-catalog` (model to context-window map), keyed by id, provider selector, and bare id. Primed off-path from the detector's existing `omp models --json` discovery and eagerly (fire-and-forget, deduped, env-aware) on local omp spawn.
- The omp parser surfaces the per-turn `model` on the usage event.
- `StdoutHandler` resolves the model against the catalog for local (non-SSH) omp, stamps the real window on `UsageStats`, and sets a new `contextWindowResolved` flag marking it authoritative.
- The renderer usage listener and the header gauge let a genuinely resolved window beat the agent-level configured fallback, while an explicit per-session override and `[1m]` model markers still rank above it. Batched usage updates preserve the flag through accumulation and flush.

The flag scoping means zero behavior change for other agents: when a parser does not report, the reported value already equals the configured window, so the reorder is a no-op. SSH remote omp agents keep the configured/fallback window since the local catalog cannot be trusted for a remote.

## Testing

- `tsc` (renderer + main + cli): clean.
- Targeted vitest suites (114 passing on this branch's rc base):
  - `omp-model-catalog` catalog set/get (id, selector, bare id, case-insensitive, unknown, zero/undefined window).
  - omp parser: per-turn `model` extraction.
  - `StdoutHandler`: local omp catalog resolve sets the window + flag; SSH omp keeps the configured window and no flag.
  - usage listener: a resolved 1M window beats a configured 200k fallback.
  - timeline store: `openPanel` records the anchor rect; anchorless open and `closePanel` clear it.

## Notes

- Very first omp turn on a fresh app run may briefly show the fallback if the catalog prime has not finished; it self-corrects immediately after.
- No `docs/releases.md` edit (handled by release tooling).

## Summary by CodeRabbit

* **New Features**
  * Added automatic Oh My Pi model detection and per-environment cataloging to resolve model-specific context-window limits.
  * Context usage now prefers provider-resolved per-turn model windows when flagged as authoritative.
  * Context timeline panel now anchors to the trigger element and reclamps within the viewport.

* **Bug Fixes**
  * Fixed context-window precedence so resolved provider values override cached/fallback windows.
  * Ensured resolved context-window state is carried through batching/updates for accurate percentages and timeline display.


pedramamini:
Hi @chr1syy, thanks a lot for this contribution! The two fixes are well-scoped and the writeup (especially the flag-scoping rationale for zero behavior change on other agents) made this easy to follow. A few things from the Greptile/CodeRabbit passes look worth addressing before we merge:

**1. Scope the omp model catalog to the binary/config that produced it (Greptile P1 + CodeRabbit Major).** `omp-model-catalog.ts` keeps a single module-level `Map` and a globally-deduped prime. If the detector or one local session primes it, a second session started within the TTL with a different `customPath`/env reuses the first catalog and gets its window stamped as `contextWindowResolved` (authoritative). Because you added the "authoritative" flag specifically to beat the configured fallback, a wrong-install resolution is worse here than the old static fallback. Keying the cache + in-flight prime by command plus a safe config identity, and carrying that identity into the lookup, would close this. This also subsumes the P2 bare-id collision note (two selectors sharing a bare id silently keep the last write).

**2. Precedence disagreement between the two hooks (CodeRabbit Major).** `useAgentUsageListener` ranks the `[1m]`/`customModel` marker above `contextWindowResolved` (and its comment says that's intended), but `useContextWindow` folds the marker into `configuredContextWindow` and lets the resolved window win first. So a session on a `[1m]` model could show one window in the header gauge and another in the timeline. Resolving the `customModel` marker synchronously ahead of the resolved window in `useContextWindow` (and adding `activeSession.customModel` to the deps) would keep the two paths consistent.

**3. Minor / optional:** the click-time panel position in `ContextTimelinePanel` does not reclamp on window resize (Greptile P1), so a shrink while it is open can leave it partly offscreen. And the fire-and-forget prime vs. first-turn usage race - I see you already call this out in the PR notes as self-correcting, so I am fine leaving it as-is unless you want to bound-await it.

One heads-up that is not on you: the red `test (windows-latest)` leg is a flaky 10s timeout in `AutoRunBlurSaveTiming.test.tsx`, which this PR does not touch. No need to chase it; a re-run should clear it. No merge conflicts on our end, so no rebase needed.

Thanks again - happy to re-review once #1 and #2 are addressed.

chr1syy:
Thanks for the reviews. Pushed `314859957` addressing all findings:

**Catalog reused across spawn environments (greptile P1 / CodeRabbit)** - The catalog is now keyed by an identity (`computeOmpCatalogKey` = resolved binary + a SHA-1 of the session's stable env overrides), stored per-identity. The key is carried on the managed process as `ompModelCatalogKey` (through `SpawnConfig` -> `ChildProcessSpawner`) and `StdoutHandler` resolves only against the catalog primed for THIS process's binary/env. The detector warms just the default identity. New tests cover two distinct local configs not cross-contaminating and a mismatched-identity process refusing to resolve.

**Bare model keys can collide (greptile P2)** - Conflicting windows on a shared normalized key are now marked ambiguous (`null`); only unambiguous full selectors/ids resolve, everything else falls back. `setOmpModelCatalog` also replaces an identity's snapshot instead of merging (since `omp models --json` is a full snapshot), so stale/ambiguous state can't accumulate across primes. Covered by a bare-id collision test and a snapshot-replace test.

**Don't race first usage against priming (CodeRabbit)** - The spawn now `await`s the prime with a 2s cap: a warm/fast catalog makes the first turn correct, and a slow prime never blocks the spawn (it finishes in the background for later turns).

**Position stays stale after resize (greptile P1)** - The panel now reclamps on `resize` (via `useEventListener`) and re-queries the gauge's live rect each render, so an open panel stays attached and never ends up partly offscreen. (The gauge lives in a fixed top header, so scroll doesn't move it; resize/layout shifts are the cases that do.)

**Custom-model marker precedence in `useContextWindow` (CodeRabbit)** - The `[1m]` custom-model marker is now resolved synchronously and ranked above the runtime-resolved window (with `customModel` added to the deps), matching `useAgentUsageListener` so the header gauge and timeline agree. Added regression tests: resolved-beats-config, and marker-beats-resolved (asserting the marker short-circuits the config fetch).

The one failing `test (windows-latest)` leg was `AutoRunBlurSaveTiming.test.tsx:688` ("very long content can be saved"), a timing-sensitive AutoRun test untouched by this PR (my catalog/StdoutHandler suites passed on that same Windows run). The full `validate:push` gate (all type checks, ESLint, Prettier, 34186 tests) is green locally on the rc base; the fresh CI run should clear the flake.

PR #1222, opened 2026-07-15
perf(session): peel App hooks off selectActiveSession

reachrazamair:
- Peel App-mounted hooks off full `selectActiveSession` / `useActiveSession` so streamed session rebuilds no longer wake `MaestroConsoleInner` through those paths. Handlers resolve the active agent at event time via `getState()`; effects keep only narrow field selectors where needed.
- Hooks covered: modal handlers, Quick Actions, Prompt Composer, session switch, Auto Run document loader, wizard, file explorer effects, session lifecycle, and merge/transfer (including dropping the `sessions` prop from the WithSessions wrappers).
- Does not grow `activeSessionChromeEquality`. Streaming is not fully fixed: App still wakes from raw `sessions[]` subscribers (worktree, starred items, window-scoped active session, Auto Run achievements, remote, Symphony) plus chrome/sidebar equality on busy/state changes.

## Summary by CodeRabbit

- **Performance**
  - Improved responsiveness across session, tab, file explorer, wizard, modal, and Quick Actions interactions.
  - Reduced unnecessary rerenders by reacting to only the needed session fields.

- **Reliability**
  - Merge, send-to-agent, and other action flows now use the latest session/tab state at the moment you click.
  - More consistent behavior when rapidly switching sessions/tabs and during concurrent wizard activity.

- **Bug Fixes**
  - Improved session navigation history, auto-run loading, file explorer updates/keyboard handling, and wizard state synchronization.
  - Modal and Quick Actions now behave more consistently during fast context changes.


pedramamini:
Thanks for this, @reachrazamair - nice, disciplined pass at peeling the App-mounted hooks off the full `selectActiveSession` subscription. The narrow-selector approach is exactly the right shape and the comments explaining *why* each hook avoids the broad subscription are appreciated.

I went through both AI reviews plus the diff. Two of the flagged items are genuine and worth addressing before merge; the other two I believe are false positives (called out below so you don't burn time on them).

### Please address: read/write identity split (both Greptile P1s are real)

`selectActiveSession` is fallback-aware - it returns `sessions.find(id === activeSessionId) || sessions[0]`. In these two hooks the field *reads* now go through that fallback while the *write/guard* path uses the raw `activeSessionId`. When `activeSessionId` is transiently stale (e.g. mid deletion/restoration) the two resolve to different agents, so the effect reads agent A's data and writes to a now-nonexistent id, dropping the update.

1. **`useAutoRunDocumentLoader.ts`** - `autoRunFolderPath` / `autoRunSelectedFile` / `autoRunSshRemoteId` come from `selectActiveSession`, but `sessionId = activeSessionId` and the `setSessions(prev => prev.map(s => s.id === sessionId ? ...))` writes at ~L206/L233 key off that raw id. Content and task counts land on the wrong (or missing) session.
2. **`useFileExplorerEffects.ts`** (pending jump) - `pendingJumpPath` is read fallback-aware, but the clear write keys off `activeSessionId`. The fallback agent's jump runs but never clears, so it can re-fire on the next tree change.

Cleanest fix that keeps the perf win: derive the resolved id from the same selector and use it in the write/guard paths, e.g.

```ts
const resolvedSessionId = useSessionStore((s) => selectActiveSession(s)?.id);
```

then use `resolvedSessionId` (not raw `activeSessionId`) wherever you key the `setSessions` update or guard. That keeps read and write on the same session without reintroducing the broad subscription.

### Likely false positives (no change needed, just my reasoning)

- **CodeRabbit "Critical" - blank file tree on `!fileExplorerExpanded`** (`useFileExplorerEffects.ts`): the prior code was `if (!activeSession || !activeSession.fileExplorerExpanded)`, so a fresh session with undefined expanded set already early-returned. Behavior is unchanged by this PR, not a regression. Fine to leave as-is (or guard on `!activeSessionId` if you want the cosmetic improvement, but it's out of scope for a perf PR).
- **CodeRabbit "Major" - stale `isWizardActiveForCurrentTab`** (`useWizardHandlers.ts`): reactivity is preserved. `isInlineWizardActiveForTab` is a `useCallback` keyed on `tabStates`, so its reference changes whenever a wizard starts/stops on any tab, which re-runs the memo via its dep array. The `getState()` read inside is just the snapshot; the callback dep supplies the reactivity.

No merge conflicts, so nothing needed there. Once the two identity-split writes key off the resolved id, I'll take another look. Thanks again!

PR #1235, opened 2026-07-17
perf(session): move Left Bar sort/starred into sidebarNavStore

reachrazamair:
- Move Left Bar sorted/nav/starred projections into `sidebarNavStore`, owned by `SidebarNavSync`, so App no longer mounts `useSortedSessions` / `useStarredItems` or subscribes to `sessionsForSidebar`.
- SessionList, cycle, and keyboard nav read the store (or getState at event time) instead of App prop plumbing.
- Stop `useSessionFilterMode` from subscribing to the full `sessions` array (restores the Left Bar streaming shield).
- Stabilize `deleteSession` and drop unused `onJumpToStarredSession` props so SessionList `memo` is not busted by App handler churn.
- Add `seedSidebarNav` test helper for SessionList / Auto Run harnesses.

## Summary by CodeRabbit

* **New Features**
  * Centralized left sidebar session sorting, visibility, navigation ordering, and starred-session activation for more consistent behavior.
  * Added unread-only filtering that keeps the active session (and its related worktrees) visible.
  * Improved keyboard navigation and session cycling using the sidebar’s current state.
* **Bug Fixes**
  * Reduced unnecessary UI updates by syncing sidebar state without relying on App-level props/subscriptions.
  * Improved delete-session modal wiring and filter-state handling.
* **Documentation**
  * Updated state-management guidance for the sidebar navigation store.
* **Tests**
  * Extended and rewired integration/component tests to seed and reset the new sidebar navigation state.

### DIFF
--- a/docs/agent-guides/STATE-PATTERNS.md
+++ b/docs/agent-guides/STATE-PATTERNS.md
@@ -20,19 +20,20 @@ All stores are in `src/renderer/stores/`.
 
 ## Store Inventory
 
-| Store                 | File                   | Hook                   | Purpose                                                                                     |
-| --------------------- | ---------------------- | ---------------------- | ------------------------------------------------------------------------------------------- |
-| **sessionStore**      | `sessionStore.ts`      | `useSessionStore`      | Sessions, groups, active session, bookmarks, worktree tracking, initialization              |
-| **uiStore**           | `uiStore.ts`           | `useUIStore`           | UI layout: sidebars, focus, notifications, search, drag-and-drop, editing                   |
-| **tabStore**          | `tabStore.ts`          | `useTabStore`          | Tab operations (CRUD, navigation, metadata), gist state. Wraps tabHelpers.ts + sessionStore |
-| **agentStore**        | `agentStore.ts`        | `useAgentStore`        | Agent detection cache, error recovery, queue processing, agent lifecycle                    |
-| **modalStore**        | `modalStore.ts`        | `useModalStore`        | Modal visibility via registry pattern. Single Map replaces 90+ boolean fields               |
-| **groupChatStore**    | `groupChatStore.ts`    | `useGroupChatStore`    | Group chat state: chats list, messages, moderator, participants, execution queue            |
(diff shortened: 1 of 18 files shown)


pedramamini:
Thanks for this, @reachrazamair - really nice piece of work centralizing the Left Bar sort/nav/starred projections into `sidebarNavStore`. The `SidebarNavSync` host owning the subscriptions is a clean way to keep App from busting `SessionList` `memo`, and the `seedSidebarNav` test helper is a good addition. Before we approve, a few things from the Greptile / CodeRabbit passes look worth addressing:

**1. Duplicate `useEffect` import (must fix)** - `src/__tests__/integration/AutoRunSessionList.test.tsx:24` re-imports `useEffect`, which is already pulled in on line 13. I confirmed this is a real `TS2300: Duplicate identifier` error locally; CI's `lint-and-format` stays green only because the test tsconfig isn't type-checked in that job. Please drop the line 24 import.

**2. Scope the projection to the owning window.** `SidebarNavSync` passes the full `sessions` list into `computeSortedSessions`, so the store's `sortedSessions` / `visibleSessions` / `navIndexMap` are unscoped. `SessionList` re-scopes via `scopeSessionsToWindow` before rendering, but keyboard nav (`useKeyboardNavigation`, `useCycleSession`) and the App cycle context read those fields straight from `useSidebarNavStore.getState()` without re-scoping. In a multi-window setup that looks like it could let Cmd+[ / ] cycling or Opt+Cmd+NUMBER jumps land on another window's agent. Could you either apply `scopeSessionsToOwningWindow` before computing the projection, or confirm the nav consumers already re-scope and this is a no-op?

**3. Initial projection timing.** The projection is published in a post-commit effect, so the first committed render observes the store's empty initial projection and nav refs stay empty until a subsequent update. Greptile and CodeRabbit both flagged this. If it's a real gap for preloaded sessions, seeding from `useSidebarNavStore.getState()` at subscribe time (or a synchronous initial publish) would close it. If you've already verified it's benign, a note on why would help.

No merge conflicts, so no rebase needed - just the fixes above (and the minor `bookmarksCollapsed` seed sync CodeRabbit noted in the test harness, if you agree). Happy to re-review once these are in. Thanks again!

PR #960, opened 2026-05-05
fix(codex): place -C working directory flag before exec subcommand

pedramamini:
## Summary

- Codex treats `-C`/`--cd` as a root-level global flag — when it appears after the `exec` subcommand the CLI silently ignores it and falls back to its own directory resolution. On Windows that often lands on the wrong project entirely (codex#9671 + Maestro #959 repro).
- `buildAgentArgs` now prepends `workingDirArgs` so the directory flag lands at position 0, ahead of any `batchModePrefix`.
- `spawnJsonLineAgent` (CLI/playbook spawn path) reorders the same way so codex batch invocations from playbooks pick up the cwd.
- Updated the integration test fixtures and the Codex command-shape comment in `definitions.ts` to match the new ordering.

Other agents that define `workingDirArgs` are unaffected: gemini-cli has no `batchModePrefix`, and factory-droid's `--cwd` is accepted in either position.

## Test plan

- [x] `npm run lint` (tsc projects) — passes
- [x] `npx vitest run` full suite — 22534 passed / 107 skipped
- [x] `prettier --check` on all changed files — clean
- [ ] Manual verification on Windows with the repro from #959 (a Codex agent in a non-default project directory)

closes #959

## Bug Fixes
* Fixed working directory configuration for Codex agent to be properly recognized and applied.

Issue #959, opened 2026-05-05
Bug: Codex `-C` working directory flag placed after `exec` subcommand - ignored on Windows


jcr211:
## Summary

Codex agents execute in the wrong working directory because the `-C` (working directory) flag is placed **after** the `exec` subcommand in the CLI args. The Codex CLI treats `-C`/`--cd` as a root-level global flag that must appear **before** any subcommand. When it appears after `exec`, it's silently ignored and Codex falls back to its own directory resolution — which on Windows can land on the wrong project entirely (see [codex#9671](https://github.com/openai/codex/issues/9671)).

## Current behavior

```
codex exec --dangerously-bypass-approvals-and-sandbox --skip-git-repo-check --json -C D:\Projects\MyProject -- "prompt"
```

`-C` comes after `exec` → ignored → Codex runs in `C:\` or a cached project directory.

## Expected behavior

```
codex -C D:\Projects\MyProject exec --dangerously-bypass-approvals-and-sandbox --skip-git-repo-check --json -- "prompt"
```

`-C` before `exec` → respected → Codex runs in the correct directory.

## Root cause

Two code paths build the args with `workingDirArgs` appended **after** `batchModePrefix` (`exec`):

### 1. `dist/main/utils/agent-args.js` — `buildAgentArgs()` (desktop app: interactive, Auto Run, Cue)

```javascript
// batchModePrefix is PREPENDED
if (agent.batchModePrefix && inBatchMode) {
    finalArgs = [...agent.batchModePrefix, ...finalArgs];
}
// ...
// workingDirArgs APPENDED — ends up after exec
if (agent.workingDirArgs && options.cwd) {
    finalArgs = [...finalArgs, ...agent.workingDirArgs(options.cwd)];
}
```

### 2. `maestro-cli.js` — `spawnJsonLineAgent()` (CLI path)

```javascript
const preOverrideArgs = [];
if (def?.batchModePrefix) preOverrideArgs.push(...def.batchModePrefix);  // "exec" first
// ... batchModeArgs, jsonOutputArgs, resumeArgs ...
if (toolType === "codex" && def?.workingDirArgs) {
    preOverrideArgs.push(...def.workingDirArgs(cwd));  // -C after exec
}
```

### `agent-args.js` — move `workingDirArgs` right after `batchModePrefix`, using **prepend** so it lands before the subcommand:

```javascript
if (agent.batchModePrefix && inBatchMode) {
[...]

PR #1243, opened 2026-07-18
fix(cli): place Codex -C working-dir flag before exec subcommand

chr1syy:
## Problem

`maestro-cli send` (and Maestro Relay follow-up messages) route Codex through `spawnJsonLineAgent` in `src/cli/services/agent-spawner.ts`. That function pushed `workingDirArgs` (`-C <dir>`) onto the arg list **last**, after the `exec` subcommand and after `resume <id>`:

```
codex exec --dangerously-bypass-approvals-and-sandbox --skip-git-repo-check --json resume <id> -C <dir> -- "prompt"
```

Codex requires `-C` as a **root-level global flag** that must precede the `exec` subcommand. On a resume it lands after `resume <id>`, where the `resume` subcommand rejects it and Codex hard-fails:

```
error: unexpected argument '-C' found
```

This broke Relay follow-up messages to Codex agents. Fresh spawns tolerated the misplaced flag (though `exec` may silently ignore the cwd per #959); only the resume path crashed.

The desktop path (`src/main/utils/agent-args.ts`) already prepends `workingDirArgs` before the batch-mode prefix; only the CLI spawner diverged. This is a regression of #960, reintroduced during the additional-directories refactor.

## Fix

Move the Codex `workingDirArgs` push to the front of `preOverrideArgs`, before `batchModePrefix`. Final order is now `-C <dir> exec ... resume <id> -- "prompt"`, matching the desktop path.

## Verification

Reproduced against `codex-cli 0.142.5`:

- **Old order** (`... resume <id> -C <dir>`): `error: unexpected argument '-C' found` (matches the report).
- **New order** (`-C <dir> exec ... resume <id>`): clears argument parsing (fails only on a non-existent test session id, as expected).

Strengthened the existing spawner test to enforce the `-C < exec < resume` ordering invariant. The prior test only checked that `-C` was present, so it passed while the ordering was broken; the new assertion fails against the old ordering and passes with the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Codex working-directory handling by placing the directory option before the command and resume arguments.
  * Improved Codex startup and resume reliability when running from a specified working directory.
  * Added regression coverage to verify command-line argument ordering.

### DIFF
--- a/src/__tests__/cli/services/agent-spawner.test.ts
+++ b/src/__tests__/cli/services/agent-spawner.test.ts
@@ -2275,6 +2275,16 @@ Some text with [x] in it that's not a checkbox
 			// resume args are ['resume', '<id>']
 			expect(args).toContain('resume');
 			expect(args).toContain('codex-thread-123');
+
+			// Ordering is load-bearing: `-C` is a ROOT-level global flag that MUST
+			// precede the `exec` subcommand, which in turn precedes `resume <id>`.
+			// Placing `-C` after `resume` makes Codex hard-fail with
+			// "unexpected argument '-C' found", breaking Relay follow-up messages
+			// to Codex agents (regression of #960). Enforce -C < exec < resume.
+			const execIdx = args.indexOf('exec');
+			const resumeIdx = args.indexOf('resume');
+			expect(c).toBeLessThan(execIdx);
+			expect(execIdx).toBeLessThan(resumeIdx);
 		});
 
 		it('unsupported agent type returns a failure result', async () => {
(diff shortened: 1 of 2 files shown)


pedramamini:
@chr1syy Thanks for the contribution - this is a well-scoped fix with a clear root-cause writeup.

Verified locally on the PR branch:

- The relocation matches the desktop path, which prepends `workingDirArgs` ahead of everything else (`src/main/utils/agent-args.ts:135`). CLI and desktop now agree on `-C <dir> exec ... resume <id>`.
- `-C` is pushed exactly once - the old trailing push is fully removed, no duplicate flag.
- The strengthened test genuinely enforces the invariant (`-C` < `exec` < `resume`) rather than just asserting presence, so it will catch a re-regression of #960.
- Full `agent-spawner.test.ts` suite passes: 127/127.
- No merge conflicts, and both Greptile and CodeRabbit came back clean.

Appreciate the reproduction against `codex-cli 0.142.5` and the note that the prior test passed while the ordering was broken - that context made this quick to confirm. Approving.

PR #1244, opened 2026-07-19
feat(cli): dispatch backgrounds by default, add --focus opt-in

chr1syy:
## What

`maestro-cli dispatch` now delivers a prompt **in the background by default**: it no longer switches to or focuses the target agent/tab. Pass the new `--focus` flag to bring the target to the foreground (the previous default). `--new-tab` likewise creates the fresh tab in the background unless `--focus` is given.

```bash
dispatch <agent> <msg>                     # background (default), no focus
dispatch <agent> <msg> --focus             # foreground, switches to & focuses the tab
dispatch <agent> <msg> --new-tab           # fresh tab, background
dispatch <agent> <msg> --new-tab --focus   # fresh tab, focused
```

## Why

Dispatching (especially from automation like Cue, Pianola, and the orchestrator) was yanking the user's view to the target agent on every call. Background delivery is the sensible default for autonomous, non-interactive dispatch; foregrounding becomes an explicit opt-in.

## How

A `background` bit is threaded end to end and only its final consumers act on it:

- **CLI** (`src/cli/index.ts`, `src/cli/commands/dispatch.ts`): `--focus` maps to Commander `focus === true`; `runDispatch` sets `background = options.focus !== true` and adds `background: true` to the `send_command` / `new_ai_tab_with_prompt` payloads.
- **Transport** (`web-server/handlers/messageHandlers.ts`, `types.ts`, `managers/CallbackRegistry.ts`, `WebServer.ts`, `web-server-factory.ts`, `preload/process.ts`, `renderer/global.d.ts`): optional `background` param on `ExecuteCommandCallback` / `NewAITabWithPromptCallback` and the IPC bridges.
- **Renderer** (`renderer/hooks/remote/useRemoteIntegration.ts`): when `background`, skips `setActiveSessionId` and creates the new tab via `createTab({ activate: false })`.
- **`createTab`** (`renderer/utils/tabHelpers.ts`): new `activate?: boolean` option (default `true`); when `false`, appends the tab but preserves every active-tab id, `activeGroupId`, and `inputMode` so the visible view never changes.

Programmatic `runDispatch` callers (Pianola watcher, orchestrator) pass no focus preference, so they now background by default too, which is the desired behavior for autonomous dispatch.

## Testing

Extended and passing: `dispatch`, `messageHandlers`, preload `process`, `useRemoteIntegration`, and `tabHelpers` suites (payloads now carry `background: true` by default; `--focus` omits it; renderer suppresses focus and background-creates tabs; `createTab({ activate: false })` preserves focus fields). Prettier and ESLint clean on all touched files; the 16 changed files add zero TypeScript errors.

## Summary by CodeRabbit

* **New Features**
  * Enabled background-by-default dispatch for CLI and remote command flows, keeping the current view in place.
  * Added `--focus` to switch to and focus the target agent/tab during dispatch.
  * Added background creation for new AI tabs, inserting them without changing the user-visible active panel state.
* **Bug Fixes**
  * Made “background” behavior consistent across command execution, image-based sends, and remote integrations, including correct foreground behavior when `--focus` is provided.

### DIFF
--- a/src/__tests__/cli/commands/dispatch.test.ts
+++ b/src/__tests__/cli/commands/dispatch.test.ts
@@ -204,6 +204,34 @@ describe('dispatch command', () => {
 				'new_ai_tab_with_prompt_result'
 			);
 		});
+
+		it('forwards tabId but omits background when --tab is combined with --focus', async () => {
+			vi.mocked(resolveAgentId).mockReturnValue('agent-abc-123');
+			const mockSendCommand = vi.fn().mockResolvedValue({
+				type: 'command_result',
+				success: true,
+				tabId: 'tab-xyz',
+			});
+			vi.mocked(withMaestroClient).mockImplementation(async (action) => {
+				const mockClient = { sendCommand: mockSendCommand };
+				return action(mockClient as never);
+			});
+
+			// The options.tab spread runs alongside the background omission; this
+			// covers that distinct payload-building path (tabId present, no background).
+			await dispatch('agent-abc', 'Follow up', { tab: 'tab-xyz', focus: true });
+
+			expect(mockSendCommand).toHaveBeenCalledWith(
+				{
+					type: 'send_command',
+					sessionId: 'agent-abc-123',
+					command: 'Follow up',
+					inputMode: 'ai',
+					tabId: 'tab-xyz',
+				},
+				'command_result'
+			);
+		});
 	});
 
 	describe('--tab <tabId> flow', () => {
--- a/src/__tests__/cli/commands/dispatch.test.ts
+++ b/src/__tests__/cli/commands/dispatch.test.ts
@@ -33,7 +33,7 @@ describe('dispatch command', () => {
 	});
 
 	describe('default (active tab) flow', () => {
-		it('sends send_command with no tabId and surfaces the desktop-supplied tabId', async () => {
+		it('sends send_command in the background by default (no tabId) and surfaces the desktop-supplied tabId', async () => {
 			vi.mocked(resolveAgentId).mockReturnValue('agent-abc-123');
 			const mockSendCommand = vi.fn().mockResolvedValue({
 				type: 'command_result',
@@ -53,6 +53,7 @@ describe('dispatch command', () => {
 					sessionId: 'agent-abc-123',
 					command: 'Hello world',
 					inputMode: 'ai',
+					background: true,
 				},
 				'command_result'
 			);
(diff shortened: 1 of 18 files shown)


pedramamini:
@chr1syy Thanks for the contribution! The design here is genuinely nice: threading a single `background` bit end to end with `message.background === true` guards at every hop keeps this fully backwards compatible, and the `createTab({ activate: false })` path that preserves every active-tab id, `activeGroupId`, and `inputMode` is exactly the right way to avoid disturbing the user's view. Backgrounding autonomous dispatch by default is the correct call.

Before we can merge, though, all three CI legs are red, and the failures do trace back to this branch. No merge conflicts, so no rebase needed.

**1. Prettier (`lint-and-format`)**

Three touched files need reformatting:

- `src/__tests__/renderer/hooks/useRemoteIntegration.test.ts`
- `src/main/web-server/handlers/messageHandlers.ts`
- `src/main/web-server/WebServer.ts`

`npx prettier --write` on those should clear it.

**2. Two test suites were not updated for the new trailing argument (`test` on both ubuntu and windows)**

Adding `background` as a new trailing parameter to `executeCommand` broke 10 existing exact-arity `toHaveBeenCalledWith` assertions in two suites that this PR does not touch:

- `src/__tests__/main/web-server/web-server-factory.test.ts` (4 failures in `executeCommandCallback behavior`)
- `src/__tests__/main/web-server/managers/CallbackRegistry.test.ts` (6 failures in `setExecuteCommandCallback / executeCommand()`)

Each fails the same way, for example:

```
AssertionError: expected "vi.fn()" to be called with arguments: [ 'remote:executeCommand', ...(6) ]
```

The assertions pin the full argument list, so the extra `background` arg makes them mismatch. I confirmed these are yours rather than pre-existing: both suites pass locally on `rc` at `fbc3fa371` (151/151), which is the exact base this PR was reviewed against. Updating those assertions to expect the new trailing argument should be all that is required.

**Optional, not a blocker**

Greptile flagged that there is no test for `--tab` combined with `--focus`. Worth adding while you are in there, but happy to merge without it.

Once CI is green, ping me and I will take another look.

chr1syy:
Thanks for the detailed CI triage @pedramamini - all three items are addressed in `9d1b694`:

**1. Prettier (`lint-and-format`)**
Reformatted the three files with prettier **3.8.3** (the version pinned in `package-lock.json` / `npm ci`). Root cause of my earlier "clean locally" claim: my local install had resolved `^3.8.0` to 3.9.5, which formats these union types inline where 3.8.3 wraps them. Formatting with the pinned 3.8.3 and re-checking is now clean.

**2. Broken exact-arity assertions**
Updated all 10 `toHaveBeenCalledWith` assertions for the new trailing `background` argument:
- `src/__tests__/main/web-server/web-server-factory.test.ts` (4, in `executeCommandCallback behavior`)
- `src/__tests__/main/web-server/managers/CallbackRegistry.test.ts` (6, in `setExecuteCommandCallback / executeCommand()`)

Both suites now pass locally (151/151). Good catch - those two suites don't import the changed code so they weren't on my radar during the original pass.

**3. Greptile `--tab` + `--focus`**
Added `forwards tabId but omits background when --tab is combined with --focus` to `dispatch.test.ts`, covering the payload path where `options.tab` spreads alongside the background omission (tabId present, no `background` field).

Full local run across the affected + previously-failing suites: **709 passing**, prettier 3.8.3 clean, ESLint clean. Ready for another look.

chr1syy:
@greptile re-review please

PR #1248, opened 2026-07-19
fix(tab-naming): don't scrape a live TUI transcript for the tab name

### LINKED ISSUE #1110: Bug: Local Claude Code agent streams response into tab name, not chat
## Summary
Local Claude Code agent streams response into tab name, not chat

## Environment
- Maestro version: 0.18.0-RC

## Steps to Reproduce
1. Configure a local Claude Code agent using the local binary in Maestro.
2. Send any prompt to the agent.
3. Observe that the streaming tokens (thinking + response) populate the tab name/title rather than the chat message area, and the chat body remains empty. (SSH remote agents are unaffected and stream into the chat correctly.)

## Expected Behavior
Streamed output from the local Claude Code binary (both thinking and the final response) should render in the chat message area / response body of the conversation.

## Actual Behavior
The streamed output renders into the tab name/title instead of the chat body. Both the thinking and the final response tokens are misrouted to the tab name. The chat body stays empty — the text never appears there, not even after streaming completes.

## Additional Context
Issue is isolated to the local Claude Code binary; SSH remote agents stream into the chat normally, which points to a stream routing/rendering bug specific to the local binary integration. Video proof: https://drive.google.com/file/d/1s0WC5hAlU1fIsB7B-L2l7WG7LL38RP6P/view?usp=sharing . Environment: Maestro 0.18.0-RC, macOS.

## Screenshots / Recordings
None

## Support Package
[maestro-debug-package.zip](https://raw.githubusercontent.com/felipeggv/maestro-feedback-attachments/main/feedback/1781898424427-maestro-debug-2026-06-19T19-47-03.zip)

pedramamini:
Closes #1110

## Problem

On the maestro-p TUI path (Claude token mode = TUI/dynamic), maestro-p strips the headless-only flags, including the `--tools ""` guard that keeps tab naming from becoming a real agentic run. The naming spawn therefore runs a full turn and emits a plain terminal transcript rather than stream-json.

`extractTabNameFromOutput` tried the agent's parser first, and when that found no JSON it fell back to scraping the raw output. On this path the raw output is a live transcript of thinking and response prose, so an arbitrary sentence fragment landed on the tab. That is what the reporter saw as "the response is streaming into the tab name."

`STRUCTURAL_NOISE_RE` (added in 4a308b31) catches the tool-call markup and `(no content)` variants of this leak, but it cannot catch plain prose - a thinking fragment has no angle brackets and clears the 40-char line filter.

This also explains the reported local-vs-SSH asymmetry: an SSH remote without `maestro-p` on its PATH collapses back to `claude --print`, which yields clean stream-json.

## Fix

Require structured output when the spawn resolved to interactive mode. If the parser cannot lift a clean result there, return null instead of scraping the buffer. Returning null is cheap: the send-side trigger retries naming on the next message, so the tab keeps its placeholder and self-heals rather than sticking a fragment.

The gate is `claudeSpawnDecision?.mode === 'interactive'`, which covers both the local maestro-p wrap and the remote maestro-p case. Headless `claude --print` and non-Claude agents are untouched and keep the raw-output fallback.

## Note on scope

The issue also reports the chat body staying empty. I could not find a code path where stream chunks are routed away from the chat: `useAgentDataListener` is the sole consumer and only ever appends to the log, and the naming spawn runs under a disjoint `tab-naming-*` id that cannot match `REGEX_AI_TAB`. This PR fixes the tab-name half, which is the part I could reproduce from the code. If the empty chat body persists for the reporter after this lands, that is worth a separate issue with a debug package.

## Testing

- Two regression tests in the `Claude token-source resolution` block: a plain-text TUI transcript now resolves to null, and a stream-json result on the same path still yields the name.
- Full file: 46/46 passing. `npm run lint` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic tab naming for interactive sessions by using only structured results.
  * Prevented plain-text progress or conversation output from being incorrectly selected as a tab name.
  * Preserved valid tab names when structured result data is available.

* **Tests**
  * Added regression coverage for ignoring unstructured output and correctly handling structured results.

### DIFF
--- a/src/__tests__/main/ipc/handlers/tabNaming.test.ts
+++ b/src/__tests__/main/ipc/handlers/tabNaming.test.ts
@@ -1158,6 +1158,80 @@ describe('Tab Naming IPC Handlers', () => {
 			await resultPromise;
 		});
 
+		// Regression for #1110: on the TUI path maestro-p strips the headless-only
+		// flags (including the `--tools ""` guard), so naming runs a real agentic
+		// turn and its raw terminal transcript is thinking/response prose, not
+		// stream-json. Scraping that buffer put a sentence fragment on the tab,
+		// which users saw as the response streaming into the tab name. Prose has no
+		// angle brackets, so STRUCTURAL_NOISE_RE can't catch it - we must decline.
+		it('returns null instead of scraping a plain-text TUI transcript for a name', async () => {
+			mockAgentDetector.getAgent.mockResolvedValue(interactiveClaudeAgent);
+			let onDataCallback: ((sessionId: string, data: string) => void) | undefined;
+			let onExitCallback: ((sessionId: string, code?: number) => void) | undefined;
+			mockProcessManager.on.mockImplementation(
+				(event: string, callback: (...args: any[]) => void) => {
+					if (event === 'data') onDataCallback = callback;
+					if (event === 'exit') onExitCallback = callback;
+				}
+			);
+
+			const resultPromise = invokeHandler('tabNaming:generateTabName', {
+				userMessage: 'Help me implement a login form',
+				agentType: 'claude-code',
+				cwd: '/test/project',
+				enableMaestroP: true,
+				maestroPMode: 'interactive',
+				maestroPPath: '/bundled/maestro-p.js',
+			});
+
+			await vi.waitFor(() => {
+				expect(mockProcessManager.spawn).toHaveBeenCalled();
+			});
+
+			// A realistic maestro-p TUI transcript: short prose lines that clear the
+			// 40-char filter and carry no structural noise.
+			onDataCallback?.(
+				'tab-naming-mock-uuid-1234',
+				'I should look at the auth flow\nLet me start with the form\n'
+			);
+			onExitCallback?.('tab-naming-mock-uuid-1234', 0);
+
+			await expect(resultPromise).resolves.toBeNull();
+		});
+
+		it('still accepts a stream-json name on the TUI path', async () => {
+			mockAgentDetector.getAgent.mockResolvedValue(interactiveClaudeAgent);
+			let onDataCallback: ((sessionId: string, data: string) => void) | undefined;
+			let onExitCallback: ((sessionId: string, code?: number) => void) | undefined;
+			mockProcessManager.on.mockImplementation(
+				(event: string, callback: (...args: any[]) => void) => {
+					if (event === 'data') onDataCallback = callback;
+					if (event === 'exit') onExitCallback = callback;
+				}
+			);
+
+			const resultPromise = invokeHandler('tabNaming:generateTabName', {
+				userMessage: 'Help me implement a login form',
+				agentType: 'claude-code',
+				cwd: '/test/project',
+				enableMaestroP: true,
+				maestroPMode: 'interactive',
+				maestroPPath: '/bundled/maestro-p.js',
+			});
+
+			await vi.waitFor(() => {
+				expect(mockProcessManager.spawn).toHaveBeenCalled();
+			});
+
+			onDataCallback?.(
+				'tab-naming-mock-uuid-1234',
+				`${JSON.stringify({ type: 'result', subtype: 'success', result: 'Login Form' })}\n`
+			);
+			onExitCallback?.('tab-naming-mock-uuid-1234', 0);
+
+			await expect(resultPromise).resolves.toBe('Login Form');
+		});
+
 		it('spawns plain claude when the agent is API-only (enableMaestroP false)', async () => {
 			mockAgentDetector.getAgent.mockResolvedValue(interactiveClaudeAgent);
 			const finish = wireProcessEvents();
--- a/src/main/ipc/handlers/tabNaming.ts
+++ b/src/main/ipc/handlers/tabNaming.ts
@@ -382,6 +382,13 @@ export function registerTabNamingHandlers(deps: TabNamingHandlerDependencies): v
 						});
 					}
 
+					// A TUI (maestro-p) naming spawn emits a plain terminal transcript,
+					// not stream-json, so the raw-output fallback in extraction would
+					// scrape live thinking/response prose onto the tab. Require
+					// structured output on that path. Covers both the local wrap above
+					// and the remote maestro-p case, which share the same decision mode.
+					const requireStructuredOutput = claudeSpawnDecision?.mode === 'interactive';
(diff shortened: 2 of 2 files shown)


chr1syy (commented):
Reviewed this one closely because the diagnosis direction is right and the spelunking behind it is clearly real - but I think the patch is aimed one layer too high, and it landed on a branch that's missing its prerequisite. **I believe #1110 still reproduces at `a1c6fb8`.** Details and a runnable repro below.

CI is 7/7 green and the PR is `MERGEABLE`/`CLEAN`, so this is purely about the logic.

---

## 1. Greptile's P1 is a false positive - safe to dismiss

> *Interactive Tabs Never Self-Name ... maestro-p normally emits the plain transcript this gate rejects.*

maestro-p doesn't emit a plain transcript on stdout. `JsonEmitter.writeLine()` -> `process.stdout` (`src/maestro-p/json-emitter.ts:153-155`) is the **only** stdout writer in the binary; the claude TUI's terminal output goes to the pty data handler (`tui-driver.ts:202`) and never reaches stdout. maestro-p's stdout is exclusively stream-json envelopes.

So `extractAgentResponseText()` always sees JSON on this path, `structuredOnly` never trips, and tabs don't get stuck on the placeholder. No fix needed for that concern.

## 2. ...but that same fact is why the gate never fires

`structuredOnly` only engages when `extractAgentResponseText()` returns `null`, i.e. when *no* JSON was seen. On the interactive path JSON is always seen, so the new gate is skipped in exactly the scenario it was written for.

Worse, the thinking text is *inside* those envelopes. `emitAssistantMessage()` is called unconditionally (`src/maestro-p/index.ts:456`) and forwards the JSONL `.message` **verbatim, thinking blocks included**. `ClaudeOutputParser.parseJsonObject()` then turns those into `{ type: 'text' }` events (`claude-output-parser.ts:153-168`, `contentToEmit = thinkingText || text`), which `extractAgentResponseText()` accumulates into `assistantText`. The early-extract interval resolves from that partial buffer before the terminating `result` event ever arrives.

Running the real `ClaudeOutputParser` plus the extraction helpers as they stand at `a1c6fb8`, against what maestro-p actually writes:

```js
const stdout = [
  '{"type":"system","subtype":"init","session_id":"a","model":"m","cwd":"/p"}',
  JSON.stringify({ type: 'assistant', message: { role: 'assistant', content: [
    { type: 'thinking', thinking: 'The user wants a login form. I should check the existing auth flow first. Let me look at the components directory.' }
  ]}}),
].join('\n');
```

```
responseText           => "The user wants a login form. I should check the existing
                           auth flow first. Let me look at the components directory."
gate fires?            => false
WITH gate (what ships) => { name: 'Let me look at the components directory', reason: 'ok' }
```

That's #1110, verbatim, with the fix applied.

## 3. Root cause: this is on the wrong base branch

`rc` already carries the actual root-cause fix for this extraction path, and `main` does not:

```diff
-		} else if (event.type === 'text') {
+		} else if (event.type === 'text' && !event.isReasoning) {
+			// Reasoning deltas ... stream BEFORE the answer, so mining them would let
+			// early extraction resolve with a thinking fragment ...
```

Same function, same bug class, same reasoning. Re-running the probe on **`rc`'s** extraction logic, the gate finally earns its keep:

| Turn shape | rc, gate OFF | rc, gate ON |
|---|---|---|
| thinking-only | `'I should check the auth flow first'` :x: | `null` :white_check_mark: |
| response prose | `'It will live in src/components/auth'` :x: | `'It will live in src/components/auth'` :x: |

So the gate **is** a genuine improvement - but only stacked on top of `rc`'s `isReasoning` guard. Merged into `main` as-is it ships a fix that doesn't fix the reported bug.

CONTRIBUTING.md:938 says bug fixes target `main`, so this followed policy; the policy just doesn't survive this particular dependency. Suggest rebasing onto `rc`.

Note the second row: even on `rc` the response-prose half survives, because early extraction over a genuinely agentic turn is unsound by construction. My instinct is that the interactive path should require the terminal `result` event rather than mining `assistantText` at all - happy to be talked out of it, but a partial-buffer scrape can't distinguish "the name" from "a sentence of the answer."

## 4. The tests pass because the fixture can't occur

Both new tests feed **raw prose with no JSON envelope**:

```js
onDataCallback?.('tab-naming-mock-uuid-1234',
  'I should look at the auth flow\nLet me start with the form\n');
```

maestro-p structurally cannot produce that - every stdout line is a JSON envelope. The fixture lands on `structured_only_no_json`, so the suite goes green while the real path stays broken. A fixture built from `{"type":"assistant","message":{...}}` envelopes (as in §2) would fail today, which is exactly the regression test this needs.

## 5. The stated mechanism doesn't hold up

The PR body, the new 10-line comment on `extractTabNameFromOutput`, and the explanation already posted to the reporter in #1110 all rest on:

> maestro-p strips the headless-only flags, **including the `--tools ""` guard**

Running the real parser (`parseArgs(['--print','--verbose','--output-format','stream-json','--tools','','a prompt'])`):

```
warn: maestro-p: ignoring --verbose - headless-mode flag, not forwarded to the TUI.
warn: maestro-p: ignoring --output-format - headless-mode flag, not forwarded to the TUI.
passThroughArgs (forwarded to the claude TUI): ["--tools",""]
```

`--tools` is **passthrough**, not stripped. The stripped set is `--output-format` / `--input-format` / `--verbose` (`args.ts:105-110`). The turn goes agentic because the interactive TUI ignores a `--print`-mode flag, not because maestro-p removed it.

Cosmetic on its own, but it's the load-bearing premise of the explanation in three places including the one the reporter has already read - worth correcting so it doesn't get built on later.

## 6. `Closes #1110` covers only half the report

The issue reports two symptoms: text lands in the tab name, **and** "the chat body stays empty ... the text never appears there, not even after streaming completes." This PR addresses the tab-name half only. As written it'll auto-close the issue with the reporter's primary complaint still open - suggest downgrading to `Refs #1110` unless the empty-chat-body half is handled separately.

---

**Summary:** §2/§3 are the blocking pair - the gate can't fire on the path it targets, and the prerequisite guard lives on `rc`. A rebase onto `rc` gets the thinking half working immediately; the response-prose half needs the separate call in §3. §4 explains why CI didn't catch any of this.

Verified by running the actual `ClaudeOutputParser` and maestro-p `parseArgs` against the envelopes the binary really emits, not by reading alone - happy to share the full probe script if useful.

Issue #1231, opened 2026-07-16
Feature: Provide more insight to long running "thinking" tasks


tseven:
### Summary

Unable to see what the agent is doing behind the scenes

### Problem or opportunity

The GUI hides helpful information about what's currently happening like:

- Number of background shell commands and agents running
- Natural status updates of the agent thinking and "almost done thinking" etc.

The task name above the text input looks and acts like a clickable element but doesn't do anything.
I have no visibility into what is happening.

<img width="1494" height="1220" alt="Image" src="https://github.com/user-attachments/assets/c758ab4c-3ccc-45ae-a869-caad202dc72f" />

### Desired outcome

Ability to zoom in and see what claude code is actually doing. Replicate what you see in the terminal.
Doesn't need to be fancy, just would like to know what's happening.
Is there an API issue? Network connectivity issue?
Is this thread broken?
I've experienced threads getting corrupted and cannot continue. This shows up as "API timeout" or something, but never resolves. 

### Additional context
Here is how CC running in a terminal indicates background tasks are running
<img width="1746" height="372" alt="Image" src="https://github.com/user-attachments/assets/e079b10b-b3cd-49c9-8454-4682fe9b9335" />

### Maestro version

0.17.3

### Operating system

macOS

### Install source

None

### Product area

None

### Agent/provider involved

None

### Is SSH remote execution part of this request?

None

pedramamini:
Thanks for the detailed writeup and screenshots, @tseven - the "the task name above the input looks clickable but does nothing" observation was spot on.

That element is the status pill's session/tab name, and it was styled to look clickable but its click only re-focused the already-active session, so it appeared inert. We put up a proposed fix in #1238 that wires it (and the "+N running processes" dropdown rows) to open Maestro's **live Thought Stream** for that session - a real-time view of what the agent is reasoning about, which is the "zoom in and see what Claude is actually doing / replicate the terminal" view you're after. It reuses the existing Thought Stream capture pipeline (previously only reachable from the Auto Run panel), so interactive sessions now get an entry point too.

We scoped this PR to that concrete first step. The other parts of your request - a count of background shell commands / running tool calls, and "almost done thinking"-style status text - are good follow-ups we'd tackle separately.

Does this direction match what you had in mind? Any feedback on the Thought Stream as the "zoom in" surface would be very welcome before we finalize.

tseven:
Yeah that is definitely in the right direction. 

[...]

PR #1254, opened 2026-07-20
Agent Flow plugin + tool.executed event and ui.panelPost host channel (host API 1.13.0)

chr1syy:
## What this is

Adds **Agent Flow**, a tier-2 example plugin that visualizes agent execution live as a node graph (tool-call lanes, timing, pan/zoom, inspector, timeline, and an activity/health overlay for issue #1231). To make it possible, it also lands two small **additive** host-API surfaces, both at **host API `1.13.0`**.

Built as a 7-phase playbook; this PR is the whole effort.

## Host-API additions (both additive / MINOR)

- **`tool.executed` plugin event topic** (`src/shared/plugins/events.ts`) - a metadata-only tool-call lifecycle event: `sessionId`, `toolName`, `timestamp`, optional `toolCallId` and `phase`. Never carries tool arguments, results, or the tool `state` object. Emitted alongside the existing renderer forward in `forwarding-listeners.ts`.
- **`maestro.ui.panelPost(panelId, data)` host-to-panel push** (`src/shared/plugins/rpc-protocol.ts`, cap `MAX_PANEL_POST_BYTES = 64 KB`) - the first host-to-panel data channel. Own declared panels only, JSON only, size-capped, one-way (no reply channel). Delivered to the panel page as a `maestro:panelData` window message.

`HOST_API_VERSION` bumped `1.12.0 -> 1.13.0`; the vendored `@maestro/plugin-sdk` contracts and its drift-guard test are synced; `CLAUDE-PLUGINS.md` and `docs/agent-guides/PLUGIN-DEVELOPMENT.md` updated.

## The plugin (`examples/plugins/agent-flow/`)

- `main.js` (sandbox entry): subscribes to `tool.executed` plus agent/session/usage topics, maintains a per-session tool-call graph, and pushes coalesced snapshots (one per 250 ms, under 64 KB) to its `flow` panel.
- `panel.html`: one self-contained file (vanilla JS + inline SVG/CSS, no external references) rendering the node graph, pan/zoom, inspector, timeline, session tabs, and the issue #1231 activity/health overlay (fleet counts, per-lane coarse status, running-tool counts, live elapsed timer, stall warning, error badge).

## Security invariants (audited, all PASS)

Recorded in the plugin README "Security notes" section:

1. `tool.executed` payload has no tool arguments/results/`state` at the emit site.
2. `ui.panelPost` is broker-gated on `ui:panel`, own-panels-only, JSON + size-capped, and fails closed when its sink is unwired.
3. The panel guest preload relays only the two fixed shapes, exposes no `ipcRenderer`, and has no reply channel.
4. `panel.html` has no external references (only the inert SVG `xmlns`).
5. No read path calls `PluginManager.refresh()`.
6. The activity/health overlay renders metadata only (counts, coarse status, timing, error type/recoverable) - never thinking prose, tool arguments, or results.

## Validation

- `npm run lint` (all three tsc configs): clean.
- `eslint src/`: clean.
- Scoped tests: 196 root (forwarding-listeners, plugin-event-bus, plugin-host-handlers, plugin-host-deps-wiring, permission-broker, events, contributions, rpc-protocol, rpc-and-signing) + 29 `packages/plugin-sdk` (drift guard + type checks). All green.
- `node --check main.js` passes; `panel.html` structurally validated.

Full-suite and cross-OS coverage will run in CI here.

## Known limitations (by design)

- Metadata only: no tool arguments/results ever reach the plugin (event payloads are metadata-only by contract).
- Coarse health, not thinking text: the overlay shows a coarse status plus derived stall/error signals, not the provider's free-form thinking prose.
- No subagent nesting yet: `parent_tool_use_id` is not parsed by the claude output parser.
- Snapshots capped at 64 KB (per-lane node history trimmed oldest-first under load).

## Summary by CodeRabbit

* **New Features**
  * Added the “Agent Flow” example plugin with an execution-graph UI, timelines, lane health badges, and interactive inspection.
  * Plugins can now receive one-way, size-limited JSON data pushes into their own panels via `ui.panelPost` / `plugins.onPanelData` (64 KB cap).
  * Added the `tool.executed` metadata-only event for tool-call phase tracking.
  * Updated host API/SDK support to **1.14.0**.
* **Bug Fixes**
  * Added fail-closed permission checks and hardened payload validation (closed schema, JSON-serializable data, size limits; unsupported fields like tool `state` are stripped).
* **Documentation**
  * Expanded plugin guides/security notes for `ui.panelPost` and `tool.executed`.
* **Tests**
  * Added/updated coverage for the new event, panel data channel, and `ui.panelPost` behavior.


chr1syy:
Thanks for the review. Addressed in `511291b24` (verified against the host event contracts in `src/shared/plugins/events.ts` and the emit sites):

- **Agent events update wrong/ghost lane** (`main.js`): `agent.statusChanged` / `agent.awaiting` now apply the coarse agent-level status to *every* existing lane matching the `agentId` and no longer synthesize a nodeless `agentId`-keyed ghost lane. `tabId` isn't used for precise per-tab routing because the `sessionId`->tab mapping is host-internal (documented as a limitation).
- **Usage deltas replace totals** (`main.js`): confirmed `usage.updated` emits per-turn counts (the host keeps its own running totals only for the terminal `agent.completed`), so the plugin now accumulates tokens/cost across turns; `contextWindow` takes the latest value.
- **Running events never reach terminal state** (`forwarding-listeners.ts`): mitigated on the consumer side. On a terminal `agent.completed` / `agent.exited` / `run.completed`, the plugin closes any still-open tool nodes and zeroes the running count, so a producer that emits `running` without a terminal event no longer leaves a lane "working" forever. The host-parser fidelity itself (a producer emitting only `running`) is pre-existing and tracked as future work (tool-result status fidelity).
- **Success retains error state** (`main.js`): `lastError` is now cleared on *any* tool activity (open, terminal, or phase-less), not only when opening a new node.
- **Minimum node cap still oversizes** (`main.js`): after the per-lane node-cap shrink, if even one node per lane still exceeds the cap, the least-recently-active lanes are dropped so the host accepts a reduced snapshot instead of rejecting the post. The fleet `summary` still counts every lane.
- **Late panel mount loses state** (`main.js` + `plugin.json` + `panel.html`): added a `sync` command the panel invokes on load and on becoming visible, so a panel opened after activity ended pulls the current snapshot instead of sitting empty.
- **Unused `storage:*` permissions** (`plugin.json`): removed `storage:read` / `storage:write` (the plugin never touches `sdk.storage`).
- **Blanket catch** (`PluginPanelFrame.tsx`): the host-to-panel send now drops only the expected not-attached / tearing-down races and reports anything else via `captureException`, so real regressions still reach Sentry.

Validation: `prettier --check` clean, `eslint src/` clean, `tsc` (3 configs) clean, manifest validates, `main.js` parses and passes a stub-driven behavior check, `panel.html` stays self-contained, and the affected renderer tests are green.

PR #1256, opened 2026-07-20
fix(sidebar): show errored agents in the unread agents filter

chr1syy:
## Problem

Turning on **Filter unread agents** (the bell in the Left Bar) hid any agent in an `error` state. The filter keeps agents visible when they are unread, busy, auto-running, or stuck auto-retrying an outage; an errored agent matched none of those predicates, so it silently disappeared exactly when the user most needs to see it.

## Fix

Add `state === 'error'` alongside the existing `busy` check in every parallel "needs attention" predicate, for both parents and worktree children, so the rendered list, bell badge, jump numbers, and keyboard cycling all agree:

- `useSessionCategories` - core sidebar filter (parent + worktree children)
- `computeSortedSessions` - Opt+Cmd+number jump-badge / nav visibility
- `SessionList` - `hasUnreadAgents` bell badge, worktree-child filter, and empty-state copy ("No unread, working, or errored agents")
- `SkinnySidebar` - collapsed sidebar visibility
- `useCycleSession` - Cmd+[/] cycling order under the filter

Keeping all of these in sync matters: otherwise an errored agent could appear in the sidebar but be skipped by keyboard navigation, or vice versa.

Scope note: only the `error` state is added. `connecting` and `waiting_input` are intentionally left out.

## Tests

- `useSessionCategories.test.ts`: an errored agent (no unread tabs) stays visible; an errored worktree child keeps its parent visible.
- `computeSortedSessions.test.ts` (new): errored agent appears in `visibleSessions` (jump-badge path); errored worktree child keeps parent visible; idle agent is excluded.

## Validation

- Affected suites green locally: `useSessionCategories`, `computeSortedSessions`, `useCycleSession`, `SessionList` (272 tests).
- `prettier --check` and `eslint` clean on all touched files.
- Rebased onto latest `rc` and adapted to the sidebarNavStore refactor (#1235): the `useSortedSessions` predicate now lives in `computeSortedSessions`.

_Note: the full local `validate:push` / repo-wide `tsc` is not green due to two pre-existing, unrelated dependency errors (`languageLoader.ts` nested `@codemirror` duplicate; `GitGraphView.tsx` missing `@gitgraph/*` modules) present on `rc` and untouched here; none of the changed files produce type errors. CI is the authoritative gate._

### DIFF
--- a/src/__tests__/renderer/hooks/computeSortedSessions.test.ts
+++ b/src/__tests__/renderer/hooks/computeSortedSessions.test.ts
@@ -1,0 +1,53 @@
+import { describe, it, expect } from 'vitest';
+import { computeSortedSessions } from '../../../renderer/hooks/session/computeSortedSessions';
+import { createMockSession } from '../../helpers';
+
+describe('computeSortedSessions - unread filter jump-badge visibility', () => {
+	it('keeps an errored agent in visibleSessions under the unread filter', () => {
+		const errored = createMockSession({ id: 'e1', name: 'Errored', state: 'error' });
+		const idle = createMockSession({ id: 'i1', name: 'Idle' });
+
+		const { visibleSessions } = computeSortedSessions({
+			sessions: [errored, idle],
+			groups: [],
+			bookmarksCollapsed: false,
+			showUnreadAgentsOnly: true,
+		});
+
+		const names = visibleSessions.map((s) => s.name);
+		expect(names).toContain('Errored');
+		expect(names).not.toContain('Idle');
+	});
+
+	it('keeps a parent visible when one of its worktree children is errored', () => {
+		const parent = createMockSession({ id: 'p1', name: 'Parent' });
+		const child = createMockSession({
+			id: 'c1',
+			name: 'Worktree Child',
+			parentSessionId: 'p1',
+			state: 'error',
+		});
+
+		const { visibleSessions } = computeSortedSessions({
+			sessions: [parent, child],
+			groups: [],
+			bookmarksCollapsed: false,
+			showUnreadAgentsOnly: true,
+		});
+
+		expect(visibleSessions.map((s) => s.name)).toContain('Parent');
+	});
+
+	it('excludes idle agents with no unread/errored children under the unread filter', () => {
+		const idle = createMockSession({ id: 'a1', name: 'Idle Alone' });
+
+		const { visibleSessions } = computeSortedSessions({
+			sessions: [idle],
+			groups: [],
+			bookmarksCollapsed: false,
+			showUnreadAgentsOnly: true,
+		});
+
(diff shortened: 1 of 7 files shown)


chr1syy:
@coderabbitai Good catch, and the observation is accurate: `computeSortedSessions` (the Opt+Cmd+number jump-badge / nav projection) keeps only unread, busy, and now errored agents visible under the unread filter - it does not consult `activeBatchSessionIds` (auto-running) or the outage signature (stuck).

However, this is a **pre-existing, systemic divergence** rather than something this PR introduces:

- Before this PR, `passesUnreadFilter` in `computeSortedSessions` already checked only `unread || busy || <children>`. This PR only adds `error` to that existing set; it does not remove or regress auto-running/stuck handling (they were never there).
- The same omission exists in **`SkinnySidebar`** (collapsed rail) and **`useCycleSession`** (Cmd+[/] cycling). Only the rendered list in `useSessionCategories` (and the `SessionList` worktree-child filter) consults `activeBatchSessionIds` / `stuckOutageSignature`. So the real gap is between the rendered list and *all* of the nav/jump/skinny/cycle predicates, not this one call site.
- Fixing it correctly means threading batch + outage state into each of those predicates (`useCycleSession` in particular reads at event time via `sidebarNavStore`/`getState`, so it needs App-level wiring), i.e. a shared "needs attention" predicate unifying every surface. Doing that partially - e.g. only in `computeSortedSessions` - would make the jump-badge path *newly* inconsistent with the skinny rail and cycle order.

This PR is intentionally scoped to the reported bug: **errored agents should appear under the "Filter unread agents" toggle**, applied consistently across every surface for the `error` state. The auto-running/stuck jump-badge gap is orthogonal and predates this change; it's best resolved by a dedicated follow-up that consolidates all of these predicates into one shared `sessionNeedsAttention(...)` helper. Keeping it out of this PR avoids widening scope into the `sidebarNavStore` refactor and avoids introducing a new asymmetry.

chr1syy:
@coderabbitai the follow-up is done in #1260. It introduces the shared `sessionNeedsAttention(...)` predicate (`src/renderer/utils/sessionAttention.ts`, with `sessionOrChildrenNeedAttention` and `outageIdsFromSignature`) and routes every Left Bar surface through it, so `{ unread, busy, error, auto-running (batch), stuck (outage) }` plus worktree children are now treated identically across the rendered list (`useSessionCategories`), the bell badge and worktree-child filter (`SessionList`), the Opt+Cmd+number jump-badge / nav projection (`computeSortedSessions` via `SidebarNavSync`), the collapsed rail (`SkinnySidebar`), and Cmd+[ / Cmd+] cycling (`useCycleSession`). That closes the pre-existing cross-surface divergence flagged here; no separate follow-up issue needed.

PR #1260, opened 2026-07-21
fix(sidebar): unify the "needs attention" predicate across Left Bar surfaces

chr1syy:
## Problem

CodeRabbit flagged (Major) on #1256 that the Left Bar "needs attention" predicate is duplicated across ~5 surfaces and only the rendered list treats auto-running (AUTO badge) and stuck (outage auto-retry) agents as needing attention. The jump badges, collapsed rail, and keyboard cycling do not, so an auto-running or stuck agent shows in the full sidebar but is skipped by Opt+Cmd+number jumps, hidden in the skinny rail, and skipped by Cmd+[ / Cmd+]. This is a pre-existing systemic divergence that #1256 intentionally left out of its focused scope.

## Fix

Introduce one shared predicate and route every surface through it.

`src/renderer/utils/sessionAttention.ts` (new):

- `sessionNeedsAttention(session, ctx)` - unread | busy | error | batch | stuck
- `sessionOrChildrenNeedAttention(session, children, ctx)` - parent stays visible when a worktree child needs attention
- `outageIdsFromSignature(sig)` - parse the comma-joined outage signature into a lookup set
- `AttentionContext = { batchSessionIds, stuckOutageIds }`

Surfaces routed through the helper:

- `useSessionCategories` (rendered list): refactored onto the helper, no behavior change (it already covered all five states).
- `computeSortedSessions` (Opt+Cmd+number jump badges / nav): now accepts `activeBatchSessionIds` / `stuckOutageSessionIds`; `SidebarNavSync` subscribes to `batchStore` and the outage signature and threads them into the projection.
- `SkinnySidebar` (collapsed rail): now accepts `stuckOutageSignature`; `SessionList` passes it at the call site.
- `useCycleSession` (Cmd+[ / Cmd+]): reads batch/outage at event time via `getState()` (matching how it already reads sessions / UI / nav), with optional test-override deps. Adds a non-reactive `getActiveOutageSessionIds()` to `retryStore`.
- `SessionList`: the `hasUnreadAgents` bell badge and the worktree-child filter now include batch.

The active-session carve-out stays per-surface (the semantics differ: parent-of-active rollup, active group chat, per-row), so there is no behavior change for the rendered list; the other surfaces gain batch + stuck visibility to match it.

## Tests

- New `sessionAttention` unit suite covering the helper directly: idle / unread / busy / error / batch / stuck, the parent-or-child rollup, and empty-signature parsing.
- `computeSortedSessions`: auto-running and stuck cases for both a parent and a worktree child.
- `SkinnySidebar`: auto-running and stuck agents stay visible under the unread filter, and idle agents are hidden (the collapsed rail filters rows individually, so these are per-row cases).
- `useCycleSession`: auto-running and stuck via both the deps-override and the event-time store-read paths (the latter exercises `getActiveOutageSessionIds()` and `selectActiveBatchSessionIds`), plus a worktree-child rollup case (an auto-running child keeps its parent in the cycle).

## Validation

- Touched suites green: `npx vitest run` on `sessionAttention`, `computeSortedSessions`, `SkinnySidebar`, `useCycleSession`, `useSessionCategories`, and `SessionList` (311 tests).
- `prettier --check` and `eslint` clean on touched files.
- `tsc -p tsconfig.lint.json`: no new type errors in any touched file. The two pre-existing unrelated errors present on `rc` (`languageLoader.ts` nested `@codemirror` duplicate; `GitGraphView.tsx` missing `@gitgraph/*`) remain untouched. CI is the authoritative gate.

## Note on the stack

This was originally stacked on #1256, which adds the `error` state to the same predicates. #1256 has since merged into `rc`, and this branch has been rebased onto it, so the PR is now a single commit against `rc` containing only this change.

Follow-up resolving the CodeRabbit thread on #1256.

## Summary by CodeRabbit

* **Enhancements**
  * Improved the “Unread agents only” filter to keep sessions active in batch runs or affected by stuck outages.
  * Parent sessions remain visible when a related worktree child requires attention.
  * Sidebar session lists, categories, and session cycling now apply consistent attention rules.

* **Bug Fixes**
  * Sidebar visibility now updates when batch activity or outage status changes.

* **Tests**
  * Added coverage for unread filtering, batch sessions, stuck outages, worktree relationships, and session cycling.

### DIFF
--- a/src/__tests__/helpers/seedSidebarNav.ts
+++ b/src/__tests__/helpers/seedSidebarNav.ts
@@ -32,6 +32,8 @@ export interface SeedSidebarNavFromSessions {
 	activeSessionId?: string | null;
 	bookmarksCollapsed?: boolean;
 	showUnreadAgentsOnly?: boolean;
+	activeBatchSessionIds?: string[];
+	stuckOutageSessionIds?: string[];
 	starredItems?: StarredItem[];
 	activateStarredItem?: ActivateStarred;
 }
@@ -54,6 +56,8 @@ export function seedSidebarNav(input: SeedSidebarNavInput): void {
 			bookmarksCollapsed: input.bookmarksCollapsed ?? ui.bookmarksCollapsed,
 			showUnreadAgentsOnly: input.showUnreadAgentsOnly ?? ui.showUnreadAgentsOnly,
 			activeSessionId: input.activeSessionId,
+			activeBatchSessionIds: input.activeBatchSessionIds,
+			stuckOutageSessionIds: input.stuckOutageSessionIds,
 		});
 		useSidebarNavStore.setState({
 			...projection,
--- a/src/__tests__/renderer/components/SessionList/SkinnySidebar.test.tsx
+++ b/src/__tests__/renderer/components/SessionList/SkinnySidebar.test.tsx
@@ -4,6 +4,7 @@ import { SkinnySidebar } from '../../../../renderer/components/SessionList/Skinn
 import type { Session, Group, Theme } from '../../../../renderer/types';
 
 import { mockTheme } from '../../../helpers/mockTheme';
+import { createMockSession } from '../../../helpers';
 
 let idCounter = 0;
 function makeSession(overrides: Partial<Session> = {}): Session {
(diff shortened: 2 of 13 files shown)


chr1syy:
@coderabbitai addressed in d4af16d23: extracted a shared `collectActiveOutageSessionIds(outages)` helper in `retryStore.ts`, and both `getActiveOutageSessionIds()` (non-reactive, event-time) and `useActiveOutageSessionSignature()` (reactive) now call it, so the active-outage scan lives in one place. Thanks for the catch - it fits the PR's anti-divergence goal.

PR #1253, opened 2026-07-20
Tool display: parse tool_result, nest Task subagents, answer AskUserQuestion, per-tab Show Tools

chr1syy:
## Summary

Makes agent tool activity a first-class, visible thing across providers. Four phases, each its own commit:

1. **Parse claude-code `tool_result`** so tool badges reach a terminal state. Previously claude-code tools stayed stuck "running" because the parser never consumed the follow-up `tool_result` block. Now an assistant `tool_use` followed by a user `tool_result` completes the badge (with `is_error: true` -> `failed`), unknown ids fall back to a name, and array-content is flattened.
2. **Nest `Task` subagent activity** via `parent_tool_use_id`, threaded through the main-process pipeline (`ParsedEvent`/`ToolExecution` -> `StdoutHandler` -> IPC/web broadcast) into renderer tool logs, and rendered as collapsible children under their parent Task in the transcript (`groupSubagentToolLogs` + a reusable `ToolBadge`).
3. **Answer `AskUserQuestion` through the permission relay.** The relay recognizes `AskUserQuestion` in standard permission mode, parses its `questions` into a typed shape, and a new `QuestionPrompt` UI renders the options (radio / checkbox / free-text) and returns the selection as the tool answer via the proven `deny` + `message` shape. The system prompt is relaxed to permit `AskUserQuestion` only in interactive standard-permission sessions.
4. **Decouple tool badges from Show Thinking.** A per-tab `showTools` toggle (Wrench icon next to Thinking) now governs tool visibility independently of reasoning display. It falls back to the tab's `showThinking` state when absent, so upgrades see no behavior change; new tabs default to on. Adds an end-to-end pipeline section to `AGENT-INFRA.md` with a per-provider support matrix. Factory Droid was investigated (its `-o stream-json` carries no structured tool events; the binary isn't available to capture samples) and left unchanged with findings recorded.

## Provider tool-badge support after this PR

| Provider | Live tool badges |
| --- | --- |
| claude-code | Yes (incl. `tool_result` completion + Task nesting) |
| codex, opencode, copilot-cli, pi, omp | Yes |
| grok | No (tool activity only in on-disk session files) |
| factory-droid | No (`stream-json` folds tool activity into assistant text) |

## Testing

[...]

PR #1263, opened 2026-07-21
fix(chat): keep AI transcript stuck to the bottom during tool streaming

chr1syy:
## Problem
In claude-code chats the AI transcript stops following the bottom when new messages and tool badges stream in. Follow-up to #1253 (which made claude-code emit tool badges).

## Root cause
`useTerminalOutputScroll`'s count-effect re-measured the scroll container the instant the log **count** grew. That runs before the `MutationObserver`'s rAF scroll, so a newly appended entry taller than the 50px slack (a `running`/`completed` tool badge) reads as "not at bottom" and calls `setIsAtBottom(false)`, pausing auto-follow. On `rc` before #1253 claude-code emitted no such badges, so this latent bug stayed hidden.

## Fix
Trust the tracked follow state (`isAtBottomRef.current`) instead of an instantaneous pre-scroll measurement. While following, mark the new content read and let the observer pin to the new bottom; only raise the "new messages" pill when genuinely scrolled up.

## Tests
Adds `useTerminalOutputScroll.test.ts`: (1) stays following when a tall entry arrives at bottom, (2) raises the pill when scrolled up. Typecheck + touched suites green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal output scrolling behavior when new log entries arrive.
  * Auto-scroll stays enabled for users already at the bottom, even with newly added tall entries.
  * When scrolling up, the “new messages” indicator and unread count now update reliably.
* **Tests**
  * Added a regression test covering mid-stream updates to ensure scrolling follow/paused states behave consistently.

### DIFF
--- a/src/__tests__/renderer/components/TerminalOutput/useTerminalOutputScroll.test.ts
+++ b/src/__tests__/renderer/components/TerminalOutput/useTerminalOutputScroll.test.ts
@@ -1,0 +1,84 @@
+import { describe, it, expect } from 'vitest';
+import { renderHook, act } from '@testing-library/react';
+import { useTerminalOutputScroll } from '../../../../renderer/components/TerminalOutput/hooks/useTerminalOutputScroll';
+
+/**
+ * Regression coverage for the stick-to-bottom follow behaviour when the log
+ * COUNT grows (a new tool badge or message appears) mid-stream.
+ *
+ * The container is stubbed to report "not at bottom" (an instantaneous,
+ * pre-scroll measurement). The count-effect must NOT use that measurement to
+ * pause a user who is already following: doing so is what made a tall tool
+ * badge kill auto-follow (the MutationObserver's rAF jump had not run yet). It
+ * must trust the tracked follow state instead.
+ */
+function makeContainer(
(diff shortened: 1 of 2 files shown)


pedramamini:
Thanks for the contribution, @chr1syy - much appreciated! 🙏

I reviewed the PR and the automated feedback:

- **CodeRabbit**: no actionable comments.
- **Greptile**: 5/5 confidence, "safe to merge".

The root-cause writeup is spot on: trusting the tracked follow state (`isAtBottomRef.current`) instead of the instantaneous pre-scroll measurement is the right call, since the MutationObserver's rAF jump hasn't run yet when the count-effect fires. The regression tests cover both the stay-following and scrolled-up-pill cases nicely. No merge conflicts either.

One heads-up (not a blocker for this PR): the `Test (…, shard 2/2)` legs are red, but that's a **pre-existing failure unrelated to your change** - `SessionList > Pianola Pinned Agent > renders Pianola above the unread-filter empty state`, which also fails on `rc` directly. Your change only touches `useTerminalOutputScroll`, so this isn't on you.

Approving. ✅

PR #1264, opened 2026-07-21
fix(tools): stop tool-badge flicker by render-gating the per-tab Tools toggle

chr1syy:
## Problem
Follow-up to #1253. Disabling and re-enabling the per-tab **Tools** toggle during an active run makes the chat flicker on tool calls.

## Root cause
The toggle physically **deleted** `source:'tool'` logs on disable (`useAITabHandlers`), and `useAgentToolExecutionListener` **dropped** tool events while off. So an off/on cycle mid-run churned the transcript (badges deleted, then re-appended out of order as events streamed) and broke `running`->`completed` correlation.

## Fix
Move to **always record, hide at render**:
- `useAgentToolExecutionListener` always records tool events (correlation survives a mid-run toggle).
- `TerminalOutput` filters `source:'tool'` at render time via the tab's effective setting (`toolLogsRecorded(showTools, showThinking)`).
- The Tools toggle and the three thinking-cycle surfaces (toolbar, quick-actions, keyboard) flip flags without mutating logs.
- Exit/interrupt/batched lifecycle cleanup is intentionally unchanged.

Also resolves the greptile/CodeRabbit P1s on #1253 (lost tool events when off; thinking-toggle wiping tool logs).

## Tests
Updated `useAgentToolExecutionListener.test.tsx` to the always-record contract; added a `showTools:false` render test in `TerminalOutput.test.tsx` (tools hidden, not dropped). Typecheck + touched suites green (156 tests).

## Summary by CodeRabbit

- **Bug Fixes**
  - Tool execution events are now consistently recorded even when tool and/or thinking visibility is hidden.
  - Collapsed terminal output now omits tool-origin entries when the active tab’s tool visibility is off.
  - Toggling thinking to **off** clears only thinking logs (tool logs are retained).
  - Toggling tools updates display visibility without removing previously recorded tool events.
- **Tests**
  - Updated/expanded coverage for showTools/showThinking rendering, hidden tool boundaries, and log-recording expectations across terminal output and agent/tool listeners.

### DIFF
--- a/src/__tests__/renderer/components/TerminalOutput.test.tsx
+++ b/src/__tests__/renderer/components/TerminalOutput.test.tsx
@@ -134,6 +134,7 @@ const createDefaultSession = (overrides: Partial<Session> = {}): Session => ({
 			agentSessionId: 'claude-123',
 			logs: [],
 			isUnread: false,
+			showTools: true,
 		},
 	],
 	activeTabId: 'tab-1',
@@ -214,7 +215,9 @@ describe('TerminalOutput', () => {
 			];
 
 			const session = createDefaultSession({
-				tabs: [{ id: 'tab-1', agentSessionId: 'claude-123', logs, isUnread: false }],
+				tabs: [
+					{ id: 'tab-1', agentSessionId: 'claude-123', logs, isUnread: false, showTools: true },
+				],
 				activeTabId: 'tab-1',
 			});
 
@@ -228,7 +231,9 @@ describe('TerminalOutput', () => {
 			const logs: LogEntry[] = [createLogEntry({ text: 'User input here', source: 'user' })];
 
 			const session = createDefaultSession({
-				tabs: [{ id: 'tab-1', agentSessionId: 'claude-123', logs, isUnread: false }],
+				tabs: [
+					{ id: 'tab-1', agentSessionId: 'claude-123', logs, isUnread: false, showTools: true },
+				],
 				activeTabId: 'tab-1',
 			});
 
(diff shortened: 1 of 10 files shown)


pedramamini:
Thanks for the follow-up, @chr1syy - the always-record / hide-at-render approach is the right call, and cleaning up the #1253 P1s in the same pass is much appreciated. 🙏

One change before this is good to merge, flagged independently by both Greptile (P2) and CodeRabbit (Major) and confirmed against the code:

**Preserve tool boundaries before hiding tool entries** (`src/renderer/components/TerminalOutput/TerminalOutput.tsx`, ~L126-140).

`collapseAiResponseLogs` treats `source: 'tool'` as a flush boundary. Because `visibleLogs` strips tool entries *before* collapse runs, a turn shaped like [assistant text] -> [tool call] -> [assistant text] loses its boundary when Tools is off, and the two assistant segments fold into a single bubble. That diverges from the Tools-on layout, so the transcript grouping changes depending on the toggle.

Fix is just a reorder - collapse first, then filter:

```ts
const collapsedLogs = useMemo(() => collapseAiResponseLogs(activeLogs), [activeLogs]);
const visibleLogs = useMemo(
	() => (toolsVisible ? collapsedLogs : collapsedLogs.filter((l) => l.source !== 'tool')),
	[collapsedLogs, toolsVisible]
);
```

Downstream (`groupSubagentToolLogs`, rendering) should then consume `visibleLogs`. Please also add a regression test with assistant text -> tool entry -> assistant text asserting the two responses stay separate when `showTools` is false.

No merge conflicts, so nothing to rebase - just this one fix and a re-run of the touched suites. Thanks again!

PR #1265, opened 2026-07-21
feat(plugins): add a hide/collapse control for docked plugin panels

chr1syy:
## Problem
Follow-up to #1254. A plugin panel with `placement:'right'` (e.g. Agent Flow) docks as a full-height column via `PluginPanelSlot` with no chrome, so the only way to stop showing it was disabling the whole plugin.

## Change
- Host-drawn header per docked panel with a hide control.
- Hidden panels collapse to a slim vertical **reopen rail** (one click to restore); when other panels stay docked, the hidden one becomes a "Show <title>" chip.
- Collapsed panel ids persist in `uiStore.hiddenPluginPanels` (settings write-through, mirrors `hiddenQuotaAccounts`) and hydrate on startup.
- The shared `PluginPanelFrame` and its non-suppressible provenance line are untouched, so the trusted-chrome guarantees (`PROTECTED_UI_SURFACES`) are preserved.

## Tests
`PluginPanelSlot.test.tsx`: hide -> collapse-rail -> restore; and hide-one-while-another-stays-docked (reopen chip). Store reset in `beforeEach`. Typecheck + suite green.

## Summary by CodeRabbit

* **New Features**
  * Added hide/show controls for docked plugin panels, including compact reopen chips.
  * If all panels in a slot are hidden, the slot collapses into a slim reopen rail.
  * Hidden panel visibility now persists across restarts.
* **Bug Fixes**
  * Improved multi-panel hide/reopen behavior so only the intended panel collapses/restores.
  * Added validation for persisted hidden-panel settings to avoid malformed data breaking panel visibility.

### DIFF
--- a/src/renderer/components/plugins/PluginPanelSlot.tsx
+++ b/src/renderer/components/plugins/PluginPanelSlot.tsx
@@ -12,14 +12,21 @@
  * Renders nothing when the `plugins` Encore flag is off (then
  * `usePluginContributions` returns empty buckets) or when no panel targets this
  * slot, so the slot stays invisible until a plugin docks here.
+ *
+ * Each docked panel carries a host-drawn header with a hide control; hidden
+ * panels collapse to a slim reopen rail. The set of collapsed panel ids lives
+ * in `uiStore.hiddenPluginPanels` (persisted). The frame's non-suppressible
+ * provenance line stays in `PluginPanelFrame` and is never touched here.
  */
 
 import { useMemo } from 'react';
+import { PanelRightClose, PanelRightOpen } from 'lucide-react';
 import type { Theme } from '../../types';
 import type { PanelContribution, PanelPlacement } from '../../../shared/plugins/contributions';
 import { usePluginContributions } from '../../hooks/usePluginContributions';
 import { mergePluginContributions } from '../../utils/pluginContributionMerge';
 import { MODAL_PRIORITIES } from '../../constants/modalPriorities';
+import { useUIStore } from '../../stores/uiStore';
 import { PluginPanelFrame } from './PluginPanelFrame';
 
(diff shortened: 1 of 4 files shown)


pedramamini:
Thanks for the follow-up, @chr1syy - nice, self-contained addition, and I appreciate that you kept `PluginPanelFrame` and the non-suppressible provenance line untouched so the `PROTECTED_UI_SURFACES` guarantees stay intact.

One change before this is good to merge, flagged independently by both Greptile (P1) and CodeRabbit (Major):

**Validate `hiddenPluginPanels` before hydrating it** (`src/renderer/stores/settingsStore.ts`):

```ts
if (allSettings['hiddenPluginPanels'] !== undefined)
    useUIStore.setState({
        hiddenPluginPanels: allSettings['hiddenPluginPanels'] as string[],
    });
```

The `!== undefined` guard plus the bare `as string[]` cast lets a malformed persisted value (`null`, an object, or an array with non-string entries) reach `uiStore`. `PluginPanelSlot` then calls `hiddenPluginPanels.includes(...)` during render, which throws and takes down every docked panel slot instead of falling back to the visible default. Please accept only arrays and retain only string ids, e.g.:

```ts
if (Array.isArray(allSettings['hiddenPluginPanels']))
    useUIStore.setState({
        hiddenPluginPanels: allSettings['hiddenPluginPanels'].filter(
            (panelId): panelId is string => typeof panelId === 'string'
        ),
    });
```

Optional (CodeRabbit nitpick, take it or leave it): the multi-panel test only asserts the "Show <title>" chip appears - you could extend one case to `placement="right"`, click the chip, and assert both frames re-render and the id is cleared from `hiddenPluginPanels`.

Once the hydration validation is in, I'm happy to approve. Thanks again!

PR #1266, opened 2026-07-21
feat(plugins): trusted bundled first-party plugins + Agent Flow seeding

chr1syy:
## What
Follow-up to #1254. Establishes how a **sandboxed folder plugin shipped inside the app** (Agent Flow) becomes `trusted` - and therefore runnable - without a manual install. Today no sandboxed plugin can be trusted (`pluginTrustedKeys` defaults empty; only native first-party features are trusted-by-construction), so this is the first bundled trusted plugin.

## How
- **`src/shared/plugins/publisher-keys.ts`** - a baked `MAESTRO_PUBLISHER_KEYS` anchor, merged with the user's `pluginTrustedKeys` via `resolveTrustedKeys()`. This is the only built-in trust anchor.
- **`src/main/plugins/bundled-plugins.ts`** - `seedBundledPlugins()` copies bundled plugins (`extraResources -> <resources>/plugins/<id>`) into `pluginsDir()` at startup, **only when the signature verifies `trusted`**. Symlink-safe, version-aware, idempotent; also replaces a manual unsigned install of the same id with the trusted bundled copy.
- **`src/main/index.ts`** - merges the anchor into `trustedKeys`, seeds before `pluginManager.refresh()`.
- **`package.json`** - bundles `examples/plugins/agent-flow` for mac/win/linux.

## Security / key handling
The matching **private** key is a maintainer/CI secret and is **never committed**. `MAESTRO_PUBLISHER_KEYS` ships **empty** here, and the seeder is **trust-gated**, so with an empty anchor **nothing is auto-installed** - no orphaned, auto-installed-but-non-runnable plugin. Agent Flow activates once a release signs it with the publisher key and adds that public key to the anchor.

## Tests
`publisher-keys.test.ts` (merge/dedupe) and `bundled-plugins.test.ts` (trust gate, replace-untrusted-same-version, skip-trusted-same, refresh-on-newer). `tsc` main + lint clean.

## Summary by CodeRabbit

* **New Features**
  * Bundled first-party plugins are now included in desktop app builds for macOS, Windows, and Linux.
  * On startup, trusted bundled plugins are automatically seeded into the user plugins directory when missing or outdated.
  * Configured plugin trust keys are merged with built-in publisher trust keys (whitespace/duplicates/empty values ignored).
* **Security**
  * Untrusted bundled plugins are not installed; trusted ones are installed or refreshed only when the bundled version is newer.
  * Updates avoid symlinks and use safe replace/rollback to reduce risk of failed installs.
* **Tests**
  * Added coverage for bundled plugin seeding behavior and trusted-key resolution edge cases.

### DIFF
--- a/package.json
+++ b/package.json
@@ -141,6 +141,10 @@
 				{
 					"from": "src/prompts/bmad",
 					"to": "prompts/bmad"
+				},
+				{
+					"from": "examples/plugins/agent-flow",
+					"to": "plugins/agent-flow"
 				}
 			]
 		},
@@ -192,6 +196,10 @@
 				{
 					"from": "src/prompts/bmad",
 					"to": "prompts/bmad"
+				},
+				{
+					"from": "examples/plugins/agent-flow",
+					"to": "plugins/agent-flow"
 				}
 			]
 		},
@@ -235,6 +243,10 @@
 				{
 					"from": "src/prompts/bmad",
 					"to": "prompts/bmad"
+				},
+				{
+					"from": "examples/plugins/agent-flow",
+					"to": "plugins/agent-flow"
 				}
 			]
 		},
--- a/src/__tests__/main/plugins/bundled-plugins.test.ts
+++ b/src/__tests__/main/plugins/bundled-plugins.test.ts
@@ -21,6 +21,13 @@ vi.mock('../../../main/plugins/plugin-store-main', () => ({
 	isSafePluginFolderName: (name: string) => /^[a-z][a-z0-9._-]*$/i.test(name),
 }));
 
+// Real fs everywhere, but a controllable cpSync so a copy failure can be forced
+// to exercise the atomic-replace rollback.
+vi.mock('fs', async (importOriginal) => {
+	const actual = await importOriginal<typeof fs>();
+	return { ...actual, cpSync: vi.fn(actual.cpSync) };
+});
+
 import { seedBundledPlugins } from '../../../main/plugins/bundled-plugins';
 
 let tmp = '';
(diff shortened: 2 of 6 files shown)


pedramamini:
Thanks for the follow-up @chr1syy, this is a clean, well-documented approach to bundled first-party trust and the trust-gated-on-empty-anchor design is a nice safety property. Before we merge, I'd like to see a few points from the AI reviewers addressed - I agree they're worth doing:

1. **Bundle identity vs. directory name** (Greptile, `bundled-plugins.ts:88`). Seeding targets `pluginsDir()/<manifest.id>` without requiring the manifest `id` to match the bundled directory name. A packaging mistake (dir `foo` shipping manifest id `bar`) could overwrite an unrelated installed `bar` while leaving `foo` behind. Since these are first-party bundles we control, adding an `entry.name === id` guard is a cheap invariant that catches the mistake at seed time.

2. **Destructive replace with no rollback** (Greptile, `bundled-plugins.ts:119`). `fs.rmSync(dest)` runs before `fs.cpSync(src, dest)`, so if the copy fails partway (disk full, permissions, unreadable entry) the previous install is gone and the dir is left partial. It's self-healing on the next boot since the bundled source is still present, but consider copying to a temp dir and renaming into place so a failed copy never destroys the existing install.

3. **`readManifest` swallows unexpected errors** (CodeRabbit, `bundled-plugins.ts:52-58`). The catch converts every failure to `null`, including unexpected I/O and validation errors. Our error-handling policy (CLAUDE.md) is to handle absent/malformed manifests explicitly but let unexpected exceptions bubble to `onError`/Sentry rather than silently skip. Worth narrowing that catch.

Separately, CI shard 2/2 is red on both legs, but the two failures (`SessionList` Pianola and `TerminalOutput` auto-scroll) are unrelated to this PR and look pre-existing/flaky on `rc`. No action needed from you there beyond a re-run to confirm green; just flagging so it isn't a surprise.

No merge conflicts, so no rebase needed. Once the above are addressed I'm happy to approve.

PR #1268, opened 2026-07-21
Honor JSON mode for movement numeric errors

stevenmgordon:
## Summary

- Route movement numeric parse failures through the existing JSON-aware error envelope.
- Apply the shared behavior to add, begin, update, move, and progress numeric flags.
- Add regressions for non-numeric Concerto position and phase subdivision values in JSON mode.

This follows up on the unresolved CodeRabbit finding from #1259.

## Validation

- `src/__tests__/cli/commands/movement.test.ts`: 19 passed
- `tsc -p tsconfig.cli.json --noEmit`
- Targeted ESLint passed
- Targeted Prettier check passed
- `git diff --check` passed

## Note

The repository-wide pre-push formatting check currently reports 74 unrelated baseline files, including untracked local mockups, so the validated push used `--no-verify`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JSON error responses for invalid numeric movement command options.
  * Invalid values now return consistent structured errors without attempting to connect to the movement service.
  * Added validation coverage for non-numeric position, step count, and step values.

### DIFF
--- a/src/__tests__/cli/commands/movement.test.ts
+++ b/src/__tests__/cli/commands/movement.test.ts
@@ -209,6 +209,17 @@ describe('movement begin command', () => {
 		);
 		expect(withMaestroClient).not.toHaveBeenCalled();
 	});
+
+	it('prints a JSON error for a non-numeric position', async () => {
+		await expect(
+			movementBegin('startup', { title: 'Loopline startup', x: 'left', json: true })
+		).rejects.toThrow('__exit__');
+
+		expect(console.log).toHaveBeenCalledWith(
+			JSON.stringify({ success: false, error: '--x must be a number' })
+		);
+		expect(withMaestroClient).not.toHaveBeenCalled();
+	});
 });
 
 describe('movement progress command', () => {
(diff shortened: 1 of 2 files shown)


pedramamini:
@stevenmgordon Thanks for the contribution, and thanks for circling back to close out the unresolved CodeRabbit finding from #1259 rather than letting it drift.

Reviewed the diff alongside the Greptile and CodeRabbit passes. This looks good to me:

- `failMovementCommand` is typed `never`, so threading it through `parseNum` keeps the control flow sound and the `return n` line stays unreachable on the failure path.
- Non-JSON behavior is byte-identical to before (`Error: --x must be a number` on stderr, exit 1), so this is a pure additive fix for JSON mode rather than a behavior change for existing callers.
- All five call sites (add, begin, update, move, progress) consistently pass their own `options.json`, and the tests assert that `withMaestroClient` is never reached, which is the part that actually matters here.
- Test coverage hits both a non-numeric position and a non-numeric `--steps`, and the `--steps` case slots cleanly into the existing table-driven phase subdivision cases.

CI is green across both matrix legs (ubuntu and windows), lint-and-format passed, and the branch merges cleanly into `rc` with no conflicts.

One entirely optional follow-up, not a blocker for this PR: `movementMove`'s "requires --x and --y" error at the missing-flag check is still a raw `console.error` + `process.exit(1)`, so it will bypass JSON mode the same way the numeric errors used to. There are a handful of other validation paths in the file in the same shape (`--type`, `--html-file`/`--body-file`, `op`). Converting those to `failMovementCommand` would finish the job, but it is fairly clearly outside the scope you drew for this change and is fine as a separate PR.

Approving. Nice, tight, well-scoped fix.

pedramamini:
Re-validated against current `rc` before merging: this branch was 375 commits behind, so its green CI reflected a July base rather than today's tree. Merged `origin/rc` into it locally - no conflicts, `tsc -p tsconfig.cli.json` at exact parity with the rc baseline (the one pre-existing `ignoreDeprecations` config error, unchanged), and `movement.test.ts` 19/19 green.

Merging. Thanks again.

PR #1280, opened 2026-07-22
fix(spawn): decide prompt delivery in main, never in the renderer

chr1syy:
## The bug

Every message sent from **web-desktop to an Oh My Pi agent** fails with:

> Oh My Pi exited without producing a response. The agent process ended early (for example right after startup) before sending any output. Please send your message again.

and once it happens the tab is permanently broken - every later turn dies with `Oh My Pi exited with code 1: Error: Session "<id>" not found.`

## Root cause: the client tells the host what OS it is

1. `src/web-desktop/bootstrap.ts` fabricates `process.platform` from the **browser's** user agent, so `window.maestro.platform` is `win32` whenever the web UI is opened from a Windows machine - while the host that actually spawns agents is Linux or macOS.
2. The renderer's `getStdinFlags()` read that value and set `sendPromptViaStdinRaw: true` - the Windows-only workaround for the ~32K `CreateProcess` command-line limit.
3. `ChildProcessSpawner` then keeps the prompt **out of argv** and writes it to the child's stdin.
4. **omp only accepts the prompt as a positional argument.** It ignores stdin, so it starts with no prompt at all: it prints its `{"type":"session"}` line and exits 0 without ever calling the model.
5. Maestro records that session id on the tab, but omp never persisted the session - so every subsequent `--resume <id>` fails. That is the "every message fails" part.

This is also exactly the asymmetry that made the earlier reports confusing ("why does it work on desktop but not web-desktop?"): every main-side decision used the real host platform, only the renderer-side one lied.

### Evidence

Against the real CLI - a prompt on stdin is silently discarded:

```console
$ echo "Reply with exactly the word: pong" | omp -p --mode json --cwd /tmp
{"type":"session","version":3,"id":"019f8b2f-...","timestamp":"...","cwd":"/tmp"}
$ echo $?
0
```

Driving a `process:spawn` over the web-desktop WS bridge with `sendPromptViaStdinRaw: true` reproduces the failing turn deterministically:

```
[event] process:session-id ["...","019f8b2f-e1bb-7000-b70f-0fe920a71f8c"]
[event] agent:error [..."Oh My Pi exited without producing a response."...]
[event] process:exit [...,0,null]
```

And loading the live web-desktop bundle in a browser with a Windows user agent reproduces it end to end through the real UI (`platform seen by renderer: win32`, session id then exit 0, no answer) - while the tab-naming turn in the same run answered normally, because tab naming computes its stdin flag in the main process.

The UUIDv7 timestamps of the poisoned `agentSessionId`s on the affected tabs decode to the exact times of the reported failures, confirming the same promptless spawn produced them.

## The fix

Move the decision to where the facts live.

- **`handleProcessSpawn` is now authoritative.** It computes `sendPromptViaStdin` / `sendPromptViaStdinRaw` from the host `isWindows()`, the SSH flag, the agent's capabilities and whether images are attached, and overrides whatever the caller sent. This also fixes the mirror case: a macOS browser driving a Windows host previously lost the argv-length workaround.
- **New `supportsPromptViaStdin` capability** records whether an agent's CLI actually reads the prompt from stdin. `omp` is `false` - which also fixes omp on a *genuine* Windows host, plus its tab naming and group-chat spawns, where the same silent-no-prompt failure applied. Agents that were not re-verified keep their existing behavior (`true`), with the comment saying so.
- **The renderer no longer computes stdin transport at all.** `getStdinFlags()` and the two spawn-config fields are removed, so a client can no longer state an opinion about the host it is driving. The renderer-side tests that locked in that behavior are removed with it; equivalent coverage now sits on the main side.

## Tests

- `process.test.ts` - new `prompt delivery (argv vs stdin)` block: a caller asking for stdin delivery on a non-Windows host is ignored (the web-desktop regression); the prompt stays in argv on Windows for positional-only agents; raw stdin is enabled on a Windows host even when the caller did not ask; stream-json stdin with images; SSH leaves both off.
- `tabNaming.test.ts` and `group-chat-config.test.ts` - regression tests for an agent that never reads stdin.
- `capabilities.test.ts` - the new capability is required on every agent.

`npm run lint` (all three configs) is clean; ESLint clean on the changed files; the touched main and renderer suites pass (2340 tests).

## Follow-up, deliberately not in this PR

`src/web-desktop/bootstrap.ts` still derives `process.platform` from the user agent, which is right for keyboard-modifier hints and wrong for anything host-shaped (path separators, `isWindows()` in shared path helpers). Fixing that properly means splitting "client platform" from "host platform" across the renderer; this PR removes the one place where it was actively corrupting agent spawns.

## Summary by CodeRabbit

* **New Features**
  * Added a new agent capability flag indicating whether the agent can accept prompts via standard input.
* **Bug Fixes**
  * Improved prompt delivery on Windows by automatically choosing stdin vs argv based on agent capability and whether image data is included; agents that don’t support stdin keep receiving prompts via argv.
* **Documentation**
  * Updated agent capability and platform guides to describe the new capability and Windows stdin behavior.
* **Tests**
  * Expanded coverage to validate stdin/argv routing across spawning flows and agent capability variations.


src/main/ipc/handlers/process/handle-spawn.ts
```diff
@@ -189,6 +189,32 @@ export async function handleProcessSpawn(
...
+	// not decide it: a web-desktop client is a browser that can sit on a
+	// completely different OS than the host (`window.maestro.platform` there is
+	// derived from the browser's user agent), so a Windows browser driving a Linux
+	// host used to request stdin delivery. Agents that only accept a positional
+	// prompt then started with no prompt at all - omp emitted its session line and
+	// exited 0, which left the tab holding an agent session id omp never persisted,
+	// so every later turn failed to resume it too.
+	// ========================================================================
+	// An agent that has not declared the capability keeps its prompt in argv: an
+	// over-long command line fails loudly at spawn, while stdin delivery to a CLI
```
chr1syy, line 207: Deliberate, and I made it explicit in the code (`ee651d5`).

An agent with no declared `supportsPromptViaStdin` now keeps its prompt in argv. The two failure modes are not symmetric:

- **argv on a too-long prompt** fails *loudly* at spawn (the process never starts, the user gets an error).
- **stdin to a CLI that ignores stdin** fails *silently*: the agent runs with no prompt, prints its session banner, exits 0, and Maestro stores a session id the agent never persisted, so every later turn fails to resume. That is exactly the bug this PR fixes.

Conservative-by-default also matches `DEFAULT_CAPABILITIES`, where every flag is `false` for unknown agents, and `getAgentCapabilities()` already hands unknown ids the full default object - so this is the same treatment they get for resume, JSON output, session storage and the rest. Plugin-contributed runtime agents are not spawnable through this path yet (`agent-registry.ts`: "SPAWNING a runtime agent ... is NOT enabled by registration alone"), so nothing reachable regresses today; when that wiring lands, the contribution should declare the flag like any other capability.

pedramamini:
@chr1syy thanks for this - it's a really thorough fix, and the writeup tracing the promptless-spawn from the user-agent-derived `platform` all the way to the poisoned `agentSessionId` (UUIDv7 timestamps and all) made it easy to follow. Moving the argv-vs-stdin decision into `handleProcessSpawn` where the real host facts live is the right call, and the new `supportsPromptViaStdin` capability plus the regression tests on the main side are solid.

I went through the Greptile and CodeRabbit feedback and reviewed the diff. My take on their points:

- **Greptile's "unknown agents lose the Windows stdin fallback"**: real behavioral change, but I think defaulting an unmapped agent to the positional-arg path (`?? false`) is the correct, conservative choice here. It's exactly consistent with the thesis of this PR - guessing stdin support is what silently broke omp - and it only affects the narrow case of a custom agent, on Windows, with a near-32K prompt, that happens to read stdin. Comfortable keeping it as-is.
- **CodeRabbit's Hermes `supportsPromptViaStdin: true`**: this preserves the pre-existing Windows behavior (every agent already went via raw stdin on Windows before this change), so it's status-quo rather than a regression, and the `// Unverified - keeps the pre-existing Windows behavior` comments make the intent clear. Fine to defer verification of individual agents to a follow-up.
- **CodeRabbit's doc count `23 -> 24`**: minor, and actually the "23 boolean flags" figure in `CLAUDE-AGENTS.md` was already stale before this PR (the interface has more than that), so I wouldn't bother bumping it to a number that's still wrong. Not blocking.

Approving. Nice work.

PR #836, opened 2026-04-13
feat: configurable utility agent for auxiliary tasks

pedramamini:
## Summary

Closes #835

- Adds `utilityAgentId` and `utilityModelId` settings so users can route lightweight auxiliary tasks (tab naming, context grooming) to a cheaper/faster agent instead of burning primary session tokens on housekeeping
- When unset, behavior is unchanged (uses the session's own agent)
- UI dropdown in Settings > General (right below "Automatic Tab Naming") populated from detected available agents, with optional model override text field

## Files changed
- `src/shared/settingsMetadata.ts` — new setting definitions
- `src/main/stores/defaults.ts` — defaults (both `null`)
- `src/renderer/stores/settingsStore.ts` — Zustand state, setters, and load logic
- `src/renderer/hooks/settings/useSettings.ts` — hook interface
- `src/main/ipc/handlers/tabNaming.ts` — resolves utility agent before spawning
- `src/main/ipc/handlers/context.ts` — resolves utility agent for groomContext; added `settingsStore` dependency
- `src/main/utils/context-groomer.ts` — accepts optional `modelId` and passes to `buildAgentArgs`
- `src/main/index.ts`, `src/main/ipc/handlers/index.ts` — wire `settingsStore` to context handlers
- `src/renderer/components/Settings/tabs/GeneralTab.tsx` — Utility Agent UI section
- `src/__tests__/main/ipc/handlers/tabNaming.test.ts` — fix mock to return proper defaults

## Test plan
- [x] All 22,451 existing tests pass (537 suites)
- [x] TypeScript type-check clean across all configs
- [x] ESLint clean (0 errors)
- [ ] Manual: open Settings, verify Utility Agent dropdown shows detected agents
- [ ] Manual: select a utility agent, send a message to a new tab, verify tab naming uses the selected agent
- [ ] Manual: set model override, verify it's passed through to the agent
- [ ] Manual: clear both fields, verify default behavior (session agent) is restored

## Release Notes

* **New Features**
  * Added Utility Agent configuration in settings. Users can now designate a specific agent for auxiliary tasks with an optional model override. When configured, auxiliary operations use the selected utility agent instead of the default session agent.

* **Documentation**
  * Normalized release notes formatting for improved readability.

PR #1282, opened 2026-07-22
feat(agents): configurable utility agent for auxiliary tasks

### LINKED ISSUE #835: Feature: Configurable Utility Agent for Auxiliary Tasks (Tab Naming, Summaries, Toasts)**
**Problem**

Today, auxiliary tasks — tab naming, context summaries, and notification toasts — spawn the same agent (and model) as the primary session. If a user runs Claude Code with Opus, every tab rename quietly burns Opus tokens on a task that Haiku could handle. There's no way to decouple the "thinking" model from the "housekeeping" model.

This is especially relevant now that Copilot CLI exposes free-tier models. A user might run Opus for real work but happily let GPT-4o-mini or Haiku generate a 3-word tab title.

**Proposal**

Add a **Utility Agent** setting that lets users pick which agent (and implicitly which model) handles lightweight auxiliary tasks:

- Tab naming (`tabNaming.ts`)
- Context summarization (`context.ts` / `contextSummarizer.ts`)
- Any future low-stakes LLM calls (toast copy, auto-tagging, etc.)

When unset, behavior stays as-is (uses the session's own agent). When set, auxiliary tasks route to the configured utility agent instead.

**Where it fits in Settings**

A single dropdown in the General or Agents section:

```
Utility Agent: [Default (same as session)] ▾
               Claude Code
               Copilot CLI
               OpenCode
               ...
```

Optionally, a model override field (for agents that support model selection) so users can pin e.g. `haiku` or `gpt-4o-mini` explicitly.

**Implementation scope**

- New setting in `settingsMetadata.ts`: `utilityAgent` (agent ID or `null` for default)
- Optional: `utilityModelOverride` (string, passed as `modelId` to `buildAgentArgs`)
- `tabNaming.ts` and `context.ts` handlers check for the utility agent setting before resolving which agent to spawn
- Prompt files (`tab-naming.md`, `context-summarize.md`) are already agent-agnostic — no changes needed there

**Example savings**
| Task | Frequency | Current (Opus) | With Utility Agent (Haiku) |
|------|-----------|-----------------|---------------------------|
| Tab naming | Every new tab | ~500 tokens | ~500 tokens at ~1/30th cost |
| Context summary | Per grooming cycle | ~2-4k tokens | ~2-4k tokens at ~1/30th cost |

The token count stays the same — the cost doesn't.

**Prompt**

```
Implement a "Utility Agent" setting that lets users configure a separate agent for auxiliary LLM tasks (tab naming, context summarization, and future lightweight calls).

## Context

Currently, `tabNaming.ts` and `context.ts` (groomContext) resolve the agent from the active session. We want to add a setting that optionally overrides this with a dedicated utility agent + model.

## Requirements

1. Add two new settings in `src/shared/settingsMetadata.ts`:
   - `utilityAgentId`: `string | null` — an agent ID from `agentIds.ts`, or null to keep current behavior (default: `null`)
   - `utilityModelId`: `string | null` — optional model override passed to `buildAgentArgs()` (default: `null`)

2. Add defaults in `src/main/stores/defaults.ts`.

3. In `src/main/ipc/handlers/tabNaming.ts`:
   - When `utilityAgentId` is set, resolve that agent instead of the session's agent
   - When `utilityModelId` is set, pass it as `modelId` to `buildAgentArgs()`
   - Fall back to current behavior when either is null

4. In `src/main/ipc/handlers/context.ts` (groomContext):
   - Same override logic as tab naming

5. Add UI in `SettingsModal.tsx` under a "Utility Agent" section:
   - Agent dropdown (populated from available/detected agents)
   - Model text input (free-form, since valid values vary by agent)
   - Both fields show placeholder text explaining they default to the session agent when empty

6. Respect SSH remote config: if the utility agent needs to run remotely, apply the same `wrapSpawnWithSsh` pattern as the primary agent.

## Files to modify
- `src/shared/settingsMetadata.ts` — setting definitions
- - `src/main/stores/defaults.ts` — defaults
- `src/renderer/stores/settingsStore.ts` — store integration
- `src/main/ipc/handlers/tabNaming.ts` — utility agent resolution
- `src/main/ipc/handlers/context.ts` — utility agent resolution
- `src/renderer/components/SettingsModal.tsx` — UI for the new settings

## Out of scope
- Per-task agent selection (e.g. different agent for tab naming vs summaries) — single setting covers all auxiliary tasks for now
- Validating model IDs against agent capabilities — user is trusted to enter a valid model string
```

pedramamini:
## Summary

Fresh reimplementation of #836 on the latest `main`. The original branch went stale and could not be rebased after the settings-system refactor that split `GeneralTab.tsx` into `GeneralTab/` sub-sections, so this re-homes the UI and re-threads the backend from scratch. Closes #835.

Adds two settings that route auxiliary/background tasks (tab naming and context grooming) to a cheaper or faster agent with an optional model override, instead of always using the tab's own session agent:

- `utilityAgentId` - agent to use for auxiliary tasks. Null = use the session agent (fully backward compatible).
- `utilityModelId` - optional model override for the utility agent. Null = agent default model.

When both are unset, behavior is unchanged.

## UI

New Utility Agent section under Settings > General (right after Tab Behavior, since tab naming is one of the auxiliary tasks it governs). The agent dropdown is populated from `window.maestro.agents.detect()` (available, non-hidden, non-terminal agents); the model-override field only appears once an agent is selected.

## Design

- Both aux IPC handlers resolve the effective agent up front (`utilityAgentId || sessionAgent`) and thread the model override through `buildAgentArgs` (tab naming) and `groomContext` (grooming). The model override is only applied when a utility agent is actually configured.
- SSH remote execution and per-agent config overrides (customPath/customArgs/customEnvVars) are preserved and follow the resolved effective agent - nothing is hardcoded to `claude-code`. The Claude token-mode resolution in `tabNaming.ts` keys off the resolved agent's id, so a non-Claude utility agent naturally skips it.

## Files changed

Settings plumbing (all layers wired):
- `src/shared/settingsMetadata.ts` - metadata for both settings
- `src/main/stores/types.ts`, `src/main/stores/defaults.ts` - type + defaults (both null)
- `src/renderer/stores/settingsStore.ts` - state, setters, load, action export
- `src/renderer/hooks/settings/useSettings.ts` - hook interface
- `src/renderer/components/Settings/searchableSettings.ts` - search entry (+ `data-setting-id="general-utility-agent"` on the rendered section)

UI:
- `src/renderer/components/Settings/tabs/GeneralTab/components/UtilityAgentSection.tsx` (new)
- `GeneralTab.tsx` + `components/index.ts` - wire the section in

Backend:
- `src/main/ipc/handlers/tabNaming.ts` - resolve utility agent for tab naming
- `src/main/ipc/handlers/context.ts` - resolve utility agent for grooming (+ `settingsStore` dep)
- `src/main/utils/context-groomer.ts` - accept optional `modelId`
- `src/main/index.ts` + `src/main/ipc/handlers/index.ts` - pass `settingsStore` to the context handlers

Tests:
- `src/__tests__/main/ipc/handlers/tabNaming.test.ts` - two new tests (routes naming to the configured utility agent; falls back to the session agent when unset) + mock updated to return the null default for the new keys

## Deviation from #836

The old PR added a Utility Agent block inline in the now-deleted monolithic `GeneralTab.tsx`. This version implements it as a dedicated `UtilityAgentSection` component to match the current per-section GeneralTab structure, and adds the `searchableSettings` entry + `data-setting-id` wrapper that the original PR review flagged as missing.

## Validation

- `tsc -p tsconfig.main.json` and `-p tsconfig.json`: clean
- ESLint on changed files: clean
- `tabNaming.test.ts`: 48 passed (2 new); context-groomer / settingsStore / searchableSettings / defaults suites: green
- Pre-push full suite: 34,052 passed, 108 skipped

Reimplemented from scratch; ready for review. Do not merge.

## Summary by CodeRabbit

* **New Features**
  * Added Utility Agent settings for routing auxiliary tasks to a selected agent.
  * Added an optional model override for Utility Agent tasks.
  * Settings persist across sessions and default to the active session agent when unset.
  * Added agent discovery, unavailable-agent warnings, and searchable General settings support.

* **Bug Fixes**
  * Improved context grooming and tab naming to consistently use the configured utility agent and model.
  * Preserved session-specific settings when the session and utility agents match.


pedramamini:
Rebased onto current rc and fixed both findings. The rebase was not mechanical - **the settings system moved underneath this PR again**, which is the same thing that killed its predecessor #836.

### What actually conflicted

rc split `settingsMetadata.ts` and `searchableSettings.ts` into per-category modules, and moved the tab-naming registration out of `main/index.ts` into `ipc/bootstrap/index.ts`. The conflicts looked enormous (~3,000 lines) but this PR's own contribution to those files is **+51/-6**, so I took rc's side wholesale and re-applied the 51 lines at their new homes:

- `settingsMetadata.ts` -> `settingsMetadataCore.ts`
- `searchableSettings.ts` -> `searchableSettingsGeneral.ts`
- `main/index.ts` -> **redundant.** rc had already added `settingsStore` to `registerTabNamingHandlers` independently, so that hunk is gone.

### The findings

**Session executable overrides followed the agent switch.** Both `context.ts` and `tabNaming.ts` passed the SESSION agent's `customPath` / `customArgs` / `customEnvVars` through even after resolving to a different utility agent. That launches the wrong binary, or hands one agent's flags and API keys to another - a session pinned to a custom `claude` path would have run *that* path while claiming to run the utility agent. Dropped when the two agents differ; the utility agent's own settings already come from `agentConfigValues`, which is keyed by the effective id.

`sshRemoteConfig` is deliberately **not** dropped: it says *where* the work runs, not *which* binary runs it, and grooming a remote agent's context against the local filesystem would be its own bug.

**Stale persisted agent id.** A `utilityAgentId` whose agent is no longer installed matched no `<option>`, and a `<select>` with an unmatched value renders as its **first** option - so the UI read "Default (same as session)" while tab naming and grooming kept routing to the missing binary and failing. Now surfaced as an explicit `(not installed)` option plus a warning, so the setting the user is running is the setting they can see.

### Two rules rc added since this PR was written

The rebase also turned up `settingsStyleGuide.test.ts`, which did not exist when this branch was cut:

- The description stacked `opacity-50` on `theme.colors.textDim`, multiplying the dimming and dropping contrast below 3:1 in 17 of 21 themes (rule 1).
- The section heading was hand-rolled instead of using the shared `SettingsSectionHeading` primitive (rule 3).

Both fixed. Worth noting neither would have been caught by a green CI on the stale base.

Local: `tsc` clean on all four configs; 4,530 tests green across `main/ipc/`, `renderer/components/Settings/`, and `shared/`.

pedramamini:
Good catch - fixed.

The `catch` silently swallowed a failed `agents.detect()` and never set `agentsLoaded`, so the stale-agent branch could not fire at all. The select then fell back to its first option and read "Default" while the persisted agent was still what tab naming and grooming actually used. Same lie as the missing-agent case, reached a different way.

The fix tracks detection failure separately, because the two states are not the same claim:

- **Detection succeeded, agent absent** -> it is genuinely gone. Shows `(not installed)` and warns that auxiliary tasks will fail until the user picks another agent or returns to Default.
- **Detection failed** -> we do not know. Shows `(unverified)` and only notes the selection is still in use.

Collapsing them would have meant labelling a perfectly good agent "not installed" every time an IPC call hiccuped, which is its own bad outcome.

Local: `tsc` clean, 510 tests green in `renderer/components/Settings/`.

@greptileai please re-review.

PR #1284, opened 2026-07-22
fix(autorun): reconcile CLI Auto Run summary with cumulative session stats

pedramamini:
## Summary

Fresh reimplementation of #735 on latest `main`. The original PR became unrebasable after the batch-processor rewrite (`src/cli/services/batch-processor.ts` +688/-573 and the `useBatchProcessor.ts` split +140/-1810). Closes #734.

## Bug still reproduces (verified on current main)

The renderer path already gained history reconciliation during the rewrite (`src/renderer/hooks/batch/internal/batchFinalSummary.ts` via `aggregateAutoRunHistoryTotals` + `mergeFinalSummaryTotals`), so the desktop half of #735 is already covered by a more robust implementation.

The CLI path did NOT. On current main, `createAutoRunSummary` in `src/cli/services/batch-processor.ts` and the `complete` JSONL event were built purely from in-memory counters (`totalCompletedTasks`, `totalInputTokens`, `totalOutputTokens`, `totalCost`, `Date.now() - batchStartTime`). Those reset when a run spans an app/CLI restart, a resume, or a mid-run kill, so the final summary and the `complete` event undercount cumulative work. That is the remaining live half of the bug.

## Fix

Reconcile the in-memory counters with persisted on-disk history before writing the summary and emitting the `complete` event:

- Read history for the session, reconstruct cumulative totals from task entries written after the most recent final `Auto Run ...` summary (so it spans restarts but does not absorb earlier completed runs on the same session), and take `Math.max` of (in-memory, history-derived).
- History-read failure is an expected/recoverable mode: fall back to the in-memory counters and `logger.warn` (does not bubble to Sentry).

To avoid duplicating the renderer's battle-tested logic (per CLAUDE.md dedup rules), the pure aggregate/merge helpers were extracted from `batchFinalSummary.ts` into a new shared module `src/shared/autoRunHistoryReconciliation.ts`. Both the renderer and the CLI now import it; `batchFinalSummary.ts` re-exports it so existing renderer importers and tests are unchanged. CLI per-task history entries now also stamp `completedTaskCount` so the shared aggregation counts checkboxes exactly (matching desktop Auto Run), rather than approximating one task per run entry.

## Design deviations from #735

- #735 hand-rolled a bespoke reconciliation loop in both the renderer and the CLI. Here the renderer is already done (better), so this PR only fixes the CLI and does so by reusing the shared canonical logic instead of hand-rolling a second copy.
- Added `completedTaskCount` on CLI task entries for exact cross-restart counts (a small accuracy improvement #735 lacked).

## Files changed

- `src/shared/autoRunHistoryReconciliation.ts` (new) - shared pure reconciliation helpers
- `src/renderer/hooks/batch/internal/batchFinalSummary.ts` - moved helpers to shared, re-export
- `src/cli/services/batch-processor.ts` - reconcile before summary/complete; stamp `completedTaskCount`
- `src/__tests__/cli/services/batch-processor.test.ts` - restart-reconciliation, no-double-count, no-undercount, read-failure-fallback

## Validation

- `tsc` clean on all four configs: `tsconfig.main.json`, `tsconfig.json`, `tsconfig.cli.json`, `tsconfig.lint.json`
- eslint clean on changed files
- Full pre-push suite green: 34,054 passed / 108 skipped

Ready for review. Do not merge.

## Summary by CodeRabbit

* **Bug Fixes**
  * Auto Run summaries now retain accurate task counts, elapsed time, token usage, and costs after restarts.
  * Prevented previously completed runs from being counted again.
  * Final summaries are now recorded for completed and halted runs.
  * Improved handling when saved run history is delayed, empty, or temporarily unavailable.
  * Added a warning and safe fallback when run history cannot be read.
  * Task progress is now preserved in history to improve cumulative totals across resumed runs.


pedramamini:
Rebased onto current rc (clean) and fixed all three findings.

### 1. Halt bypassed reconciliation

`haltMarker.halted` returned at `batch-processor.ts:689` before ever reaching `reconcileTotals()`, so a run that crossed a restart and then halted reported only this process's in-memory slice. Now reconciles like any other end-of-run.

### 2. Non-loop runs left no boundary - this was the sneaky one

`createAutoRunSummary` had `if (!playbook.loopEnabled && loopIteration === 0) return;`.

That row is not just cosmetic: **it is the run boundary** that `aggregateAutoRunHistoryTotals` scans back to. Skipping it meant the *next* run had nothing to stop at, so its aggregation swept up the previous run's task rows and, because the merge is `Math.max`, the inflated number won. Two consecutive 5-task non-loop runs reported the second as 10.

Desktop never had this gate - `buildFinalSummary` in `useBatchRunner.ts` is unconditional - so removing it is also CLI/desktop parity, which is what this PR is for.

### 3. Legacy rows

Left as `?? 1`, now documented as a deliberate floor. A row written before `completedTaskCount` existed may cover several tasks and nothing on disk says how many, so one-per-row can undercount an old run but never overcount one. Recovering the true number is not possible from data that was never written.

### One thing the fix forced

The halt row is `Auto Run halted: <reason>`, which did **not** match `FINAL_AUTORUN_SUMMARY_RE`. Writing it without extending that regex would have produced a summary row that silently is not a boundary - the exact bug in #2, reintroduced. `halted` is now in the pattern, with a comment saying that adding a terminal outcome means adding it there in the same change.

Local: `tsc` clean on all four configs; 3,573 tests green across `__tests__/cli/`, `__tests__/shared/`, and `renderer/hooks/batch/`, including two new cases (boundary written for a non-looping run; halt reconciles and writes its boundary).

Worth noting this closes the loop on **#734**, which I closed on 2026-07-25 telling @kringle25 that the desktop half was fixed and this PR was "the one to watch" for the CLI.

PR #1287, opened 2026-07-22
feat(cli): add --visible flag to goal-run for desktop-owned Auto Runs

### LINKED ISSUE #1286: Feature: CLI: add --visible flag for desktop Goal/Spec Auto Runs
## Summary
CLI: add --visible flag for desktop Goal/Spec Auto Runs

## Environment
- Maestro version: 0.18.5-RC
- Install source: Dev build

## Desired Outcome
The CLI should launch a desktop-visible Goal/Spec Auto Run with parity to the UI Go/Spec buttons. Recommended: `maestro-cli goal-run --visible --new-tab [--tab <id>] [--focus/--no-focus] [--wait] --json --exit-criteria "..." <agent-id> "<goal>"`. The run must be owned by the desktop app (not a headless subprocess), appear as an active visible Auto Run in the same UI surface used by UI-launched Go/Spec, and be openable/monitorable via a returned `maestro://session/<agent-id>/tab/<tab-id>` URI. With --json it returns stable identifiers (ok, mode, visible:true, agent_id, tab_id, session_id, run_id, status). It should be stoppable via `stop-auto-run -a <agent-id>` and appear in `session list --json`. Ideally a shared `startVisibleGoalRun({agentId, goal, exitCriteria, maxIterations, tabId, newTab, focus, writeHistory})` service is called by both the UI Go button and the new CLI IPC handler.

## Details
`maestro-cli goal-run` runs headlessly. Its only options are --exit-criteria, --max-iterations, --no-history, --json, --verbose. There is no --visible / --new-tab / --tab / --focus option, so CLI-launched Goal runs are real but only observable via CLI/history artifacts, never in the agent tab / Auto Run surface. `dispatch --new-tab` opens a visible tab and returns ids but is a one-shot prompt dispatch, not the Goal/Spec Auto Run loop (no exit-criteria/lifecycle semantics). `auto-run --launch` is the visible route only for document/playbook Auto Runs, and `run-doc` is explicitly headless. Net effect: a root/orchestrator agent cannot launch a monitored, UI-visible free-text Goal worker without handing control back to the user.

## Additional Context
Backward-compatibility is a hard requirement: existing headless `goal-run` must stay the default unless --visible is passed; --visible must fail closed (clear error) when the desktop app is unreachable rather than silently falling back to headless; busy-agent behavior must be deterministic (clear busy error, or wait only with --wait); --new-tab and --tab <id> should be mutually exclusive. Non-goals: don't change headless goal-run, don't make run-doc visible, don't replace dispatch --new-tab, and don't force users to create a document/playbook just to launch a visible free-text Goal. Acceptable alternative to extending goal-run: explicit `go`/`spec` subcommands with --visible, since the key requirement is UI parity, not the command name. Test plan supplied covers CLI parser unit tests, mocked-IPC integration tests, and manual QA (VISIBILITY_TEST_OK visible run + HEADLESS_TEST_OK regression). Interim workarounds: `dispatch --new-tab` for visible one-shot prompts, `auto-run --launch` for document-based visible Auto Runs. Motivation: multi-agent orchestration needs workers that the user can see and trust are running in the intended worktree via the same UI surface as manual Go/Spec runs.

## Screenshots / Recordings
None

pedramamini:
Closes #1286

## What

Adds `--visible` to `maestro-cli goal-run`:

```
maestro-cli goal-run --visible [--exit-criteria "..."] [--max-iterations N] [--json] <agent-id> "<goal>"
```

With `--visible`, the Goal-Driven Auto Run is handed to the **running desktop app** so it executes as a live, desktop-owned Auto Run in the **same UI surface as the Go/Spec buttons**, instead of running headless in the CLI process. This lets a root/orchestrator agent launch a monitored, UI-visible free-text Goal worker without handing control back to the user.

## How

The visible path threads a new `launch_goal_run` message through the **existing** CLI → desktop WebSocket bridge (the same four-layer path `dispatch --new-tab` and `auto-run --launch` already use) and terminates in the renderer at the exact `startBatchRun({ goalConfig })` entry point the UI **Go** button uses. There is no new run engine: the desktop owns spawning, busy arbitration, `stop-auto-run`, and `session list` visibility, so those requirements come for free.

Layers touched (all additive):
- **CLI**: `--visible` flag + `runVisibleGoalRun()` in `goal-run.ts`, registration in `index.ts`
- **Web server**: `launch_goal_run` message handler, callback registry/types, WebServer wiring
- **Main → renderer**: `remote:launchGoalRun` IPC bridge in `web-server-factory.ts`, preload subscription
- **Renderer**: `maestro:launchGoalRun` remote event listener that builds a `goalConfig` and calls `startBatchRun`

## Behavior / contract

- **Backward compatible**: headless `goal-run` stays the default; `--visible` is purely additive (no deletions).
- **Fails closed**: `--visible` errors with `MAESTRO_NOT_RUNNING` when the desktop app is unreachable rather than silently falling back to headless.
- **Deterministic busy handling**: a busy agent returns a clear `AGENT_BUSY` error.
- **`--json`** returns stable identifiers plus a deep link:
  ```json
  { "ok": true, "mode": "visible", "visible": true, "agent_id": "...", "session_id": "...", "tab_id": "...", "status": "launched", "uri": "maestro://session/<id>/tab/<id>" }
  ```

## Scope notes / open questions for the author

The issue's "recommended" API also listed `--new-tab` / `--tab` / `--focus/--no-focus` / `--wait` and a `run_id`. This PR intentionally implements the **core UI-parity MVP** the issue calls the key requirement, and leaves those out for now because:
- Goal-Driven Auto Runs are modeled **per-agent** (batch state keyed by session/agent id), not per-tab, so `--new-tab`/`--tab` don't map cleanly onto the existing engine. The run attaches to the agent and surfaces on its active AI tab, which is what we return as `tab_id`.
- `run_id` / provider `session_id` aren't known synchronously at launch (the provider session is assigned after the first agent turn), so they're omitted rather than faked.
- Busy is handled as a deterministic clear error, which the issue lists as acceptable in lieu of `--wait`.

Happy to extend any of these if you'd prefer the fuller surface.

## Tests

- New `src/__tests__/cli/commands/goalRunVisible.test.ts` covers the visible path: message shape, JSON identifiers + deep link, fail-closed (`MAESTRO_NOT_RUNNING`), `AGENT_BUSY`, generic rejection, and human-readable output.
- Updated existing web-server-factory and useRemoteIntegration mocks for the new bridge method.
- `npm run lint` (all tsconfigs) and the CLI + web-server + goalDriven + remote-integration suites pass locally.

> Note: local validation is single-OS; CI (ubuntu + windows) is the source of truth before merge.

## Summary by CodeRabbit

* **New Features**
  * Added `goal-run --visible` to launch Goal-Driven Auto Runs in the desktop interface.
  * Displays a deep link to the launched run in JSON or human-readable format.
  * Added validation and clear errors for unavailable, busy, or rejected launches.
  * Added desktop-to-CLI communication for starting visible runs and reporting results.

* **Tests**
  * Added coverage for successful launches, deep-link formatting, missing sessions, busy agents, desktop connectivity, and rejection errors.

### DIFF
--- a/src/__tests__/cli/commands/goalRunVisible.test.ts
+++ b/src/__tests__/cli/commands/goalRunVisible.test.ts
@@ -1,0 +1,130 @@
+/**
+ * @file goalRunVisible.test.ts
+ * @description Tests for `maestro-cli goal-run --visible`, which hands the run
+ * to the desktop app so it appears as a live Auto Run (parity with the UI Go
+ * button) instead of running headless in the CLI process.
+ */
+
+import { describe, it, expect, vi, beforeEach, type MockInstance } from 'vitest';
+
+vi.mock('../../../cli/services/maestro-client', () => ({
+	withMaestroClient: vi.fn(),
+}));
+
+import { runVisibleGoalRun } from '../../../cli/commands/goal-run';
+import { withMaestroClient } from '../../../cli/services/maestro-client';
+
+const GOAL_CONFIG = { goal: 'Ship the thing', exitCriteria: 'tests pass', maxIterations: 5 };
+
+describe('goal-run --visible (runVisibleGoalRun)', () => {
+	let logSpy: MockInstance;
+	let errorSpy: MockInstance;
+	let exitSpy: MockInstance;
+
+	beforeEach(() => {
+		vi.clearAllMocks();
+		logSpy = vi.spyOn(console, 'log').mockImplementation(() => {});
+		errorSpy = vi.spyOn(console, 'error').mockImplementation(() => {});
+		exitSpy = vi.spyOn(process, 'exit').mockImplementation(() => undefined as never);
+	});
+
+	function mockDesktop(response: unknown): MockInstance {
+		const sendCommand = vi.fn().mockResolvedValue(response);
+		vi.mocked(withMaestroClient).mockImplementation(async (action) =>
+			action({ sendCommand } as never)
+		);
+		return sendCommand;
+	}
+
+	it('sends launch_goal_run and prints stable JSON identifiers with a deep link', async () => {
+		const sendCommand = mockDesktop({
+			type: 'launch_goal_run_result',
+			success: true,
+			tabId: 'tab-77',
+		});
+
+		await runVisibleGoalRun('agent-1', GOAL_CONFIG, true);
+
+		expect(sendCommand).toHaveBeenCalledWith(
+			{
+				type: 'launch_goal_run',
+				sessionId: 'agent-1',
+				goal: 'Ship the thing',
+				exitCriteria: 'tests pass',
+				maxIterations: 5,
+			},
+			'launch_goal_run_result'
+		);
+
+		const output = JSON.parse(logSpy.mock.calls[0][0]);
+		expect(output).toMatchObject({
+			ok: true,
+			mode: 'visible',
+			visible: true,
+			agent_id: 'agent-1',
+			tab_id: 'tab-77',
+			status: 'launched',
+			uri: 'maestro://session/agent-1/tab/tab-77',
+		});
+		expect(exitSpy).not.toHaveBeenCalled();
+	});
+
+	it('returns tab_id null and a session-only deep link when the desktop omits a tab id', async () => {
+		mockDesktop({ type: 'launch_goal_run_result', success: true });
+
+		await runVisibleGoalRun('agent-1', GOAL_CONFIG, true);
+
+		const output = JSON.parse(logSpy.mock.calls[0][0]);
+		expect(output.tab_id).toBeNull();
+		expect(output.uri).toBe('maestro://session/agent-1');
+	});
+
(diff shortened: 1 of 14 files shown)


pedramamini:
Closing this. Borderline call, so writing down the reasoning.

The findings are fixable rather than architectural: startup errors are raised *after* the success response has already been sent, and busy-checking does not atomically reserve the agent before acknowledging the launch, so two visible runs can overlap on one agent. Both are real, neither is deep.

What tips it is the drift: +594 across 14 files, conflicting, **305 commits behind rc**, for a niche CLI flag. The renderer integration and the WebSocket/IPC bridge it threads through have both been reworked since (`#1352` decomposed `web-server-factory`, `#1357` reworked CLI dispatch), so the rebase is not mechanical - it is re-deriving the integration against a bridge that has moved.

If you are actually using desktop-visible goal runs, say so and I will rebuild it against current rc rather than untangle this. Branch is kept.

PR #1288, opened 2026-07-23
feat(ui): single thinking pill + global tool-call visibility setting

chr1syy:
## Summary

The chat-box toolbar now shows a **single Thinking pill**. The per-tab **Tools pill** is removed, and tool-call visibility becomes a **global toggle under Settings -> General** ("Show tool calls in responses", default on).

This is a display control, not a permission control: agents still execute tools regardless. When off, `source:'tool'` log entries are hidden from the transcript. Tool events are still recorded unconditionally, so toggling visibility never mutates the transcript or breaks running -> completed correlation.

## What changed

**New global setting `showToolCalls` (default `true`)**
- `src/shared/settingsMetadata.ts` - metadata (boolean, default `true`)
- `src/renderer/stores/settingsStore.ts` - state field, `setShowToolCalls`, load patch, and the `getSettingsActions` export
- `src/renderer/hooks/settings/useSettings.ts` - `UseSettingsReturn` interface
- `src/renderer/components/Settings/tabs/GeneralTab/components/ToolCallsSection.tsx` - new "Tool Calls" section (Wrench icon, single toggle), exported and wired into `GeneralTab` after the thinking-mode section
- `src/renderer/components/Settings/searchableSettings.ts` - `general-tool-calls` search entry

It persists through the generic string-key settings path, exactly like `defaultShowThinking` / `enterToSendAI` (none of which live in `main/stores/defaults.ts` / `MaestroSettings`).

**Removed the per-tab Tools pill and its whole chain (clean cutover)**
- `ToolbarControls.tsx` - deleted the Tools pill button, its two props, and the now-unused `Wrench` import (Thinking pill untouched)
- Removed `tabShowTools` / `onToggleTabShowTools` through `InputArea`, `MainPanel` / `MainPanelContent`, `useMainPanelProps`, `App.tsx`, and `tabs/internal/types.ts`
- Removed `AITab.showTools` (`types/index.ts`), the `createTab` option (`tabHelpers.ts`), the `handleToggleTabShowTools` handler (`useAITabHandlers.ts`), and the now-orphaned `toolLogsRecorded` helper (`thinkingLogs.ts`)
- `TerminalOutput.tsx` now reads `useSettingsStore((s) => s.showToolCalls)` to gate `source:'tool'` rendering

**Docs/tests**
- Updated the stale `AGENT-INFRA.md` line describing the removed per-tab gating
- Tests: TerminalOutput hide-tests now drive the global setting; obsolete `showTools` listener/`createTab` tests removed; all `showTools` fixtures stripped; added GeneralTab toggle tests and a `settingsStore` default+setter test

## Testing

- `npm run lint` (all three tsc configs): clean
- `npm run lint:eslint` on changed files: clean
- `npm run format:check:all`: clean
- Touched test suites (TerminalOutput, GeneralTab, settingsStore, useAgentToolExecutionListener, tabHelpers, useAITabHandlers): all pass
- Full `validate:push` gate runs via the pre-push hook

## Summary by CodeRabbit

* **New Features**
  * Added a global **Tool Calls** setting to General preferences.
  * Added search support for the new **Tool Calls** setting.
* **Changes**
  * Removed the per-tab Tools visibility toggle; tool-call visibility is now controlled globally.
  * Tool-call events are always recorded; hiding tool-call output is a render-only behavior.
* **Tests**
  * Extended coverage for the new setting across settings UI, terminal output, and tool-execution listener.
* **Documentation**
  * Updated the tool execution pipeline documentation to reflect the new tool log visibility behavior.


chr1syy:
Addressed review feedback in `8170b6def`:

- **ToolCallsSection keyboard double-fire** (CodeRabbit): the clickable row's `onKeyDown` now ignores keydowns whose target is a descendant (`e.target !== e.currentTarget`), so a focused nested `ToggleSwitch` no longer toggles twice (row handler + the button's native Enter/Space click). Click was already single-fire via `ToggleSwitch`'s `stopPropagation`. Added a test asserting the switch click fires the setter exactly once, plus a row keyboard-activation test.
- **AGENT-INFRA wording** (CodeRabbit): scoped the "always recorded" claim to the `showToolCalls` setting and noted that the `showThinking` lifecycle can still drop stored `thinking`/`tool` entries (e.g. on exit when not `sticky`), independently of render visibility.

### CI note: the one red check is pre-existing on `rc`, not from this PR

`Test (windows-latest, shard 1/2)` fails in `src/__tests__/main/stats/token-usage/token-usage-accounts.test.ts` with a Windows path-normalization assertion (`D:\home\u\.claude` vs `/home/u/.claude`).

- That test is **not in this PR's diff** and lives in the stats/token-usage subsystem, unrelated to this tool-call visibility change.
- It was last modified on `rc` by `fb7df880f` (feat(stats): restore Tokens dashboard tab), and the current `rc` head (`55a04f5fc`, this PR's base) is itself red on CI - so the failure is inherited from the base branch, not introduced here.
- All checks this change actually affects are green: `lint-and-format`, both `ubuntu-latest` shards, and `windows-latest` shard 2/2.

I've left this out of scope to avoid mixing an unrelated cross-platform stats fix into a UI PR. Happy to fix it in a separate PR to `rc` if preferred.

pedramamini:
@chr1syy Thanks for this one, and thanks especially for the thorough writeup and the follow-up commit in `8170b6def`. Moving tool-call visibility from a per-tab pill to a global setting is the right call, and the cutover is genuinely clean.

## What I verified

**Setting registration is complete.** All four required touchpoints are there per `CLAUDE.md`: `settingsMetadata.ts`, `settingsStore.ts` (state + setter + `loadAllSettings` patch + `getSettingsActions`), `searchableSettings.ts`, and the `data-setting-id="general-tool-calls"` wrapper on the rendered control. `ToolCallsSection` matches the existing sibling-section pattern (`role="button"` row + nested `ToggleSwitch`), so it is consistent with `RenderingSection`, `TabBehaviorSection`, and friends.

**The cutover really is clean.** `git grep` for `showTools`, `onToggleTabShowTools`, and `toolLogsRecorded` across `src/` and `docs/` on your branch returns zero hits. No orphaned props, no dead helper, no stale doc line.

**Both CodeRabbit actionable comments are addressed.** The `onKeyDown` guard (`e.target !== e.currentTarget`) is the right fix for the double-fire, and the `AGENT-INFRA.md` rewording correctly separates "recorded regardless of `showToolCalls`" from "the `showThinking` lifecycle can still drop stored entries."

**Tests pass locally.** Ran `TerminalOutput` (135), `GeneralTab`, `settingsStore`, `useAgentToolExecutionListener`, and `tabHelpers` (584 combined) against your branch: all green.

## On the Greptile P1

Greptile flagged that `showThinking` cleanup deletes `source:'tool'` entries unless the mode is `sticky`, and scored the PR 3/5 on that basis. I traced it and I think the "should not merge" verdict is overstated for this PR specifically:

- `useBatchedSessionUpdates.ts:223-229` and `applyExitThinkingPolicy` in `exitTabCleanup.ts` are both untouched by your diff.
- On the base branch, `toolsVisible` resolved to `toolLogsRecorded(showTools, showThinking)`, and new tabs defaulted `showTools` to `true`. So tool logs were already being purged on exit and on new assistant text under exactly the same conditions.

The behavior is identical before and after, so this is not a regression you are introducing. Your scoped `AGENT-INFRA.md` wording already documents the interaction honestly, which is the right response.

That said, it is a real gap worth a follow-up: `defaultShowThinking` defaults to `'off'` while `showToolCalls` defaults to `true`, so out of the box a user watches tool calls stream in and then vanish once the turn completes. The new setting promises a bit more than the lifecycle currently delivers. Decoupling tool-log retention from thinking mode is squarely out of scope here, but it would be worth its own issue.

## On CI

Your analysis of `Test (windows-latest, shard 1/2)` is correct. Base `rc` (`55a04f5fc`) fails the identical four `discoverClaudeAccounts` assertions in `token-usage-accounts.test.ts`. Inherited, unrelated, and right to leave out of a UI PR.

`Test (windows-latest, shard 2/2)` had also gone red on `8170b6def`, failing `TerminalOutput > auto-scroll when at bottom > keeps sticking to the bottom after clicking the pin button while scrolled up`. That test is not in your diff, and `8170b6def` only touched three unrelated files, so I re-ran the job: it came back green. Confirmed flake, no action needed from you.

## Optional nit (not blocking)

CodeRabbit's nitpick about hydration coverage is fair. `loadAllSettings()` reading `showToolCalls` is the one piece of new code without a test. A few lines mocking `getAll()` to return `showToolCalls: false` and asserting the store stays disabled after load would close that. Entirely up to you whether it rides along here or not.

## Verdict

Approved. Nice, disciplined refactor: complete cutover, docs updated alongside the code, and tests migrated rather than merely deleted. The only red check left is inherited from `rc`.

chr1syy:
Follow-up in `e28cca87c`: made the `TerminalOutput` follow-scroll assertions deterministic.

Growing `GeneralTab.test` shifted vitest's shard balancing, which moved `TerminalOutput.test.tsx` onto a Windows shard and exposed a latent flake: the auto-follow scroll is armed by a `MutationObserver` microtask that then schedules a `requestAnimationFrame`, so advancing fake timers synchronously could race ahead of the observer on a slow runner and miss the `scrollTo`. The three rerender-driven follow-scroll assertions now use `vi.advanceTimersByTimeAsync`, which drains microtasks between timer steps. `windows-latest` shard 2/2 is green again.

### Remaining red check is pre-existing on `rc`

The only failing check is `Test (windows-latest, shard 1/2)`, entirely in `src/__tests__/main/stats/token-usage/token-usage-accounts.test.ts` (a Windows path-normalization assertion, `D:\home\u\.claude` vs `/home/u/.claude`). It is not in this PR's diff, and `rc`'s own head (`55a04f5fc`) is already red on it - so it is inherited from the base branch, not introduced here. Happy to fix it in a separate PR to `rc`.

PR #1290, opened 2026-07-23
ci(release): build Linux x64 on ubuntu-22.04 and guard the glibc floor

chr1syy:
## Problem

The Linux x64 native modules (`better-sqlite3`, `node-pty`) are compiled from source on the release runner (`npm run postinstall` -> `electron-rebuild`, with prebuilds deliberately removed to force from-source), so **the runner's glibc sets the symbol-version floor of the shipped binaries**.

The x64 release leg runs on `ubuntu-latest`, a moving label that now points at Ubuntu 24.04 (glibc 2.39). Building native modules on a newer glibc can raise their floor above what Ubuntu 22.04 provides (glibc 2.35). When that happens the modules fail to load at runtime with:

```
/lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.38' not found
(required by .../better-sqlite3/build/Release/better_sqlite3.node)
```

Because both the Cue database (`src/main/cue/cue-db.ts`) and the stats database (`src/main/stats/stats-db.ts`) load `better-sqlite3`, this takes out both. `cue-recovery-service.ts` catches the init failure and logs `[CUE] Failed to initialize Cue database - engine will not start`, and `cue-engine.ts` bails out of `start()` on that failure, so the engine never runs and the UI shows paused/disabled with the toggle doing nothing. It is a system-library mismatch, not a config problem, so no `cue.yaml` change can fix it.

## Fix

1. **Pin the x64 release leg to `ubuntu-22.04`** (the oldest supported target, glibc 2.35) so the floor cannot drift. This mirrors the existing deliberate pins for `macos-15` and `ubuntu-24.04-arm`.
2. **Add `.github/scripts/verify-glibc-floor.sh`** and run it after the x64 packaging/verify step. It scans every shipped x86-64 `.node` in the unpacked resources and fails the build if any requires a glibc newer than 2.35. It skips musl-linked and foreign-arch (arm64 / Windows PE) prebuilds that `node-pty` bundles. This is a backstop that holds regardless of which image built the package.

## Verification

Ran the guard against the actual packaged binaries from a current Linux build:

- Floor `2.35`: **passes** (exit 0). Checks `better_sqlite3.node`, `pty.node`, `node-pty/bin`, `prebuilds/linux-x64/pty.node`, `keyring-linux-x64-gnu`; correctly skips the musl keyring and the darwin/arm64/win32 prebuilds.
- Floor `2.33`: **fails** (exit 1) on the `GLIBC_2.34` binaries, proving it catches violations rather than passing vacuously.

`release.yml` parses as valid YAML; the script passes `bash -n`. No em/en dashes introduced.

## Scope / follow-up

ARM64 (`ubuntu-24.04-arm`) is intentionally left unchanged in this PR to avoid destabilizing that carefully-tuned runner (fpm/ruby setup), and because the reported failure is x86_64. It has the same theoretical exposure; a matching pin (`ubuntu-22.04-arm`) plus the same guard is a sensible follow-up if ARM 22.04 support is required.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Linux release compatibility by ensuring shipped native modules remain compatible with the GLIBC version provided by Ubuntu 22.04.
  * Added automated packaging checks that verify native `.node` modules won’t require a newer GLIBC than the release target supports before a release is published.

### DIFF
--- a/.github/scripts/verify-glibc-floor.sh
+++ b/.github/scripts/verify-glibc-floor.sh
@@ -1,0 +1,96 @@
+#!/usr/bin/env bash
+# verify-glibc-floor.sh - Fail the build if any shipped x86-64 native module
+# requires a glibc symbol version newer than the supported floor.
+#
+# Why this exists: the Linux native modules (better-sqlite3, node-pty, ...) are
+# compiled from source on the CI runner, so their glibc floor is whatever the
+# runner's toolchain happens to provide. If the runner base image drifts to a
+# newer glibc, the shipped .node files start requiring newer versioned symbols
+# and silently fail to load on older-but-still-supported distros. That is what
+# broke Ubuntu 22.04 (glibc 2.35): better_sqlite3.node ended up needing a
+# GLIBC_2.38 symbol, so the Cue and stats databases failed to initialize.
+#
+# This guard is a cheap insurance net that runs regardless of which runner built
+# the package: it inspects the actual packaged binaries and fails loudly if the
+# floor was exceeded.
+#
+# Usage: ./verify-glibc-floor.sh <max-glibc> <search-dir>
+#   max-glibc:  highest allowed glibc version, e.g. 2.35 (Ubuntu 22.04)
+#   search-dir: directory scanned recursively for *.node files
+
+set -euo pipefail
+
+MAX_GLIBC="${1:?Usage: $0 <max-glibc> <search-dir>}"
+SEARCH_DIR="${2:?Usage: $0 <max-glibc> <search-dir>}"
+
+if ! command -v readelf >/dev/null 2>&1; then
+	echo "✗ ERROR: readelf not found (install binutils)" >&2
+	exit 1
+fi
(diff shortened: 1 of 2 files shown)


chr1syy:
Addressed both review findings on `verify-glibc-floor.sh` in `92ba8bcdf` (fail-closed guard):

- **find** now writes the `.node` scan to a temp file and aborts on any find error or stderr output, so a partial scan (e.g. a permission error on a subdir) can't silently drop a binary and let an over-floor module slip through.
- **readelf** is read explicitly, and a readelf failure now fails the guard - instead of `readelf ... || true` masking a parse failure as an empty result and skipping an x86-64 module as "musl/static".
- GLIBC extraction is restricted to the version-needs section (`.gnu.version_r`) so the symbol/definition tables can't skew the detected max.

Verified against real `.node` binaries (glibc pass at floor 2.35, breach at 2.30, musl skip) and a stubbed `readelf` failure (now fails closed, exit 1).

PR #1294, opened 2026-07-25
Keep composer caret visible on programmatic value changes

chr1syy:
## Finding E1 - composer multiline caret scroll

On programmatic value changes (draft restore, @-mention / slash insertion, template write-back, voice transcription) the composer textarea snapped from 176px down to 112px and hid the caret below the fold.

## Change
- Unifies the two divergent textarea max-height constants into `TEXTAREA_MAX_HEIGHT = 176`.
- Replaces the `shouldScrollTextareaToEnd` heuristic + manual `scrollTop` in the programmatic path with `scrollTextareaToCaretEnd`, pinning the view to the caret exactly like the keystroke path.
- Removes now-dead `prevInputValueRef` / `useRef` import.
- Mid-text carets (deferred @-mention placement) no longer yank the view to the bottom.

## Validation
Type check (3 tsc configs), ESLint, Prettier clean. Scoped tests `useInputAreaAutosize.test.tsx` + `textareaSizing.test.ts` -> 9/9 pass.

Fix plan: `.maestro/FIX-E1-01.md`. Authored + validated via Maestro Auto Run in an isolated worktree.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved composer textarea resizing with a consistent maximum height.
  * Preserves caret visibility by scrolling only when the caret is at/after the end of the content.
  * Keeps scroll position when the caret is in the middle of the text.
  * Corrects scroll position when restoring shorter drafts with the caret at the end.
* **Tests**
  * Updated textarea autosize/scrolling test coverage and expectations to better match real composer behavior.

### DIFF
--- a/src/__tests__/renderer/components/InputArea/hooks/useInputAreaAutosize.test.tsx
+++ b/src/__tests__/renderer/components/InputArea/hooks/useInputAreaAutosize.test.tsx
@@ -3,7 +3,15 @@ import { useRef } from 'react';
 import { describe, expect, it } from 'vitest';
 import { useInputAreaAutosize } from '../../../../../renderer/components/InputArea/hooks/useInputAreaAutosize';
 
-function Harness({ value, selectionEnd = value.length }: { value: string; selectionEnd?: number }) {
+function Harness({
+	value,
+	selectionEnd = value.length,
+	initialScrollTop = 0,
+}: {
+	value: string;
+	selectionEnd?: number;
+	initialScrollTop?: number;
+}) {
 	const ref = useRef<HTMLTextAreaElement>(null);
 	useInputAreaAutosize({ inputRef: ref, inputValue: value, activeTabId: 'tab-1' });
 
@@ -11,8 +19,14 @@ function Harness({ value, selectionEnd = value.length }: { value: string; select
 		<textarea
 			ref={(el) => {
 				if (!el) return;
+				// The real composer textarea is controlled, so its DOM value tracks
+				// inputValue. defaultValue is uncontrolled and would keep the previous
+				// render's value across a rerender, so mirror the current value here -
+				// scrollTextareaToCaretEnd reads textarea.value.length.
+				el.value = value;
 				Object.defineProperty(el, 'scrollHeight', { value: 200, configurable: true });
 				Object.defineProperty(el, 'selectionEnd', { value: selectionEnd, configurable: true });
+				el.scrollTop = initialScrollTop;
 				ref.current = el;
 			}}
 			defaultValue={value}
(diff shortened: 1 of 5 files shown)


pedramamini:
Thanks for the contribution @chr1syy, and for the detailed write-up plus the fix plan.

Unifying the two max-height constants is clearly the right call. The composer's own CSS caps at `maxHeight: '11rem'` (176px) in `src/renderer/components/InputArea/components/InputTextarea.tsx:201`, so the old `EXTERNAL_TEXTAREA_MAX_HEIGHT = 112` was the odd one out. I traced it back to `c854ad188`: the pre-refactor monolith already carried this split (112 in the external sync effect at `InputArea.tsx:434`, 176 in the keystroke path at line 1029), so this is a genuine long-standing inconsistency rather than intentional design. CI is green on both matrix legs, no merge conflicts, and the dead `prevInputValueRef` removal is clean.

One thing I would like corrected before merge. It is about the claim rather than the behavior:

### The mid-text caret preservation does not actually happen in the flows named, and the new test masks that

`scrollTextareaToCaretEnd` branches on `textarea.selectionEnd >= textarea.value.length`. In both deferred-caret flows the new comment cites, caret placement is deferred to a `requestAnimationFrame`:

- `src/renderer/components/InputArea/overlays/AtMentionPopover.tsx` `acceptItem`: `setInputValue(accept.value)`, then `requestAnimationFrame(() => { el.selectionStart = el.selectionEnd = accept.caretPos })`
- `src/renderer/hooks/input/useTemplateAutocomplete.ts` `selectVariable`: `onChange(newValue)`, then `requestAnimationFrame(() => textareaRef.current.setSelectionRange(newCursorPos, newCursorPos))`

`useInputAreaAutosize`'s effect runs in the commit phase, so it runs *before* that rAF. At that moment the caret is not mid-text. Assigning `.value` on a textarea moves the text entry cursor to the end of the control per spec, and React does not restore a mid-text selection when the same element stays focused (the familiar "controlled input caret jumps to the end" behavior).

I verified this with a throwaway probe rather than reasoning from memory: a focused controlled textarea with the caret parked at offset 5, value replaced programmatically, and the commit-phase effect observes `selectionEnd = 55` against `value.length = 55`. So the branch taken is always scroll-to-bottom, and the rAF then moves the caret mid-text with no follow-up scroll.

The new `leaves the scroll position untouched when the caret is mid-text` test passes only because the harness pins `selectionEnd` via `Object.defineProperty` before the value is applied. The real DOM will not hand you that state.

**On severity:** I do not believe this is a regression. The old `shouldScrollTextareaToEnd(55, 32, 55)` also returned `true` through its `caretWasAtEnd` arm, so runtime behavior on these paths is unchanged, and the 176px fix stands on its own merit. My concern is comment and test fidelity: someone will later read "leave mid-text carets (deferred @-mention placement) untouched" and build on a guarantee that is not there.

### Suggested resolution, either is fine

1. **Cheapest, and my preference for this PR:** soften the code comment in `useInputAreaAutosize.ts` and the corresponding PR bullet to state what the helper actually does (scroll only when the caret sits at the end of the value) without claiming the deferred @-mention flow benefits. Reframe the mid-text test as covering the helper's contract rather than that flow.
2. **Actually deliver it:** have the deferred caret setters re-run the scroll (or scroll the caret into view) after they place the caret, so mid-text placement is honored end to end. Larger scope, and very reasonable as a follow-up.

Everything else looks good to me. Ping me once the comment and test framing are updated and I will re-review promptly.

chr1syy:
Thanks for the careful trace, @pedramamini - you're right, and I confirmed it against the two setters: `AtMentionPopover.acceptItem` (`setInputValue` then caret in a `requestAnimationFrame`) and `useTemplateAutocomplete.selectVariable` (`onChange` then `setSelectionRange` in a rAF). Since `useInputAreaAutosize`'s effect runs in the commit phase, it observes `selectionEnd === value.length` at that point, so `scrollTextareaToCaretEnd` always takes the scroll-to-end branch on those paths - the mid-text guarantee was never delivered there, and you're also right the test only "passed" because the harness pins `selectionEnd` via `Object.defineProperty`.

I went with your preferred option 1 (comment + test fidelity, no behavior change, since as you noted this isn't a regression) in `adf1a4f53`:

- `useInputAreaAutosize.ts`: reworded the comment to state the helper's real contract (snap to bottom only when the caret is at the end of the value, no-op otherwise) and added a caveat that the deferred @-mention / template setters place their caret in a later rAF, so this commit-phase effect can't honor those mid-text placements - flagging that honoring them end to end would need those setters to re-scroll (your option 2, reasonable as a follow-up).
- The mid-text test is renamed to `leaves the scroll position untouched when the caret is before the end of the value` and its comment now says it exercises the helper's contract directly, explicitly noting the harness pins `selectionEnd` and that the real deferred flows don't reach the helper in that state.
- Updated the PR description bullet + CodeRabbit release-note line to match.

The 176px unification and the dead `prevInputValueRef` removal are unchanged. Scoped tests still 9/9, prettier/eslint clean. Ready for re-review whenever you have a moment.

Issue #1140, opened 2026-06-28
Bug: Auto-scroll jumps output to bottom on re-render of long content


felipeggv:
## Summary
Auto-scroll jumps output to bottom on re-render of long content

## Environment
- Maestro version: 0.18.3-RC
- Operating system: macOS (Darwin Kernel Version 25.4.0: Thu Mar 19 19:26:07 PDT 2026; root:xnu-12377.101.15~1/RELEASE_ARM64_T6031, 25.4.0)
- Install source: Packaged build (release build or locally packaged)
- Agent/provider involved: Not provided
- SSH remote execution: Not provided

## Steps to Reproduce
1. Have the AI produce a long response, ideally containing code blocks / syntax-highlighted content. 2. After generation finishes, scroll up to read an earlier section. 3. Observe the view auto-scroll back down to the bottom on its own, with no input — the jump coincides with content/text-box re-rendering (e.g. code blocks rendering in different language formats).

## Expected Behavior
When the user has scrolled up to read earlier output, the view should stay put. Auto-scroll to the bottom should only occur when the user is already at the bottom of the pane, and should never re-trigger on a content re-render after generation has finished.

## Actual Behavior
The AI output pane forcibly scrolls back down to the last line on its own, with no user input. It happens during streaming but is noticeably worse after the AI has finished generating (static content). Reading a section above the bottom (e.g. section 3) gets yanked down to the latest line (e.g. section 6), making it hard to read long responses.

## Additional Context
User reports the jump appears triggered by content refresh/re-render of text boxes that render different language formats (code blocks). The behavior is worse on already-finished (static) content than during live streaming, suggesting a scroll handler that re-anchors to the bottom on every content update rather than only when the user is pinned to the bottom.

## Screenshots / Recordings
None

## Support Package
[maestro-debug-package.zip](https://raw.githubusercontent.com/felipeggv/maestro-feedback-attachments/main/feedback/1782642195416-maestro-debug-2026-06-28T10-23-14.zip)

pedramamini:
Thanks for this one, @felipeggv - your diagnosis in the "Additional Context" section was spot on, and it's what pointed straight at the root cause.

**This is fixed on `rc`.** Your instinct that it was "a scroll handler that re-anchors to the bottom on every content update rather than only when the user is pinned to the bottom" was exactly right.

### What was wrong

The auto-scroll gate in `useTerminalOutputScroll` was:

```ts
[...]

PR #1295, opened 2026-07-25
Snap transcript to bottom on re-focus when following

chr1syy:
## Finding J1 - transcript does not restore to bottom on re-focus

On session/tab re-focus the transcript restored a stale absolute pixel offset instead of snapping to the live bottom, even when the user was following the bottom when they left.

## Change
- The mount-time restore effect now consults the persisted `isAtBottom` flag: it latches `hasRestoredScrollRef` first, then early-returns when `initialIsAtBottom !== false`, so a followed transcript snaps to the live bottom on re-focus.
- A deliberate scroll-up (`isAtBottom === false`) still restores its saved offset. Legacy tabs (`undefined`) are treated as at-bottom.
- The `initialIsAtBottom` prop is threaded types.ts -> MainPanelContent -> TerminalOutput -> hook.
- Stick-to-bottom-during-streaming (#1263 / #1233 / #1140) is preserved and covered by new gated tests.

## Validation
Type check, ESLint, Prettier clean. Scoped test `useTerminalOutputScroll.test.ts` -> 6/6 pass.

Fix plan: `.maestro/FIX-J1-01.md`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed terminal output scroll restoration when switching tabs by using the tab’s initial “at bottom” state to decide whether to restore the saved offset or snap to live output.
  * Prevented unwanted scroll jumps in cases where the user had paused following after a restore.
  * Ensured the terminal continues auto-following the bottom as new entries arrive when appropriate.
* **Tests**
  * Added deterministic coverage for mount-time restore and auto-follow behavior using controlled animation-frame flushing.

### DIFF
--- a/src/__tests__/renderer/components/TerminalOutput/useTerminalOutputScroll.test.ts
+++ b/src/__tests__/renderer/components/TerminalOutput/useTerminalOutputScroll.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect } from 'vitest';
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 import { renderHook, act } from '@testing-library/react';
 import { useTerminalOutputScroll } from '../../../../renderer/components/TerminalOutput/hooks/useTerminalOutputScroll';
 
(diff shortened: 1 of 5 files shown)


pedramamini:
@chr1syy Thanks for the contribution - the J1 diagnosis is spot on, and gating the mount-time restore on the persisted `isAtBottom` flag is the right fix. Nice touch threading it as an optional prop so legacy tabs degrade to "was at bottom", and the four gated tests read well.

One thing needs a change before this can go in, though. Both bots signed off, but I think they missed it.

### The restore effect can re-fire later with a stale offset

Adding `initialIsAtBottom` to the dependency array of the restore effect creates a new trigger, and the comment's justification for why that is safe does not hold:

```
// hasRestoredScrollRef is already latched above so a later prop
// change cannot re-trigger this. (J1)
```

The latch is undone in the same commit. The reset effect is declared *after* the restore effect:

```ts
}, [initialScrollTop, initialIsAtBottom, scrollContainerRef]);

useEffect(() => {
    hasRestoredScrollRef.current = false;
}, [sessionId, activeTabId]);
```

React runs effects in declaration order, so on mount and on every tab switch the restore effect latches the ref to `true` and then the reset effect immediately flips it back to `false`. The ref is therefore `false` whenever the user starts interacting with a freshly focused tab.

That matters because of how the two persisted values are written in `useScrollLogHandlers.ts`. `onAtBottomChange` writes `tab.isAtBottom` **synchronously** on scroll, while `onScrollPositionChange` is debounced 200ms inside the hook. So when a user scrolls up, `initialIsAtBottom` flips to `false` while `initialScrollTop` is still the **old** saved offset. The restore effect re-runs, sees an unlatched ref, and yanks them to that stale position.

### Repro

Probe run against your branch (`f8058dc`):

```ts
// mount: followed tab, stale saved offset 8000, live bottom 9000
{ initialScrollTop: 8000, initialIsAtBottom: true }
flushRaf();
expect(ref.current.scrollTop).toBe(9000); // passes - your fix works

// user scrolls up to 2000; isAtBottom persists immediately, scrollTop is still debounced
ref.current.scrollTop = 2000;
rerender({ atBottom: false });
flushRaf();

expect(ref.current.scrollTop).toBe(2000);
// AssertionError: expected 8000 to be 2000
```

The user gets thrown from 2000 back to 8000 mid-scroll. I confirmed this is new to this PR and not pre-existing: on `rc` the deps are `[initialScrollTop, scrollContainerRef]`, so an `isAtBottom` flip on its own could never re-run the effect, and any `initialScrollTop` change that did re-run it always carried a fresh value.

In the app the exposure window is the gap before the first debounced scroll save re-latches the ref, but it is a genuine window: a wheel or trackpad scroll-up in the first ~200ms after focusing a tab, or any case where the mount bottom-jump does not move `scrollTop` and so never emits a scroll event to trigger a save at all.

### Suggested fix

Move the reset effect *above* the restore effect so the latch survives the commit:

```ts
useEffect(() => {
    hasRestoredScrollRef.current = false;
}, [sessionId, activeTabId]);

useEffect(() => {
    if (initialScrollTop !== undefined && initialScrollTop > 0 && !hasRestoredScrollRef.current) {
        hasRestoredScrollRef.current = true;
        ...
```

This makes the existing comment true, and it also closes the pre-existing (benign today) re-entry on `initialScrollTop` changes. I applied it locally and ran the full `src/__tests__/renderer/components/TerminalOutput/` directory: 18/18 pass, including all four of your new gated cases and the #1263 stick-to-bottom coverage. Worth adding a regression test for the flip-with-stale-offset case too, since it is exactly the seam this PR introduces.

### Nit (non-blocking, pre-existing)

The `initialScrollTop > 0` guard means a tab parked at the very top (`scrollTop === 0`, `isAtBottom === false`) still snaps to the bottom on re-focus, so the "a deliberate scroll-up still restores its saved offset" claim in the description has that one hole. It predates this PR, so I am fine either way - just flagging it since it is adjacent to what you are fixing.

No merge conflicts and CI is green across both matrix legs. Ping me once the reorder is in and I will take another pass.

chr1syy:
@pedramamini Great catch, and thank you for the precise repro - you're exactly right that the latch was undone in the same commit by declaration order, so the comment was lying about being safe.

Fixed in `deead380e`:

- Moved the reset effect **above** the restore effect so React runs reset (`false`) then restore (`true`) in declaration order, leaving the ref latched after commit. Rewrote the comment to explain the ordering dependency instead of just asserting safety.
- Added **Case 5** as a regression test for exactly the seam this PR introduces: mount following with a stale saved offset (8000 vs live 9000), then flip `initialIsAtBottom` to `false` with `initialScrollTop` still stale while the user sits at 2000. It reproduces your `expected 8000 to be 2000` failure on the old order and passes on the reorder.

Scoped run of `src/__tests__/renderer/components/TerminalOutput/useTerminalOutputScroll.test.ts`: 7/7 pass; prettier + eslint clean on both files.

On the non-blocking nit (`initialScrollTop > 0` means a tab parked at the very top snaps to bottom on re-focus): agreed it's a real hole, but as you noted it predates this PR and it's orthogonal to the J1 restore-gate fix, so I'm leaving it out to keep this PR in scope. Happy to file a follow-up for the `scrollTop === 0 && isAtBottom === false` case.

Ready for another pass.

PR #1303, opened 2026-07-25
feat(cue): add generic webhook trigger for external events

### LINKED ISSUE #858: Feature: Add generic webhook support to trigger Cue pipelines from external ev...
## Summary
Add generic webhook support to trigger Cue pipelines from external events

## Environment
- Maestro version: 0.16.9-RC
- Operating system: macOS (Darwin Kernel Version 25.2.0: Tue Nov 18 21:09:40 PST 2025; root:xnu-12377.61.12~1/RELEASE_ARM64_T6000, 25.2.0)

## Desired Outcome
Maestro exposes a generic webhook endpoint that receives events from external services (GitHub, GitLab, Slack, CI systems, etc.) and triggers Cue pipelines with the webhook payload, allowing users to write custom logic for different event types.

## Details
Maestro currently has no webhook support for triggering Cue pipelines from external events.

## Additional Context
Example use cases include: auto-triggering code reviews when PRs are opened, automatically picking up new GitHub issues to work on, responding to CI/CD events. The implementation should be generic/extensible rather than hardcoded for specific services.

## Screenshots / Recordings
None

pedramamini:
closes #858

## Summary

Adds a `webhook.received` Cue event type so any external service that can send an HTTP POST - GitHub, GitLab, Slack, CI systems, ad-hoc scripts - can trigger a Cue pipeline.

Kept deliberately generic as the issue asked: the delivery lands in the event payload and the **existing filter engine** decides which deliveries matter, rather than hardcoding per-vendor handling. One endpoint can drive several pipelines:

```yaml
subscriptions:
  - name: pr-opened
    event: webhook.received
    webhook:
      path: gh-pr
      secret_env: GH_WEBHOOK_SECRET
      signature_header: X-Hub-Signature-256
    filter:
      webhook_event: pull_request   # from X-GitHub-Event
      body.action: opened           # any field in the JSON body
    prompt: |
      A pull request was just opened. Review it.

      {{CUE_WEBHOOK_BODY}}
```

## The listener (`src/main/cue/cue-webhook-server.ts`)

- **One process-wide listener** shared by every webhook subscription across every agent (a port can only be bound once). Refcounted, so the socket only exists while a webhook pipeline is loaded - a user with none never has a listening port.
- **Loopback by default**: binds `127.0.0.1:17997`. `MAESTRO_CUE_WEBHOOK_PORT` / `MAESTRO_CUE_WEBHOOK_HOST` override. Public delivery is expected to go through a tunnel (ngrok, cloudflared) or reverse proxy rather than a public bind.
- **A secret is mandatory.** A webhook path with no auth is a remote trigger for an agent running with the user's credentials, so a subscription without `secret`/`secret_env` fails config validation rather than quietly starting to listen. `secret_env` is preferred and the two are mutually exclusive.
- **Two auth modes**, both timing-safe:
  - presented secret via `X-Maestro-Cue-Secret` or `Authorization: Bearer`
  - HMAC-SHA256 over the raw body via a configured `signature_header`, accepting the `sha256=<hex>` form GitHub and GitLab send - so pointing a GitHub webhook at Maestro needs no glue code.
- **Secrets never reach a payload**: `Authorization`, `Cookie`, `X-Maestro-Cue-Secret`, and the configured signature header are stripped before the event is built, so they can't leak into a prompt, the activity log, or the Cue DB.
- POST only (405 otherwise), 1 MB body cap enforced before authentication, unknown paths 404 without revealing whether a secret would have matched. Subscriptions sharing a path each authenticate independently.

## Wiring

Trigger source, config normalizer + validator, template-context enricher, and five `{{CUE_WEBHOOK_*}}` variables (`BODY`, `EVENT`, `PATH`, `DELIVERY_ID`, `HEADERS`). The body is pretty-printed JSON, since senders minify and a single-line blob is harder for an agent to reason about.

Pipeline editor gets the trigger in the drawer plus a config panel (path / secret env var / signature header) that writes `secret_env`, never a literal. A hand-written literal `secret` still round-trips losslessly so opening and saving a pipeline in the editor can't silently strip a working secret off disk.

## Docs

New event section in `docs/maestro-cue-events.md`, an in-app Cue pattern, and a `CueHelpModal` entry.

## Testing

- New: 17 listener tests (routing, both auth modes, header redaction, fan-out, size cap, non-JSON bodies) and 12 trigger-source tests (secret resolution, path defaulting, payload shape, filter integration, `enabled()` gate, idempotent start/stop).
- New validator + loader coverage for the `webhook` block.
- Updated three pre-existing event-type count assertions (10 -> 11, 8 -> 9 drawer items).
- Full suite green locally: 1401 files, 34682 passing. `npm run lint`, `lint:eslint`, and `prettier --check` clean.

Local run is macOS only - the CI matrix still needs to confirm ubuntu and windows.

## Open question for the maintainers

The listener is a standalone `http.Server` rather than a route on the existing Fastify `WebServer`. That keeps webhooks working without the user enabling remote web access, and keeps the auth model per-subscription instead of per-app-token. If you'd rather have this live under `WebServer` (inheriting its rate limiting and tunnel story), the trigger source and registration API stay unchanged - only `cue-webhook-server.ts` swaps out.

## Summary by CodeRabbit

* **New Features**
  * Added the `webhook.received` Cue event type with a local webhook listener and routing to matching subscriptions.
  * Supports required authentication (shared secret or Bearer) and optional HMAC-SHA256 signature verification.
  * Added webhook-specific prompt variables and editor support (configuration, validation, YAML conversion).
  * Expanded trigger/event catalogs to include the new type.

* **Documentation**
  * Updated Cue docs and help content, including that Cue supports 11 Maestro Cue event types.
  * Added full `webhook.received` documentation with security, delivery behavior, filters, and templates.

* **Tests**
  * Expanded coverage for validation, signature/body verification, payload handling, redaction, and unregister behavior.


pedramamini:
Rebased onto current rc - **clean merge, no conflicts**, despite being 201 commits behind. `tsc` clean on all four configs; 2,188 tests green across `main/cue/` and `renderer/components/CuePipelineEditor/`.

On the two blocking findings: **both are already fixed on this branch**, and I verified each in the code rather than taking the review summary at face value.

**HMAC over transmitted bytes.** `authenticates()` takes `rawBody: Buffer` and hashes it directly at `cue-webhook-server.ts:161`. The handler passes `read.body` (the Buffer) at line 312, *before* the `toString('utf8')` decode at line 326. The reasoning is written into the function's docstring: senders sign the wire bytes, and round-tripping a non-UTF-8 body through `toString('utf8')` substitutes U+FFFD for invalid sequences, which would reject a legitimately signed delivery.

**Editor consolidating distinct webhook configs.** The trigger group key now includes the webhook block (`yamlToPipeline.ts:304-311`), so two `webhook.received` subscriptions on different paths stay separate nodes through an open-and-save cycle. Covered by *"keeps webhook triggers with different paths as separate nodes"*.

This is the largest PR left in the queue (+2,000 / 28 files) and it adds a network listener, so it is the one I'd most want a human pass on rather than landing on bot review alone. @reachrazamair if you have bandwidth after your current queue, this is the one worth your eyes - specifically the auth path in `cue-webhook-server.ts` and the reference-counted listener lifecycle.

Not merging it myself for that reason. Everything else in the rc batch is small enough to land on its own merits.

reachrazamair:
> Rebased onto current rc - **clean merge, no conflicts**, despite being 201 commits behind. `tsc` clean on all four configs; 2,188 tests green across `main/cue/` and `renderer/components/CuePipelineEditor/`.
> 
> On the two blocking findings: **both are already fixed on this branch**, and I verified each in the code rather than taking the review summary at face value.
> 
> **HMAC over transmitted bytes.** `authenticates()` takes `rawBody: Buffer` and hashes it directly at `cue-webhook-server.ts:161`. The handler passes `read.body` (the Buffer) at line 312, _before_ the `toString('utf8')` decode at line 326. The reasoning is written into the function's docstring: senders sign the wire bytes, and round-tripping a non-UTF-8 body through `toString('utf8')` substitutes U+FFFD for invalid sequences, which would reject a legitimately signed delivery.
> 
> **Editor consolidating distinct webhook configs.** The trigger group key now includes the webhook block (`yamlToPipeline.ts:304-311`), so two `webhook.received` subscriptions on different paths stay separate nodes through an open-and-save cycle. Covered by _"keeps webhook triggers with different paths as separate nodes"_.
> 
> This is the largest PR left in the queue (+2,000 / 28 files) and it adds a network listener, so it is the one I'd most want a human pass on rather than landing on bot review alone. @reachrazamair if you have bandwidth after your current queue, this is the one worth your eyes - specifically the auth path in `cue-webhook-server.ts` and the reference-counted listener lifecycle.
> 
> Not merging it myself for that reason. Everything else in the rc batch is small enough to land on its own merits.

Got it, I'll put this on my queue right after finishing up others

PR #1304, opened 2026-07-25
feat(autorun): schedule an Auto Run to start at a specific date/time

### LINKED ISSUE #716: Feature: Schedule auto-run documents to execute at a specific date and time
## Summary
Schedule auto-run documents to execute at a specific date and time

## Environment
- Maestro version: 0.16.6-RC
- Operating system: macOS (Darwin Kernel Version 25.4.0: Thu Mar 19 19:33:50 PDT 2026; root:xnu-12377.101.15~1/RELEASE_ARM64_T6050, 25.4.0)

## Desired Outcome
Users can schedule an auto-run document to execute at a specific date and time instead of immediately. The Auto Run panel includes a scheduling option (defaulting to 'now'/immediate) where users can set a future date and time. The auto-run fires once at the scheduled time.

## Details
Auto-run documents can only be executed immediately — there is no option to schedule them for a future date/time.

## Additional Context
Community feature idea from J.R. (@adsconcierge on Discord). Primary use case: scheduling auto-runs to kick off after token limits reset at a known time. One-shot scheduling only, no recurring schedules needed. UX should live in the existing Auto Run panel as a toggle/setting alongside existing options like the worktree toggle.

## Screenshots / Recordings
None

pedramamini:
Closes #716

## What

Auto Run documents could only be launched immediately. This adds a **Start** control to the Auto Run window, sitting alongside the worktree toggle exactly as the issue suggested:

- **Now** - the default, unchanged behavior.
- **At a set time** - pick a future date and time; the run fires once, then the schedule is gone.

One-shot only. No recurring schedules, per the issue.

The Go button becomes **Schedule** when a time is set. A pending schedule shows as a banner at the top of the Auto Run panel with an X to cancel it, so it isn't invisible once the modal closes.

## How

| Piece | Role |
| --- | --- |
| `scheduledAutoRunStore.ts` | One pending run per agent. Mirrored into the settings KV store, so a schedule survives an app restart - the driving use case ("start this when my token limit resets in 5 hours") routinely outlives a single app session. |
| `useScheduledAutoRunDispatcher.ts` | Mounted once from `App.tsx`. Polls the wall clock every 15s and launches due runs. |
| `ScheduleRunSection.tsx` | The Now / At-a-set-time control plus local-timezone `datetime-local` helpers. |
| `AutoRunScheduledBanner.tsx` | Pending-schedule banner + cancel, in the Auto Run panel. |

`handleStartBatchRun` parks a config carrying a future `scheduledFor` instead of launching it, and now accepts an explicit target session so the dispatcher can fire runs for agents the user has navigated away from.

## Decisions worth a second opinion

- **Polling, not `setTimeout`.** A multi-hour timer silently under-fires when the machine sleeps. A 15s tick against the wall clock is correct across sleep/wake and across a hot reload.
- **6h grace window on a missed schedule.** If Maestro was closed at the fire time, the run still starts when you reopen within 6h; past that it's dropped with a notification rather than kicking off hours late. That 6h matches the existing Cue `time.once` `grace_minutes` default, so both scheduling surfaces behave the same.
- **Busy agent doesn't block scheduling.** The "agent is thinking" and "an Auto Run is already active" gates are skipped while scheduling, since the run starts later; the dispatcher re-checks both at fire time and defers (retrying each tick) until the agent frees up.
- **Worktree prep is deferred too.** Scheduling with a create-new worktree target creates the worktree at fire time, not at schedule time.

## Testing

- `scheduledAutoRunStore.test.ts` - schedule/cancel, persistence, hydration incl. malformed entries, and the due/expired timing policy.
- `useScheduledAutoRunDispatcher.test.tsx` - parks vs. launches, fires on time, defers while busy or mid-run, drops past-grace and orphaned schedules, and confirms `scheduledFor` never reaches the runner.
- `ScheduleRunSection.test.ts` - local-timezone round-tripping of the picker value (a `toISOString()` here would shift the run by the UTC offset).
- Full suite green locally: 1401 files, 34673 tests. `npm run lint` and eslint clean.

Windows CI leg still needs to confirm - nothing here touches paths, but the local run is a single OS.

## Not included

Issue #716 also drew a comment asking for a watched folder where new documents are auto-discovered and executed. That's a different feature and Maestro Cue already covers that shape, so it's out of scope here.

## Summary by CodeRabbit

* **New Features**
  * Added one-shot Auto Run scheduling for a future date and time.
  * Added “Now” and “At a set time” options with local date/time validation.
  * Scheduled runs persist across restarts and can be cancelled from the Auto Run panel.
  * Runs automatically start when due, while waiting if the agent is busy.

* **Documentation**
  * Added scheduling guidance to the Auto Runner help modal and store documentation.

* **Tests**
  * Added coverage for scheduling, persistence, timing, cancellation, and dispatch behavior.

### DIFF
--- a/docs/agent-guides/STATE-PATTERNS.md
+++ b/docs/agent-guides/STATE-PATTERNS.md
@@ -20,20 +20,21 @@ All stores are in `src/renderer/stores/`.
 
 ## Store Inventory
 
-| Store                 | File                   | Hook                   | Purpose                                                                                           |
-| --------------------- | ---------------------- | ---------------------- | ------------------------------------------------------------------------------------------------- |
-| **sessionStore**      | `sessionStore.ts`      | `useSessionStore`      | Sessions, groups, active session, bookmarks, worktree tracking, initialization                    |
-| **uiStore**           | `uiStore.ts`           | `useUIStore`           | UI layout: sidebars, focus, notifications, search, drag-and-drop, editing                         |
-| **tabStore**          | `tabStore.ts`          | `useTabStore`          | Tab operations (CRUD, navigation, metadata), gist state. Wraps tabHelpers.ts + sessionStore       |
-| **agentStore**        | `agentStore.ts`        | `useAgentStore`        | Agent detection cache, error recovery, queue processing, agent lifecycle                          |
-| **modalStore**        | `modalStore.ts`        | `useModalStore`        | Modal visibility via registry pattern. Single Map replaces 90+ boolean fields                     |
-| **groupChatStore**    | `groupChatStore.ts`    | `useGroupChatStore`    | Group chat state: chats list, messages, moderator, participants, execution queue                  |
-| **settingsStore**     | `settingsStore.ts`     | `useSettingsStore`     | App settings (theme, font, shortcuts, agent configs, per-modal `modalSizes`, etc.)                |
-| **fileExplorerStore** | `fileExplorerStore.ts` | `useFileExplorerStore` | File explorer panel state                                                                         |
-| **sidebarNavStore**   | `sidebarNavStore.ts`   | `useSidebarNavStore`   | Left Bar sort/nav/starred projections (`SidebarNavSync` host writes; SessionList + keyboard read) |
-| **batchStore**        | `batchStore.ts`        | `useBatchStore`        | Batch/Auto Run execution state                                                                    |
-| **notificationStore** | `notificationStore.ts` | `useNotificationStore` | In-app notification queue                                                                         |
-| **operationStore**    | `operationStore.ts`    | `useOperationStore`    | Long-running operation tracking                                                                   |
+| Store                     | File                       | Hook                       | Purpose                                                                                                                                                                             |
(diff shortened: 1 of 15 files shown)


pedramamini:
Closing this. Scheduling an Auto Run for a specific date/time is a reasonable feature, but two of the findings are data-loss shaped and the branch has drifted too far to be worth repairing in place.

Greptile scored it **3/5**:

- A due schedule is **consumed before the launch is accepted**, so when deferred-launch validation fails the schedule is permanently gone. The user set a run for 6am, it silently never happens, and there is nothing left to inspect.
- The stored config resolves the Auto Run folder at launch time rather than capturing it with the schedule, so changing an agent's folder between scheduling and firing runs the agent against the wrong documents.

Plus +1,192 across 15 files, conflicting, **174 commits behind rc**.

Worth noting there is now overlap with Cue, which already does time-triggered dispatch with persistence and a visible activity log. A fresh take should probably ask whether this is a thin Cue-backed scheduler rather than a second parallel scheduling system with its own 15-second poll. Branch is kept either way.

PR #1306, opened 2026-07-25
feat(worktrees): run a post-create setup script in new worktrees

### LINKED ISSUE #409: [feature] new worktree creation setup script
Hi, my use case is to run some OS specific script after a new worktree is created. Is there anyway way to do it ?

Some other example use cases :
- after creating new worktree, copy .env.local files to the newly created worktree 
- after creating new worktree , run setup.sh to generate linux specific env files in the new worktree

pedramamini:
Closes #409

## Problem

A fresh worktree only contains what git tracks. Anything gitignored - `.env.local`, generated config, installed dependencies - is missing until you recreate it by hand, every single time. Issue #409 asked for a hook to run an OS-specific script after worktree creation, and the workaround people landed on was a shell script that wraps `git worktree add` entirely, which means giving up Maestro's worktree UI.

## What this adds

A per-agent **Setup Script** in the Worktree Configuration modal. It runs inside each worktree Maestro creates, with the new worktree as its working directory.

Available to the script:

| Variable | Value |
| --- | --- |
| `MAESTRO_WORKTREE_PATH` | Absolute path of the new worktree (also the cwd) |
| `MAESTRO_WORKTREE_BRANCH` | Branch checked out in the new worktree |
| `MAESTRO_MAIN_REPO_PATH` | Absolute path of the main repository |
| `MAESTRO_BASE_BRANCH` | Branch the new branch was based on, when specified |

So the two use cases from the issue become:

```bash
cp "$MAESTRO_MAIN_REPO_PATH/.env.local" .
cp "$MAESTRO_MAIN_REPO_PATH/.env.local" . && ./scripts/setup.sh
```

OS-specific logic stays in a checked-in script that the field points at, so it lives with the repo rather than in app config.

## Implementation

- `git:worktreeRunSetup` IPC (`src/main/utils/worktree-setup-script.ts`) runs the command through `cmd.exe /d /s /c` on Windows and the login shell elsewhere, with a 10 minute cap and stdout/stderr truncated to a 4k tail.
- Local runs inherit the login-shell PATH (same treatment `git:createPR` gives `gh`), so GUI-launched Electron on macOS doesn't lose `node`/`npm`.
- SSH remote agents run the script on the remote host via a new exported `execShellRemote` in `remote-git.ts`; the private `execRemoteShellCommand` now delegates to it.
- Wired into all four creation paths - config modal, create-worktree modal, Auto Run spawn (`worktreeSpawn.ts`), batch runner (`useWorktreeManager`) - through one renderer helper, gated on `result.created` so reused or re-attached worktrees never re-run it.
- Failures raise a toast and are swallowed: the worktree and its agent are already usable, so a broken script must not abort the spawn flow.
- `execFileNoThrow`'s `ExecOptions` form now accepts `env` alongside `timeout` (previously mutually exclusive), which the runner needs.

## Design notes

The script is stored per parent agent rather than read from a file in the repo. Auto-executing a checked-in script on `git worktree add` would mean any cloned repo could run code the moment you make a worktree; requiring the user to type the command keeps that surface closed while still letting the command be `./scripts/setup.sh`.

## Notes for review

`useWorktreeHandlers.ts` shows ~85 lines of reindentation: shortening the `handleSaveWorktreeConfig` parameter type let Prettier collapse the `useCallback(` wrapper onto one line. `git diff -w` on that file is 27 insertions / 8 deletions.

## Testing

- `src/__tests__/main/utils/worktree-setup-script.test.ts` - 13 tests: no-op paths, env exposure, local vs SSH dispatch, exit-code/timeout reporting, output truncation, PATH-probe failure
- `src/__tests__/renderer/utils/worktreeSetupScript.test.ts` - 7 tests: script resolution by session and by repo path, child-agent exclusion, toast surfaces, IPC rejection handling
- Full suite green locally: 31,674 passed / 108 skipped across 1,109 files
- `npm run lint` (all three configs) and `prettier --check` clean on the changed files

## Summary by CodeRabbit

* **New Features**
  * Added optional setup scripts for newly created Git worktrees.
  * Scripts run locally or over SSH before the worktree agent starts.
  * Added environment variables, platform-specific shell support, and execution output reporting.
  * Added a 10-minute timeout; failures show a notification without blocking worktree creation.
  * Blank scripts disable setup, and reused worktrees skip execution.

* **Documentation**
  * Documented setup scripts, configuration, environment variables, and Git integration APIs.

* **Tests**
  * Added coverage for local and remote execution, error handling, configuration, and worktree creation flows.


pedramamini:
@reachrazamair your call. **Note this is the only one of the batch with a merge conflict**, though it is a trivial one.

### Rebase status

`git merge-tree` shows exactly one conflicting file: `src/renderer/components/AppModals/AppWorktreeModals.tsx`. Your side changes 2 lines; main added 37 lines to the same file since the branch point. Genuinely a few minutes of work, not a real rebase problem.

### The finding I would not wave through

`src/main/utils/execFile.ts:108` - the `ExecOptions` vs legacy env-object discriminator can misclassify. It decides which overload it got by probing for keys, so a plain env object that happens to contain a variable literally named `env` gets treated as `ExecOptions`. Then `execOpts.env` is a string, gets passed as the child `env`, and process spawning breaks in a way that is very hard to trace back to here.

Worth knowing: **#1317 tripped over this same signature independently**, where `execFileNoThrow(command, args, undefined, env)` silently gets no timeout because the env object lacks an `input`/`timeout` key. Two PRs hitting the same overloaded function from different directions is a decent argument that the signature itself should be split into two explicitly-named functions rather than discriminated at runtime. Might be worth pulling that out as its own cleanup.

### Setup script owner is ambiguous

`worktreeSetupScript.ts:46` - when multiple parent agents share a repository with different setup scripts, the lookup takes the first matching session rather than the agent that started the batch. The batch path can then run another agent's setup command in the new worktree. Plausible in exactly the multi-agent-per-repo workflow Maestro is built for.

### Style and docs

- Em-dashes in comments at `worktreeSpawn.ts:115`, `useWorktreeManager.ts:269`, and `useWorktreeHandlers.ts:379, 383, 509`. CLAUDE.md forbids these outright.
- `docs/git-worktrees.md:123` says Windows runs the command through `cmd.exe` but then tells the user to point the field at `./scripts/setup.sh`, which `cmd.exe` cannot execute. Should recommend `.cmd`/`.bat` or explicitly invoke Bash.

**My take:** rebase (5 min), fix the em-dashes and the docs, resolve the setup-script owner lookup. The `execFile` discriminator I would rather fix properly as a standalone change touching both this and #1317.

reachrazamair:
> @reachrazamair your call. **Note this is the only one of the batch with a merge conflict**, though it is a trivial one.
> 
> ### Rebase status
> `git merge-tree` shows exactly one conflicting file: `src/renderer/components/AppModals/AppWorktreeModals.tsx`. Your side changes 2 lines; main added 37 lines to the same file since the branch point. Genuinely a few minutes of work, not a real rebase problem.
> 
> ### The finding I would not wave through
> `src/main/utils/execFile.ts:108` - the `ExecOptions` vs legacy env-object discriminator can misclassify. It decides which overload it got by probing for keys, so a plain env object that happens to contain a variable literally named `env` gets treated as `ExecOptions`. Then `execOpts.env` is a string, gets passed as the child `env`, and process spawning breaks in a way that is very hard to trace back to here.
> 
> Worth knowing: **#1317 tripped over this same signature independently**, where `execFileNoThrow(command, args, undefined, env)` silently gets no timeout because the env object lacks an `input`/`timeout` key. Two PRs hitting the same overloaded function from different directions is a decent argument that the signature itself should be split into two explicitly-named functions rather than discriminated at runtime. Might be worth pulling that out as its own cleanup.
> 
> ### Setup script owner is ambiguous
> `worktreeSetupScript.ts:46` - when multiple parent agents share a repository with different setup scripts, the lookup takes the first matching session rather than the agent that started the batch. The batch path can then run another agent's setup command in the new worktree. Plausible in exactly the multi-agent-per-repo workflow Maestro is built for.
> 
> ### Style and docs
> * Em-dashes in comments at `worktreeSpawn.ts:115`, `useWorktreeManager.ts:269`, and `useWorktreeHandlers.ts:379, 383, 509`. CLAUDE.md forbids these outright.
> * `docs/git-worktrees.md:123` says Windows runs the command through `cmd.exe` but then tells the user to point the field at `./scripts/setup.sh`, which `cmd.exe` cannot execute. Should recommend `.cmd`/`.bat` or explicitly invoke Bash.
> 
> **My take:** rebase (5 min), fix the em-dashes and the docs, resolve the setup-script owner lookup. The `execFile` discriminator I would rather fix properly as a standalone change touching both this and #1317.

Makes sense. I will rebase and fix the docs, em-dashes, and setup script once I finish the queue. Leaving execFile for a separate PR as discussed.

reachrazamair (changes requested):
Agreed on your take, will do the fixes after I am through the rest of the PR queue.

pedramamini:
Rebased and fixed. The conflicts were 16 lines, not the wall they looked like.

**Conflicts:** main added `useEffect`/`useState` to `AppWorktreeModals` while this branch added a `SessionWorktreeConfig` import (kept both), and main re-indented the `gitSubdirs` loop (took main's).

**Setup-script owner.** The fallback took the FIRST session matching the repo path. Several agents on one repo is a normal Maestro setup, and `useWorktreeManager` is the one caller that cannot pass its own session - `setupWorktree(sessionCwd, worktree)` has no session in scope - so the batch path could run a *different* agent's shell command in this worktree.

Threading a session through would have changed that signature and its callers, so the fix is at the source: auto-resolution now only happens when the candidates agree on the script, and warns instead of guessing when they disagree. Arbitrary shell commands are not a coin flip worth taking. Callers that know their own session pass it directly and never reach the fallback.

**Docs.** The Windows section said the command runs through `cmd.exe` and then, two lines later, told the user to point the field at `./scripts/setup.sh` - which `cmd.exe` cannot execute. Now names the per-platform spelling and shows the explicit-interpreter form (`bash scripts/setup.sh`) for a single cross-platform script.

**Em-dashes** stripped; main gained an eslint rule for those after this branch was cut.

On `execFile.ts`: leaving the overload discriminator alone here as we agreed. Worth noting #1317 merged with a partial fix (it now also probes for `env`), so a standalone cleanup splitting it into two named functions is still the right follow-up.

`tsc` clean on all four configs, 2,186 tests green.

PR #1307, opened 2026-07-25
fix(opencode): offer a fresh session when a provider rejects stored history

### LINKED ISSUE #307: Error " Request contains an invalid argument"
I'm using opencode with gemini, And from time to time I get this error in a session.

`**Request contains an invalid argument**`

When this occurs, I'm typically unable to recover the session, and any prompt results in the same error, The quickest fix is just to start a new session.   This behavior is the same whether in Maestor, or directly interacting with the session in Opencode.  

**Cause**
There are two situations where I have noticed that this can occur.  It doesn't always happen, but I've noticed it several times in these situations.
- If I interrupt processing in Maestro by hitting the stop button
- If I rerun a prompt that has a file reference (@filename)

**Troubleshooting**
- Compact: I've tried compacting context from Open Code CLI ,that failed with the same error.
- Export/Import: I've tried to export the session from the open code CLI and re import it, I just get the same session with the same error
- Export/New Session: If I export the session then create a new session and ask it read the exported session Log, that sort of works but if the session log is very long then it doesn't process the whole thing and have to do some selective prompting to get it back up to date.

**Thoughts**
I'm fairly new at using these tools so I'm not sure what other troubleshooting steps I can try to diagnose or recover the broken session.  Of course it would be great to find a way to fix these sessions and be able to implement it from the Maestro UI.  Also thinking some additional session management tools could be useful like deleting a session that is no longer needed or corrupted.

Here is snippet from the session where this error is occurring:
  ```
    "info": {
        "id": "msg_c2efb2d46001bn4AER2F7qpWxl",
        "sessionID": "ses_3e2e5f6ffffeGquo6vrS5CRJP6",
        "role": "user",
        "time": {
          "created": 1770314739014
        },
        "summary": {
          "title": "Availability check and plan formatting",
          "diffs": []
        },
        "agent": "plan",
        "model": {
          "providerID": "google",
          "modelID": "gemini-3-flash-preview"
        }
      },
      "parts": [
        {
          "id": "prt_c2efb2d47001SGNsFfudFvRdbx",
          "sessionID": "ses_3e2e5f6ffffeGquo6vrS5CRJP6",
          "messageID": "msg_c2efb2d46001bn4AER2F7qpWxl",
          "type": "text",
          "text": "\"hi, are you available\n\n---\n\nIMPORTANT: You are in read-only/plan mode. Do NOT write a plan file. Instead, return your plan directly to the user in beautiful markdown formatting.\"\n"
        }
      ]
    },
    {
      "info": {
        "id": "msg_c2efb2f2c002KifZ5bZwi1hpfb",
        "sessionID": "ses_3e2e5f6ffffeGquo6vrS5CRJP6",
        "role": "assistant",
        "time": {
          "created": 1770314739500,
          "completed": 1770314777099
        },
        "error": {
          "name": "APIError",
          "data": {
            "message": "Request contains an invalid argument.",
            "statusCode": 400,
            "isRetryable": false,
            "responseHeaders": {
              "alt-svc": "h3=\":443\"; ma=2592000,h3-29=\":443\"; ma=2592000",
              "content-length": "127",
              "content-type": "text/event-stream",
              "date": "Thu, 05 Feb 2026 18:06:16 GMT",
              "server": "ESF",
              "server-timing": "gfet4t7; dur=35009",
              "vary": "Origin, X-Origin, Referer",
              "x-cloudaicompanion-trace-id": "41f3084346a94eae",
              "x-content-type-options": "nosniff",
              "x-frame-options": "SAMEORIGIN",
              "x-xss-protection": "0"
            },
            "responseBody": "{\n  \"error\": {\n    \"code\": 400,\n    \"message\": \"Request contains an invalid argument.\",\n    \"status\": \"INVALID_ARGUMENT\"\n  }\n}\n",
            "metadata": {
              "url": "https://generativelanguage.googleapis.com/v1beta/models/gemini-3-flash-preview:streamGenerateContent?alt=sse"

```

### PR DESCRIPTION
**Pull request by pedramamini**
closes #307

## Problem

With OpenCode + Gemini, a session can reach a state where every prompt fails with a bare 400:

```
{ "error": { "code": 400, "message": "Request contains an invalid argument.", "status": "INVALID_ARGUMENT" } }
```

Once it happens the session never recovers. That part is not ours to fix: the reporter confirmed the same session fails identically in the OpenCode CLI with Maestro out of the loop, so the malformed transcript is already in OpenCode's session storage. OpenCode replays the whole stored conversation on every turn, so one bad entry (typically a tool call left without its matching result when a turn is aborted mid-stream) poisons every subsequent request.

What *was* ours to fix is the dead end Maestro presented:

- `matchErrorPattern` had no OpenCode pattern for this message, so it fell through to `type: 'unknown'` and the modal showed the raw provider string.
- `useAgentErrorRecovery` has no `session_not_found` case at all, so it hit the `default` branch and offered a single **Try Again** action. Retrying replays the same broken history and fails the same way. The reporter's actual workaround was to notice this and hand-start a new session.

## Change

- `src/main/parsers/error-patterns.ts`: classify the `Request contains an invalid argument` / `INVALID_ARGUMENT` family as `session_not_found` for OpenCode, with a message that names the cause ("the stored session history is unusable - start a new session to continue"). `session_not_found` is already in `NON_RETRYABLE_TYPES`, so automation won't burn quota re-poking a dead session either.
- `src/renderer/hooks/agent/useAgentErrorRecovery.tsx`: add the missing `session_not_found` case, leading with **Start New Session** (a fresh AI tab in the same agent, workspace and settings intact). This also fixes the pre-existing gap for the `session not found` / `invalid session` patterns that Claude Code and OpenCode already emit - those got the same futile retry button.

## Tests

- `error-patterns.test.ts`: new `session_not_found` block covering the bare Gemini message, the full response body with `"status": "INVALID_ARGUMENT"`, the existing `session not found` path, and false-positive guards ("the function takes three arguments" etc. stay unmatched).
- `useAgentErrorRecovery.test.ts`: asserts `session_not_found` yields exactly one primary `new-session` action and never calls `onRetry`.

Verified locally: `vitest run src/__tests__/main/parsers/ src/__tests__/renderer/hooks/useAgentErrorRecovery.test.ts src/__tests__/renderer/hooks/useModalHandlers.test.ts src/__tests__/shared/retryClassification.test.ts` (537 passed), plus `npm run lint`, eslint and prettier clean on the changed files.

## Not addressed here

Prevention. The most likely trigger is the interrupt path: on the CLI transport the stop button sends SIGINT and escalates to a kill 2s later, which can cut OpenCode off before it finalizes the aborted turn. The Encore-gated OpenCode serve transport (#1171) already aborts through `session.abort()` instead, which should be gentler - but confirming that as the cause needs a repro rather than a guess, so it stays out of this PR.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved detection of malformed or unavailable replayed sessions from Gemini/OpenCode error responses.
  - Users are now prompted to start a new session when recovery is possible.
  - Prevented unrelated invalid-argument errors, such as invalid model IDs, from being misclassified as missing sessions.

- **Tests**
  - Added coverage for provider-specific, JSON, and plain-text session errors.
  - Added safeguards against false-positive error matches.

### CODE REVIEW

**Comment by pedramamini** [comment]
@reachrazamair your call. Branch updated against main; red `test` job was the stale-base localStorage mock, not this diff.

Small PR (4 files, +108/-0) and the intent is right, but one finding lands in a failure mode this codebase has been bitten by before.

### The regex is too broad, and we have a scar from exactly this

`error-patterns.ts:387`:
```
/request contains an invalid argument|\bINVALID_ARGUMENT\b/i
```
The second alternative matches **every** OpenCode `INVALID_ARGUMENT` response - including request-level failures from an invalid model option, a malformed tool payload, a bad prompt field, or a config error. All of those get classified as "your stored session is unusable," which clears the provider session and pushes the user into fresh-session recovery. Starting a new session does not fix an invalid model id; it just loses the conversation and then fails again the same way.

This is the same shape as the broad `LIMIT_REGEX` false-positive that took Claude TUI dead-in-the-water: a status-code alternative matching far more than the specific condition it was written for. Worth being conservative here on principle.

CodeRabbit's narrowing looks right to me:
```diff
-			pattern: /request contains an invalid argument|\bINVALID_ARGUMENT\b/i,
+			pattern: /\brequest contains an invalid argument\b/i,
```
It also notes the JSON test fixture happens to contain the message text too, so the test passes through the first alternative and never actually exercises the standalone-status branch. A negative test for an unrelated `INVALID_ARGUMENT` would lock this down.

### The recovery branch may be unreachable

`useAgentErrorRecovery.tsx:171` - Greptile's read is that the normal `session_not_found` flow clears the tab error and explicitly skips opening the agent-error modal, so the modal branch that advertises "Start New Session" never actually runs. The hook test covers the branch in isolation, which is why it looks green. If that is accurate, the user-facing recovery this PR promises does not fire and the branch is dead code that will drift.

Worth verifying by hand against the real flow rather than trusting the bot, since it is the difference between "ship it" and "ships a no-op."

**My take:** narrow the regex, add the negative test, and confirm the modal branch is actually reachable. Then it is a clean merge - it is a genuinely useful recovery path if it fires.

reachrazamair (changes requested):
The regex needs narrowing. Right now it can catch errors that have nothing to do with a broken session, like a bad model setting, and wrongly tells the user to start over when that won't actually fix anything. Simple fix, already tested.

The new button never runs, a different part of the app skips the popup for this error type. Not urgent, but I'll pull it out since the existing recovery card already handles this case on its own.

reachrazamair (approved):

pedramamini:
Rebased onto current main and removed the dead branch.

The `session_not_found` arm in `getRecoveryActionsForError` never ran: `useAgentErrorListener` suppresses the agent-error modal for that type, and `SessionRecoveryCard` handles it inline instead. Its test passed because it called the function directly - which is exactly how dead code survives review.

The regex narrowing is the substance and it stands: a bare `INVALID_ARGUMENT` no longer classifies as an unusable session, so a bad model id or malformed tool payload stops pushing users into fresh-session recovery that cannot help them.

`tsc` clean on all configs, 133 tests green.

PR #1308, opened 2026-07-25
feat(opencode): per-agent OpenCode agent selection (--agent)

### LINKED ISSUE #284: Ability to utilize existing agents on Opencode (oh-my-opencode support)
One example is oh-my-opencode, I'd like to leverage the features of Maestro like easy multi-tasking, worktrees, group chat etc. But also I do not want to create an external/brand new agent as oh-my-opencode already creates a bunch of agents with its own systems.

pedramamini:
Closes #284

## Problem

Maestro spawns OpenCode with `opencode run ...` and never passes `--agent`, so every Maestro agent runs OpenCode's default primary agent. There was no way to pin a Maestro agent to a specific OpenCode agent (`build`, `plan`, or the ones plugins like [oh-my-opencode](https://github.com/code-yeongyu/oh-my-opencode) register), which is what #284 asks for: reuse the agents OpenCode already defines instead of recreating personas as Maestro nudge instructions.

## What this adds

An **OpenCode Agent** field in the agent configuration panel (New Agent, Edit Agent, Wizard, Group Chat, Encore). Setting it runs that Maestro agent as `opencode run --agent <name>`, so it keeps the OpenCode agent's persona, model, and instructions.

- Free text, not a dropdown, on purpose: plugin-contributed agents are resolvable at run time but do not show up in `opencode agent list` (per the OpenCode bug linked in the issue thread), so a discovery-driven picker would hide exactly the agents this issue is about.
- Per-agent, not per-provider: the value is stored in that agent's **Custom Arguments** (`session.customArgs`), which is per-agent, while agent config options are shared by every agent on the provider. Two-way bound, so `--agent foo` typed directly into Custom Arguments shows up in the field and vice versa.
- Because it lives in customArgs, it flows through every existing spawn path (desktop, CLI, Cue, group chat, SSH) with no new plumbing.

The workflow from the issue thread now works end to end: create "Project X: Prometheus" and "Project X: Sisyphus", point each at its OpenCode agent, and group chat across them.

## Bug this exposed

Config options and custom args are appended *after* `buildAgentArgs()` has already emitted `readOnlyArgs`. For OpenCode, plan mode is `--agent plan`, so a user's `--agent prometheus` landed after it and silently won: read-only mode was defeated with no warning. This is reachable today by anyone who puts `--agent` in Custom Arguments.

`applyAgentConfigOverrides()` now accepts `readOnlyMode` and drops any config-option or custom arg that repeats a flag the agent's `readOnlyArgs` pin (both `--flag value` and `--flag=value` spellings). Callers thread their existing read-only state through. Nothing else changes: no other agent's config option or `readOnlyArgs` overlap today, so this only bites the conflict case.

## Testing

- New `src/__tests__/shared/opencodeAgentArg.test.ts` (read/write/round-trip, quoting, duplicates, removal).
- New read-only pinning cases in `agent-args.test.ts`.
- New OpenCode field cases in `AgentConfigPanel.test.tsx`.
- Full suite green locally: 31682 passed, 1107 files.
- `npm run lint` and `npm run lint:eslint` clean.

## Notes for review

- Per-message agent switching (a picker above the input, like the model/effort pills) is deliberately out of scope here: that needs a new per-session/per-tab field threaded through ~30 files. This PR gets the per-agent binding working first. Happy to follow up if you want the pill.
- Tab naming and group chat moderator spawns declare `readOnlyMode: true`, so they now ignore a configured `--agent` rather than running the user's persona agent for a title. That seemed clearly right, flagging it in case you disagree.

## Summary by CodeRabbit

- **New Features**
  - Added an optional **OpenCode Agent** field for selecting a named OpenCode agent directly from the configuration panel.
  - Preserves other custom arguments when adding, updating, or removing the selected agent.

- **Bug Fixes**
  - Read-only mode now consistently enforces pinned settings across launches, group chats, tab naming, and session recovery.
  - Conflicting custom arguments are automatically ignored in read-only mode.

- **Documentation**
  - Documented OpenCode agent selection, including plan-mode behavior and plugin-provided agents.


pedramamini:
@reachrazamair your call. Branch updated against main; red `test` job was the stale-base localStorage mock, not this diff.

Feature itself (per-agent OpenCode `--agent` selection, #284) is clean and the plumbing threads through all the spawn surfaces correctly - CLI spawner, group chat router, tab naming, context groomer. Two findings I would want closed first, one of which is a silent-corruption class bug.

### Boolean flags eat the following argument

`agent-args.ts:98` - when stripping a repeated boolean switch for read-only Codex or Gemini runs (`--skip-git-repo-check`, `-y`), the strip also removes the **next non-flag token**. If that token belongs to an unrelated custom argument, it silently vanishes from the command line. The user gets a subtly different invocation than the one they configured, with no error and nothing in the log to point at.

This is the kind of bug that surfaces three weeks later as "why did my custom args stop working" and takes an afternoon to trace. Boolean flags need to be in a known set that the stripper consults, rather than inferred from whether the next token looks like a flag.

### Read-only Cue steps do not get plan mode

`agent-args.ts:235` - the Cue override call omits `readOnlyMode`, so a configured `--agent <name>` survives after `--agent plan` is applied and defeats plan mode. A read-only Cue step then runs with write capability. Given Cue steps fire unattended, a read-only step that is not actually read-only is worth treating as a correctness bug rather than a polish item.

### Smaller items

- `opencodeAgentArg.ts:38` - `quoteIfNeeded()` only quotes on whitespace, so an agent name containing `"` or `'` emits unescaped and breaks tokenization (`my"agent` becomes two tokens). Free-text field, so it should either quote-and-escape or sanitize.
- `opencodeAgentArg.ts:14` - module header claims the helpers rewrite only the `--agent` token and leave the rest untouched, but `writeOpenCodeAgentArg()` rebuilds via `kept.join(' ')`, which normalizes whitespace. Comment should match behavior.
- `AgentConfigPanel.tsx:689` - the "OpenCode Agent (optional)" label is not programmatically associated with its input. Needs `aria-label` or `id`/`htmlFor`.

**My take:** fix the boolean-flag stripping and the Cue `readOnlyMode` pass-through, and this is good to go. The other three are quick.

reachrazamair (approved):
Fixed the boolean-flag bug, stripFlags now checks whether a pinned flag actually takes a value by looking at the read-only args themselves, instead of guessing from what the next token looks like. Added two tests for it, all 73 pass plus everything else across the related suites.

On the Cue item, checked the whole Cue codebase by hand, it never runs in read-only mode today, so that one isn't a live bug right now, just worth keeping in mind if that changes later.

The three small ones (quote handling, a stale comment, missing aria-label) are still open, left them out since they're not blocking this pr

pedramamini:
@reachrazamair these two are the only PRs in the queue with no response from anyone - not flagging them as urgent, just that they fell through rather than being deliberately parked.

Both still have my findings from the earlier pass standing unaddressed:

- **#1308** (OpenCode `--agent`): the boolean-flag stripper eats the following token, and the Cue override path omits `readOnlyMode`, so a read-only Cue step runs with write capability.
- **#1317** (binary path chooser): the chooser persists the wrong path (blur reads stale closure state), and the differently-named wrapper case that #1048 was actually filed about is never detected.

Happy to take either one if you'd rather stay on your current queue - just say which and I'll pick it up. Otherwise they're yours whenever you get there.

reachrazamair:
> @reachrazamair these two are the only PRs in the queue with no response from anyone - not flagging them as urgent, just that they fell through rather than being deliberately parked.
> 
> Both still have my findings from the earlier pass standing unaddressed:
> 
> * **[feat(opencode): per-agent OpenCode agent selection (--agent) #1308](https://github.com/RunMaestro/Maestro/pull/1308)** (OpenCode `--agent`): the boolean-flag stripper eats the following token, and the Cue override path omits `readOnlyMode`, so a read-only Cue step runs with write capability.
> * **[feat(agents): chooser for multiple detected binary paths #1317](https://github.com/RunMaestro/Maestro/pull/1317)** (binary path chooser): the chooser persists the wrong path (blur reads stale closure state), and the differently-named wrapper case that [Persist and choose among detected Codex provider paths #1048](https://github.com/RunMaestro/Maestro/issues/1048) was actually filed about is never detected.
> 
> Happy to take either one if you'd rather stay on your current queue - just say which and I'll pick it up. Otherwise they're yours whenever you get there.

Thanks, @pedramamini! I'm taking care of both of these

reachrazamair:
> > @reachrazamair these two are the only PRs in the queue with no response from anyone - not flagging them as urgent, just that they fell through rather than being deliberately parked.
> > Both still have my findings from the earlier pass standing unaddressed:
> > 
> > * **[feat(opencode): per-agent OpenCode agent selection (--agent) #1308](https://github.com/RunMaestro/Maestro/pull/1308)** (OpenCode `--agent`): the boolean-flag stripper eats the following token, and the Cue override path omits `readOnlyMode`, so a read-only Cue step runs with write capability.
> > * **[feat(agents): chooser for multiple detected binary paths #1317](https://github.com/RunMaestro/Maestro/pull/1317)** (binary path chooser): the chooser persists the wrong path (blur reads stale closure state), and the differently-named wrapper case that [Persist and choose among detected Codex provider paths #1048](https://github.com/RunMaestro/Maestro/issues/1048) was actually filed about is never detected.
> > 
> > Happy to take either one if you'd rather stay on your current queue - just say which and I'll pick it up. Otherwise they're yours whenever you get there.
> 
> Thanks, @pedramamini! I'm taking care of both of these

Both fixed. Boolean-flag stripping held up on a second look, nothing to change there.

readOnlyMode for Cue was missing in more spots than the override step you flagged, it also never reached the arg builder or the env-override step, so even after the first pass flipping it on wouldn't have actually enforced plan mode. Wired all three up and added tests that fail without the change, confirmed that part.

Matches what you said would make this good to go. Re-approving.

pedramamini:
Rebased onto current main. Two of my earlier findings are already fixed on this branch, and one was wrong.

**Fixed by @reachrazamair:** the boolean-flag stripper now classifies value-taking flags from `readOnlyArgs` instead of assuming every flag consumes the next token, so a repeated boolean no longer eats an unrelated argument.

**My finding was wrong:** I flagged the Cue override path for omitting `readOnlyMode`. Cue spawns with `yoloMode: true` unconditionally and has no read-only concept at all (nothing in `cue-types.ts` exposes one), so there is no read-only Cue step to defeat. Withdrawn.

**Fixed here:** `quoteIfNeeded` only quoted on whitespace, so an agent name containing a quote emitted raw and split into two tokens on the next read - the trailing fragment silently becoming a separate CLI argument. `tokenize()` has no escape syntax, so no spelling of `my"agent` can round-trip; quote characters are stripped instead. An OpenCode agent name is an identifier, so that costs nothing real and keeps the arg list well-formed. Also wired the label to its input via `htmlFor`/`id` and corrected the module header, which claimed the rewrite leaves everything untouched when it re-joins tokens with single spaces.

`tsc` clean on all four configs, 1,341 tests green.

PR #1309, opened 2026-07-25
feat(worktree): merge and rebase actions in the worktree context menu

### LINKED ISSUE #263: Feature Request: Add auto commit and merge and rebase buttons.
Since the worktree is in a separate directory, I want it to be automatically git commit when the task is completed, and I want to provide a merge button so that I can merge the worktree's results into the main branch.

There's also a rebase button that rebases new code from the main branch to the worktree branch.

When creating a worktree and merging branches, I wish I could also choose which main branch to create the branch from and which branch to merge into.

When manually committing and merging, I want a pop-up window to allow me to confirm the message content and provide options to use a cheaper model to generate the commit message, like DeepSeek.

pedramamini:
Closes #263

## What

Worktree sub-agents could already create a PR or be removed, but there was no way to land their work without dropping to a terminal. This adds two entries to the right-click menu on a worktree agent:

- **Merge Branch Into...** - merge this worktree's branch into a branch you pick
- **Rebase Branch Onto...** - replay this worktree's branch on top of a branch you pick, to pull in new upstream work

Both open a confirmation modal with a branch picker that defaults to the repo's default branch. This follows the direction in the issue thread: "Adding these as options on the right-click menu for worktrees does sound like a nice idea."

## How

Two new SSH-aware IPC handlers in `src/main/ipc/handlers/git.ts`:

**`git:mergeBranch`** runs in whichever worktree currently has the target branch checked out, located via `git worktree list --porcelain`. That indirection is necessary: git will not check the same branch out twice, so the merge cannot just happen inside the source worktree. It refuses to run when that checkout has uncommitted changes, and on conflict it runs `git merge --abort` and returns the conflicting paths. A single button press should never leave someone's main checkout stuck half-merged.

**`git:rebaseBranch`** runs in place in the worktree - the base branch only needs to exist as a ref, no checkout required. Same clean-tree requirement, same abort-on-conflict behavior.

Because neither operation runs against a dirty tree, `WorktreeMergeModal` detects uncommitted changes in the worktree and offers to commit them first with an editable message, reusing the existing `git:commitAll` handler. That covers the issue's ask for "a pop-up window to allow me to confirm the message content".

Branch names get a leading-dash check before reaching git. `execGit` already uses `execFile` (no shell), so this only stops a name being parsed as a flag rather than as a ref.

## Not in this PR

Two parts of the original request are deliberately left out, and I'd rather confirm the shape before building them:

- **Auto-commit when a task completes.** This needs a decision about where it hooks in (agent-idle transition? Auto Run iteration boundary?) and whether it is opt-in per agent. Worth its own issue.
- **Generating commit messages with a cheaper model (e.g. DeepSeek).** Maestro has no separate "utility model" configuration today, so this means new settings surface rather than a small addition. The modal's message field is hand-editable in the meantime.

Choosing the base branch when *creating* a worktree already works (`CreateWorktreeModal` has a Base Branch picker); this PR adds the matching choice on the merge/rebase side.

## Testing

- 11 new handler tests in `src/__tests__/main/ipc/handlers/git.test.ts` covering: merging in the target's worktree, already-up-to-date, refusing a dirty target, abort-and-report on conflict, target branch not checked out anywhere, self-merge, and flag-like branch names - plus the rebase equivalents.
- 7 new component tests in `src/__tests__/renderer/components/WorktreeMergeModal.test.tsx` covering default-branch preselection, excluding the worktree's own branch from the picker, merge and rebase dispatch, commit-first with the confirmed message, not merging when that commit fails, and conflict/error rendering.
- Full suite: 34,665 passed, 108 skipped. `npm run lint`, `npm run lint:eslint`, and `prettier --check .` all clean.

Note: validated on macOS only. Needs both CI matrix legs green before merge.

## Summary by CodeRabbit

* **New Features**
  * Added options to merge a worktree branch into another branch or rebase it onto a selected branch.
  * Added a guided modal for selecting branches, committing pending changes, and reviewing operation results.
  * Added conflict reporting with affected file paths and clear success, error, and up-to-date messages.
  * Added support for performing merge and rebase operations on remote worktrees.
* **Bug Fixes**
  * Prevented operations on uncommitted worktrees, invalid branch names, missing branches, and self-merges.

### DIFF
--- a/src/__tests__/main/ipc/handlers/git.test.ts
+++ b/src/__tests__/main/ipc/handlers/git.test.ts
@@ -197,6 +197,8 @@ describe('Git IPC handlers', () => {
 				'git:isRepo',
 				'git:init',
 				'git:commitAll',
+				'git:mergeBranch',
+				'git:rebaseBranch',
 				'git:numstat',
 				'git:branch',
 				'git:remote',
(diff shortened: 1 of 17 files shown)


pedramamini:
Closing this. Not because the feature is wrong - merge and rebase actions in the worktree context menu are wanted - but because the carrying cost has passed the redo cost.

Where it stands: +1,251 across 17 files, conflicting, **174 commits behind rc**, and Greptile scored it **2/5**, the lowest of the 26 open PRs in this queue. The three findings are architectural rather than cosmetic:

- SSH operations lose the remote working directory, so merge/rebase run against the wrong checkout on any remote agent.
- A failed merge leaves the target checkout sitting in an unfinished merge state, which the user then has to unwind by hand.
- The modal offers remote-only refs that the merge handler cannot actually use.

Fixing those means reworking the SSH path and the failure handling, and then rebasing 174 commits of drift on top - which is most of the way to rewriting it against a `git.ts` and handler layer that have since been decomposed (#1367, #1352, #1355 all landed after this branch).

The branch is not deleted, so nothing is lost. The right next step is a fresh implementation against current rc, with `resolveGitCwd()` / `resolveGitSshRemoteId()` from `useGitAgentActions` handling the SSH resolution rather than re-deriving it, and an explicit abort path when a merge fails.

PR #1310, opened 2026-07-26
fix(sentry): stop URIError blanking markdown and spawn failures breaking tab naming

pedramamini:
Eighth triage pass on the `rc` channel. Three field issues where the bug and the fix were both unambiguous, each verified against current `rc` HEAD rather than the shipped build.

## MAESTRO-XS - `URIError: URI malformed` (6 occ, 0.18.4-RC, renderer)

`remarkFileLinks`' link visitor called bare `decodeURIComponent(href)` on the link target. Any stray `%` in an href throws `URIError` - `[see](100% done.md)`, or a Windows / non-ASCII path (the reporting user is on Windows in a Korean locale). Because the plugin runs inside the unified transform, the throw escapes react-markdown's render and **blanks the entire message body**, so this is a visible crash, not telemetry noise.

The Sentry stack is decisive: it walks `root > list > listItem > blockquote > paragraph > strong > link` and bottoms out in `decodeURIComponent`, which pins it to this exact call.

Four sibling call sites (`markdownLinkParser`, `LocalImage`, `MarkdownImage`, `markdownItAdapter`) **already** guarded this; the remark plugin was the lone outlier. Rather than add a fifth inline copy, the best existing guard was promoted to a canonical `safeDecodeURIComponent()` in `src/shared/stringUtils.ts` - it swallows only `URIError` and rethrows anything else, so a real fault still surfaces.

## MAESTRO-X4 - `Error: spawn EFTYPE` (26 occ, 0.18.5-RC, main)

Tab naming spawns an ephemeral agent purely to label a tab. Every other failure path in that function already bails to `null` (timeout, non-zero exit, extraction failure) - the spawn itself was the one unguarded step.

`child_process.spawn` throws **synchronously** for a binary that isn't executable. The call sits in the naming `Promise`'s executor, and the enclosing `try/catch` **`return`s** that promise rather than awaiting it, so its `catch` never saw the throw (the classic `return`-inside-`try` gotcha). The rejection escaped as a hard IPC failure and leaked the `data`/`exit` listeners, all for a cosmetic feature.

Now caught inside the executor and resolved to `null` like its neighbors. Sentry is skipped for the errnos that mean "the configured binary is unusable on this machine" (`ENOENT`/`EFTYPE`/`EACCES`/`EPERM`/`ENOEXEC`), following the existing `isExpectedFsError` / `isExpectedSpawnFailure` convention; anything else still reports.

## Drive-by: a self-recursive helper

`AttachmentImage.tsx`'s `safeDecodeURIComponent` called **itself** instead of `decodeURIComponent`. It blew the stack on every invocation and its own `catch` swallowed the `RangeError`, so the function silently returned its input **never decoded** - Auto Run attachment images with `%20` in the name could not resolve, and every call burned a full stack depth. It never showed up in Sentry precisely because the catch ate it. Replaced with the canonical helper; `markdownItAdapter`'s private copy was folded in too.

## Tests

14 new cases across three suites. Each was confirmed **red** against the unfixed source (`git stash push` on the source files only, re-run, confirm failure) to prove they aren't vacuous:

- `safeDecodeURIComponent`: well-formed escapes, the four malformed shapes, and a termination check guarding against the self-recursive implementation.
- `remarkFileLinks`: four malformed-target shapes render without throwing (3 of these fail without the fix), valid escapes still resolve to `maestro-file://`.
- `tabNaming`: spawn throw resolves to `null` instead of rejecting, the five expected errnos skip Sentry, an unexpected error still reports, and listeners are cleaned up. All 4 fail without the fix.

Verification: `npm run lint` reports **0** TS errors, ESLint clean on all changed files, and 82 test files / 3139 tests pass across the Markdown, AutoRun, and renderer-utils suites.

## Deliberately skipped

- **MAESTRO-V9** (`SyntaxError ... is not valid JSON`, 161 occ) - the carve-out from #1214 is present and correct at `ExitHandler.ts:469`. Its events are on release `0.19.0-RC`, a locally-built version newer than `rc`'s own `package.json` and not present on any remote branch, so the emitting code isn't inspectable. Not re-fixed.
- **MAESTRO-Q2** (`maestro-p --status`, 1980 occ) - still the noisiest, deferred a 5th time; needs a human call on `maestro-p` exit semantics.
- **MAESTRO-XV/XX** (`Rendered more hooks than during the previous render`) - real, but the stack is a minified `Modal-*` chunk with no app frames, the same situation as MAESTRO-XH/XG/XF. Needs a real sourcemap; shipping a defensive guard would hide the symptom without a diagnosis. `XX` is the `[Splash]` twin of `XV` and carries a `.worktrees/maestro-omp-deep-integration/` dev-server path.
- **MAESTRO-QZ/QY/QX** (`Failed to construct 'URL'`) - pure `vendor-react` recursion frames, no app code to act on. `QX` is the `[Splash]` twin.
- **MAESTRO-XW/XC/TX** - OMP / stats-sidecar issues whose stacks are all under `.worktrees/maestro-omp-deep-integration/`, i.e. an unmerged feature worktree, not a shipped build.
- **MAESTRO-J0** (`Failed to load required prompt`) - `group-chat-participant-request.md` is present in `src/prompts/`; broken-install class, same as MAESTRO-RE.
- **MAESTRO-RS** (GLIBC 2.38, 2718 occ) - Linux build/CI glibc target, still needs a human.
- Native/GPU crashes (TM/9K/SZ/4V/QG/HP/X8/XY/MA), `Window unresponsive` (62), and already-fixed-at-HEAD RR/RA/WH/JE.

## Summary by CodeRabbit

- **New Features**
  - Added safer handling for malformed percent-encoded links, preventing markdown content from failing to render.
  - Added a shared utility for decoding URL-encoded text without throwing on invalid escapes.

- **Bug Fixes**
  - Tab naming now handles expected process-launch failures gracefully and avoids unnecessary error reporting.
  - Unexpected tab-naming errors continue to be reported for investigation.

- **Documentation**
  - Documented the new safe URL-decoding utility and its behavior.


pedramamini:
Rebased onto current rc. One conflict, in `src/shared/stringUtils.ts`: rc added `escapeRegExp` at the same insertion point this branch added `safeDecodeURIComponent`. Both kept, no behavior change to either.

`tsc` clean on all configs; 161 tests green across `stringUtils`, `remarkFileLinks`, and `tabNaming`.

One thing worth a separate issue rather than expanding this PR: `src/shared/deep-link-urls.ts` has six unguarded `decodeURIComponent` calls on user-supplied deep-link URLs, which is the same `URIError` class this PR is fixing. Out of scope here, but it's the next one that will page.

pedramamini:
@coderabbitai yes, please open a follow-up issue for the unguarded `decodeURIComponent` calls in `src/shared/deep-link-urls.ts`. Reference this PR and `safeDecodeURIComponent` in `src/shared/stringUtils.ts` as the intended fix.

PR #1311, opened 2026-07-26
fix(sentry): restore guards lost in the main reset and quiet expected poll failures

pedramamini:
Field-crash triage against the `channel:stable` (main) population on release 0.17.3.

## The headline finding: two merged PRs are no longer on `main`

`main` was reset after the accidental rc promotion (see the `ci(guard): block the rc line from reaching main` commits). That reset also took two legitimately-merged main-line PRs with it:

| PR | Merge commit | Reachable from `origin/main`? |
| --- | --- | --- |
| #1213 | `dbc371eb2` | no |
| #1246 | `fab1daae2` | no |

Both still show `MERGED` on GitHub, so the issues they closed looked handled while the code was gone. Confirmed against the source: `sentryFilters.ts` had only the `EPIPE: broken pipe` spelling, `stats.ts` had the `isReady()` guard on `record-shortcut-usage` only, and `isExpectedGroomingFailure` did not exist. All three issues are live again on 0.17.3.

Both PRs are restored here byte-for-byte (`git apply` of their own diffs, plus cherry-pick of the `6d3d22766` regex follow-up) so the eventual rc -> main merge converges without conflict.

## Restored

- **MAESTRO-1G** (29 events/14d) - the filter only matched the libuv fs spelling `EPIPE: broken pipe`; Node emits a bare `write EPIPE` for stream/socket writes to a dead pipe, so the rule never fired.
- **MAESTRO-2S / MAESTRO-2Z** - `stats:record-session-created` / `-closed` threw `Database not initialized` when an agent was restored on boot ahead of the stats DB. Both callers are fire-and-forget and ignore the return value.
- **MAESTRO-V5** - `sessions:setMany` / `setAll` EACCES on `maestro-sessions.json`.
- **MAESTRO-KA** - `isExpectedGroomingFailure()` for the group-chat groom path, including the follow-up that scopes the regex to `Agent ... is not available`.

## New this round

- **MAESTRO-JS** (53 occ) and **MAESTRO-K7** - extend `isExpectedGroomingFailure()` to cover a groom process that would not launch (`context-groomer.ts:393`, reached from `groupChat.ts`) and revoked provider credentials. Both are user-environment conditions that `resetContext` already recovers from by starting a fresh session. Only the group-chat call site suppresses - the user-initiated groom paths in `context.ts` and `director-notes.ts` still surface, as in #1246.
- **MAESTRO-RR** (478 occ - the noisiest app-code issue on stable) - the codex quota sampler carved out 401/403 but still reported `http 503`. The sampler runs on a timer, so a single ChatGPT outage reports once per tick per install. Now skips 408/429/5xx as well; a 4xx implying we sent a malformed request still reports, because that would be our bug.
- **MAESTRO-KE** (70 occ) - `isGitHubConnectivityError` matched socket-level failures but not GitHub-side 5xx, so `HTTP 504: 504 Gateway Timeout (https://api.github.com/graphql)` paged Sentry on every Cue poll tick for the duration of a GitHub outage. The poller already retries on its own schedule.
- **MAESTRO-W9** - the `getGlobalStats` incremental loops report EACCES on transcripts under `~/.claude/projects` that belong to the agent CLI and that we merely discovered on disk. Extends the existing `RangeError` carve-out to expected fs error codes, mirroring `EXPECTED_FS_ERROR_CODES` in `shared-history-manager.ts`.

## Deliberately not touched

- **MAESTRO-RS** (616 events, the largest) - `GLIBC_2.38` for better-sqlite3. Build/CI glibc target, not app code. rc PR #1290 (`build Linux x64 on ubuntu-22.04 and guard the glibc floor`) is the real fix.
- **MAESTRO-BC** (326 events) and its cascade **W2/W7/WA/WB/W6** - `Preload script error` is a corrupt install (`dist\main\preload.js not found in ...app.asar`); `window.maestro` is then undefined and every unguarded call site throws. Optional chaining would mask a broken install across ~40 call sites.
- **MAESTRO-TX / W1 / WP** - tagged `channel:stable` but their release is the Electron version (`41.6.0`) and the paths are `AppData\Roaming\Electron`, i.e. unpackaged dev runs, not the stable release line. TX and W1 are already fixed on rc in #1247.
- **MAESTRO-CT / NM** - `spawn codex ENOENT` with no app frames in the stack. Genuine spawn failures are real signal.
- **MAESTRO-Q2** - deferred for the fifth time; still needs a human call on `maestro-p` exit semantics.
- Native/GPU crashes (SG/1F/QG/4V/9K/5A/TF/RD/VX/23/T2/1J/XQ/MA/WN/SA/62).

## Validation

- 27 regression tests added, **each verified to fail against the unfixed source** (`git stash push` of the source files only, re-run, 27 red, restore, 206 green).
- The pre-existing codex sampler test asserting HTTP 500 *is* reported was intentionally retargeted to 400, since 5xx is now an expected upstream condition.
- `npm run lint`: 112 errors, **identical to clean `main`** (missing `@codemirror/lang-*` optional deps); zero in any changed file.
- Full suite: 59 failing files with these changes vs 60 on clean `main` - **no new failures**.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced unnecessary error and telemetry reporting during expected upstream outages, GitHub connectivity issues, session transcript read failures, and known grooming failures.
  * Improved handling of expected usage-service responses, broken-pipe errors, and permission-related background failures to avoid false alarms.
  * Prevented session statistics updates from running before the statistics database is ready.
* **Tests**
  * Expanded automated coverage for expected-error filtering, connectivity failures, session and grooming classification, telemetry suppression, and database readiness behavior.


pedramamini:
All three review findings were valid and are addressed in b0b35870b.

**CodeRabbit, `cue-github-poller.ts:104` (Major) - confirmed, and it was a real gap in my fix.** `resolveRepo()` runs before the poll and has its own `captureException` at the `cue:github:resolveRepo` tag, so a `gh repo view` 5xx during a GitHub outage still paged Sentry once per tick for every auto-detect trigger. The doPoll-side fix alone left MAESTRO-KE only half covered.

I did **not** take the proposed `if (isGitHubRateLimitError(err) || isGitHubConnectivityError(err)) throw err;` though. Re-throwing routes these into doPoll's outer catch, which drags them through the first-poll seed-marker path - and that path exists to decide whether existing items count as "already seen". Changing event-loss semantics to quiet telemetry is the wrong trade. Skipping just the `captureException` keeps the existing skip-the-poll-and-return-null behavior byte for byte. Added a regression test that drives the `repo: undefined` path and verified it fails without the fix.

**CodeRabbit, `cue-github-poller.test.ts:989` - fair.** The stash-the-source check had already proven the test wasn't vacuous, but the assertion is cheap and guards the future. Added.

**Copilot, `agentSessions.ts:66` - correct.** `EXPECTED_FS_ERROR_CODES` lives in `shared-history-manager.ts` on `rc`, not on `main` (it arrived in rc PR #1141 and reaches main on promotion). The comment would have sent the next reader to a symbol that isn't there. Reworded to point at the sibling `RangeError` carve-out in the same function instead.

pedramamini:
**CI note: the `test` job is red, but it is red on `main` too, with the identical failure set.**

| | `main` @ a277af40a (this PR's base) | this PR |
| --- | --- | --- |
| Failing files | `HistoryPanel.test.tsx`, `DirectorNotes/UnifiedHistoryTab.test.tsx` | same two |
| Failing tests | 51 | 51 |
| Passing files | 1104 | 1104 |

The last three commits on `main` (a277af40a, d2085a008, bbbb0f3a0) all have a failing `CI` run. This PR neither adds to nor fixes that.

The failures look environmental rather than logical: every one is a ~1.1s `waitFor` timeout ending in `TestingLibraryElementError: Unable to find an element with the text: Test summary`, and both files pass locally (107/107). The shared ingredient is the virtualized history list, which renders no rows when the container reports zero height - a jsdom/CI layout difference, not a behavior change.

Flagging rather than fixing, since it is unrelated to this PR's scope. Worth its own issue.

Local validation for the changes here: 207/207 across the six affected suites, `npm run lint` at the pre-existing 112-error baseline with zero errors in any changed file.

PR #1313, opened 2026-07-26
feat(web-desktop): resolve web clients to a live window for window-scoped IPC

user79697:
## What this does

`FAKE_EVENT`, the synthetic `IpcMainInvokeEvent` that web-desktop clients
invoke `ipcMain` handlers through, defines `senderFrame`, `frameId`,
`processId` and `type`, but no `sender`. This adds a lazily-resolved
`sender` pointing at a live window's `webContents`.

## Why (revised, please read)

This PR was opened as a crash fix. That framing was wrong for this base
branch, and I would rather correct it than let it merge under a stale
description.

The crash I originally hit is real, but it is already handled on `rc` by
4765edae1a (2026-07-04), which added to `resolveCallingWindow`:

```ts
// A web client is not a window; resolve to "no window" instead of letting
// BrowserWindow.fromWebContents throw on the missing WebContents.
if (!event?.sender) return undefined;
```

My installed 0.19.0-RC build predates that commit, which is why I saw
`Cannot read properties of undefined (reading 'getOwnerBrowserWindow')`
and the remote page hanging on the splash screen. On current `rc` that
cannot happen.

So what is left is not a crash but a behavioral gap. With no `sender`,
every window-scoped handler resolves to "unknown caller" and silently
degrades for web clients:

- `windows:getState` returns `null`
- claiming a freshly-created agent for the calling window is a no-op
- `setPanelState` and window naming are no-ops

Giving the bridge a `sender` makes those work remotely, on the theory
that a web client is a mirror of the desktop and should act as the window
it mirrors.

**That directly contradicts the intent stated in 4765edae1a** ("a web
client is not a window"), so this is your call, not mine. If the
degradation is deliberate, close this and I will not object. I have left
the guard in `resolveCallingWindow` untouched either way, so nothing here
depends on removing it.

## Implementation notes

`sender` is a getter, not a value: the module is evaluated before any
`BrowserWindow` exists, and the correct answer changes over the app's
lifetime. It still yields `undefined` when no window is open
(headless/tray), so every handler's existing "unknown caller" path stays
intact.

Per review feedback, the comment no longer claims `getAllWindows()` is
creation-ordered, since Electron does not document that. It picks *a*
live window; with a single window open, the common case, that is the main
window. I looked at using `getMainWindow` instead, but it is dependency
injected at every call site (`web-server-factory.ts`) and is not
importable from a module-level const.

## Testing

Three regression tests: sender resolves to a live window, destroyed
windows are skipped, no-windows still yields `undefined`. Verified by
mutation (reverting the getter fails two). Full file: 11/11 passing.

Manually verified against a local web server over loopback with a CDP
probe. Zero failed `windows:getState` responses out of 11,803 successful
`bridge.response` frames, no `getOwnerBrowserWindow` in console, and the
DOM renders the real UI. The only remaining failures are pre-existing
`fs:readDir` EPERM against Windows junction points
(`My Music`/`My Pictures`/`My Videos`), unrelated to this change.

## Review feedback addressed

- Em dashes removed from the production and test comments (flagged by
  CodeRabbit, Greptile, and Codex).
- `getAllWindows()` ordering claim dropped (CodeRabbit).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability for web-initiated window actions by ensuring requests can be associated with an available application window.
  * Prevented errors when the previously active window has been closed or is unavailable.
  * Added safeguards for scenarios where no usable window exists, leaving sender information unset when appropriate.
  * Expanded test coverage to verify sender selection and skipped destroyed windows for bridge invoke handling.

### DIFF
--- a/src/__tests__/main/web-server/handlers/bridgeHandlers.test.ts
+++ b/src/__tests__/main/web-server/handlers/bridgeHandlers.test.ts
@@ -167,11 +167,11 @@ describe('handleBridgeInvoke', () => {
 		expect(callArgs.slice(1)).toEqual(['session-123', 'echo hello']);
 	});
 
-	// Regression: FAKE_EVENT used to omit `sender` entirely. Handlers that
+	// Regression: FAKE_EVENT used to omit `sender` entirely, so handlers that
 	// resolve the calling window via BrowserWindow.fromWebContents(event.sender)
-	// — windows:getState is the one web-desktop calls during boot — then made
-	// Electron throw "Cannot read properties of undefined (reading
-	// 'getOwnerBrowserWindow')", and the remote page hung on the splash forever.
+	// saw no caller. windows:getState, which web-desktop calls during boot, is
+	// the one that matters: it returned null for every web client, leaving the
+	// remote page without the window state it mirrors.
 	it('passes a live window webContents as event.sender', async () => {
 		const webContents = { id: 7 };
 		windows.push({ isDestroyed: () => false, webContents });
--- a/src/__tests__/main/web-server/handlers/bridgeHandlers.test.ts
+++ b/src/__tests__/main/web-server/handlers/bridgeHandlers.test.ts
@@ -21,13 +21,18 @@ type Handler = (event: unknown, ...args: unknown[]) => unknown;
 // `vi.hoisted` runs before the module factory so the map is initialized in
 // time for vi.mock - top-level consts would be in the temporal dead zone
 // at hoist time and the mock factory would throw.
-const { invokeHandlers } = vi.hoisted(() => ({
+const { invokeHandlers, windows } = vi.hoisted(() => ({
 	invokeHandlers: new Map<string, Handler>(),
+	// Stand-ins for live BrowserWindows; FAKE_EVENT.sender resolves through these.
+	windows: [] as { isDestroyed: () => boolean; webContents: unknown }[],
 }));
 
 vi.mock('electron', () => ({
 	ipcMain: {
 		_invokeHandlers: invokeHandlers,
+	},
+	BrowserWindow: {
+		getAllWindows: () => windows,
 	},
 }));
 
@@ -55,6 +60,7 @@ function lastSend(client: ReturnType<typeof makeClient>): Record<string, unknown
 
 beforeEach(() => {
 	invokeHandlers.clear();
+	windows.length = 0;
 });
 
 describe('handleBridgeInvoke', () => {
@@ -159,6 +165,60 @@ describe('handleBridgeInvoke', () => {
 		// First arg is the FAKE_EVENT object, then the user args.
 		const callArgs = spy.mock.calls[0];
 		expect(callArgs.slice(1)).toEqual(['session-123', 'echo hello']);
+	});
+
+	// Regression: FAKE_EVENT used to omit `sender` entirely. Handlers that
+	// resolve the calling window via BrowserWindow.fromWebContents(event.sender)
+	// — windows:getState is the one web-desktop calls during boot — then made
+	// Electron throw "Cannot read properties of undefined (reading
+	// 'getOwnerBrowserWindow')", and the remote page hung on the splash forever.
+	it('passes a live window webContents as event.sender', async () => {
+		const webContents = { id: 7 };
+		windows.push({ isDestroyed: () => false, webContents });
+
+		const spy = vi.fn().mockResolvedValue(null);
+		invokeHandlers.set('windows:getState', spy);
+
+		await handleBridgeInvoke(
+			makeClient(),
+			{ type: 'bridge.invoke', requestId: 1, channel: 'windows:getState' },
+			vi.fn()
+		);
+
+		const event = spy.mock.calls[0][0] as { sender?: unknown };
+		expect(event.sender).toBe(webContents);
+	});
+
+	it('skips destroyed windows when resolving event.sender', async () => {
+		const liveContents = { id: 2 };
+		windows.push({ isDestroyed: () => true, webContents: { id: 1 } });
+		windows.push({ isDestroyed: () => false, webContents: liveContents });
+
+		const spy = vi.fn().mockResolvedValue(null);
+		invokeHandlers.set('windows:getState', spy);
+
+		await handleBridgeInvoke(
+			makeClient(),
+			{ type: 'bridge.invoke', requestId: 1, channel: 'windows:getState' },
+			vi.fn()
+		);
+
+		const event = spy.mock.calls[0][0] as { sender?: unknown };
+		expect(event.sender).toBe(liveContents);
+	});
+
+	it('leaves event.sender undefined when no window is open', async () => {
+		const spy = vi.fn().mockResolvedValue(null);
+		invokeHandlers.set('windows:getState', spy);
+
+		await handleBridgeInvoke(
+			makeClient(),
+			{ type: 'bridge.invoke', requestId: 1, channel: 'windows:getState' },
+			vi.fn()
+		);
+
+		const event = spy.mock.calls[0][0] as { sender?: unknown };
+		expect(event.sender).toBeUndefined();
 	});
 
 	void lastSend; // helper kept available for future tests; tsc would warn if unused.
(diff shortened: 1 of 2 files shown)


user79697:
Pushed b7fbe6564 addressing the review, and I have retitled and rewritten
the description because my original framing turned out to be wrong.

**Review items**

- Em dashes removed from both the production comment and the
  regression-test comment (@coderabbitai, @greptile-apps,
  chatgpt-codex-connector all flagged these).
- Dropped the claim that `getAllWindows()` is creation-ordered.
  @coderabbitai was right that Electron does not document any ordering.
  The comment now says it picks *a* live window rather than asserting it
  picks the main one. I checked whether `getMainWindow` could be used
  instead: it is dependency injected at every call site
  (`web-server-factory.ts:139`) and never exported at module level, so it
  is not reachable from a module-level const like `FAKE_EVENT`.

**The part worth your attention**

I opened this as a fix for
`Cannot read properties of undefined (reading 'getOwnerBrowserWindow')`.
While addressing the review I found that 4765edae1a (2026-07-04) already
guards that path:

```ts
if (!event?.sender) return undefined;
```

That commit is an ancestor of both `rc` and this branch, so the crash I
described cannot occur here. I hit it because the installed 0.19.0-RC
build predates the guard; I confirmed by scanning the shipped `app.asar`
(`resolveCallingWindow` present, guard absent).

That makes this a semantics change rather than a crash fix, and it cuts
against the intent your comment states, that a web client is not a
window. What it buys is that window-scoped handlers stop silently
no-oping for web clients: `windows:getState` returns real state instead
of `null`, and agent claiming, panel state, and window naming start
working from the remote page.

Since that is a product decision about what a web client *is*, it is
yours to make. Happy to close this as superseded if the current
degradation is intentional. I deliberately did not touch the guard in
`resolveCallingWindow`, so this change composes with it rather than
replacing it.

pedramamini:
@user79697 Thanks for the contribution, and genuinely thank you for the follow-up comment. Coming back to say "my original framing was wrong, here is what this actually is, and here is why you might want to close it" is a rare and very welcome thing to see on a PR. You also nailed the review items: em dashes are gone from both comments, the `getAllWindows()` ordering claim is dropped, and I pulled the branch and confirmed 11/11 passing locally. No merge conflicts either.

To answer your direct question first: yes, the gap you found is real. `windows:getState` returning `null` to a web client is a genuine papercut, not a designed behavior, and I do not want to leave it there. But I cannot take this implementation, because giving `sender` to the shared `FAKE_EVENT` reaches a lot further than the four `windows:*` handlers you are aiming at.

## The blast radius

`FAKE_EVENT` is the event object for *every* bridged `ipcMain.handle` channel, so anything that inspects `event.sender` changes behavior. Two of those are trust guards that currently reject web clients only because `sender` is `undefined`:

**1. `plugins:request-consent` (`src/main/index.ts:2209`)**

```ts
// Open the consent window. Only the trusted main renderer may ask.
ipcMain.handle('plugins:request-consent', async (event, pluginId: unknown) => {
	if (event.sender !== mainWindow?.webContents) throw new Error('UntrustedConsentRequester');
```

Today a bridged call has no `sender`, the comparison fails, and the call is rejected. With this getter, in the single-window case (the case you correctly call common) `getAllWindows().find(...)` *is* `mainWindow.webContents`, so the guard passes and any token-holding web client can pop a native consent window for an arbitrary plugin id. This guard is load-bearing: there is no `plugins:setEnabled` channel, consent is the enable path.

The mint itself stays safe, for the record. `plugins:confirm-consent` runs `senderTokenOf(event)` through `consentMinter.confirm`, which checks the token against the consent window's own sender plus the one-time nonce, and a main-window id will not match. So this is not a capability-grant escalation. It is a remote client being able to force a native permission dialog onto the desktop user, through a check whose comment says only the main renderer may ask.

There is a second-order wrinkle here too: the `getAllWindows()` ordering non-guarantee you and CodeRabbit already discussed now sits *inside* a security comparison. With a consent window open, `sender` could resolve to the consent window's `webContents`. Nondeterminism in an equality check is worse than the same nondeterminism in `windows:getState`.

**2. `browser:clearSessionData` (`src/main/ipc/handlers/browser-session.ts:43`)**

```ts
if (event.sender.getType() !== 'window') { ...reject... }
```

Currently that throws a TypeError for bridged calls and the bridge returns it as an error. After this change it passes, so a web client can `clearStorageData()` + `clearCache()` on a tab partition. That one may well be what we want for web-desktop, but the comment right above it says it validates the sender precisely because the op is destructive, so it should be an explicit decision rather than a side effect.

(`fs:startDragOut` and the coworking sender checks are `ipcMain.on`, and the bridge only dispatches `_invokeHandlers`, so those are not affected. I checked.)

## The multi-window mutations

Separately, within `windows:*` itself, two of the four handlers are mutations with invariants stated in their comments:

- `windows:registerSession` - "Resolved from event.sender so a window only ever claims an agent into itself"
- `windows:setPanelState` - "Resolved from event.sender so a window only ever writes its own state"

With one window open, resolving a web client to it is harmless. With two or more, a web client creating an agent claims it into whichever window `getAllWindows()` happens to return first, and a remote panel collapse silently rewrites a desktop window's persisted panel state. Multi-window is the entire reason the registry exists, so that is exactly the configuration where it misbehaves.

The read-only half (`windows:getState`, `windows:getBounds`) does not have this problem. Reads answering for the main window are defensible; writes landing on an arbitrary window are not.

## What I would take instead

Scope the resolution to the handlers that want it, rather than putting it on the shared event.

The cleanest version: leave `FAKE_EVENT` without a `sender` and let `resolveCallingWindow` in `ipc/handlers/windows.ts` opt in. The event already carries `type: 'bridge'`, so that function can recognize a bridge event and resolve the main window itself, replacing the current `if (!event?.sender) return undefined` early-out with a bridge-aware branch. Every other sender guard in the codebase keeps its current behavior, and the change lands in exactly the file whose comment states the policy you are amending.

On reaching the real main window: your point about `getMainWindow` being dependency-injected at `web-server-factory.ts:139` and unreachable from a module-level const is right, but the bridge module already has an injection seam in `installWebContentsBridgeHook`. Threading `getMainWindow` through it would beat `getAllWindows().find()` and would retire the ordering caveat entirely. Either shape works for me.

If you would rather split it, a PR that fixes only the read-only handlers is an easy yes from me, and we can decide the `registerSession` / `setPanelState` semantics separately.

Happy to keep this open while you consider which way you want to go - the investigation here is solid and I would like to land the fix, just with a narrower reach.

pedramamini:
@user79697 Following up on this after a month. Your call to hand the semantics decision back was the right one, and the decision is: **the gap is real and should be fixed, but not by giving `FAKE_EVENT` a `sender`.**

I have implemented the narrower shape from my earlier comment in #1420, so this does not sit open indefinitely.

## What went in

`resolveCallingWindow` takes an `allowBridge` option that only the two **read** handlers pass. `windows:getState` and `windows:getBounds` now answer a bridge caller with the primary window; `registerSession` and `setPanelState` keep degrading, because both document that a window only ever mutates itself.

Two details worth flagging since they came out of your investigation:

1. **The ordering caveat is gone.** Rather than `getAllWindows()[0]`, it resolves the registry's `isMain` entry. You were right that `getMainWindow` is unreachable from a module-level const, but the registry already knows which window is primary, so neither is needed. The test registers a secondary window *first* specifically so an order-based lookup would fail it.

2. **`FAKE_EVENT` staying senderless is what keeps this contained.** I verified both guards from my earlier comment are still closed: `plugins:request-consent` still rejects with `UntrustedConsentRequester`, and `browser:clearSessionData` still rejects on `event.sender.getType()`. The tests also fail if `allowBridge` is extended to the write handlers, so the narrow scope is pinned rather than merely intended.

## Closing this one

Closing as superseded by #1420, not as rejected. You found the gap, characterized it correctly, and then came back to say your own original framing was wrong once you found `4765edae1a` - that last part is genuinely rare and it is why the fix is shaped the way it is. You are credited in the PR body and the commit message.

If you would rather have carried it yourself, say so and I will hand it back - the door is open on the `registerSession` / `setPanelState` semantics too, which stay unresolved by design.

PR #1314, opened 2026-07-26
feat(thought-stream): live plain-language tool-call feed during runs

### LINKED ISSUE #1312: Feature: Show live agent step/tool-call feed during runs (like Claude Code)
## Summary
Show live agent step/tool-call feed during runs (like Claude Code)

## Environment
- Maestro version: 0.18.4-RC
- Operating system: macOS (Darwin Kernel Version 25.3.0: Wed Jan 28 20:53:31 PST 2026; root:xnu-12377.91.3~2/RELEASE_ARM64_T8103, 25.3.0)

## Desired Outcome
Maestro shows a live, streaming feed of agent activity while a run is in progress — each tool call (file read, shell command, file edit) rendered as one short line in plain, easy language rather than raw JSON or verbose logs. Similar to how Claude Code surfaces steps as they happen.

## Details
No live step-by-step visibility into what the agent is doing during a run, so the user cannot follow progress in real time.

## Additional Context
Primary motivation: token cost control. With live steps visible, the user can spot an agent stuck in a loop or an unproductive task and interrupt it early instead of letting it burn tokens. Requested detail level: full tool-call detail (reads, commands, edits) but summarized in easy language and short text — not verbose raw output.

pedramamini:
Closes #1312

## Problem

The Thought Stream captured only `process:thinking-chunk`, so it showed an agent's **reasoning** but never its **actions**. Tool calls did render in the chat transcript, but only when a tab's `showThinking` mode was switched on, and the in-chat listener (`useAgentToolExecutionListener`) matches the streaming id with `REGEX_AI_TAB` alone - so an Auto Run, which spawns as `{sessionId}-batch-{ts}`, had its tool calls dropped entirely.

Net effect for the reporter: no surface anywhere that shows what an agent is actually *doing* while it works.

This is the follow-up explicitly deferred in #1231 / #1238, where the reasoning half shipped and "a count of background shell commands / running tool calls" was called out as separate work.

## What this does

Tool calls now stream into the Thought Stream **interleaved with the reasoning**, each reduced to ONE short plain-language line:

```
3:42:07 PM  ⟳ Ran npm test
3:42:04 PM  ✓ Read src/renderer/components/ThoughtStreamPanel.tsx
3:42:01 PM  ✓ Searched for THOUGHT_BLOCK_GAP_MS
3:41:58 PM  ! Edited src/renderer/constants/themes.ts
```

A spinner marks a call still in flight; a check or a warning marks how it ended. The point is scannability, which is what the reporter's stated motivation needs: **cost control** - spotting an agent looping or grinding on an unproductive task and interrupting it before it burns more tokens.

## Changes

**`src/renderer/utils/toolActivityLabel.ts`** (new)
`describeToolActivity(toolName, input)` normalizes tool names across Claude Code (`Read`/`Bash`/`MultiEdit`), OpenCode (lowercase `read`/`bash`), Codex (`shell`/`apply_patch`/`update_plan`), Copilot (`write_to_file`), and MCP (`mcp__server__tool`) onto plain English. Unknown tools fall back to `Used <name>` rather than being silently dropped, so a new provider tool still shows up. Handles the raw-string input shape (Codex `apply_patch`) that would otherwise be iterated character-by-character, and argv-array commands.

This is deliberately **not** `summarizeToolInput()` - that one builds the verbose in-chat cell (every input key as `key=value`, untruncated command, output preview). Both are documented in `SHARED-UTILS.md` with a note on which to reach for.

**`src/renderer/stores/thoughtStreamStore.ts`**
- `activities` buffer + `appendToolActivity()`. A tool call arrives as two or more events; the completion is merged into its `running` entry so the feed lists **actions, not state transitions**. Matched by `toolCallId` where the provider sends one, else by newest still-running call with the same tool name in the same tab (the rule the in-chat listener already uses for Codex).
- Merged entries keep their **start** timestamp, so a line doesn't jump to the top of the feed when it finishes.
- Capped at `MAX_ACTIVITIES_PER_SESSION` (2000) with the existing `trimmed` flag.
- `buildActivityFeed()` merges thought blocks and tool calls into ONE chronological list, so a tool call appears between the two halves of the reasoning that produced it. Ties break toward the tool (a dispatch instant is when the deciding block stops growing).

**`src/renderer/hooks/agent/internal/useThoughtStreamToolListener.ts`** (new)
Taps `process:tool-execution` and resolves the id with `parseSessionId`, so Auto Run `-batch-` and synopsis spawns are captured too. Same cheap `capturing[sessionId]` early-out as the thinking listener - does nothing unless a panel is open or minimized. No rAF coalescing: tool calls arrive per agent action, not per frame. Normalizes provider status wording (`error` -> `failed`, missing -> still running).

**`ThoughtStreamPanel.tsx`**
Renders tool rows, search matches both the rendered line and the raw tool name, header counts "N thoughts · M actions". Search highlighting is done inline rather than through `<Markdown>` - a shell command or glob pattern is not markdown and would be mangled by it.

**Copy/docs:** Right Panel button becomes **View Activity**; in-app Auto Run help and `docs/autorun-playbooks.md` updated.

## Testing

- `describeToolActivity`: 20 cases across all five provider naming styles, plus the never-throws cases (`''`, `null`, `undefined`, array input).
- `appendToolActivity`: capture gating, merge-by-id, merge-by-newest-running, no cross-tab merge, start-timestamp preservation, cap/trim, parallel-run isolation.
- `buildActivityFeed`: interleaving, tie-break ordering, stable same-timestamp order.
- `useThoughtStreamToolListener`: `-batch-` capture (the case the in-chat listener misses), `-ai-` tab capture, completion merge, status normalization, not-capturing drop, cross-session isolation, unmount cleanup.
- `useAgentListeners` registration counts updated: `onToolExecution` now has two subscribers, exactly mirroring the existing `onThinkingChunk` pattern. Added a note at the test mock that last-registration-wins the captured handler, so a future reorder has a breadcrumb.

Full suite green: **34,720 passed, 108 skipped**. `prettier --check .`, all four `tsc` configs, and `eslint src/` clean.

## Note on the base branch

Based on **`rc`**, not `main`. The Thought Stream does not exist on `main` at all - `rc` is 794 commits ahead and the whole surface this extends is rc-only, so `main` is not a viable base.

## Reviewer notes

- Capture starts when the panel is opened, so the feed covers activity from that point forward - unchanged from how the thinking capture already behaves, but worth confirming that's the right default rather than always-on capture.
- The status-pill entry point from #1238 opens this same panel, so the reporter's screenshot surface gets the feed for free.

## Summary by CodeRabbit

* **New Features**
  * Auto Run’s Thought Stream now displays a unified, chronological activity feed combining agent thoughts and tool calls.
  * Tool calls show plain-language descriptions, targets, timestamps, and running, completed, or failed statuses.
  * Search now covers both thoughts and tool-call activity, including raw tool names.
  * Renamed the activity viewer action to **View Activity** and updated related help text.

* **Documentation**
  * Expanded guidance for using and understanding the Thought Stream and activity-labeling utilities.


pedramamini:
Closing this. The idea is good - a plain-language tool-call feed during runs is worth having - but the chronology is structurally wrong, and the fix is a redesign rather than a patch.

Greptile scored it **3/5** on exactly that: `groupThoughtsIntoBlocks` ignores intervening tool calls, and the feed sorts by assigning each *complete* thought block its **start** time. So a tool call that happened in the middle of a block can never render between the reasoning chunks that surround it. A fast tool call displays *after* the reasoning that actually followed it. For a feature whose entire value is "watch what the agent is doing, in order," getting the order wrong is the whole thing.

Repairing that means changing the grouping model so blocks can be split by interleaved events, which touches most of what this PR adds. On top of that: +1,332 across 15 files, conflicting, **160 commits behind rc**, and `thoughtStreamStore` has moved since.

Branch is kept. A fresh pass should interleave on a single timeline of timestamped events rather than merging two independently-sorted sequences after the fact.

PR #1316, opened 2026-07-27
Clamp window minimum size to the work area so maximize works on small/scaled screens

chr1syy:
## Problem

On a small or heavily-scaled display, the main window cannot be maximized - the native "maximize" control silently does nothing.

**Root cause:** the window is created with a fixed `minWidth: 1000, minHeight: 600` (`src/main/app-lifecycle/window-manager.ts`). Those are logical (DIP) values, but a real display's *work area* (screen minus panel/dock, also in DIPs) can be smaller. A 1366x768 laptop at 125% scale has a work area of only ~1093x593 DIPs. Since the design-min height (600) exceeds the work-area height (593), the window can never shrink to fit, and native maximize (which resizes the window to the work area) has no satisfiable size, so it no-ops.

Reproduced live on Linux/XFCE at `Xft.dpi 120`: the WM's `_NET_WM_ALLOWED_ACTIONS` *does* include maximize, and Maestro sets no `maximizable: false` - the pure size conflict is the whole cause. Setting `Xft.dpi 96` (work area 1366x741 > 600) makes maximize work again, confirming the diagnosis.

## Fix

Add `resolveWindowSizeConstraints()` in `window-position.ts` (mirrors the existing `resolveVisibleWindowPosition` helper, uses `screen`):

- **Relax the minimum to the work area.** `minWidth = min(1000, workArea.width)`, `minHeight = min(600, workArea.height)`, computed against the display the window will actually occupy (`screen.getDisplayMatching`, same rule the position resolver uses, so a window on a small secondary monitor is clamped to that monitor). On any display large enough to hold the design minimum, it is unchanged - **no behavior change for the common case.**
- **Clamp the initial size down to the work area** so a saved (or default) size larger than the current screen does not spawn oversized. Clamped down only, never enlarged - a smaller saved size is preserved and Electron's `minWidth`/`minHeight` supplies the floor exactly as before.
- **Re-clamp on `display-metrics-changed`** so a live rescale, or dragging the window to a smaller monitor, cannot leave a stale minimum that blocks maximize again. The listener is removed on window `closed`.

`workAreaSize` is already in DIPs with the panel subtracted, so this needs no manual DPI math and is correct cross-platform and multi-monitor.

## Not done (deliberately)
- The design minimum stays 1000x600 as the aspiration; it is only relaxed when the hardware forces it. A squished-but-usable window beats an unmaximizable one.
- No switch to a frameless/custom titlebar - out of scope for a sizing bug.

## Tests
Extended `src/__tests__/main/app-lifecycle/window-manager.test.ts` (79 pass, 5 new): design minimum kept on a large display; min height relaxed + size clamped on a 1093x593 scaled display; both minimums relaxed on a tiny display; re-clamp fires `setMinimumSize` on `display-metrics-changed`; and the metrics listener is removed after `closed`. Type-check (main + lint configs), ESLint, and Prettier are clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved window sizing by calculating initial and minimum dimensions from the target display’s work-area constraints.
  * Ensured minimum sizes never exceed available screen space, with reclamping when display metrics change or when the window moves to a smaller display.
  * Added safeguards to avoid redundant minimum-size updates and to stop reclamping after the window is closed.
* **Tests**
  * Expanded Electron/window and screen mocks for deterministic shutdown and metrics-driven scenarios.
  * Added coverage for initial clamping, reclamping on metrics/work-area changes, move behavior, and post-close listener cleanup.

### DIFF
--- a/src/__tests__/main/app-lifecycle/window-manager.test.ts
+++ b/src/__tests__/main/app-lifecycle/window-manager.test.ts
@@ -15,7 +15,7 @@ import type { WindowManagerDependencies } from '../../../main/app-lifecycle/wind
 
 // Track event handlers
 let windowCloseHandler: (() => void) | null = null;
-// The window's `move` listener (drives the cross-display min-size re-clamp).
+// The window's `moved` listener (drives the cross-display min-size re-clamp).
 let windowMoveHandler: (() => void) | null = null;
 // The factory registers several `closed` listeners; collect them all so a test
 // can fire the full close sequence rather than guessing which one is last.
@@ -89,7 +89,12 @@ const mockWindowInstance = {
 	on: vi.fn((event: string, handler: () => void) => {
 		if (event === 'close') windowCloseHandler = handler;
 		if (event === 'closed') windowClosedHandlers.push(handler);
-		if (event === 'move') windowMoveHandler = handler;
+		if (event === 'moved') windowMoveHandler = handler;
+	}),
+	// The reclamp cleanup on 'closed' removes the window's own `moved` listener,
+	// so the mock must expose removeListener to mirror Electron's EventEmitter.
+	removeListener: vi.fn((event: string) => {
+		if (event === 'moved') windowMoveHandler = null;
 	}),
 };
 
@@ -110,6 +115,7 @@ class MockBrowserWindow {
 	getBounds = mockWindowInstance.getBounds;
 	webContents = mockWindowInstance.webContents;
 	on = mockWindowInstance.on;
+	removeListener = mockWindowInstance.removeListener;
 
 	constructor(options: unknown) {
 		lastBrowserWindowOptions = options as Record<string, unknown>;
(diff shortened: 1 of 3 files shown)


pedramamini:
@chr1syy Thanks for this one, and thanks in particular for the writeup. The root-cause analysis is excellent: identifying that the fixed `minHeight: 600` exceeds the ~593 DIP work area on a 1366x768 panel at 125% scale, then confirming it by flipping `Xft.dpi` back to 96 and watching maximize start working again, is exactly the kind of evidence that makes a sizing bug reviewable. Ruling out `_NET_WM_ALLOWED_ACTIONS` and `maximizable: false` up front saved a lot of guessing. The shape of the fix is right too: relaxing the minimum only when the hardware forces it, clamping down but never up, and reusing the same `getDisplayMatching` rule the position resolver already uses.

Before we merge, one thing needs addressing, and I have two smaller observations.

### 1. Cross-display moves keep a stale minimum (agreeing with Greptile's P1)

`display-metrics-changed` only fires when a display's own geometry or scale factor changes. Dragging a window from a large monitor onto an already-connected smaller one fires no such event, so the window keeps the minimum it was created with and maximize keeps no-opping on the destination. That is the same class of bug this PR is fixing, just reached by a different route, and it is a common laptop-plus-external-monitor setup.

`reclampMinimumSize` already does the right thing from `getBounds()`, so this should be small:

```ts
screen.on('display-metrics-changed', reclampMinimumSize);
browserWindow.on('moved', reclampMinimumSize);
```

Two things worth folding in while you are there:

- Guard the actual `setMinimumSize` call on the values having changed. On GTK, `moved` can fire repeatedly through a drag, and re-asserting the same minimum on every tick is churn we do not need.
- Consider calling `reclampMinimumSize()` once immediately after creation. That also covers point 2 below without any extra logic.

A test in the same style as the existing ones (swap `mockScreen.state.displays` for a two-display setup, point `getBounds` at the small one, fire the window's `moved` handler) would round out the coverage.

### 2. Constraints and position can be resolved against different displays

In `createBrowserWindow` the constraints are computed from the raw saved `x`/`y`, but `resolveVisibleWindowPosition` can then reject those coordinates (the Windows `-32000` minimized sentinel, or an unplugged monitor) and re-center the window on the primary display. When that happens the minimum was derived from whichever display `getDisplayMatching` landed on, not the one the window actually opens on. If that display is the larger of the two, the minimum stays at 1000x600 while the window opens on a smaller primary, which is the original bug surviving in a narrow case. The Windows-minimized path is the realistic one here, since `-32000` matches to the leftmost display rather than the primary.

I would not restructure the ordering to fix this. Swapping it would break the centering, which currently relies on the clamped size. The post-creation `reclampMinimumSize()` call from point 1 handles it: once the window exists, `getBounds()` is the truth and the minimum gets recomputed against the display it really landed on.

### 3. Listener accumulation across windows (minor)

`createBrowserWindow` is shared by `createSecondaryWindow`, so every open window adds its own `screen` listener. Cleanup on `closed` is correct and I am not worried about a leak, but past ten concurrent windows Node will start emitting `MaxListenersExceededWarning` on the screen emitter. Not a blocker, and only worth restructuring into a single shared listener if you think that many windows is realistic. A brief note in the comment that this is intentionally per-window would be enough.

Also for visibility: no merge conflicts, and CI is green apart from one Windows shard still running at the time of writing. Point 1 is the one I would like resolved before merge. Points 2 and 3 come along for free if you take the post-creation reclamp. Thanks again.

chr1syy:
Good catch. Fixed in 5c636f66d: the window now re-clamps its minimum size on its own `move` event as well, debounced so a live drag collapses into a single recompute once the window settles on the new display. This covers the cross-display drag case where `display-metrics-changed` never fires. Added a test that drags onto a smaller second display and asserts the minimum relaxes to (1000, 593).

chr1syy:
@pedramamini Thank you, this was a great review and the point-by-point framing made it easy to act on. Pushed in 92739e76b.

**1. Cross-display moves keeping a stale minimum.** Switched from the debounced `move` listener to the `moved` event exactly as you suggested:

```ts
screen.on('display-metrics-changed', reclampMinimumSize);
browserWindow.on('moved', reclampMinimumSize);
```

I folded in both of your asks:

- **Change-guard on `setMinimumSize`.** `reclampMinimumSize` now tracks the last-applied `minWidth`/`minHeight` and returns early when they are unchanged, so GTK firing `moved` repeatedly through a drag collapses to one `setMinimumSize` call per real change rather than churning every tick.
- **`moved` listener is removed on `closed`** alongside the existing `display-metrics-changed` removal.

New tests in the existing style: a two-display setup where `getBounds()` points at the smaller monitor asserts the minimum relaxes to (1000, 593) on the `moved` handler, and a churn-guard test fires `moved` twice with the same bounds and asserts `setMinimumSize` is called exactly once.

**2. Constraints vs. position resolved against different displays.** Handled for free by adding a single `reclampMinimumSize()` call immediately after creation, as you predicted. I did not touch the constraint/position ordering, so the centering that relies on the clamped size is unchanged; once the window exists, `getBounds()` is the truth and the minimum is recomputed against the display it actually opened on (including the Windows `-32000` minimized-sentinel case). There is a dedicated test asserting the one post-create clamp fires with the design minimum.

**3. Per-window screen listener.** Left as-is and added a comment noting the per-window `display-metrics-changed` listener is intentional, that cleanup on `closed` is correct, and that past ~10 concurrent windows Node's `MaxListenersExceededWarning` is acceptable rather than restructuring into a shared listener.

Type-check, ESLint, and Prettier are clean on the changed files, and the scoped `window-manager.test.ts` suite passes (82 tests). Thanks again for the thorough pass.

PR #1317, opened 2026-07-28
feat(agents): chooser for multiple detected binary paths

### LINKED ISSUE #1048: Persist and choose among detected Codex provider paths

pedramamini:
Closes #1048

Supersedes #1050, which was closed after its branch picked up an unrelated 57k-line test-coverage dump. This is the same approach, rebuilt cleanly on top of `main`, plus a fix for the custom-wrapper case and renderer test coverage.

## Problem

Agent detection stopped at the first binary it found. Users with more than one valid install - most commonly an nvm-managed global `codex` alongside an auth wrapper like `codex-multi-auth-codex` - got whichever one happened to sort first, and had to paste the correct path by hand for every new agent.

## What changed

**Detection enumerates everything** (`src/main/agents/path-prober.ts`)

New `findAllBinaryPaths()` combines direct probes of known install locations (Homebrew, npm-global, `~/.local/bin`, nvm/fnm/volta bins) with everything `which -a` / `where` returns against the expanded shell PATH. Results are de-duplicated by canonical resolved path via `fs.realpath`, so a symlink alias and its target collapse to one entry. Case-insensitive on Windows.

The existing `probeUnixPaths` / `probeWindowsPaths` are now thin first-match wrappers over new `*All` variants, so single-path callers are unaffected.

**Detector populates `allPaths`** (`src/main/agents/detector.ts`)

Only set when more than one install is found. The currently active path (custom or detected) is always merged in first, so the chooser reflects what is actually in use. Failures are non-fatal and logged at debug - the chooser is an enhancement, and single-path detection still works if the extra probe throws. Skipped for `bash`, which is on every system and not user-selectable.

**Chooser in the agent config panel** (`AgentConfigPanel.tsx`)

Renders under the Path field only when `allPaths.length > 1`. Selecting a path writes through `onCustomPathChange` and commits immediately via `onCustomPathBlur`, which persists to the existing per-agent `customPath` store - so it survives restarts and becomes the default for the next agent.

Hidden for SSH-backed agents, where the path field means a remote command rather than a local binary.

**Custom wrappers stay first-class.** A hand-typed path that detection did not find (not on PATH, or a `~`-relative path that detection reports expanded) will not match any `<option>`. Rather than let the browser silently render the first option - misrepresenting which binary is active - it is surfaced as an explicit `Custom: <path>` entry. The text input remains the way to enter one.

## Mapping to the request

| Asked for | Status |
| --- | --- |
| Detect multiple installs incl. PATH, Homebrew, system node/npm, nvm | Yes |
| Let the user choose when multiple are found | Yes, dropdown under the Path field |
| Persist last selection as the default for future agents | Yes, via existing `customPath` store |
| Keep manual entry for custom wrappers | Yes, plus an explicit `Custom:` entry |
| Show chooser only on first multi-detection or explicit Detect | Partial - see below |

The chooser stays visible whenever multiple installs exist rather than appearing once and hiding. It is a compact collapsed `<select>`, so it is not intrusive, and a persistently visible control is discoverable later when a user adds a second install. Happy to gate it behind the existing Detect button if you'd rather have the original behavior.

## Testing

- `path-prober.test.ts`: 4 new cases - priority ordering, symlink de-duplication, empty result, and `which` failure falling back to probe results. 28 pass.
- `AgentConfigPanel.test.tsx`: 8 new cases - hidden at 0/1 paths, lists all options, preselection with and without a custom path, immediate persistence on change, the `Custom:` entry for an undetected wrapper, and hidden under SSH. 33 pass.
- `npx eslint` clean on all changed files; prettier reports no changes.

Note: `src/__tests__/main/agents/claude-usage-startup.test.ts` fails 20 tests on this branch. It fails identically on `main` at `39c9a6744` - pre-existing and unrelated.

## Summary by CodeRabbit

* **New Features**
  * Detects multiple local installations of supported agents.
  * Adds an installation chooser when multiple paths are available.
  * Preserves custom paths and clearly labels them as custom selections.
  * Automatically selects the detected installation currently in use.

* **Bug Fixes**
  * Removes duplicate installation paths, including equivalent filesystem targets.
  * Continues detection successfully when optional path discovery fails.

* **Tests**
  * Added coverage for path discovery, duplicate handling, chooser behavior, custom paths, and SSH configurations.


pedramamini:
@reachrazamair your call. Branch updated against main (red `test` job was the stale-base localStorage mock, not this diff).

**Summary of my read: the feature is good and wanted, but two of the findings mean it does not currently deliver its headline use case.**

### The chooser can persist the wrong path

`AgentConfigPanel.tsx:576` - `onChange` calls `onCustomPathChange(next)` and then immediately `onCustomPathBlur()`. React state updates are async and batched, so `onCustomPathBlur` runs against the closure's *previous* `customPath`. At call sites that persist on blur (the onboarding wizard's `handleCustomPathBlur` in `useAgentConfigurationPanel.ts`), this writes the old path to `agentConfigsStore` and subsequent detection and launches keep using the old binary.

So: user picks installation B, Maestro saves installation A. Both Greptile and Copilot landed on this independently. Copilot's suggested fix is the right shape - change the contract so the commit handler receives the value (`onCustomPathBlur(next)` / `onCustomPathCommit(next)`) rather than reading it back out of state.

### The wrapper case the feature exists for is not detected

`path-prober.ts:678` - the direct probes only construct `codex` paths, and `which -a codex` / `where codex` only search that canonical name. A differently-named wrapper like `codex-multi-auth-codex` never enters `allPaths`, so the chooser does not offer it. That is the primary scenario #1048 was filed about, so worth confirming against the original report before merge.

### Two dedup / robustness items

- `detector.ts:212`: `active` is merged into `found` with a raw `String.includes` check. A symlink alias or a Windows casing difference resolves to the same binary but compares unequal, so `allPaths` shows a phantom second install and the chooser appears when there is only one. `findAllBinaryPaths` already does realpath canonicalization internally for its own candidates - this final merge step just does not use it.
- `path-prober.ts:713`: `execFileNoThrow(command, args, undefined, env)` passes `env` as the 4th arg, which is only treated as `ExecOptions` (and therefore only honors `timeout`) when the object has an `input` or `timeout` key. So no timeout is applied to the `which`/`where` lookup. `doDetectAgents` awaits this sequentially once per detected agent, so a broken PATH entry pointing at an unresponsive network mount can hang agent detection indefinitely. Worth fixing regardless of what happens to the rest of this PR.

Note that same `ExecOptions` discriminator ambiguity is what Copilot flagged over on #1306 (`src/main/utils/execFile.ts`) - an env object that happens to contain an `env` key gets misclassified. Two PRs tripping over the same overloaded signature suggests the signature itself is the problem and might deserve a separate cleanup.

### Minor

`AgentConfigPanel.tsx:600` - the helper copy promises the selection is saved as the default for future agents, but `EditAgentModal` deliberately makes `onCustomPathBlur` a no-op and saves on modal save instead. Copy is misleading in that flow.

**My take:** fix the state-timing bug and confirm the wrapper detection actually covers #1048, and this is a clean merge. The timeout item I would take even as a standalone.

reachrazamair:
> @reachrazamair your call. Branch updated against main (red `test` job was the stale-base localStorage mock, not this diff).
> 
> **Summary of my read: the feature is good and wanted, but two of the findings mean it does not currently deliver its headline use case.**
> 
> ### The chooser can persist the wrong path
> 
> `AgentConfigPanel.tsx:576` - `onChange` calls `onCustomPathChange(next)` and then immediately `onCustomPathBlur()`. React state updates are async and batched, so `onCustomPathBlur` runs against the closure's _previous_ `customPath`. At call sites that persist on blur (the onboarding wizard's `handleCustomPathBlur` in `useAgentConfigurationPanel.ts`), this writes the old path to `agentConfigsStore` and subsequent detection and launches keep using the old binary.
> 
> So: user picks installation B, Maestro saves installation A. Both Greptile and Copilot landed on this independently. Copilot's suggested fix is the right shape - change the contract so the commit handler receives the value (`onCustomPathBlur(next)` / `onCustomPathCommit(next)`) rather than reading it back out of state.
> 
> ### The wrapper case the feature exists for is not detected
> 
> `path-prober.ts:678` - the direct probes only construct `codex` paths, and `which -a codex` / `where codex` only search that canonical name. A differently-named wrapper like `codex-multi-auth-codex` never enters `allPaths`, so the chooser does not offer it. That is the primary scenario #1048 was filed about, so worth confirming against the original report before merge.
> 
> ### Two dedup / robustness items
> 
> * `detector.ts:212`: `active` is merged into `found` with a raw `String.includes` check. A symlink alias or a Windows casing difference resolves to the same binary but compares unequal, so `allPaths` shows a phantom second install and the chooser appears when there is only one. `findAllBinaryPaths` already does realpath canonicalization internally for its own candidates - this final merge step just does not use it.
> * `path-prober.ts:713`: `execFileNoThrow(command, args, undefined, env)` passes `env` as the 4th arg, which is only treated as `ExecOptions` (and therefore only honors `timeout`) when the object has an `input` or `timeout` key. So no timeout is applied to the `which`/`where` lookup. `doDetectAgents` awaits this sequentially once per detected agent, so a broken PATH entry pointing at an unresponsive network mount can hang agent detection indefinitely. Worth fixing regardless of what happens to the rest of this PR.
> 
> Note that same `ExecOptions` discriminator ambiguity is what Copilot flagged over on #1306 (`src/main/utils/execFile.ts`) - an env object that happens to contain an `env` key gets misclassified. Two PRs tripping over the same overloaded signature suggests the signature itself is the problem and might deserve a separate cleanup.
> 
> ### Minor
> 
> `AgentConfigPanel.tsx:600` - the helper copy promises the selection is saved as the default for future agents, but `EditAgentModal` deliberately makes `onCustomPathBlur` a no-op and saves on modal save instead. Copy is misleading in that flow.
> 
> **My take:** fix the state-timing bug and confirm the wrapper detection actually covers #1048, and this is a clean merge. The timeout item I would take even as a standalone.

Makes total sense, agree on shipping this now and handling wrapper detection as a follow-up. All three fixes are ready (chooser staleness with a real component test, the EncoreTab config fix, dedup, and timeout). I will open the follow-up issue for wrapper binaries once this merges.

reachrazamair (approved):
Approving. Chooser staleness, phantom duplicates, and the missing timeout are all fixed and verified, details in the comment above. Wrapper detection is a real gap but not a regression, manual entry already covers it same as before this PR. Opening that as its own follow-up issue instead of blocking here.

pedramamini:
@reachrazamair these two are the only PRs in the queue with no response from anyone - not flagging them as urgent, just that they fell through rather than being deliberately parked.

Both still have my findings from the earlier pass standing unaddressed:

- **#1308** (OpenCode `--agent`): the boolean-flag stripper eats the following token, and the Cue override path omits `readOnlyMode`, so a read-only Cue step runs with write capability.
- **#1317** (binary path chooser): the chooser persists the wrong path (blur reads stale closure state), and the differently-named wrapper case that #1048 was actually filed about is never detected.

Happy to take either one if you'd rather stay on your current queue - just say which and I'll pick it up. Otherwise they're yours whenever you get there.

PR #1319, opened 2026-07-28
fix(tabs): keep the in-progress indicator on parallel threads

### LINKED ISSUE #1318: Bug: Thread missing yellow in-progress indicator with multiple threads per...
## Summary
Thread missing yellow in-progress indicator with multiple threads per agent

## Environment
- Maestro version: 0.18.5-RC
- Operating system: macOS (Darwin Kernel Version 25.5.0: Mon Apr 27 20:39:42 PDT 2026; root:xnu-12377.121.6~2/RELEASE_ARM64_T6031, 25.5.0)

## Steps to Reproduce
1. Start multiple threads (3) on a single agent
2. Have all three actively running at the same time
3. Observe the thread list — only two of the three show the yellow in-progress icon

## Expected Behavior
Every thread that is actively running should display the yellow in-progress indicator icon, regardless of how many threads are running under the same agent.

## Actual Behavior
With three threads running under a single agent, only two show the yellow in-progress icon. The third running thread has no indicator.

## Additional Context
Reporter did not confirm whether the indicator appears when clicking into the affected thread, or whether it is missing in all views. Unclear if this is specific to a certain thread count or ordering.

## Support Package
[maestro-debug-package.zip](https://raw.githubusercontent.com/ronaldeddings/maestro-feedback-attachments/main/feedback/1785201474360-maestro-debug-2026-07-28T01-17-53.zip)

pedramamini:
Closes #1318

## Problem

With multiple threads running under a single agent, a thread could keep working while its tab chip showed no yellow in-progress dot. The reporter saw 3 threads running and only 2 indicators.

The reporter's support package confirms the setup: three live agent processes on one agent (`{sessionId}-ai-{tabId}`, uptimes 26m / 12m / 9m). Their screenshot shows the indicator missing specifically on the **focused** tab, while two background tabs still had theirs.

## Root cause

Both causes are in the Auto Run exit reducer inside `spawnAgentForSession` (`src/renderer/hooks/agent/useAgentExecution.ts`).

**1. Dequeue branch never marked the target tab busy.**
When the exit dequeued a queued item, it appended the user log and set `activeTabId: target.tabId`, but never set that tab's `state: 'busy'` / `thinkingStartTime`. The turn ran with no indicator - and because the same branch focuses the tab, the affected thread was the one in view, matching the screenshot exactly.

The three sibling dispatch paths all mark the tab busy; this copy was the outlier:

| Path | Marks tab busy? |
| --- | --- |
| `useQueueProcessing.dispatchQueuedItem` | yes, via `markTabRunningQueuedItem` |
| `useAgentExitListener` onExit | yes, via `markTabRunningQueuedItem` |
| `useBatchHandlers` queue drain | yes, inline |
| `useAgentExecution` onExit | **no** |

Now routed through the shared `markTabRunningQueuedItem` helper, which also picks up the `forceParallel`/`readOnly` log flags the hand-rolled entry dropped.

**2. Empty-queue branch force-idled every busy tab.**
With no queued items, the exit set `state: 'idle'` on ALL busy tabs and on the session. A batch task spawns under its own `{sessionId}-batch-{ts}` process id and, per the note at its own spawn site, deliberately never marks a tab busy. So any tab busy at that moment has its own live `-ai-{tabId}` agent and is cleared by that agent's own `onExit`. Session state is now derived from whether a tab is still busy, mirroring the same pattern already used in `useAgentExitListener`.

## Testing

New `useAgentExecution.exitQueueState.test.tsx` covers the empty-queue, dequeue, and orphaned-tab branches. Three of its four cases fail against the previous reducer:

```
× leaves still-running parallel tabs busy when the batch task exits
  AssertionError: expected [ 'idle', 'idle', 'idle' ] to deeply equal [ 'busy', 'busy', 'busy' ]
× marks the dequeued item's target tab busy and logs its prompt
× marks an orphaned (closed) target tab busy without touching live tabs
```

One assertion in the existing `useAgentExecution.test.ts` seeded an artificially busy tab and asserted the batch exit cleared it - that codified cause 2. Updated to seed an idle tab (the real batch case, since Auto Run never marks a tab busy), with a pointer to the new file for the still-running case.

- `npm run lint` (all three tsconfigs): clean
- `npx eslint` on changed files: clean
- Full renderer suite: **678 files passed, 18469 tests passed, 1 skipped**

Local validation is macOS-only; leaving the Ubuntu and Windows CI legs to confirm.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Auto Run session handling when batch processes exit.
  - Parallel AI tabs remain busy while other work is running.
  - Sessions return to idle when all active work is complete.
  - Queued prompts resume in the appropriate tab with accurate activity status and logs.
  - Improved handling of queued work for tabs closed before execution resumes.

- **Tests**
  - Added regression coverage for Auto Run exits, queued prompts, parallel tabs, and closed tabs.


pedramamini:
@reachrazamair your call on this one. Branch updated against main; the red `test` job was the stale-base localStorage mock issue, not this diff.

**The good news:** I dug into the P1 and I think it is **not a regression introduced here**. Greptile flags that `anyTabStillBusy` at `useAgentExecution.ts:442` only checks `s.aiTabs` and ignores `s.orphanedThinkingTabs`, so a batch exit can mark the session idle while work owned by a since-closed tab is still running. That is true - but the same blind spot already exists at roughly ten other sites on main:

```
useAgentExitListener.ts:375, 434, 496, 571, 665, 706
useAgentErrorListener.ts:195
useAgentCommandExitListener.ts:30
tabHelpers.ts:867
terminalTabHelpers.ts:71
```

Every one of them derives session busy-ness from `aiTabs.some(tab => tab.state === 'busy')` and none consult `orphanedThinkingTabs`. So this PR inherits the pattern rather than creating it, and blocking it on that finding would be holding it to a standard main does not currently meet.

**What I would actually do:** land this, then fix the blind spot once in a shared helper - something like `isSessionBusy(session)` in `tabHelpers.ts` that unions `aiTabs` and `orphanedThinkingTabs` - and sweep all eleven call sites onto it. That is the version worth doing given how much duplicate busy-derivation logic is sitting in the exit listeners already.

**One thing that is specific to this diff and worth a look before merge:** line 446 reads

```ts
state: anyTabStillBusy ? s.state : ('idle' as SessionState)
```

whereas `useAgentExitListener.ts:574` - the closest sibling path - reads

```ts
state: anyTabStillBusy ? ('busy' as SessionState) : ('idle' as SessionState)
```

Copilot raised this too. Preserving `s.state` means that if the session is somehow already `idle` while a tab is `busy`, this path will not correct it, where the sibling path would. Probably harmless in practice, but the inconsistency is the kind of thing that makes the next busy-state bug hard to trace. Cheap to align.

Minor: the test file overwrites `window.maestro` without capturing the original for restoration in `afterEach` (lines 21 and 127) - cross-test pollution risk.

pedramamini:
Rebased onto current main and aligned the one thing specific to this diff.

`useAgentExecution.ts:446` now reads `anyTabStillBusy ? 'busy' : 'idle'`, matching its sibling at `useAgentExitListener.ts:574`. Preserving `s.state` meant an already-idle session with a still-busy tab would never get corrected, and the inconsistency between two paths doing the same job is what makes the next busy-state bug hard to trace.

Also captured and restored `window.maestro` in the new suite so the wholesale stub cannot leak into other tests in the same worker.

The `orphanedThinkingTabs` blind spot Greptile flagged stands as written in my earlier comment: it is inherited from ~10 sites on main, not introduced here, and is worth one shared `isSessionBusy(session)` helper rather than blocking this PR.

`tsc` clean; 13 tests green across both `useAgentExecution` suites.

PR #1320, opened 2026-07-28
Fix manual transcript reading position

felipeggv:
## Summary

- treat manual scroll-away intent as authoritative before the throttled React state update
- prevent settled final-response DOM mutations from pulling the transcript back to the bottom
- resume follow mode when the user returns to the bottom or uses the existing jump-to-bottom action
- add regression coverage for the race between manual scrolling and an internal DOM mutation

## Validation

- 34,702 tests passed; 108 skipped
- renderer production build passed
- TypeScript, ESLint, and Prettier checks passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved manually selected reading positions in terminal output when new content or DOM updates appear.
  * Prevented automatic scrolling from unexpectedly returning the view to the bottom after scrolling upward.
  * Improved scroll-position handling when switching sessions, restoring views, or resuming output.


pedramamini:
@felipeggv Thanks for the contribution, and for the thorough writeup. This is a good catch and a clean fix.

## Verification

I pulled the branch and confirmed the regression coverage is real: reverting only `useTerminalOutputScroll.ts` back to base makes `Case 6` fail with `expected 9000 to be 2000` (the transcript gets yanked back to the bottom), and it passes with the fix in place. That's exactly the bug described. All 8 tests in the file pass on the PR head.

CI is green on all four legs (`ubuntu-latest` and `windows-latest`, both shards), plus `lint-and-format`. No merge conflicts, base is `rc`, and both Greptile and CodeRabbit came back with no actionable findings.

## On the approach

The mechanism holds up. The root problem is that `isAtBottomRef.current = isAtBottom` runs on every render, so any synchronous write to that ref gets clobbered by the throttled React state on the next commit. A dedicated ref that nothing reassigns from state is the right way out, and splitting the intent capture into an unthrottled wrapper in front of the throttled handler keeps the leading/trailing-edge behavior intact. The `atBottom` / `parkedAtProgrammaticTarget` branching in the new `handleScroll` matches `handleScrollInner` exactly, so the two cannot disagree today, and every new gate is a strict `&&` tightening, which keeps the failure mode on the safe side (show the jump-to-bottom pill rather than steal the scroll position).

## Non-blocking follow-ups

Nothing here needs to change before merge, just flagging for whenever this file is next touched:

1. The at-bottom measurement is now duplicated verbatim between `handleScroll` and `handleScrollInner`, including the bare `50` slack. Since correctness depends on those two staying identical, a small shared `measureScrollState(container)` helper plus an `AT_BOTTOM_SLACK_PX` constant next to `PROGRAMMATIC_TARGET_EPSILON_PX` would make future divergence impossible rather than merely unlikely.
2. This hook now carries four overlapping notions of follow state: `isAtBottom`, `isAtBottomRef`, `userScrolledAwayRef`, and `autoScrollPaused`. Each was added for a real bug (#1140, J1, and now this one), but they are getting hard to reason about together. Worth a consolidation pass at some point.

Approving and tagging `approved`. Nice work.

pedramamini:
Coming back to land this. The branch was 298 commits behind `rc`, so the green CI above was measured against a late-July tree - I re-validated locally and pushed the result to this branch (`8d8ece360`).

## Still needed, confirmed

The bug has not been fixed by anything that landed since. Reverting only `useTerminalOutputScroll.ts` to today's `origin/rc` makes `Case 6` fail with `expected 9000 to be 2000`; with your fix it passes. So this is still live on `rc` today.

## One conflict, resolved

`rc` grew a `jumpInFlightRef` guard (#1323 / #1348) on the same rAF re-check your change gates with `userScrolledAwayRef`:

```ts
if (scrollContainerRef.current && isAtBottomRef.current && !jumpInFlightRef.current) {
```

The two guards are orthogonal - one means a cross-tab search jump is in flight, the other means the user deliberately scrolled away - so both now gate it rather than one replacing the other.

## Applied the follow-up from my review

Since the file conflicted anyway, I took item 1 from my earlier comment: `handleScroll` and `handleScrollInner` had the at-bottom test and the programmatic-target test duplicated verbatim, including a bare `50`. Correctness depends on those two agreeing, so they now share a `measureScrollState(container, isProgrammatic, targetTop)` helper and an `AT_BOTTOM_SLACK_PX` constant next to `PROGRAMMATIC_TARGET_EPSILON_PX`. Behavior is unchanged; the divergence is now impossible rather than merely unlikely. (Item 2, consolidating the four follow-state refs, is still open for a future pass.)

## Validation

```
full suite (pre-push)   1559 files, 37556 passed, 108 skipped
TerminalOutput suites   37 passed
tsc --noEmit            clean
eslint / prettier       clean
```

Merging. Thanks for the fix and the clear writeup.

PR #1328, opened 2026-07-29
feat(dispatch): dispatch --notify-on-complete (dispatch with callback)

chr1syy:
## What

Adds `maestro-cli dispatch --notify-on-complete <caller-agent>`: when that specific dispatch finishes, Maestro starts a **real turn in the calling agent's live tab** carrying the result plus a handle to the full output.

Phase 1 (correlated single-dispatch callback) and Phase 2 (Auto Run finality signal) ship together, because Phase 1 alone would misfire once per task on any Auto Run.

## Why

`dispatch` is fire-and-forget by contract. It returns a tab id and exits, and nothing tells the caller when the delegated run finished. In practice a human watches the target tab and pokes the orchestrator with "it's done, carry on" - which breaks every multi-stage workflow (build -> review -> fix -> verify -> merge) into a chain of manual handovers. The orchestrator has the plan; it just cannot observe the completion of its own delegations.

### Gap analysis - why the existing building blocks do not cover this

Each of these was checked against the code, not assumed:

1. **`agent.completed` has no dispatch-level correlation.** `exit-listener.ts` calls `cueEngine.notifyAgentCompleted(sessionId, ...)`; matching in `cue-completion-service.ts` is `sources.some(src => src === sessionId || src === completingName)`. A subscription fires on *any* completion in the source session. `AgentCompletionData` has no dispatch identity at all.
2. **Cue subscriptions are persistent YAML only.** There is no ephemeral subscription API - `getSessionConfigs()` reads exclusively from the registry snapshot populated by `cue-yaml-loader.ts`. `time.once` self-destruct is a YAML *rewrite*, so an ad-hoc dispatch callback would mean write + watcher reload + rewrite + reload per delegation, racing every other writer on that file.
3. **An N-task Auto Run fires N completions.** `notifyAgentCompleted` runs once per process exit, and Auto Run spawns one process per task.
4. **A Cue callback would not give the orchestrator a real turn.** This is the finding that kills the "just use a Cue chain" answer even if 1-3 were solved: `cue-run-manager.ts` spawns a **fresh headless child process** via `buildSpawnSpec()`. The orchestrator's accumulated context is not there. What is needed is a turn in the caller's *existing* tab - which is what `dispatch` itself already does.
5. **`pianola orchestrate`** is the existing workaround and it is a **poll loop** (`list_sessions` + per-tab busy/idle inference, plus an extra transcript check because the desktop collapses `waiting_input` to `idle`). That an orchestrator has to poll and disambiguate idle states is the clearest evidence the push-based primitive is missing.

## Design

**Correlation.** An entry is keyed on the composite process session id the exit listener already sees (`{agentId}-ai-{tabId}`) and only arms once the dispatched process actually **spawns**. That start gate is what makes `--queue` safe: a queued prompt can sit behind another turn for minutes, and that predecessor's exit must not fire the callback. For `--new-tab` the registry may adopt a spawn observed in the last 60s, since the tab id is brand new and the renderer can spawn before the ack round-trips back.

**Delivery.** Through the renderer's authoritative execution queue (the same `enqueue_command` path `dispatch --queue` uses), so a busy caller queues rather than being rejected or interleaved - and it lands as a real turn in the live tab, with context.

**Auto Run finality (Phase 2).** New `AutoRunStateTracker` fed **unconditionally** from `web:broadcastAutoRunState`, before the `getWebServer()` null check. Previously Auto Run finality was only reachable through `broadcastService.broadcastAutoRunState()`, and the IPC handler dropped the state entirely when the web server was off - so with Live Mode disabled the main process could not tell "the process exited and the agent is done" from "the process exited and task 4 is about to spawn". The tracker also treats a `null` state as end-of-run, which the web broadcaster's transition detection misses (the renderer clears Auto Run state to `null` when a batch ends, and the existing `autorun_complete` detection only looks at the non-null branch). This is 5.2(a) from the design doc, and it is the same fix Cue's `agent.completed` will want for its own N-fires-per-Auto-Run problem.

## API

```bash
maestro-cli dispatch <reviewer-agent> "review the diff on feat/x" \
  --new-tab \
  --notify-on-complete <orchestrator-agent> \
  [--callback-tab <tabId>] \
  [--callback-prompt "<wrapper>"] \
  [--callback-timeout <seconds>]
```

Response gains `callbackId` and `notifyOnComplete`. Template variables available in `--callback-prompt`: `DISPATCH_CALLBACK_ID`, `DISPATCH_TARGET_ID`, `DISPATCH_TARGET_NAME`, `DISPATCH_TAB_ID`, `DISPATCH_STATUS`, `DISPATCH_EXIT_CODE`, `DISPATCH_DURATION`, `DISPATCH_OUTPUT`, `DISPATCH_OUTPUT_TRUNCATED`, `DISPATCH_TASKS_COMPLETED`, `DISPATCH_TASKS_TOTAL`, `DISPATCH_PROMPT`.

## Decisions taken on the open questions

The design doc left six questions open. Decisions made so the build could finish; all are cheap to reverse at review:

1. **Ship Phase 1 alone?** No - 1+2 together, as instructed. Phase 1 alone is a footgun with Auto Run.
2. **Auto Run signal approach** - took 5.2(a), the first-party main-process tracker. It is the correct fix and it also unblocks `only_on_final` for Cue later.
3. **Flag name** - kept `--notify-on-complete`. Explicit beats short for a flag that arms a background side effect. Easy to add `--then` as an alias if preferred.
4. **Encore gating** - **not gated**. This is core CLI dispatch surface, not Cue; it does not touch the Cue engine, and gating it would make orchestration scripts depend on an Encore toggle.
5. **`DISPATCH_OUTPUT` cap** - kept Cue's **5000 chars**. The tab handle is the escape hatch for the rest, and matching Cue keeps one number to reason about for "agent output landing in another agent's prompt".
6. **Restart behaviour** - entries are in-memory only; a desktop restart drops them silently. Persisting armed entries with `status: interrupted` was Phase 3 in the design and is deliberately out of scope here.

Additional decisions worth flagging:

- **`--notify-on-complete` requires `--new-tab` or `--tab`.** Without an explicit tab the dispatch lands in whichever tab happens to be active, and there is nothing stable to correlate a completion against. Rejected with `INVALID_OPTIONS` rather than silently guessing.
- **`DISPATCH_*` variables are NOT registered in `shared/templateVariables.ts`.** `substituteTemplateVariables()` needs a full `TemplateContext` (session record, git branch, agent paths) that the main-process registry does not have at fire time, and these variables are meaningless outside a callback body. Rendering is a small single-pass substituter in `dispatch-callback-prompt.ts`. Single-pass matters: agent output containing a literal `{{DISPATCH_...}}` cannot expand into another variable.
- **Output source is best-effort.** `HistoryEntry` carries no `tabId`, so the newest history entry written after the callback was armed is used, floor-gated on the arming timestamp. The tab handle in the prompt is the authoritative path to the real transcript.
- **The spawn/exit hook is its own listener** (`process-listeners/dispatch-callback-listener.ts`) rather than an edit inside `exit-listener.ts`. Correlation is exact-match on ids the registry itself minted, so group-chat / batch / synopsis / terminal session ids can never match - this path needs none of the group-chat containment machinery that makes `exit-listener.ts` load-bearing.
- **Runaway guards included** (they were Phase 3 in the design but are ~10 lines each): 200-entry arm cap, `CALLBACK_ALREADY_ARMED` rejection instead of stacking callbacks on one tab, and self-callback-loop rejection in both the CLI and the server. A full A -> B -> A chain-depth cap in the style of Cue's `maxChainDepth` is still open.

## Scope

23 files, +2252/-17. New: `src/main/dispatch-callbacks/` (registry, prompt renderer, service, types), `src/main/autorun/autorun-state-tracker.ts`, `src/main/process-listeners/dispatch-callback-listener.ts`. Edited: CLI dispatch + flag wiring, web-server message handlers (`send_command`, `new_ai_tab_with_prompt`, `enqueue_command`), `WebServer`, `web:broadcastAutoRunState`, main wiring.

## Tests

New suites for registry lifecycle (start gate, fire-once, Auto Run parking, grace window, timeout, cancel, spawn lookback, cap, duration), prompt rendering (truncation, fencing, custom templates, no re-substitution of placeholders in agent output), the Auto Run tracker (including the `null`-clear edge), the service wiring (queued delivery, Auto Run finality, output failures), CLI validation/payload wiring, and the web-server arming guards.

`npx vitest run` on all affected suites: 815 passing, 0 failing. `tsc` clean on all three configs (`tsconfig.lint`, `tsconfig.main`, `tsconfig.cli`). `npm run build` exits 0. ESLint clean on every changed source file.

Local validation is single-OS; both CI matrix legs still need to be green before merge.

## Follow-ups not in this PR

- Persist armed entries so a restart wakes the caller with `status: interrupted` (design Phase 3).
- Cancel callbacks on target-tab close (the registry has `cancelForTab`, but nothing calls it yet - entries currently expire on timeout instead).
- `maestro-cli dispatch callbacks list|cancel <id>` for registry inspection.
- Chain-depth cap for A -> B -> A callback loops.
- `only_on_final: true` for Cue's `agent.completed`, now that the tracker exists.
- Point `pianola orchestrate` at the push signal instead of its poll loop.

## Summary by CodeRabbit

* **New Features**
  * Added `--notify-on-complete` to dispatch, enabling exactly-once wake-ups to a specified caller when delegated work finishes.
  * Added callback targeting (`--new-tab`/`--tab`), prompt override, and timeout controls, including support for queued dispatches and Auto Run-aware delivery.
  * Dispatch responses can now include `callbackId`/`notifyOnComplete`, and delivered callbacks include status, duration, output (with truncation), and task progress.

* **Documentation**
  * Expanded CLI reference and added a dedicated “Dispatch with Callback” guide with examples, templates, and error cases.

* **Tests**
  * Added comprehensive test coverage for CLI validation, callback prompting, registry/service behavior, WebSocket handling, and Auto Run state tracking.


chr1syy:
## Review round 1 addressed

Pushed as a single fix commit on `feat/dispatch-notify-on-complete`.

### Fixed

| Finding | Source | Fix |
|---|---|---|
| Agent-wide Auto Run parking delays/pollutes a tab-specific callback | greptile P1 | Auto Run is agent-scoped with no tab dimension, so there is no tab-level signal to correlate against. Instead the registry now parks by **start time**: `autoRunRunningSince(agentId)` (new `AutoRunStateTracker.getRunningSince()`) replaces the boolean, and an entry only parks when the batch began at/after its own turn started. A batch already running in another tab is by definition not ours, so it neither delays the callback nor donates its task counts. |
| Pre-registration exits are lost (`--new-tab` fast exit) | greptile P1 | The registry now keeps a `recentExits` lookback alongside `recentSpawns`. When a `--new-tab` registration adopts a recent spawn, an exit recorded after that spawn is replayed, so a short turn reports its real result instead of a bogus `timeout`. Exits that predate the adopted spawn are ignored (predecessor turn). |
| CLI reports `notifyOnComplete` without confirmation it was armed | coderabbit Major | Both the direct and queued paths now fail with a dedicated `CALLBACK_NOT_ARMED` code when the desktop acks without a `callbackId` (e.g. an older build that ignores the field), rather than claiming a wake-up that will never arrive. The tab id is still returned so callers can poll or re-dispatch. |
| Callback target validated after the tab is created | coderabbit Major | Extracted `validateCallbackRequest()` (registry availability, caller exists, self-target) and run it **before** `newAITabWithPrompt`. Only `hasArmedFor(tabId)` stays post-dispatch, since it genuinely needs the fresh tab id. No more `success: false` alongside a live turn in an orphaned tab. |
| `send_command` rejection log-only | coderabbit Major | Now routed through the existing `reportHandlerError` helper. |
| `catch {}` swallows `onFinal` listener exceptions | coderabbit Major | Reported via `captureException` with `area: autorun-state-tracker`; remaining listeners still run. |
| Non-finite `timeoutMs` -> `NaN` expiry that never sweeps | coderabbit nit | Guarded with `Number.isFinite`, falling back to the default window. |
| `docs/cli.md` continuation lines use spaces | coderabbit nit | Tabs, matching the rest of the file. |
| Test `overrides` untyped / no `enqueue_command` arming coverage | coderabbit nits | `Partial<DispatchCallbackRegistration>`, plus an `enqueue_command` arming test. |

### Declined

- **Retry/backoff on callback delivery failure** (coderabbit, "heavy lift"). Out of scope for this PR and arguably wrong here: delivery already goes through the renderer's authoritative execution queue, which is the durable-ish path - a failure there means the caller agent or its tab is gone, which a retry cannot fix. A blind retry would risk double-waking the caller, and the existing timeout already bounds the "no wake-up arrived" case. Worth revisiting only with real field data showing lost callbacks.

### New tests

Registry: pre-existing-Auto-Run non-parking, exit replay (success + failure exit code), no-replay for a pre-spawn exit. CLI: `CALLBACK_NOT_ARMED` on both direct and queued paths. Web handlers: `enqueue_command` arming, plus new-tab pre-validation asserting `newAITabWithPrompt` is never called.

Local: `tsc --noEmit` clean on all four configs, prettier/eslint clean, 868 tests green across the affected suites.

chr1syy:
### Round 2 (CodeRabbit, outside-diff)

**`docs/cli.md:191` - "document `callbackTimeout` as unset when omitted"** - declining. The observation is correct about the mechanism (the CLI omits the field so the main process applies `DEFAULT_CALLBACK_TIMEOUT_MS`), but the doc line describes the user-visible contract, and that default is 3600s. Rewording it to "unset when omitted, then whatever the desktop decides" would be less accurate for a reader of the CLI reference, not more.

All other findings from both reviewers are fixed in `e0a1a5a`-onward; the three still-open threads anchor to code that the fix commit changed (the Sentry one now reads `this.reportHandlerError(...)`, and the greptile new-tab thread anchors to the arming site while the actual fix - the `recentExits` replay - lives in the registry).

**CI: all 5 checks green on both matrix legs** (ubuntu shards 1/2 + 2/2, windows shards 1/2 + 2/2, lint-and-format). `mergeStateStatus: CLEAN`.

PR #1334, opened 2026-07-31
feat(resilience): Provider Failover to Anthropic-compatible backup endpoints

### LINKED ISSUE #139: Feature Request: Automatic Failover to Custom Anthropic-Compatible Endpoints (via Env Var Swap)
### Is your feature request related to a problem? Please describe.
Yes. Currently, if the primary Anthropic API connection stalls—due to rate limits (429), credit exhaustion (402), or strict daily reset windows (e.g., the "resets at 1am" message)—the `claude` process simply hangs or exits. This interrupts long-running autonomous workflows and requires manual intervention to swap API keys or waiting for the reset window.

Many users now have access to **Anthropic-compatible backup endpoints** (such as local LLMs via vLLM/Ollama, Z.AI, or enterprise proxies) that function as drop-in replacements but are often preferred only as a secondary fallback.

### Describe the solution you'd like
I propose a **Failover & Recovery System** for the process manager that leverages the standard `claude` CLI environment variables.

When Maestro detects a "stalled" state or a specific billing/rate-limit error from the CLI, it should trigger the following workflow:

**1. Hot-Swap Failover (The "Spare Tire")**
* Instead of exiting, Maestro restarts the process using a secondary, user-defined configuration.
* **Mechanism:** Overwrite the `ANTHROPIC_BASE_URL` and `ANTHROPIC_API_KEY` environment variables with values from a "backup" block in the config.
* **Context Handoff:** Inject a "handoff prompt" into the new session (e.g., *"Previous session stalled. Resume task: [Task Summary]"*) so the new provider knows where to pick up.

**2. Polling & Auto-Recovery (Return to Primary)**
* The system should not stay on the backup provider indefinitely.
* **Time-Based Reset:** Allow users to define a "Reset Time" (e.g., `01:00 PST`) or a "Retry Interval" (e.g., every 60 minutes).
* When the reset condition is met, Maestro attempts to spawn the next agent process using the **Primary (Official)** credentials again.
* If the Primary fails again, it immediately reverts back to the Backup.

### Technical Implementation Note
This feature does **not** require writing a new API client or integration.
Since the `claude` CLI natively accepts a custom Base URL via environment variables, this is purely a process-spawning configuration change:

* **Official Mode:** Spawn process with standard `env`.
* **Failover Mode:** Spawn process with modified `env`:
    ```javascript
    env = {
      ...process.env,
      ANTHROPIC_BASE_URL: "[https://api.z.ai/api/anthropic](https://api.z.ai/api/anthropic)", // or local/custom endpoint
      ANTHROPIC_AUTH_TOKEN: "sk-custom-key..."
    }
    ```
This allows Maestro to be **infrastructure agnostic**—supporting any provider (Local or Cloud) that mimics the Anthropic API structure—with minimal code changes.

### Proposed Configuration Example
```json
{
  "provider_strategy": {
    "failover_enabled": true,
    "trigger_errors": ["credit balance is too low", "rate limit exceeded", "stalled"],
    "primary": "anthropic_official",
    "backup": {
      "env": {
        "ANTHROPIC_BASE_URL": "[https://my-custom-endpoint.com/v1](https://my-custom-endpoint.com/v1)",
        "ANTHROPIC_AUTH_TOKEN": "sk-custom-key..."
      }
    },
    "recovery_strategy": {
      "method": "schedule_or_poll",
      "reset_time": "01:00 PST",
      "poll_interval_minutes": 60
    }
  }
}

### PR DESCRIPTION
**Pull request by pedramamini**
Closes #139

## What this does

Agent Resilience today answers an upstream failure by waiting and resending to the **same** provider. That is right for a 60-second "Overloaded" blip, but poor for plan quota exhaustion: the agent idles until the reset window (potentially hours) and any long-running autonomous work stalls with it.

This adds the other half. An agent can carry an ordered list of **Anthropic-compatible backup endpoints** (local vLLM/Ollama, Z.AI, an enterprise proxy, or simply a second account). When resilience decides an error is retryable *and* failover is armed, Maestro swaps the endpoint's env vars into the next spawn after a 3s handover instead of waiting out the clock.

As #139 proposed, endpoints are plain env-var bundles, because the `claude` CLI already reads its base URL and token from the environment. No new API client, no per-provider integration - just a different `env` on the spawn.

## Design decisions worth a look

**1. Each endpoint carries an optional model override.** The original proposal only swaps `ANTHROPIC_BASE_URL` + `ANTHROPIC_AUTH_TOKEN`. In practice backup providers publish their own model ids (Z.AI wants `glm-4.6`), so carrying the primary's model across would trade a quota error for an unknown-model error.

**2. The swap is applied in the main process at the spawn IPC choke point,** not threaded through the ~20 renderer call sites that build spawn payloads. That means Auto Run, Cue, tab naming and background synopsis inherit failover for free, and the next spawn surface someone adds cannot forget it.

**3. Runtime pin state is in-memory only.** This mirrors the existing retry engine's rule that a closed app should not silently keep routing prompts to a backup provider. Every agent comes back on its primary after a restart. The endpoint *list* persists on the session as usual.

**4. Fail-back is a lazy probe, not a timer.** After a configurable dwell time (default 60m), the next turn re-tests the primary; if it is still down the normal failover path moves the agent off again. A background timer would burn quota on an idle agent just to discover the window had not reopened.

**5. Off by default,** and the checkbox only appears once an endpoint exists. Failing over changes who sees the user's prompts and what the output costs, so it should be an explicit opt-in.

## Where it lives

| Layer | File |
| --- | --- |
| Pure model + selection logic | `src/shared/providerFailover.ts` |
| Main-process overlay registry | `src/main/process-manager/failover-overlay.ts` |
| Applied at spawn | `src/main/ipc/handlers/process.ts` |
| Renderer runtime + swap/fail-back | `src/renderer/stores/failoverStore.ts` |
| Retry-engine seam | `src/renderer/stores/retryStore.ts` |
| Config UI | `src/renderer/components/NewInstanceModal/AgentFailoverSection.tsx` |

## Testing

41 new tests across the three new modules, plus a round-trip test through the edit dialog so a dropped arg in the save chain cannot silently disarm the feature. Notably covered: the overlay reaching main **before** the resend spawns, a failed overlay write degrading to a plain retry rather than swallowing the turn, and endpoint exhaustion falling back to the normal quota wait.

Full suite green locally: 31,930 passed, 0 failed. `npm run lint`, eslint and prettier all clean.

## Open questions for review

- **Where should endpoint config live?** Per #139's comment thread the UX is "when you define or edit an agent, choose your primary provider, then define fall back providers". This PR does the **edit** half only - new agents start with no failover config. Adding it to the create flow means threading a 17th positional arg through the create chain too; happy to do that if wanted.
- **Credential storage.** Auth tokens land in the session record like any other `customEnvVars` value, i.e. plaintext in the store. That matches how users configure `ANTHROPIC_API_KEY` today, but a dedicated endpoint list is a more obvious target and may warrant keychain storage.
- **Docs.** Agent Resilience itself is not documented in `docs/` or the agent guides, so I left this undocumented to match. Say the word and I will write up both.
- **Naming.** #139 mentions a "Virtuoso"/"Ensemble" feature that would cover this use case. I could not find any such feature in the codebase, so I built this as a direct extension of Agent Resilience. If Ensemble is a planned broader design, this may want to fold into it rather than ship standalone.

## Summary by CodeRabbit

* **New Features**
  * Added configurable provider failover with ordered backup endpoints for agent sessions.
  * Added endpoint validation, model overrides, environment variables, and automatic return-to-primary behavior.
  * Integrated failover into retries with brief handover delays and endpoint switching.
  * Added failover configuration to the agent editor, including saving and restoring settings.
  * Improved credential and environment handling when switching providers.
  * Standardized duration displays across the app and CLI with clearer human-readable formats.

* **Tests**
  * Expanded coverage for failover behavior, validation, retries, overlays, configuration persistence, and duration formatting.

### CODE REVIEW

**Comment by pedramamini** [comment]
@reachrazamair your call, but I want to flag this one hard: **I do not think it should merge in its current state**, and one of the findings is a credential-exposure issue rather than a correctness nit.

Branch has been updated against main (the red `test` job was the stale-base localStorage mock, not this diff). 1,436 additions across 20 files - the largest PR in the queue.

### The one I would not ship

**Primary credentials reach backup hosts** (`src/shared/providerFailover.ts:156`). When a backup endpoint specifies only `ANTHROPIC_BASE_URL`, the env merge preserves whatever `ANTHROPIC_AUTH_TOKEN` / `ANTHROPIC_API_KEY` was inherited from the session, the global shell, or the process environment. The spawned CLI then presents **the user's primary Anthropic credential to a third-party endpoint**. Greptile traced this through overlay merging into both local and SSH spawn environments.

Failover means "send my traffic somewhere else when Anthropic is down." A URL-only backup entry is the most natural thing a user would configure, and it is exactly the case that leaks the key. The overlay needs to treat auth as all-or-nothing per endpoint: if a backup supplies a base URL, it must supply its own credential or the primary credential must be stripped, not inherited.

### Two teardown paths that are dead code

I verified these directly rather than taking the bot's word:

```
git grep -n "clearAllFailoverOverlays" pr/1334 -- src/ | grep -v test
  -> only the definition at failover-overlay.ts:72

git grep -n "clearFailover(" pr/1334 -- src/ | grep -v test | grep -v failoverStore.ts
  -> no results
```

Both functions are defined, documented as the mechanism that prevents stale pins, and **never called**. Consequences:
- Disabling failover or deleting its endpoints leaves the agent pinned to the backup (`useSessionLifecycle.ts:228` updates `failoverConfig` but never clears the renderer state or the main-process overlay).
- A renderer reload resets the renderer store while main stays pinned, so prompts keep routing to the backup with no UI indication.

That is the feature's off switch not being wired up. Combined with the credential finding, a user who tries failover once and turns it off can keep shipping prompts and their primary key to a backup provider indefinitely.

### Also worth resolving

- **Primary model survives failover** (`process.ts:312`): a truthy guard leaves `sessionCustomModel` in the spawn payload when the backup omits its optional model override, so the backup gets a primary-provider model id and rejects the turn.
- **Blank env values clobber real ones** (`process.ts:307`): the UI seeds `ANTHROPIC_AUTH_TOKEN: ''` by default, and the overlay merge does not filter blanks - inconsistent with `resolveFailoverEnv`, which does skip them.
- **Failover fires too eagerly** (`retryStore.ts:311`): `scheduleRetryForError` schedules failover for any retryable error including short-lived overload/5xx/throttling blips. The module's own docs say waiting is preferable there. Restricting to the token-exhaustion long-wait case matches the stated design.

### My recommendation

Split it. The failover *plumbing* is good work and the feature is one we want. But the credential handling and the teardown wiring are not polish items - they are the difference between a feature that fails safe and one that fails open. I would fix the credential isolation and wire both clear paths before anything lands, then treat the model-override and eagerness items as follow-ups.

Happy to take the credential-isolation piece myself if you would rather not context-switch onto it - just say the word.

reachrazamair:
> @reachrazamair your call, but I want to flag this one hard: **I do not think it should merge in its current state**, and one of the findings is a credential-exposure issue rather than a correctness nit.
> 
> Branch has been updated against main (the red `test` job was the stale-base localStorage mock, not this diff). 1,436 additions across 20 files - the largest PR in the queue.
> 
> ### The one I would not ship
> **Primary credentials reach backup hosts** (`src/shared/providerFailover.ts:156`). When a backup endpoint specifies only `ANTHROPIC_BASE_URL`, the env merge preserves whatever `ANTHROPIC_AUTH_TOKEN` / `ANTHROPIC_API_KEY` was inherited from the session, the global shell, or the process environment. The spawned CLI then presents **the user's primary Anthropic credential to a third-party endpoint**. Greptile traced this through overlay merging into both local and SSH spawn environments.
> 
> Failover means "send my traffic somewhere else when Anthropic is down." A URL-only backup entry is the most natural thing a user would configure, and it is exactly the case that leaks the key. The overlay needs to treat auth as all-or-nothing per endpoint: if a backup supplies a base URL, it must supply its own credential or the primary credential must be stripped, not inherited.
> 
> ### Two teardown paths that are dead code
> I verified these directly rather than taking the bot's word:
> 
> ```
> git grep -n "clearAllFailoverOverlays" pr/1334 -- src/ | grep -v test
>   -> only the definition at failover-overlay.ts:72
> 
> git grep -n "clearFailover(" pr/1334 -- src/ | grep -v test | grep -v failoverStore.ts
>   -> no results
> ```
> 
> Both functions are defined, documented as the mechanism that prevents stale pins, and **never called**. Consequences:
> 
> * Disabling failover or deleting its endpoints leaves the agent pinned to the backup (`useSessionLifecycle.ts:228` updates `failoverConfig` but never clears the renderer state or the main-process overlay).
> * A renderer reload resets the renderer store while main stays pinned, so prompts keep routing to the backup with no UI indication.
> 
> That is the feature's off switch not being wired up. Combined with the credential finding, a user who tries failover once and turns it off can keep shipping prompts and their primary key to a backup provider indefinitely.
> 
> ### Also worth resolving
> * **Primary model survives failover** (`process.ts:312`): a truthy guard leaves `sessionCustomModel` in the spawn payload when the backup omits its optional model override, so the backup gets a primary-provider model id and rejects the turn.
> * **Blank env values clobber real ones** (`process.ts:307`): the UI seeds `ANTHROPIC_AUTH_TOKEN: ''` by default, and the overlay merge does not filter blanks - inconsistent with `resolveFailoverEnv`, which does skip them.
> * **Failover fires too eagerly** (`retryStore.ts:311`): `scheduleRetryForError` schedules failover for any retryable error including short-lived overload/5xx/throttling blips. The module's own docs say waiting is preferable there. Restricting to the token-exhaustion long-wait case matches the stated design.
> 
> ### My recommendation
> Split it. The failover _plumbing_ is good work and the feature is one we want. But the credential handling and the teardown wiring are not polish items - they are the difference between a feature that fails safe and one that fails open. I would fix the credential isolation and wire both clear paths before anything lands, then treat the model-override and eagerness items as follow-ups.
> 
> Happy to take the credential-isolation piece myself if you would rather not context-switch onto it - just say the word.

Agree with all of it. The credential leak and the two functions that never get called are real problems and need fixing before this merges.

Let's split it like you said. Fix the credential leak and get those two functions wired up first, then handle the model, blank fields, and retry timing after.

If you can take the credential fix, go ahead. I will fix the two functions that clear it.

reachrazamair (changes requested):
Requesting changes, credential leak and the two failover clear functions never being called, both confirmed in the comment above.

pedramamini:
@reachrazamair taking the credential half as you suggested. Pushed. **The two teardown functions are still yours** - I have not touched `clearFailoverOverlay` / `clearFailover`.

### What the leak actually was

Worse than a merge-order nit. The credential can reach the child from **three independent layers**:

1. the agent's own `customEnvVars`,
2. the global shell vars from Settings,
3. the inherited `process.env` of whatever shell launched Maestro.

The old code only merged (1). Even a perfect merge there leaves (2) and (3) intact, because `buildChildProcessEnv` starts from `{ ...process.env }`. **A merge cannot express a removal.**

### The fix

Auth is now all-or-nothing per endpoint. A backup that redefines `ANTHROPIC_BASE_URL` is pointing the agent at a different operator, so any credential it does not supply itself is stripped:

- `failoverUnsetEnvKeys(endpointEnv)` (new, in `shared/providerFailover.ts`) returns the credential keys to remove. It returns keys to REMOVE, not keys to set, precisely because the value may not be in the object being merged.
- `buildChildProcessEnv` takes `unsetEnvKeys` and applies it **last**, after every layer. That also covers the Windows branch at `process.ts:922`, which re-materializes `process.env` into `customEnvVars` via `buildExpandedEnv` - a strip done any earlier would be undone there.
- SSH is covered by the `resolveFailoverEnv` half, since the remote script only exports what is in `customEnvVars`.

A backup with no credential now fails to authenticate. That is the correct outcome - it is what the user configured - and the stripped keys are logged so it is diagnosable rather than mysterious.

### One thing worth flagging

The old behavior was **encoded as a passing test**. `providerFailover.test.ts:119` was named *"skips blank endpoint values so a half-filled row cannot clobber a working var"* and asserted `resolved?.ANTHROPIC_AUTH_TOKEN === 'primary-token'` for an endpoint that redirects the base URL. That is the leak, asserted as correct. Rewritten, plus cases for each of the three layers.

Also routed the spawn handler through `resolveFailoverEnv` instead of an inline spread - the shared helper had **zero call sites** and was drifting as dead code while the real merge lived in `process.ts`.

Local: `tsc` clean on all four configs, 1,904 tests green across `main/process-manager/`, `main/ipc/handlers/`, and `shared/providerFailover`.

@greptileai @coderabbitai please re-review the credential path specifically.

PR #1341, opened 2026-08-02
fix(sentry): repair rc IPC handler registration, bridge socket path, and routing noise

pedramamini:
Sentry triage on the `rc` channel. Four issues, all confirmed against current rc code and releases 0.18.4-RC / 0.18.5-RC.

## MAESTRO-YA - duplicate IPC handler aborts startup registration (the important one)

`Error: Attempted to register a second handler for 'director-notes:getRichOverviewStats'` - 9 events, first seen ~19h after the back-merge landed, release **0.18.5-RC (= current `package.json` at HEAD)**.

`070c2eee3 merge: back-merge origin/main into rc` (2026-07-31) resolved `db3c65946` by keeping **both** branches' copies of the rich-stats handler:

- line 651 - rc's version: tallies `AGENT`-typed history entries (`b6a052c39` gave cross-agent consults their own entry type)
- line 778 - main's version: hardcodes `agentEntryCount = 0` with a stale `// Generate AI synopsis via batch-mode agent` comment

`ipcMain.handle` throws on the second registration of a channel, and that throw escapes `registerDirectorNotesHandlers` into `setupIpcHandlers()` (no try/catch). Everything after it is skipped: `director-notes:generateSynopsis`, plus the **~40 `register*Handlers` calls that follow at `index.ts:3338+`** - cue, agents, process, persistence, system, claude, group chat, plugins, coworking, windows, stats, ssh-remote, filesystem, notifications and the rest.

Removed the main-side duplicate. The rc version is the correct one for this branch.

## MAESTRO-WH - coworking bridge unusable when userData is deep

`Error: listen EINVAL: invalid argument .../coworking.sock` - 9 events, `channel:rc`.

`getBridgeSocketPath()` returned `<userData>/coworking.sock` unconditionally on POSIX, but `sockaddr_un.sun_path` is a fixed-size array: **104 bytes on macOS/BSD, 108 on Linux**. The reported path is 123 bytes, so `bind()` rejected the address and coworking silently stayed unavailable for that install for good.

Over the limit, the socket now falls back to a short temp-dir path keyed by the same userData SHA-1 the Windows named pipe already uses (Windows solved this exact problem by hashing). The bridge and the agent-facing `COWORKING_SOCKET_ENV_VAR` both resolve through the one function, so they stay in sync automatically.

## MAESTRO-YB / MAESTRO-YC - `sshRemotes.find is not a function`

Surfaced as failed `git:status` and `git:numstat` invokes, `channel:rc`, release 0.18.4-RC.

`settings.json` is a plain file in userData that users hand-edit and sync tools rewrite, and electron-store only substitutes the `[]` default when the key is **absent** - not when it holds a non-array. A malformed value took down every caller of the canonical `getSshRemoteById()`, including the Right Bar's git polls on agents that use no SSH remote at all.

Treat an unusable value as "no remotes configured" and `logger.warn` once so the bad file stays diagnosable rather than silently swallowed.

Note: five other sites read `sshRemotes` with their own inline copy of this lookup (`marketplace.ts`, `agents.ts`, `autorun.ts`, `web-server-factory.ts`, `ssh-remote-resolver.ts`). None of them has reported a crash, so they are left alone here - flagging for a future dedup pass.

## MAESTRO-M4 - `Group chat not found` reported as a crash

Participants keep running after the user deletes their group chat, so the exit that fires later routes into a chat `loadGroupChat` can no longer find. The exit listener already recovers (participant marked done, buffer cleared) **and retries once through the fallback path**, so a single deleted chat produced two Sentry events per participant.

Follows the existing `isExpectedGroomingFailure` convention: a narrow predicate, applied only at the two exit-listener catches. Deliberately does not match `Participant '...' not found` (the chat still exists) or `Group chat not found after participant update` (that one means a broken write).

## Tests

19 new assertions across 4 files. Every one confirmed failing against the unfixed source per the usual vacuity check (`git stash push -- <source only>`): **14 red before the fix, 54 green after**.

The MAESTRO-YA test needed two angles - the existing suite's `ipcMain` mock is a `Map`, so a duplicate registration just overwrites the entry and goes unnoticed (which is why the bug shipped with tests passing). One test asserts on the raw call list, another re-runs registration against a mock that throws like the real `ipcMain`.

## Validation

- `npx tsc --noEmit`: **0 errors**
- Affected suites (`coworking/`, `stores/`, director-notes, git, groupChat, process-listeners, coworking-injection): **621 passed, 0 failed**
- ESLint: 7 pre-existing errors in `examples/plugins/agent-flow/main.js`, untouched by this branch

## Investigated and deliberately skipped

- **MAESTRO-JG / S4** (`MarketplaceFetchError: Network error fetching manifest: fetch failed`) - **already fixed, just unshipped.** The `sentryFilters.ts` rule matches, and the LinkedErrors haystack fix that made it live is `6e5236a1c` (2026-07-14). `git merge-base --is-ancestor 6e5236a1c v0.18.4-RC` = false, and all events are on 0.18.4-RC or stable 0.17.3. Re-fixing correct code was avoided.
- **MAESTRO-Y5** (`sessions:setMany` EBADF) - 1 event. `EBADF` on `open` can indicate a real fd leak, so adding it to the existing ENOSPC/EACCES/EPERM IPC-write rule could mask a genuine bug. Needs more signal first.
- **MAESTRO-Q0** (`gh pr list` failed) - per-issue breakdown is `channel:stable`, release 0.17.3. Not an rc issue.
- **MAESTRO-Q2** (1783 occ, `maestro-p --status`) - deferred a 6th time; still needs a human call on maestro-p exit semantics.
- **MAESTRO-RS** (GLIBC_2.38 / better-sqlite3) - Linux build-infra, still needs a human.
- Native/GPU crashes (SG, QG, TM, SZ, HQ, Y4, Y7, Y8, Y9, E1), `Window unresponsive` (62), renderer crashes (5A), minified `vendor-react` issues with no app frames (Y2/Y3 - and Y2 is the `[Splash]` twin of Y3), spawn EPERM/EINVAL/EFTYPE (NM, R0, X4 - genuine spawn signal).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved socket path handling across macOS, Linux, and Windows, including fallback paths for platform limits.
  * Prevented malformed SSH remote settings from causing lookup errors.
  * Reduced unnecessary error reporting for expected deleted group-chat failures.
* **Tests**
  * Added coverage for socket resolution, IPC registration, group-chat error handling, and malformed SSH remote configurations.


pedramamini:
Rebased onto current rc and fixed the one finding.

**The red Windows shards resolve themselves.** Both were failing on `execFile.test.ts > resolves with the spawn error code when the binary is missing` (expected `'ENOENT'`, got `1`). That test is not touched by this branch, and rc fixed it in `7724dc08f` ("Fix error handling race in execFileStreaming", 2026-08-03), which landed *after* this branch's base. The merge picks it up.

**Greptile's finding, fixed.** The `Array.isArray` guard stopped `sshRemotes.find is not a function`, but the same corrupted `settings.json` also produces `[null, {...}]` or `['remote-1']`, where reading `.id` off an entry throws identically. `getSshRemoteById` now skips non-object entries, warns once with a distinct detail string, and still finds a valid remote sitting alongside junk. Four new cases in `getters.test.ts` cover null/string/undefined entries plus the mixed-array case.

Local: `tsc` clean on all configs, 245 tests green across `stores/getters`, `coworking`, `directorNotes-richStats`, `process-listeners`, and `utils/execFile`.

pedramamini:
Correction to my previous comment: the rebase fixed the Windows failure I had looked at (`execFile.test.ts`, resolved by rc's `7724dc08f`), but a **second, different** Windows failure was hiding behind it - this one in this branch's own new test file.

`coworking-socket-path.test.ts` stubs `process.platform` to `darwin`/`linux`, but `path.join` stays win32 on a Windows runner, so the module produced `\Users\dev\...\coworking.sock` where the test expected forward slashes.

Fixed at the source rather than in the test: `getBridgeSocketPath` now builds with `path.posix.join`. The win32 branch returns a named pipe and exits above this point, so every path that reaches the join is a Unix domain socket path, which is POSIX by definition - `path.join` was only ever emitting POSIX separators in production anyway. Being explicit also keeps the `sun_path` byte count honest, and the temp-dir fallback still resolves correctly under a win32 `path.dirname` (verified: `path.win32.dirname('C:\\...\\Temp/x.sock')` returns `C:\...\Temp`).

Local: `tsc` clean, 142 tests green across `src/__tests__/main/coworking/`.

PR #1347, opened 2026-08-04
fix(thinking): drop stale thinking chunks that outlive their turn's answer

chr1syy:
Closes finding T1.

## Problem

A thinking log could survive **below** the final answer, duplicating it. Two writers hit the same `tab.logs` array on different schedules:

| Writer | Path | Schedule |
| --- | --- | --- |
| thinking chunks | `useAgentThinkingListener` | `requestAnimationFrame`, ~16ms |
| final answer | `useBatchedSessionUpdates` | batched flush, 200ms |

The mid-turn clear point lives only in the slow writer. The fast writer's non-continuation branch appended unconditionally, so a chunk buffered just before a flush was applied *after* it, landing below the answer it duplicated.

## Fix

Refuse to append a thinking log when the tab's last log is already the turn's `stdout` answer - a late chunk for an answered turn is stale by definition.

Deliberately NOT changed: the 200ms flush interval (documented as load-bearing for input latency), and claude-code's forward-all-partials behavior in `StdoutHandler` (its comment records that gating on `isReasoning` was already tried and silenced the thinking display for ordinary turns).

## Testing

Scoped tests only, per repo convention. Adds 100 lines to `useAgentThinkingListener.test.tsx` covering the race directly: buffered chunk, stdout flush first, then the rAF fires - asserts no thinking log is appended after the answer. Also asserts the normal path and continuation coalescing still work.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented outdated thinking messages from reappearing after an answer is displayed.
  * Preserved new thinking updates received after intermediate answers.
  * Ensured correct thinking behavior in standard and sticky modes, including after prompts or tool activity.
  * Improved handling of batched thinking and answer updates.

* **Tests**
  * Added regression coverage for stale thinking messages, intermediate responses, and message coalescing.

### DIFF
--- a/src/__tests__/renderer/hooks/agent/internal/useAgentThinkingListener.test.tsx
+++ b/src/__tests__/renderer/hooks/agent/internal/useAgentThinkingListener.test.tsx
@@ -129,12 +129,29 @@ describe('useAgentThinkingListener', () => {
 	// Regression: the fast (rAF) writer must not append a thinking log below the
 	// turn's answer, which the slow (200ms batched) writer flushed in between.
 	describe('stale straggler after the turn answer landed', () => {
-		function makeLog(source: LogEntry['source'], text: string, id: string): LogEntry {
-			return { id, timestamp: 1, source, text };
+		function makeLog(
+			source: LogEntry['source'],
+			text: string,
+			id: string,
+			timestamp = 1
+		): LogEntry {
+			return { id, timestamp, source, text };
 		}
 
-		it('drops the chunk when the last log is the turn stdout answer', () => {
-			const logs = [makeLog('user', 'prompt', 'l1'), makeLog('stdout', 'final answer', 'l2')];
+		/**
+		 * Reproduce the real interleaving: the chunk is buffered FIRST, then the
+		 * 200ms batch flushes the answer, then the rAF fires. The stdout log must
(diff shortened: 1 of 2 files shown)


chr1syy:
Thanks - the P1 is real and is now fixed in `ca2937d3e`.

You're right that a source-only guard cannot distinguish the turn's final answer from an intermediate assistant message. Since claude-code and factory-droid stream at message granularity, an intermediate message flushes as `stdout` mid-turn, and the guard silenced every later reasoning chunk for the rest of that turn.

The distinguishing signal is **when the chunk was buffered relative to the stdout log**:

- buffered BEFORE the stdout landed -> it belonged to the block the inline clear point already removed, so it is stale
- buffered AFTER -> it is new reasoning following an intermediate message, and must be appended

The rAF buffer now records `bufferedAt` (first chunk of the coalesced batch wins) and the guard drops only when `lastLog.timestamp >= bufferedAt`.

Test changes worth calling out: the two original stale-straggler cases stamped the answer at `timestamp: 1`, which predates the buffer - so they were actually encoding your intermediate-message scenario rather than the race. They now place the answer after buffering via a `landAnswerAfterBuffering` helper, which is the real interleaving. Added a case covering reasoning that arrives after an intermediate stdout message, and verified it fails against the old source-only guard.

PR #1348, opened 2026-08-04
fix(transcript): persist the scroll offset alongside the at-bottom flag

chr1syy:
Closes finding Y1.

## Problem

A deliberately scrolled-up transcript snapped back to the live bottom after an agent swap, instead of holding the position the user chose. This is the inverse of J1/O1 (#1323), which were about failing to *reach* the bottom.

`useTerminalOutputScroll` restores a saved offset only when **both** `initialScrollTop > 0` and `initialIsAtBottom === false` are persisted. The two halves were written on different schedules:

- `onAtBottomChange` fires synchronously on a transition
- `onScrollPositionChange` is debounced 200ms, and #1323 deliberately **drops** the pending save on unmount

So a user who scrolled up and swapped within ~200ms persisted `isAtBottom: false` with no matching `scrollTop`. The restore was skipped and the mount-time bottom jump won.

## Fix

Persist the offset alongside the synchronous at-bottom transition, so both halves are written together and can never disagree. This deliberately does **not** flush the debounce on unmount - #1323 rejected that explicitly, and its reasoning still holds.

## Testing

Scoped tests only. Adds 208 lines to `useTerminalOutputScroll.test.ts`, including a test that pins the race by unmounting before the 200ms debounce fires. Existing #1323 follow-the-bottom coverage and the #1140 scroll-event guard bookkeeping are untouched.

## Human verification

Confirmed on a local RC build: following the bottom, swapping away and back, still lands at the live bottom (the O1 behaviour must not regress).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transcript scroll-position preservation when switching agents after scrolling up.
  * Restored saved scroll offsets, including position 0, when returning to a transcript.
  * Kept saved position and bottom-of-transcript state synchronized during scrolling.
  * Prevented unintended position persistence when the user has not manually scrolled.

### DIFF
--- a/src/__tests__/renderer/components/TerminalOutput/useTerminalOutputScroll.test.ts
+++ b/src/__tests__/renderer/components/TerminalOutput/useTerminalOutputScroll.test.ts
@@ -675,5 +675,50 @@ describe('scrolled-up persistence across unmount (Y1)', () => {
 			expect(restored.result.current.isAtBottom).toBe(false);
 			expect(restored.result.current.autoScrollPaused).toBe(true);
 		});
+
+		// Review follow-up: the restore gate used to require a POSITIVE offset, so
+		// the one position a user reaches by scrolling all the way up - offset 0
+		// with `isAtBottom: false` - was unrestorable and snapped back to the live
+		// bottom. Zero is a real saved position, not a missing one.
+		it('restores a transcript left scrolled to the absolute top (offset 0)', () => {
+			const restoreRef = { current: makeRestoreContainer(9000) };
+			const restored = renderHook(() =>
+				useTerminalOutputScroll({
+					scrollContainerRef: restoreRef,
+					initialScrollTop: 0,
+					initialIsAtBottom: false,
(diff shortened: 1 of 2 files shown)


chr1syy:
Good catch - fixed in `2f08827d4`.

Scrolling to the absolute top of an overflowing transcript persists `scrollTop: 0` with `isAtBottom: false`, which is a perfectly ordinary deliberate position. Requiring `initialScrollTop > 0` made that one spot unrestorable, so returning to a tab left at the very top snapped it to the live bottom - the exact Y1 symptom this PR set out to fix.

`initialIsAtBottom !== false` is the real gate; the offset only needs to exist, so the check is now `>= 0`.

Two tests added: offset 0 with the flag `false` restores to the top and pauses auto-scroll, and offset 0 with the flag absent (a legacy tab that never persisted it) still falls through to the mount-time bottom jump - so the other half of the gate cannot silently regress. Verified the first fails against the old `> 0` gate.

PR #1349, opened 2026-08-04
fix(tabs): allow an agent to have zero AI tabs

### LINKED ISSUE #1062: Allow agents to have no open session tabs
## Summary

Maestro appears to require that a new session tab is always open. Users should be able to close all tabs for an agent and leave the tab bar empty until they choose to create a new tab.

## Desired behavior

Users should be able to have no open session tabs for an agent. When they need a new tab, they should be able to press the plus button and choose what kind of tab/session they want to open.

## Current behavior

A new session tab appears to always have to remain open.

## Why this matters

Some workflows do not need a default new session tab sitting open. Letting the agent have zero tabs keeps the workspace cleaner and gives users explicit control over when and what to open next.

## Environment

- App: Maestro
- Area: Session tab management
- Date reported: 2026-05-31

pedramamini:
Closes #1062

## Problem

Closing the last AI tab always spawned a fresh replacement, so an agent could never sit at zero AI tabs. As @jeffscottward put it in the issue, that leaves you with a dead tab you cannot get rid of:

> I hate when I have 3 or 4 open and one of them is essentially a dead tab. Bugs the hell out of me. Forced to give up screen real estate. Also I can't clear out my tabs for the day when I'm done.

## Approach

This keeps the original intent behind the sticky tab (@pedramamini in the issue: "i left this on purpose so that folks don't have to find how to create a new tab and can easily always at least chat") while unblocking the complaint.

The replacement tab is now created **only when the agent would otherwise be left with no tabs at all**:

- A brand new agent still always opens with a chat tab, and closing its only tab still hands you a fresh one. Nobody can end up staring at a workspace with no obvious way forward.
- Once you have other tabs open (terminal / file / browser), the last AI tab closes cleanly and the agent sits at **zero AI tabs**. The tab strip and its `+` menu stay on screen, so you explicitly choose what to open next.

## Changes

Core behavior in `closeTab()` (`src/renderer/utils/tabHelpers.ts`), plus the supporting fixes needed to make the newly reachable zero-AI-tab state actually safe:

| File | Why |
| --- | --- |
| `utils/tabHelpers.ts` | Only auto-create the replacement when no other tab kind survives. Clear `activeTabId` when no AI tab remains - a dangling id made a later switch back to AI mode render an input area bound to a tab that no longer exists. Guard the unread-filter and out-of-sync fallbacks, which indexed `updatedTabs[i]` and would throw on an empty array. |
| `components/MainPanel/MainPanel.tsx` | The tab strip was gated on `aiTabs.length > 0`. At zero AI tabs that hid the entire strip **including the `+` button**, stranding the user in whatever view was last active. Now gated on any tab kind. |
| `hooks/session/useSessionRestoration.ts` | Empty `aiTabs` was treated as data corruption: on every restart it rebuilt a default tab, flagged the agent `error`, and wiped `filePreviewTabs`/`browserTabs`/`unifiedTabOrder`. Restricted to sessions with no tabs whatsoever; a missing array is still normalized to `[]`. |
| `hooks/remote/useRemoteIntegration.ts` | Bailed out on empty `aiTabs`, so remote/mobile clients kept rendering tabs the user had already closed. |
| `QuickActionsModal/commands/tabCommands.ts` | Dropped the now-inaccurate `(creates new tab)` subtext from Close All Tabs. |

`performCloseAllTabs` needed no change: it snapshots tab ids up front, so it ends at zero AI tabs when other tabs exist and still yields one blank tab otherwise.

## Testing

- 5 new cases in `tabHelpers.test.ts` covering: zero AI tabs beside a terminal tab, beside a browser tab, under the unread filter (the crash path), closing a non-active sole tab, and the preserved fresh-tab behavior when nothing else is open.
- Full suite: **32,546 passed**, 0 failures.
- `npm run lint` (all three tsconfigs), ESLint, and Prettier clean.

Verified on CI matrix rather than local single-OS pass before merge.

## Summary by CodeRabbit

* **Bug Fixes**
  * Closing the last AI tab no longer creates an unnecessary replacement when file, terminal, or browser tabs remain.
  * Tab selection and active-state handling remain consistent when AI tabs are closed.
  * Sessions with no AI tabs are restored correctly when other restorable tabs exist.
  * Remote views stay synchronized after all AI tabs are closed.
  * Closing the final tab still creates a fresh AI tab when no other tabs remain.

* **Improvements**
  * The tab bar remains available when non-AI tabs are open.
  * Close-tab actions now use clearer wording for one versus multiple tabs.

### DIFF
--- a/src/__tests__/renderer/hooks/useSessionRestoration.test.ts
+++ b/src/__tests__/renderer/hooks/useSessionRestoration.test.ts
@@ -494,6 +494,46 @@ describe('restoreSession — Corruption recovery', () => {
 		expect(restored!.state).toBe('error');
 	});
 
+	// Zero AI tabs is only survivable if some other tab actually comes back.
+	// A terminal with no startup command is dropped during restoration, so
+	// counting the raw array here would skip recovery and leave no tabs at all.
+	it('recovers when the only remaining tab is a non-persistent terminal', async () => {
+		const session = createMockSession({
+			aiTabs: [],
+			activeTabId: null,
+			terminalTabs: [{ id: 'term-1', name: 'Terminal', pid: 0, state: 'idle' }] as any,
+		});
+		const { result } = renderHook(() => useSessionRestoration());
+
+		let restored: Session;
+		await act(async () => {
+			restored = await result.current.restoreSession(session);
+		});
+
+		expect(restored!.aiTabs).toHaveLength(1);
+		expect(restored!.state).toBe('error');
+	});
+
+	it('leaves zero AI tabs alone when a terminal with a startup command persists', async () => {
+		const session = createMockSession({
+			aiTabs: [],
+			activeTabId: null,
+			terminalTabs: [
+				{ id: 'term-1', name: 'Terminal', pid: 0, state: 'idle', startupCommand: 'npm run dev' },
+			] as any,
+		});
+		const { result } = renderHook(() => useSessionRestoration());
+
+		let restored: Session;
+		await act(async () => {
+			restored = await result.current.restoreSession(session);
+		});
+
+		expect(restored!.aiTabs).toHaveLength(0);
+		expect(restored!.terminalTabs).toHaveLength(1);
+		expect(restored!.state).not.toBe('error');
+	});
+
 	it('sets up unifiedTabOrder for recovered session', async () => {
 		const session = createMockSession({ aiTabs: [], activeTabId: null });
 		const { result } = renderHook(() => useSessionRestoration());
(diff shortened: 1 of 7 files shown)


pedramamini:
Fixed Greptile's blocking finding.

The corruption check counted `session.terminalTabs?.length` on the raw persisted array, but restoration drops every terminal without a startup command. So an agent whose only remaining tab was a plain terminal passed the "has tabs" check, skipped recovery, and *then* lost that terminal, landing on exactly the zero-tab state the check exists to prevent.

Both the check and the restoration filter now go through one `terminalTabSurvivesRestart` predicate, so they cannot disagree about how many tabs the agent is about to have. Two tests: a non-persistent terminal recovers, a terminal with a startup command correctly keeps zero AI tabs.

Local: `tsc` clean, 83 tests green in `useSessionRestoration.test.ts`, 275 in `tabHelpers.test.ts`.

PR #1351, opened 2026-08-05
fix(context): stop the context gauge latching at 0% after an accumulated turn

chr1syy:
Closes finding Q1 (severity high). **Bottom of a 3-PR stack** - see the stack note at the end.

## Problem

The context gauge and Context Details popover read `0%` on a session that is demonstrably deep into its context: 2.4M input tokens against a 1M window, but `Context Tokens 0` and `Usage 0%`.

A claude-code *turn* fans out into many internal API calls and the parser sums them, so one tool-heavy turn legitimately reports more input than the window. The intended mitigation is to fall back to the last known-good percentage, but that fallback could never arm:

1. `estimateContextUsage` returns `null` on overflow.
2. The `null` branch only estimates growth when a baseline already exists (`if (currentUsage > 0)`), so a session whose **first** usage event overflows never establishes one.
3. `calculateContextDisplay` then reads `fallbackPercentage = 0`, which **passes** its `fallbackPercentage >= 0` guard, and returns `0 tokens / 0%` flagged `trustworthy: true`.
4. Because the frame is flagged trustworthy, `lastGoodRef` caches `0` as a good value. The gauge is pinned at 0% for the life of the session.

Step 3 is the ugly part: the `trustworthy` flag exists specifically to stop bogus readings (#762), and here it launders a zero into a "good" value.

## Fix

- Require `fallbackPercentage > 0`. A zero fallback is the *absence* of a measurement, not a measurement of zero, so the frame is now correctly untrustworthy and the caller holds its previous value.
- Stop the last-good cache from storing a zero-fallback overflow frame.
- Establish a first baseline from claude-code's last-call occupancy snapshot, attached as `absoluteUsage` in the main process and consumed in the renderer. This is the structurally correct fix: the renderer cannot un-sum what the parser already added together, so the per-call figure has to come from the parser.

## Testing

Scoped tests only. The highest-value regression: `calculateContextDisplay` with `raw > contextWindow` and `fallbackPercentage = 0` must return `trustworthy: false` (it returned `true` with `0/0`). Plus parser-level coverage for the occupancy snapshot and renderer coverage for the baseline bootstrap.

## Stack

1. **this PR** - Q1, the numerator collapsing to zero
2. `fix/p1-context-window-precedence` - P1, which *denominator* the gauge divides by
3. `fix/r1-context-overlimit-display` - R1, how over-limit context renders

They all edit the same block of `useAgentUsageListener.ts`, so they must land bottom-up in that order. A fourth layer (S1, Context Timeline persistence) follows later.

## Summary by CodeRabbit

- **New Features**
  - Added more accurate context-usage snapshots for Claude sessions.
  - Context gauges and timelines now reflect the latest available occupancy data.
  - Initial snapshots can populate context tracking, including custom context windows.
  - Timeline history preserves per-turn usage while displaying current occupancy.

- **Bug Fixes**
  - Preserved the last reliable context values when overflow data reports zero usage.
  - Prevented unavailable zero-value fallbacks from appearing as trustworthy measurements.
  - Maintained aggregated usage totals independently from per-call context snapshots.

### DIFF
--- a/src/__tests__/main/parsers/claude-output-parser.test.ts
+++ b/src/__tests__/main/parsers/claude-output-parser.test.ts
@@ -271,6 +271,165 @@ describe('ClaudeOutputParser', () => {
 		});
 	});
 
+	describe('absoluteUsage occupancy snapshot (finding Q1)', () => {
+		// A fresh parser per test: the snapshot is per-turn state on the instance,
+		// and the shared `parser` above has been fed unrelated messages.
+		const assistantMessage = (usage: Record<string, number>, extra: Record<string, unknown> = {}) =>
+			JSON.stringify({
+				type: 'assistant',
+				session_id: 'sess-q1',
+				message: {
+					id: 'msg-1',
+					role: 'assistant',
+					content: [{ type: 'text', text: 'working' }],
+					usage,
+				},
+				...extra,
+			});
+
+		const resultMessage = (modelUsage: Record<string, unknown>) =>
+			JSON.stringify({
+				type: 'result',
+				result: 'done',
+				session_id: 'sess-q1',
+				modelUsage,
+				total_cost_usd: 0.5,
+			});
+
+		it('attaches the LAST assistant call usage, not the sum and not the first call', () => {
+			const p = new ClaudeOutputParser();
+
+			p.parseJsonLine(
+				assistantMessage({
+					input_tokens: 2,
+					output_tokens: 90,
+					cache_read_input_tokens: 15007,
+					cache_creation_input_tokens: 9468,
+				})
+			);
+			p.parseJsonLine(
+				assistantMessage({
+					input_tokens: 3,
+					output_tokens: 40,
+					cache_read_input_tokens: 20000,
+					cache_creation_input_tokens: 100,
+				})
+			);
+			p.parseJsonLine(
+				assistantMessage({
+					input_tokens: 2,
+					output_tokens: 12,
+					cache_read_input_tokens: 24475,
+					cache_creation_input_tokens: 109,
+				})
+			);
+
+			// The CLI's own per-model sum across all three calls - the quantity that
+			// overflows the window and pinned the gauge at 0%.
+			const event = p.parseJsonLine(
+				resultMessage({
+					'claude-opus-5': {
+						inputTokens: 7,
+						outputTokens: 142,
+						cacheReadInputTokens: 59482,
+						cacheCreationInputTokens: 9677,
+						contextWindow: 1000000,
+					},
+				})
+			);
+
+			const usage = p.extractUsage(event!);
+			// Top-level fields stay the CLI's turn totals (token spend).
+			expect(usage?.inputTokens).toBe(7);
+			expect(usage?.cacheReadTokens).toBe(59482);
+			// The snapshot is the third call only: 2 / 12 / 24475 / 109.
+			expect(usage?.absoluteUsage).toEqual({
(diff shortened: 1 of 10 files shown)


pedramamini:
@chr1syy Thanks for this one, and for the write-up. The problem statement is unusually good: walking the four steps from `estimateContextUsage` returning `null` all the way to `lastGoodRef` caching a zero as known-good made this reviewable in one pass.

I verified the fix independently rather than taking the AI reviews at face value:

- **The `> 0` guard is the right call.** Treating a zero fallback as the absence of a measurement rather than a measurement of zero is correct, and the edge case you called out in the comment (a session legitimately under 0.5%) really does render identically, so nothing is lost. The `useContextWindow` regression test is the one that convinced me, since it exercises the actual latching path rather than just the pure function.
- **Sourcing the baseline from the parser is the structurally correct fix.** You are right that the renderer cannot un-sum what the parser already added together. Pulling the last main-transcript call's `message.usage` is a genuine occupancy figure, and excluding subagent messages matters, since a Task subagent's own window says nothing about the main transcript.
- **Turn-boundary reset ordering is correct.** `this.lastCallOccupancy = undefined` lands after `extractUsageFromRaw`, so the result event still carries the snapshot and the next turn starts clean. The `does not leak a snapshot across turns` test locks that in.
- **Uncapped bootstrap is consistent.** Skipping `maxEstimate` in the new branch matches the non-null `contextPercentage` path, which is not capped either. A measurement should be allowed to trip the yellow warning, an extrapolation should not.
- **Blast radius is contained.** `absoluteUsage` now being populated for claude-code only reaches the gauge and the timeline, both intentionally updated here. The `isOutputOnlyDelta` behavior change is acknowledged in your comment and covered by a test.

Locally: the four scoped suites pass (152 tests), `tsc --noEmit` is clean, and CI is green on both the ubuntu and windows legs.

**One note, not blocking.** CodeRabbit's inline comment about `contextPercentage` being computed before `resolvedWindow` applies the session override is a real inconsistency. With a nonzero reported window and a different `customContextWindow`, `estimateContextUsage` ranks the reported window first while the timeline point uses the configured window, so the two displays can disagree. That said, it is pre-existing (the call sat at the same position before this PR and the denominator precedence is unchanged), and it is precisely the scope of `fix/p1-context-window-precedence`, the next layer of your stack. Leaving it here keeps this PR focused on the numerator. Just make sure P1 picks it up so it does not fall through the crack between the two.

CodeRabbit's docstring coverage warning is a metric artifact given how heavily commented this diff already is. Ignoring it.

Merges cleanly into `rc`, no conflicts. Approving. Since this is the bottom of the stack, please land it before P1 and R1.

chr1syy:
Fixed in `516813342`.

You're right - `contextPercentage` was computed before `resolvedWindow` existed, so it divided by the event's **reported** window while the timeline point stored a percentage against the **configured** one. That is exactly the gauge/timeline disagreement #1221 fixed, reintroduced through ordering.

Moved the computation below `resolvedWindow` and hand that window in. `estimateContextUsage` prefers `stats.contextWindow` when positive and otherwise falls back to the capability snapshot and the static table, so the precedence is now identical on both surfaces.

One note on the test, since it caught me out: my first attempt passed against the *unfixed* code. An event carrying `absoluteUsage` reaches the D1 bootstrap branch, which already divides by `resolvedWindow` and therefore masks the defect. The committed test uses a plain in-window turn with no snapshot, so it exercises `estimateContextUsage` directly - verified failing against the pre-fix ordering.

PR #1361, opened 2026-08-09
fix(sentry): stop unreadable provider transcripts from paging Sentry

pedramamini:
## What

Field triage of `channel:rc` Sentry issues. Two live issues fixed, both the same root cause at read sites that PR #1311 missed.

| Issue | Site | Events |
| --- | --- | --- |
| [MAESTRO-YH](https://smash-labs.sentry.io/issues/MAESTRO-YH) | `storage/claude-session-storage.ts` -> `parseSessionFile` | 3 |
| [MAESTRO-YJ](https://smash-labs.sentry.io/issues/MAESTRO-YJ) | `ipc/handlers/claude.ts` -> `getProjectStats` / `getGlobalStats` | 3 |

## Root cause

Transcripts under `~/.claude/projects` belong to the Claude CLI, not to Maestro. A tree we can't read (restrictive umask, a home dir owned by another user) or a file deleted between the directory listing and the read is environmental, never a Maestro bug. Every discovery pass reported the errno.

The reporting is also amplified: one unreadable tree pages a Sentry event *per transcript, per refresh*. The field events came from a single Linux user whose `~/.claude/projects/-home-marce-backup/` tree was not readable.

#1311 established the `EXPECTED_SESSION_READ_ERROR_CODES` carve-out and wired it into the two `getGlobalStats` loops in `agentSessions.ts`, but three sibling sites walking the same files were left bare:

- `claude-session-storage.ts` `parseSessionFile` - the `readFile` behind `listSessions`/`listSessionsPaginated`. Only `RangeError` was carved out.
- `claude-session-storage.ts` `listSessions` - the outer `fs.stat` catch on the same failure path. Caught by grepping the whole file for other `captureException` calls on the same path, per the lesson from #1311's review.
- `claude.ts` both stats loops.

## Where the predicate lives

Moved from `ipc/handlers/agentSessions.ts` to a dependency-free `utils/session-read-errors.ts`. `src/main/storage` cannot import from `src/main/ipc/handlers` without a require cycle (`handlers -> agents -> storage`). `agentSessions.ts` re-exports it so existing importers and the main-side copy still resolve, keeping the rc->main merge surface small.

## Deliberately not widened

- `EMFILE` still reports - descriptor exhaustion is our bug, not the environment's.
- `RangeError` (oversized transcript) keeps its own separate carve-out.
- Remote/SSH read paths untouched; they use `isExpectedRemoteError` and don't surface errno objects.

## Triaged and skipped

Already fixed at HEAD but not yet shipped in `0.18.4-RC` (the release all these events carry) - verified with `git merge-base --is-ancestor`:

- **MAESTRO-YG** (same EACCES cluster, `agentSessions.ts`) - guard landed 2026-08-06, hours after the events.
- **MAESTRO-RR** (codex usage 5xx), **MAESTRO-JS** (grooming spawn), **MAESTRO-XJ** (marketplace fetch), **MAESTRO-V9** (batch-mode JSON).

Covered by open PRs, not duplicated here: **MAESTRO-XS**/**X4** (#1310), **MAESTRO-YA**/**WH**/**YB**/**YC**/**M4** (#1341). Both are still unmerged.

Not our bug: **MAESTRO-RE**/**YF** (bmad prompts `ENOENT` from broken/partial installs - the files are verifiably in the shipped artifacts), **MAESTRO-W2/W4/W5/W7/W8** (preload failed to load, so `window.maestro` is undefined - masking it with optional chaining would hide a corrupt install).

No obvious fix: **MAESTRO-Q2** (`maestro-p --status`, needs a human call on exit semantics), native/GPU crashes, minified `vendor-react` stacks with no sourcemap, genuine spawn failures.

## Testing

- New `session-read-errors.test.ts` covering the predicate and the re-export.
- Call-site regression tests at both fixed sites, each asserting the errno codes stay silent **and** that `EMFILE` still reports.
- Every new test confirmed red against the unfixed source (`git stash` the source only, re-run, confirm failure).
- `npx tsc --noEmit` clean; full main-process suite green (376 files, 9828 tests).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of missing or inaccessible session transcripts and session data.
  * Expected file access issues now show local warnings without generating unnecessary error reports.
  * Unexpected failures continue to be logged and reported for investigation.

* **Documentation**
  * Added guidance for handling expected session-transcript read failures.


pedramamini:
Rebased onto current rc and addressed what the rebase surfaced.

**The merge was not mechanical.** rc's transcript parse cache (#1363) replaced the inline listing this branch was guarding, and the cache's own read sites went in unguarded. So the branch as written would have merged cleanly and guarded code that no longer runs. `isExpectedSessionReadError` now covers:

- `statProjectSessionFiles` -> `statSessionFile`: the `fs.stat` races the directory listing, same boundary as before.
- `parseSessionFilesCached`'s parse callback: the file can vanish between stat and read.
- `SessionInfoCache.loadScope`: this one guarded only `ENOENT`, so an unreadable `userData` dir paged Sentry once per scope per listing. Same MAESTRO-YH class, different directory.

**Test harness:** the `fs/promises` mock in `claude-session-storage.test.ts` only exported a `default`, but the cache uses a namespace import, so its `fsp.readFile` was undefined. Mock now exports both shapes, and an `electron` mock supplies the `app.getPath` the cache needs at construction.

Local: `tsc` clean on all four configs, 266 tests green across `src/__tests__/main/storage/`, `session-read-errors`, and `ipc/handlers/claude`.

pedramamini:
Greptile was right, and I had the guard one level too shallow.

The EACCES/EPERM that matters most fails in `statProjectSessionFiles`' `readdir` - the project directory itself - which reported and rethrew *before* any transcript file was touched. That is precisely the another-user's-`~/.claude` case this PR exists for, so the guards were covering the inner reads while the outer one kept paging.

Fixed:

- `statProjectSessionFiles` still **throws** on a directory it cannot read (reporting an unreadable tree as "zero sessions" would silently hide the user's transcripts, which is the wrong trade) but the expected environmental set is a local warn instead of a Sentry event.
- `agentSessions.ts`'s `getAllNamedSessions` aggregator got the same guard. It walks every provider's transcript tree on every call, so an unreadable one was re-paging from the caller side even once the storage layer went quiet.

Three new cases: EACCES and EPERM on the directory read stay local and still reject, EIO still reports.

Local: `tsc` clean on all configs, 1,915 tests green across `main/storage/`, `main/ipc/handlers/`, and `session-read-errors`.

PR #1369, opened 2026-08-10
Refactor Symphony handlers into modular domain structure

reachrazamair:
## Summary by CodeRabbit

- **New Features**
  - Added a complete contribution workflow, including repository setup, branching, draft pull requests, completion, cancellation, and manual credits.
  - Added dashboard views for active and completed contributions, contributor statistics, progress metrics, and streaks.
  - Added discovery and GitHub synchronization for repositories, issues, documents, pull requests, merges, closures, and recovery.
  - Added caching and resilient handling of temporary network or service errors.

- **Bug Fixes**
  - Improved document filename safety and duplicate-name handling.
  - Strengthened contribution validation and cancellation cleanup safeguards.

- **Documentation**
  - Updated contributor and system guides to reflect the reorganized Symphony documentation and handler structure.


pedramamini:
@reachrazamair Thank you for taking this on! Splitting a 3,315-line handler module into domain-scoped files is exactly the kind of unglamorous work that pays off every time someone has to touch Symphony afterward, and the domain boundaries you picked (discovery / dashboard / lifecycle / start / finish / sync) are sensible ones.

I verified this is behavior-preserving rather than taking the refactor on faith:

- **All 20 `symphony:*` IPC channels are preserved exactly**, registered once each, no additions or omissions between `rc` and this branch.
- **Normalized line-by-line diff of the old module against the concatenated new modules** comes back clean. The only deltas are `function X` becoming `export function X` for the shared helpers, the import/export plumbing between the new files, and `enrichWithStars()` taking `app` as an explicit parameter instead of capturing it from the enclosing closure. All equivalent.
- **`createSafeSend(getMainWindow)` now runs in three modules instead of one.** That is safe here: it takes a window *getter* and returns a closure, holding no state of its own, so per-module construction is equivalent to the single shared instance.
- **No module-level mutable state** in either the old or new code, so the split cannot fragment an in-memory cache. Symphony state and cache both round-trip through disk.
- Clean merge into `rc`, and all CI is green on both matrix legs.

## On the CodeRabbit findings

CodeRabbit raised 16 inline comments, including three marked Critical (`contributionId` path traversal, `doc.name` path traversal, unvalidated `localPath` reaching `fs.rm`). **These are all real, but every one of them is pre-existing in `symphony.ts` on `rc`, not introduced here.** I confirmed this by comparing occurrence counts of each flagged pattern between the base file and your new modules: `contributionId` 60/60, `doc.name` 8/8, `rev-list` 2/2, `recursive: true, force: true` 4/4, `AbortSignal` 0/0. The `readState()` catch block that CodeRabbit flagged for swallowing parse errors is byte-identical to the version on `rc`. Greptile reached the same conclusion independently, with 5/5 confidence and no changed-code regression.

So I am **not** asking you to fix them in this PR. Per our scope discipline guideline, folding security hardening into a pure code-motion change is how a reviewable refactor turns into an unreviewable one, and it would destroy the clean move-only diff that makes this PR verifiable in the first place.

The findings are worth acting on, just separately. The path-traversal trio in particular deserves its own PR where the fix and its tests can actually be reviewed on their merits. If you would like to pick that up as a follow-up we would welcome it, but it is entirely optional and in no way a condition of merging this one.

Approving. Nice work.

reachrazamair:
> @reachrazamair Thank you for taking this on! Splitting a 3,315-line handler module into domain-scoped files is exactly the kind of unglamorous work that pays off every time someone has to touch Symphony afterward, and the domain boundaries you picked (discovery / dashboard / lifecycle / start / finish / sync) are sensible ones.
> 
> I verified this is behavior-preserving rather than taking the refactor on faith:
> 
> * **All 20 `symphony:*` IPC channels are preserved exactly**, registered once each, no additions or omissions between `rc` and this branch.
> * **Normalized line-by-line diff of the old module against the concatenated new modules** comes back clean. The only deltas are `function X` becoming `export function X` for the shared helpers, the import/export plumbing between the new files, and `enrichWithStars()` taking `app` as an explicit parameter instead of capturing it from the enclosing closure. All equivalent.
> * **`createSafeSend(getMainWindow)` now runs in three modules instead of one.** That is safe here: it takes a window _getter_ and returns a closure, holding no state of its own, so per-module construction is equivalent to the single shared instance.
> * **No module-level mutable state** in either the old or new code, so the split cannot fragment an in-memory cache. Symphony state and cache both round-trip through disk.
> * Clean merge into `rc`, and all CI is green on both matrix legs.
> 
> ## On the CodeRabbit findings
> CodeRabbit raised 16 inline comments, including three marked Critical (`contributionId` path traversal, `doc.name` path traversal, unvalidated `localPath` reaching `fs.rm`). **These are all real, but every one of them is pre-existing in `symphony.ts` on `rc`, not introduced here.** I confirmed this by comparing occurrence counts of each flagged pattern between the base file and your new modules: `contributionId` 60/60, `doc.name` 8/8, `rev-list` 2/2, `recursive: true, force: true` 4/4, `AbortSignal` 0/0. The `readState()` catch block that CodeRabbit flagged for swallowing parse errors is byte-identical to the version on `rc`. Greptile reached the same conclusion independently, with 5/5 confidence and no changed-code regression.
> 
> So I am **not** asking you to fix them in this PR. Per our scope discipline guideline, folding security hardening into a pure code-motion change is how a reviewable refactor turns into an unreviewable one, and it would destroy the clean move-only diff that makes this PR verifiable in the first place.
> 
> The findings are worth acting on, just separately. The path-traversal trio in particular deserves its own PR where the fix and its tests can actually be reviewed on their merits. If you would like to pick that up as a follow-up we would welcome it, but it is entirely optional and in no way a condition of merging this one.
> 
> Approving. Nice work.

Thanks for the thorough review. I will keep those path-traversal fixes on my radar and open a separate PR for them next.

PR #1379, opened 2026-08-12
fix(codex): surface event_msg error events instead of swallowing them

### LINKED ISSUE #1378: Bug: Codex agent stops responding after remote compact returns 404
## Summary
Codex agent stops responding after remote compact returns 404

## Environment
- Maestro version: 0.18.5-RC

## Steps to Reproduce
1. Start a Codex session in Maestro.
2. Continue the conversation until remote compaction is triggered.
3. Observe the 404 error from the /codex/responses/compact endpoint.
4. Agent stops responding; retry the prompt.
5. Same 404 recurs — session is permanently broken.
6. Start a new session — works normally.

## Expected Behavior
A failed remote compaction should be recoverable — the session should continue, surface a retryable error, or fall back to local/client-side compaction. The agent should not become permanently unresponsive.

## Actual Behavior
Mid-conversation, Codex emits: `Error running remote compact task: unexpected status 404 Not Found: {"detail":"Not Found"}, url: https://chatgpt.com/backend-api/codex/responses/compact, cf-ray: a2a245597813f94c-DUS, request id: 83f44823-691b-4721-b52b-b7faf5e78617`. After this, the agent stops responding entirely. Retrying reproduces the identical 404 — the session never recovers.

## Additional Context
Failure is session-scoped: new sessions work fine, so it is not an install/auth-wide outage. Endpoint returning 404 (`{"detail":"Not Found"}`) suggests the upstream compact route may have moved or been removed, but the Maestro-side concern is the lack of graceful degradation — a 404 from compaction should not permanently disable the agent. Unclear whether compaction was auto-triggered or manual; user did not confirm.

## Screenshots / Recordings
None

pedramamini:
Closes #1378

## Problem

A Codex session went silent mid-conversation after remote compaction returned a 404:

```
Error running remote compact task: unexpected status 404 Not Found: {"detail":"Not Found"},
url: https://chatgpt.com/backend-api/codex/responses/compact
```

The 404 itself is upstream (OpenAI's compaction service), and we can't fix that. What we *can* fix is the reason it looked like the session was permanently dead: **Maestro never surfaced the error at all.**

## Root cause

Codex v0.111.0+ reports turn failures through an `event_msg` envelope that carries its text in `payload.message`:

```json
{"type":"event_msg","payload":{"type":"error","message":"Error running remote compact task: ..."}}
```

Both sides of the Codex parser only understood the legacy shapes (top-level `error`, `turn.failed`):

- `transformEventMsg` handled only `agent_message` and `token_count`, so an error payload fell through to the default `{type: 'system'}` - invisible.
- `detectErrorFromParsed` read only `obj.error`. For an `event_msg` there is no top-level `error`, so it returned `null` and **no `agent-error` was ever emitted**.

Net effect: no error entry in the transcript, no error state on the session, no hint to the user. The agent just stopped talking. Because the session's stored state was unchanged, retrying hit the same upstream 404, which is why it read as "permanently broken" while a fresh session worked.

## Changes

- **`codex-output-parser.ts` / `transformEventMsg`** - map `payload.type === 'error'` to an error event.
- **`codex-output-parser.ts` / `detectErrorFromParsed`** - read the `event_msg` error payload, and fall back to a top-level `message` when `error` is absent instead of discarding the event as `"Unknown error"` (this also fixes the bare `{"type":"error","message":"..."}` exec-JSON shape).
- **`error-patterns.ts`** - classify remote-compact failures and `unexpected status <code>` as **recoverable** `network_error`s with actionable text, so the session presents as retryable rather than fatal.

### On `stream_error`

Deliberately **not** treated as an error, and this is the one judgement call worth flagging. Codex emits `stream_error` for transient failures it retries itself (`retrying 1/5`). Raising one would pause a session that is about to recover on its own, and it would burn `StdoutHandler`'s once-only `errorEmitted` latch - hiding the *real* error if the retries did run out. It's surfaced as ordinary progress text instead, so the user can still see the retries happening.

## Scope

This does not make failed compaction succeed, and it does not add a fallback to local compaction - both are Codex-side behaviour. It fixes the Maestro-side defect the reporter actually asked about: the lack of graceful degradation, where a server-side failure produced silence instead of a surfaced, retryable error.

## Testing

- 6 new `detectErrorFromParsed` cases (including the verbatim 404 message from the report) and 3 new `event_msg` parse cases.
- `src/__tests__/main/parsers` + `src/__tests__/main/process-manager`: **711 passed**.
- `tsc --noEmit`, `eslint`, and `prettier --check` clean on the changed files.

Only CI can confirm the Windows leg.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of Codex server errors and retryable stream interruptions.
  * Classifies conversation compaction failures and upstream HTTP errors as recoverable network issues.
  * Distinguishes authentication failures, including forbidden-access errors, from other server issues.
  * Displays clearer error details, including fallback messages and relevant HTTP status codes.
  * Preserves partial response text when recoverable stream errors occur.
  * Prevents empty or unrelated error messages from being surfaced incorrectly.


pedramamini:
Both findings were real. Fixed.

**401/403 misclassified as transient.** `network_error` is matched before `auth_expired` (see `ERROR_TYPES_BY_HIT_FREQUENCY`), so the generic `unexpected status <code>` pattern claimed auth failures and told the user to retry a bad API key - advice that can never work. Those statuses are now excluded from the generic matcher.

Excluding 403 alone would have made it match nothing at all: the Codex auth bank only had `/unauthorized/i`, which catches 401's reason phrase but not "Forbidden". So 403 gets its own auth entry - a valid key without access to that model, organization, or region - rather than falling through unclassified.

**Retry diagnostics could become the answer.** `stream_error` was accumulating into `streamedText`, which `ExitHandler` emits as the final result when a turn ends without a result message. A turn that retried and then died handed the user `stream error: ...; retrying 1/5` as the agent's response. Marked `isReasoning` - the existing flag for text that shows as progress but is deliberately excluded from the result buffer - rather than inventing a new mechanism.

`tsc` clean on all four configs, 697 tests green across parsers and process-manager.

PR #1380, opened 2026-08-12
feat(images): one right-click menu for every image, wired nowhere

pedramamini:
Right-click any image anywhere in Maestro and get **Copy Image**, **Save to Project...**, and **Save As...**, at the pointer. Markdown embeds, transcript attachment thumbnails, the lightbox, agent-authored inline SVG, Mermaid charts - all of it.

## Why this shape

The menu used to hang off individual components. Every new image surface shipped without it until somebody remembered to wire one up, which is exactly how `main` and `rc` ended up with two different implementations covering two different subsets of surfaces.

`ImageContextMenuHost` is now mounted once in `App.tsx` and runs **one delegated `contextmenu` listener** that resolves the image from the click target. A surface is covered the moment it renders an image. There is no per-surface wiring left to forget - the old wiring is deleted from chat markdown, `MermaidRenderer`, `TerminalOutput`, and `LightboxModal`, and `useImageContextMenu` goes with it.

Net effect on wiring code is negative despite adding two features.

## What is not an image

`resolveImageFromEvent()` skips three things, or the menu would be useless everywhere:

- anything inside a `[data-no-image-menu]` subtree (surfaces owning their own gestures, e.g. `AnnotatorCanvas`)
- lucide icons, which are `<svg>` but carry the `lucide` class
- anything under 32px rendered - favicons, inline badges

Menus that already handled the right-click (`LinkContextMenu`, `FileContextMenu`, terminal selection) call `preventDefault()`, so `defaultPrevented` keeps two menus from opening on one click.

## Save destinations

- **Save to Project...** opens `ImageDestinationModal`: folder, file name, SVG/PNG, live path preview. Defaults to `.maestro/diagrams/` so a diagram lands with the project that produced it. Works over SSH; an existing name gets a `-2` suffix rather than being overwritten.
- **Save As...** opens the native OS dialog (unchanged).

## Folded in from rc

Two fixes `rc` had been carrying alone, so the branches stop diverging here:

- `serializeSvg()` stamps the measured size onto the clone. A Mermaid chart sized `width="100%"` was rasterizing cropped at the browser's 300x150 default.
- The copy path reports whether an image or only markup/URL reached the clipboard, instead of always claiming an image.

`joinPath()` and `DIAGRAMS_DIR` move down into `shared/` where both branches can use them.

## Validation

- `tsc --noEmit` clean on `tsconfig.json`, `tsconfig.main.json`, `tsconfig.cli.json`
- `eslint` clean, `prettier --check` clean
- Full suite via the pre-push hook: **32,821 passed, 108 skipped, 0 failed**
- New coverage: `resolveImageFromEvent` filter cases (icons, favicons, opt-out subtree, click-on-child-node) and `saveImageToProject` (path building, custom folder, collision suffix, raster-vs-UTF-8 write routing, SSH id threading, write-failure surfacing)

Docs updated in `CLAUDE.md` and `UI-PATTERNS.md` to say "wire up nothing" rather than pointing at the deleted hook.

## Summary by CodeRabbit

- **New Features**
  - Added app-wide right-click actions for images and diagrams.
  - Copy images or text fallbacks with clear success and failure notifications.
  - Save images to a project with selectable folders, filenames, and formats.
  - Save SVG images as SVG or PNG with improved sizing and export quality.
  - Added automatic filename suggestions and duplicate-name handling.

- **Bug Fixes**
  - Improved image detection while excluding icons, tiny images, and annotation areas.

- **Documentation**
  - Updated image context-menu guidance for the unified experience.


pedramamini:
Thanks both - all four findings were real and are fixed in e8f0d80. CodeRabbit and Greptile independently flagged the first three, which is a good signal.

**Path traversal (both bots).** `relativeDir` and `fileName` are free text and `joinPath` preserves `..`, so `../../etc` escaped the project root - and on an SSH remote, escaped it there. Both inputs are now validated before any filesystem call: absolute paths and `..` segments refused, path separators refused in the file name. An escape throws rather than being clamped, since silently rewriting the path would drop the file somewhere the user did not ask for. Ordinary nested folders (`docs/img/diagrams`) still work, covered by a test.

**Extension mismatched the encoding (both bots).** Correct: `defaultExtensionFor(saveTarget)` was called without `sourceDataUrl`, so a JPEG seeded a `.png` name while `format: 'original'` wrote JPEG bytes. The extension is now derived from what `encodeForSave` actually produced and forced onto the final name, so neither direction can lie. `defaultExtensionFor` additionally reads the element's own `src` (query string stripped, `jpeg` normalized to `jpg`) rather than assuming png, so the seeded name is honest before the write too.

**Save destination changed with the agent (both bots).** Real race. The host read the live active session at confirm time, so switching agents with the modal open saved into the new agent's project on the new agent's host. The image, project root, SSH id, and session id are now captured together as a `PendingSave` when the menu opens, and that record drives the modal, the write, and the toast's `clickAction`. Regression test switches the active session between opening the modal and confirming, asserting the write still targets project A.

**`-100` overwrite (CodeRabbit).** Correct - the loop tested through `-99` then wrote `base-100` unchecked. Every candidate is now tested before selection, and exhausting all 100 throws instead of clobbering.

One thing I did not add: a test asserting the SVG-to-PNG rename direction. It routes through `svgToPngDataUrl`, and jsdom never fires `` onload for an SVG data URL, so the call hangs rather than failing. Left a comment at the call site noting why, consistent with the existing note about the rasterizing helpers being uncovered for the same reason.

Validation: tsc clean on all three configs, eslint and prettier clean, full suite 32,836 passed / 108 skipped / 0 failed.

PR #1383, opened 2026-08-13
fix(execfile): unambiguous options detection, forward env on stdin path

reachrazamair:
## Problem

CodeRabbit flagged a bug in `execFile.ts`'s `execFileNoThrow` during review of [#1317](https://github.com/RunMaestro/Maestro/issues/1317) (already merged), which added an `env` field to `ExecOptions`:

- The legacy/`ExecOptions` signature discriminator only checked whether a key named `input`/`timeout`/`env` was present, not its value's shape. A real environment variable can be named any of those, so a legacy bare-env call containing one got misread as the structured form, and its actual entries got dropped or mangled.

Follow-up review also caught that `{ timeout: undefined }` (valid `ExecOptions`) was still misread as a legacy env dict when relying on value-type checks alone.

## What Changed

`resolveExecOptions()` extracts signature detection into a pure function. Key presence alone is ambiguous (a real env var can be named `input`/`timeout`/`env`), and value type alone isn't enough either (`{ timeout: undefined }` is valid `ExecOptions`). The discriminator treats "every key present is a known `ExecOptions` field" as structured options - a real legacy env dict always carries neighbours like `PATH`/`HOME` - and keeps value-shape checks as a secondary signal.

`EXEC_OPTIONS_FIELDS` is derived from `Record<keyof ExecOptions, true>` so adding a field to the interface without updating the set is a compile error, not a silent misclassification.

## Testing

* Added signature-discrimination regression tests (reserved-name collisions, `{ timeout: undefined }`, structured form validation).
* Kept a real-process stdin+env regression test as a guard on that existing path (not new behavior in this PR).
* Verified red against pre-fix code.
* Full test suite passing, `tsc` and `eslint` clean.

### DIFF
--- a/src/__tests__/main/utils/execFile.test.ts
+++ b/src/__tests__/main/utils/execFile.test.ts
@@ -92,6 +92,24 @@ describe('execFile.ts', () => {
 		// the child_process mock only replaces execFile, not spawn.
 		const NODE = process.execPath;
 
+		// A lone `{ input }` with no other fields is a real call shape (git.ts
+		// pipes gist content to `gh` this way) - the discriminator fix for the
+		// legacy-env-named-"input" collision must not break it.
+		it('treats a lone { input } as ExecOptions and delivers it to stdin', async () => {
+			const { execFileNoThrow } = await import('../../../main/utils/execFile');
+
+			const result = await execFileNoThrow(
+				NODE,
+				['-e', 'process.stdin.on("data", (d) => process.stdout.write(d))'],
+				undefined,
+				{
+					input: 'hello from stdin',
+				}
+			);
+
+			expect(result.stdout).toBe('hello from stdin');
+		});
+
 		// Regression: when `input` is set, execFileNoThrow hands off to
 		// execFileWithInput, which never received or forwarded `env` - a caller
 		// passing { input, env } silently got process.env in the child instead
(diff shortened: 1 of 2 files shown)


pedramamini:
Thanks for the contribution @reachrazamair, this is a careful fix.

The `allKeysAreExecOptionsFields` check is the right insight: a real legacy env dict always carries neighbours like `PATH`/`HOME` alongside anything that collides with a reserved name, so "every key present is a known field" is a genuinely unambiguous signal in a way that key presence or value type alone is not. The doc comment is honest about where it still can't disambiguate (a lone `{ input: 'x' }` env dict), which is the right call. You also picked up Greptile's `{ timeout: undefined }` case with a regression test, which the value-type checks alone would have missed.

I checked every legacy bare-env call site in the repo (`cliDetection.ts`, `path-prober.ts`, `agents/detector.ts`, `maestro-cli-manager.ts`) and they all pass a full `getExpandedEnv()` / `process.env` copy, so none of them can trip the all-known-keys rule. Behaviour for `{}` is unchanged from before, so no regression there either.

Two small things before I approve:

**1. `EXEC_OPTIONS_FIELDS` is hand-maintained and can silently drift.** If someone adds a fourth field to `ExecOptions` and forgets to add it to the Set, the discriminator starts routing structured options back down the legacy path and swallows them as an environment - exactly the class of silent bug this PR exists to fix. Deriving the set from the type makes the compiler catch it:

```ts
/**
 * Keyed by `keyof ExecOptions` so adding a field to the interface without
 * listing it here is a compile error, not a silent misclassification.
 */
const EXEC_OPTIONS_FIELD_NAMES: Record<keyof ExecOptions, true> = {
	input: true,
	timeout: true,
	env: true,
};
const EXEC_OPTIONS_FIELDS = new Set(Object.keys(EXEC_OPTIONS_FIELD_NAMES));
```

**2. The second "What Changed" bullet is stale.** `execFileWithInput` already accepts and forwards `env`: both the `env?: NodeJS.ProcessEnv` parameter and `env` in the `spawn()` call are in this PR's merge base, and the diff doesn't touch that function. The regression test you added for it is still worth keeping, it just isn't covering new ground. Since the PR body becomes the squash commit message, could you trim that bullet so the history reads accurately? `resolveExecOptions()` is the real change here and it stands on its own.

Nothing else from me. CI is still running, so please give `test` and `lint-and-format` a look once they land, then ping me and I'll take another pass.

reachrazamair:
> Thanks for the contribution @reachrazamair, this is a careful fix.
> 
> The `allKeysAreExecOptionsFields` check is the right insight: a real legacy env dict always carries neighbours like `PATH`/`HOME` alongside anything that collides with a reserved name, so "every key present is a known field" is a genuinely unambiguous signal in a way that key presence or value type alone is not. The doc comment is honest about where it still can't disambiguate (a lone `{ input: 'x' }` env dict), which is the right call. You also picked up Greptile's `{ timeout: undefined }` case with a regression test, which the value-type checks alone would have missed.
> 
> I checked every legacy bare-env call site in the repo (`cliDetection.ts`, `path-prober.ts`, `agents/detector.ts`, `maestro-cli-manager.ts`) and they all pass a full `getExpandedEnv()` / `process.env` copy, so none of them can trip the all-known-keys rule. Behaviour for `{}` is unchanged from before, so no regression there either.
> 
> Two small things before I approve:
> 
> **1. `EXEC_OPTIONS_FIELDS` is hand-maintained and can silently drift.** If someone adds a fourth field to `ExecOptions` and forgets to add it to the Set, the discriminator starts routing structured options back down the legacy path and swallows them as an environment - exactly the class of silent bug this PR exists to fix. Deriving the set from the type makes the compiler catch it:
> 
> ```ts
> /**
>  * Keyed by `keyof ExecOptions` so adding a field to the interface without
>  * listing it here is a compile error, not a silent misclassification.
>  */
> const EXEC_OPTIONS_FIELD_NAMES: Record<keyof ExecOptions, true> = {
> 	input: true,
> 	timeout: true,
> 	env: true,
> };
> const EXEC_OPTIONS_FIELDS = new Set(Object.keys(EXEC_OPTIONS_FIELD_NAMES));
> ```
> 
> **2. The second "What Changed" bullet is stale.** `execFileWithInput` already accepts and forwards `env`: both the `env?: NodeJS.ProcessEnv` parameter and `env` in the `spawn()` call are in this PR's merge base, and the diff doesn't touch that function. The regression test you added for it is still worth keeping, it just isn't covering new ground. Since the PR body becomes the squash commit message, could you trim that bullet so the history reads accurately? `resolveExecOptions()` is the real change here and it stands on its own.
> 
> Nothing else from me. CI is still running, so please give `test` and `lint-and-format` a look once they land, then ping me and I'll take another pass.

Thanks @pedramamini  - both closed.

1. EXEC_OPTIONS_FIELDS is now derived from `Record<keyof ExecOptions, true>`, so adding a field without updating the set is a compile error.
2. Dropped the stale env-forwarding bullet from the PR body.

All green. Please take another pass when you have a moment.

PR #1384, opened 2026-08-14
fix(symphony): close three path traversal issues at the IPC boundary

reachrazamair:
Follow-up to #1369, which deliberately left these out to keep that refactor a clean code-motion diff.

**What changed**
- `contributionId` is validated as a single path segment before it is joined into the Symphony directory in `startContribution` and `createDraftPR`
- external document names are reduced to a bare file name before being joined onto the documents cache
- `symphony:cancel` confirms the stored `localPath` is this contribution's clone before removing it recursively

**Where I diverged from the review suggestions**
- CodeRabbit proposed containing `localPath` to the repositories directory. The renderer builds it as `workingDirectory || /tmp/symphony/...`, and `workingDirectory` is user-chosen in the Symphony modal, so containment would break the feature. The destructive operation is guarded instead.
- Rejecting document names containing separators would refuse real input. `parseDocumentPaths` takes the name from markdown link text, so an ordinary `[docs/architecture.md](...)` yields a name with a slash. Those are reduced, not refused.

**Behaviour notes**
- `docs/architecture.md` previously hit ENOENT inside the cache directory and was silently dropped, it now downloads correctly
- `cancel` skips cleanup when the stored path is not a clone and logs it, the contribution is still cancelled
- no guard in `symphony-runner`'s cleanup, it clones into that path itself in the same call so it provably owns the directory, and a half-finished clone still has to be removable

**Verification**
- Symphony suites 291 passing, up from 290
- integration failing set byte identical to baseline, the 24 failures there are pre-existing on rc
- lint, type-check and production build clean
- confirmed in the running app, `../../etc/passwd` is rejected where it previously reached the filesystem

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive contribution workflows, including repository setup, draft pull requests, completion, cancellation, synchronization, and recovery.
  * Added dashboards for active contributions, completed contributions, and contributor statistics.
  * Added repository and issue discovery with caching, progress tracking, and GitHub metadata.
  * Added safer handling for contribution IDs, document filenames, and repository cleanup.

* **Bug Fixes**
  * Improved cleanup behavior to preserve unrelated directories and handle invalid repositories safely.
  * Prevented document name collisions from overwriting files.

* **Documentation**
  * Updated Symphony handler architecture, lifecycle, and runner documentation.

### DIFF
--- a/src/__tests__/integration/symphony.integration.test.ts
+++ b/src/__tests__/integration/symphony.integration.test.ts
@@ -23,6 +23,7 @@ import {
 	registerSymphonyHandlers,
 	SymphonyHandlerDependencies,
 } from '../../main/ipc/handlers/symphony';
+import { toSafeDocumentFileName } from '../../main/ipc/handlers/symphony/shared';
 import {
 	REGISTRY_CACHE_TTL_MS,
 	ISSUES_CACHE_TTL_MS,
@@ -2667,6 +2668,82 @@ error: failed to push some refs to 'https://github.com/owner/protected-repo.git'
 
 			expect(result.success).toBe(false);
 			expect(result.error).toContain('GitHub');
+		});
+
+		// ----------------------------------------------------------------------
+		// contributionId is joined into the Symphony directory by startContribution
+		// and createDraftPR. The cases above only ever passed a clean ID, so these
+		// sinks had no coverage.
+		// ----------------------------------------------------------------------
+
+		it('should reject a contributionId containing traversal in startContribution', async () => {
+			const result = (await invokeHandler(handlers, 'symphony:startContribution', {
+				contributionId: '../../../evil',
+				sessionId: 'session-id-traversal',
+				repoSlug: 'owner/repo',
+				issueNumber: 1,
+				issueTitle: 'ID Traversal Test',
+				localPath: path.join(testTempDir, 'id-traversal-repo'),
+				documentPaths: [],
+			})) as { success: boolean; error?: string };
+
+			expect(result.success).toBe(false);
+			expect(result.error).toContain('Invalid contribution ID');
+		});
+
+		it('should reject a contributionId containing a separator in createDraftPR', async () => {
+			const result = (await invokeHandler(handlers, 'symphony:createDraftPR', {
+				contributionId: '../../etc/passwd',
+			})) as { success: boolean; error?: string };
+
+			expect(result.success).toBe(false);
+			// Asserting the specific message matters here: without validation this
+			// handler still fails, but with "metadata not found" after it has
+			// already built and read an out-of-tree path.
+			expect(result.error).toContain('Invalid contribution ID');
+		});
+
+		it('should accept the generated contribution ID formats', async () => {
+			// Positive control. A valid ID must pass validation and reach the real
+			// logic, otherwise the rejection tests above could pass simply because
+			// every ID is refused.
+			for (const id of ['contrib_abc123_def456', 'manual_42_1700000000000']) {
+				const result = (await invokeHandler(handlers, 'symphony:createDraftPR', {
+					contributionId: id,
+				})) as { success: boolean; error?: string };
+
+				expect(result.error).not.toContain('Invalid contribution ID');
+				expect(result.error).toContain('metadata not found');
+			}
+		});
+
+		// ----------------------------------------------------------------------
+		// doc.name is joined onto the documents cache directory when an external
+		// document is downloaded. validateContributionParams checks doc.path but
+		// never checked doc.name.
+		// ----------------------------------------------------------------------
+
+		// doc.name is the link text of a markdown link in the issue body, and it is
+		// joined onto the documents cache directory when an external document is
+		// downloaded. It is reduced to a bare file name at that join, so these
+		// exercise the reduction directly.
+		it('should reduce traversal in a document name to a bare file name', () => {
+			expect(toSafeDocumentFileName('../../../evil.json')).toBe('evil.json');
+			expect(toSafeDocumentFileName('..\\..\\evil.json')).toBe('evil.json');
+			expect(toSafeDocumentFileName('/etc/passwd')).toBe('passwd');
+			// Nothing usable is left, so the document is skipped rather than written.
+			expect(toSafeDocumentFileName('../../..')).toBeNull();
+			expect(toSafeDocumentFileName('.')).toBeNull();
+			expect(toSafeDocumentFileName('')).toBeNull();
+		});
+
+		it('should keep real issue-body document names working', () => {
+			// Regression guard. An ordinary markdown link such as
+			// [docs/architecture.md](https://...) yields a name with a separator,
+			// which must still resolve to a usable file name rather than be refused.
+			expect(toSafeDocumentFileName('docs/architecture.md')).toBe('architecture.md');
+			expect(toSafeDocumentFileName('My Design Doc.md')).toBe('My Design Doc.md');
+			expect(toSafeDocumentFileName('spec.md')).toBe('spec.md');
 		});
 	});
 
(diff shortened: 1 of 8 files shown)


pedramamini:
@reachrazamair Thanks for this one, and thanks in particular for the write-up. The "where I diverged from the review suggestions" section is exactly the right way to handle bot feedback, and both of those calls are correct: containing `localPath` would break user-chosen working directories, and refusing names with separators would refuse ordinary markdown link text. The three sinks are validated in the right place, before the join rather than after, and the positive-control test on the generated ID formats is a nice touch.

I checked the two bot reviews against the code. Most of what CodeRabbit flagged as "Major" is pre-existing in `rc`'s `symphony.ts` and just moved by the #1369 refactor riding along in this branch: the unchecked `git rev-list` exit code, the `getStats` orphan filter, the `readState`/`writeState` races, and the `repoSlug` values reaching GitHub URLs in `discovery.ts` are all present at `origin/rc:src/main/ipc/handlers/symphony.ts` today. Those are real, but they are not this PR's job. Please leave them alone here so this stays a tight security diff, and let's file them separately.

Also, CodeRabbit's "critical" call on `symphony.integration.test.ts:2688` is a false positive. The `})) as { success: boolean; error?: string };` is the normal close for the `(await invokeHandler(...))` cast and matches the surrounding cases. The file parses, and both ubuntu shards are green. No action needed there.

Two things in the genuinely new code before I'm happy, though:

**1. Blocker: the origin match in `isContributionClone` is a substring test, and it gates a recursive delete.**

`src/main/ipc/handlers/symphony/shared.ts:155`

```ts
return result.stdout.toLowerCase().includes(repoName.toLowerCase());
```

Both bots landed on this independently and they're right. The origin URL contains the host and the owner, not just the repository, so `repoName` can match on the wrong part of the string entirely. For `repoSlug` `owner/maestro`, an origin of `https://github.com/maestro/some-other-project.git` passes, and so does `git@host:other/maestro-tools.git`. That then reaches `fs.rm(..., { recursive: true })` on somebody else's checkout with their uncommitted work in it.

Given the guard exists precisely because `localPath` cannot be constrained, it needs to be exact. Compare the final segment of the origin, with the optional `.git` suffix stripped:

```ts
const originRepo = result.stdout
    .trim()
    .replace(/\.git$/i, '')
    .split(/[/:]/)
    .pop()
    ?.toLowerCase();
return originRepo === repoName.toLowerCase();
```

While you're in there: if `repoSlug` ever arrives without a `/`, `repoName` is undefined and the function returns `true` on the bare `.git` check alone. `ActiveContribution.repoSlug` is typed as required so this shouldn't happen in practice, but for a predicate authorizing a recursive delete I'd rather it fail closed than open.

**2. Should fix: flattening can silently overwrite a document.**

`src/main/services/symphony-runner.ts:208` and the matching loop in `contributionStart.ts:609`

Reducing to the basename is the right call, but two distinct references such as `docs/architecture.md` and `spec/architecture.md` now both resolve to `architecture.md`, and the second write wins. That's new behaviour: before this change the first one hit ENOENT and got dropped, so we traded a visible-ish drop for a silent overwrite. Tracking the names already used in the loop and disambiguating on collision (or just skipping with a `logger.warn` like the unusable-name branch does) would close it. A test with two references sharing a basename would be worth having.

One process note, not a code issue: this branch carries all of #1369, which is still open. If #1369 merges first this will need a rebase, and if this merges first #1369 becomes redundant. Whichever order you and I settle on is fine, just flagging it so it doesn't surprise us at merge time.

No merge conflicts against `rc`, and CI is green apart from one Windows shard still running. Ping me when the origin match is exact and I'll take another look.

reachrazamair:
> @reachrazamair Thanks for this one, and thanks in particular for the write-up. The "where I diverged from the review suggestions" section is exactly the right way to handle bot feedback, and both of those calls are correct: containing `localPath` would break user-chosen working directories, and refusing names with separators would refuse ordinary markdown link text. The three sinks are validated in the right place, before the join rather than after, and the positive-control test on the generated ID formats is a nice touch.
> 
> I checked the two bot reviews against the code. Most of what CodeRabbit flagged as "Major" is pre-existing in `rc`'s `symphony.ts` and just moved by the #1369 refactor riding along in this branch: the unchecked `git rev-list` exit code, the `getStats` orphan filter, the `readState`/`writeState` races, and the `repoSlug` values reaching GitHub URLs in `discovery.ts` are all present at `origin/rc:src/main/ipc/handlers/symphony.ts` today. Those are real, but they are not this PR's job. Please leave them alone here so this stays a tight security diff, and let's file them separately.
> 
> Also, CodeRabbit's "critical" call on `symphony.integration.test.ts:2688` is a false positive. The `})) as { success: boolean; error?: string };` is the normal close for the `(await invokeHandler(...))` cast and matches the surrounding cases. The file parses, and both ubuntu shards are green. No action needed there.
> 
> Two things in the genuinely new code before I'm happy, though:
> 
> **1. Blocker: the origin match in `isContributionClone` is a substring test, and it gates a recursive delete.**
> 
> `src/main/ipc/handlers/symphony/shared.ts:155`
> 
> ```ts
> return result.stdout.toLowerCase().includes(repoName.toLowerCase());
> ```
> 
> Both bots landed on this independently and they're right. The origin URL contains the host and the owner, not just the repository, so `repoName` can match on the wrong part of the string entirely. For `repoSlug` `owner/maestro`, an origin of `https://github.com/maestro/some-other-project.git` passes, and so does `git@host:other/maestro-tools.git`. That then reaches `fs.rm(..., { recursive: true })` on somebody else's checkout with their uncommitted work in it.
> 
> Given the guard exists precisely because `localPath` cannot be constrained, it needs to be exact. Compare the final segment of the origin, with the optional `.git` suffix stripped:
> 
> ```ts
> const originRepo = result.stdout
>     .trim()
>     .replace(/\.git$/i, '')
>     .split(/[/:]/)
>     .pop()
>     ?.toLowerCase();
> return originRepo === repoName.toLowerCase();
> ```
> 
> While you're in there: if `repoSlug` ever arrives without a `/`, `repoName` is undefined and the function returns `true` on the bare `.git` check alone. `ActiveContribution.repoSlug` is typed as required so this shouldn't happen in practice, but for a predicate authorizing a recursive delete I'd rather it fail closed than open.
> 
> **2. Should fix: flattening can silently overwrite a document.**
> 
> `src/main/services/symphony-runner.ts:208` and the matching loop in `contributionStart.ts:609`
> 
> Reducing to the basename is the right call, but two distinct references such as `docs/architecture.md` and `spec/architecture.md` now both resolve to `architecture.md`, and the second write wins. That's new behaviour: before this change the first one hit ENOENT and got dropped, so we traded a visible-ish drop for a silent overwrite. Tracking the names already used in the loop and disambiguating on collision (or just skipping with a `logger.warn` like the unusable-name branch does) would close it. A test with two references sharing a basename would be worth having.
> 
> One process note, not a code issue: this branch carries all of #1369, which is still open. If #1369 merges first this will need a rebase, and if this merges first #1369 becomes redundant. Whichever order you and I settle on is fine, just flagging it so it doesn't surprise us at merge time.
> 
> No merge conflicts against `rc`, and CI is green apart from one Windows shard still running. Ping me when the origin match is exact and I'll take another look.

Thanks Pedram, both done.

**1. Origin match.** It now compares the final segment of the origin with the optional `.git` stripped, so it handles both https and scp-style remotes. Your two cases are covered as tests, plus repo-name-in-host and repo-name-as-suffix. It also fails closed when `repoSlug` arrives without an owner, rather than returning `true` on the bare `.git` check.

I ran the guard in the app against real checkouts with uncommitted files in them: the three decoy origins survive, and the three genuine ones (including a fork-rewritten origin) still get cleaned up.

**2. Collisions.** Disambiguated rather than dropped, so `docs/architecture.md` and `spec/architecture.md` land as `architecture.md` and `architecture-2.md`. Both loops go through one helper. The test drives `startContribution` with two colliding references and asserts two distinct destinations, and I checked it goes red against the old behaviour.

Left the pre-existing findings alone as you asked, and agreed on the integration test false positive.

On the process note: I retargeted this PR's base to `refactoring`, so it now shows only the three security commits. It should auto-retarget to `rc` when #1369 lands, and I'll rebase onto `rc` then. Happy for #1369 to go first.

PR #1385, opened 2026-08-15
feat/provider-auth-recovery: 35 tasks across 2026-08-15-Provider-Auth-Recovery/Phase-01-Credential-Identity-Model, 2026-08-15-Provider-Auth-Recovery/Phase-02-Main-Probe-Layer-And-Startup-Pass +4 more

pedramamini:
## Auto Run Summary

**Documents processed:**
- 2026-08-15-Provider-Auth-Recovery/Phase-01-Credential-Identity-Model
- 2026-08-15-Provider-Auth-Recovery/Phase-02-Main-Probe-Layer-And-Startup-Pass
- 2026-08-15-Provider-Auth-Recovery/Phase-03-Renderer-State-And-Left-Bar-Surfacing
- 2026-08-15-Provider-Auth-Recovery/Phase-04-Auth-Recovery-Modal
- 2026-08-15-Provider-Auth-Recovery/Phase-05-Rewire-Existing-Surfaces-And-Auto-Resend
- 2026-08-15-Provider-Auth-Recovery/Phase-06-SSH-Remotes-Docs-And-Validation

**Total tasks completed:** 35

## Changes

- MAESTRO: stop the auth flow from claiming a verdict it never earned
- MAESTRO: document the provider auth feature for users and for the next agent
- MAESTRO: make the recovery modal say which machine it is signing in on
- MAESTRO: make the auth probe answer for the host that owns the credential
- MAESTRO: join the auth recovery units with an end-to-end test
- MAESTRO: add the Provider Accounts settings section
- MAESTRO: resume the prompts an expired login killed, after the user confirms
- MAESTRO: point the wizard's auth errors at the in-app recovery flow
- MAESTRO: open the auth recovery modal from the agent error modal
- MAESTRO: cover the auth recovery modal with component tests
- MAESTRO: open the auth recovery modal from the Left Bar, the toast, and the palette
- MAESTRO: clear every agent's auth error when a provider login succeeds
- MAESTRO: run the provider login command in a PTY the recovery modal owns
- MAESTRO: add the provider auth recovery modal shell
- MAESTRO: resolve per-provider login commands for auth recovery
- MAESTRO: cover provider auth store and Left Bar auth indicator with tests
- MAESTRO: announce logged-out identities with one throttled toast
- MAESTRO: surface provider auth state in the Left Bar
- MAESTRO: mark credential identity on auth_expired
- MAESTRO: add renderer provider auth store
- MAESTRO: add provider auth probe and startup pass tests
- MAESTRO: wire provider auth startup pass and IPC surface
- MAESTRO: add provider auth startup probe pass
- MAESTRO: add provider auth probe layer
- MAESTRO: add provider auth snapshot store
- MAESTRO: add credential identity model for provider auth
- MAESTRO: add provider auth architecture survey
- feat(queue): save a queued message edit with Cmd+Enter
- feat(preview): zoom the reading font in the file preview
- feat(tabs): hide terminal and browser tabs under the unread filter
- feat(queue): send a queued message now from the Execution Queue view
- fix(media): queue behind the loaded track instead of a leftover one
- fix(media): list what plays next in the queue, not what is already playing
- fix(media): keep playing through minimize, and stop listing the loaded track as history
- fix(chat): stop losing unsent drafts in AI tabs
- fix(media): play a chat media link in Maestro, not the OS default app
- feat(media): make minimize and close mean different things
- ci: catch rc silently falling behind main
- fix(opencode): offer a fresh session when a provider rejects stored history (#1307)
- feat(worktrees): run a post-create setup script in new worktrees (#1306)
- fix(codex): surface event_msg error events instead of swallowing them (#1379)
- fix(deep-links): guard percent-decoding against malformed input (#1373)
- feat(opencode): per-agent OpenCode agent selection (--agent) (#1308)
- fix(media): stop the now-playing pill from crowding the Left Bar header
- feat(agents): chooser for multiple detected binary paths (#1317)
- feat(resilience): Provider Failover to Anthropic-compatible backup endpoints (#1334)
- fix(process-manager): don't attribute a killed process's exit to its replacement (#1302)
- feat(media): show how long each queued file runs, and how much is left
- feat(media): size the player to whatever is playing, not to a remembered box
- feat(media): a queue that survives restarts, and a note in the header when the player is hidden

---
*This PR was automatically created by [Maestro](https://runmaestro.ai) Auto Run.*

## Summary by CodeRabbit

- **New Features**
  - Added provider account authentication monitoring, status indicators, and startup checks.
  - Added in-app sign-in and credential recovery for supported providers, including remote sessions.
  - Blocked prompts can be reviewed, discarded, or resent after authentication is restored.
  - Added Provider Accounts settings with account status, re-check actions, and startup-check control.
  - Added Quick Actions and notifications for authentication recovery.
  - Improved provider-specific error detection and recovery guidance.

- **Documentation**
  - Added provider authentication setup, recovery, troubleshooting, and architecture guidance.

- **Tests**
  - Added comprehensive authentication, recovery, settings, notification, and prompt-resumption coverage.


pedramamini:
## Merge measurement against `main` (dry run, 2026-08-20)

Measured, not estimated. A scratch worktree was created off `origin/main`, `git merge --no-commit --no-ff b472611b9` was run, the result recorded, then `git merge --abort` and the worktree removed. Nothing was merged, committed, or pushed.

### Distance

| | vs `origin/main` | vs `origin/rc` |
|---|---|---|
| branch ahead | 27 | 27 |
| branch behind | **33** | **960** |

Merge base: `ae452bc8a` (2026-08-15), five days old. The branch has not rotted against `main` - `rc` ran away from it. `main` is by far the cheaper target.

### What merges cleanly

**84 files touched. 71 auto-merge: 14,832 insertions, 1,432 deletions.** 35 files are brand new and land without a fight, including `docs/provider-auth.md`, `docs/architecture/provider-auth/design.md`, and `docs/architecture/provider-auth/survey.md`.

### What conflicts

**13 files, 28 hunks** - all in one subsystem (auth error recovery and the modal layer that surfaces it). Five of the thirteen are tests.

| file | hunks |
|---|---|
| `src/renderer/stores/modalStore.ts` | 4 |
| `src/__tests__/renderer/hooks/agent/internal/useAgentErrorListener.test.tsx` | 4 |
| `src/renderer/components/AppModals/AppAgentModals.tsx` | 3 |
| `src/__tests__/renderer/stores/agentStore.test.ts` | 3 |
| `src/renderer/components/QuickActionsModal/QuickActionsModal.tsx` | 2 |
| `src/renderer/hooks/agent/useAgentErrorRecovery.tsx` | 2 |
| `src/renderer/hooks/modal/useModalHandlers.ts` | 2 |
| `src/__tests__/renderer/hooks/useAgentErrorRecovery.test.ts` | 2 |
| `src/__tests__/renderer/hooks/useModalHandlers.test.ts` | 2 |
| `src/renderer/stores/agentStore.ts` | 1 |
| `src/renderer/hooks/agent/internal/useAgentErrorListener.ts` | 1 |
| `src/renderer/constants/modalPriorities.ts` | 1 |
| `src/__tests__/renderer/components/AppAgentModals.test.tsx` | 1 |

### Why it conflicts

**`1d9930042` and `940daf8bd` are not ancestors of this branch.** Those two commits ("re-authenticate a provider without leaving Maestro" and "scope the re-auth prompt to the provider, not the first agent") landed on `main` on Aug 18 and Aug 19, after this branch's Aug 15 merge base.

So `main` and this branch each carry an independent implementation of "your provider login expired." Every one of the 28 hunks is those two designs disagreeing. Resolving them is not a textual exercise - it means deciding which design survives inside `modalStore.ts`, `agentStore.ts`, `useAgentErrorRecovery.tsx`, and `modalPriorities.ts`, which every modal and every agent error path runs through. A wrong call there fails quietly, later, when a token actually expires.

### Recommendation on timing

**Merge after the `v0.17.4` cut, not before.**

`git tag --list v0.17.4` is empty - the tag is cut from `main` HEAD at press time. Merging now would put ~14.8k lines of never-shipped code inside the `v0.17.4` build, and `v0.17.4`'s release note for re-auth describes the *shipped* implementation, not this one. Cutting first costs nothing: against `main` this branch is only 33 behind and is not decaying.

pedramamini:
Merged `main` in, reconciled the duplicate auth system, and worked all 11 findings. Summary for the re-review below.

## The structural problem first

`main` shipped its own provider re-authentication while this branch was open (`1d9930042`, `940daf8bd`, `2b0e6923d`) - independently reaching the same conclusion this branch did, that recovery keys by CREDENTIAL rather than by whichever agent failed first. Merged naively, **both fired on one `auth_expired`**: two dialogs, two identity models, one expired token. Every one of the 13 conflicts sat on that seam.

This branch's model won because it is the superset: `main` keyed a credential as `toolType@sshRemoteId`, this one carries the credential *kind* (oauth / api-key / gateway / cloud-provider), a secret fingerprint, and the host. That is what makes proactive probing possible and what lets `buildAuthAction` decline to offer "sign in" for an API key no login can repair - `main`'s identity cannot express either.

Retired, consumers rewired: `authOutageStore.ts`, `ReauthModal.tsx`, `providerAuthIdentity.ts` (+ their two test files).

Kept from `main`, because this branch had no equivalent:
- **Auto-open on failure.** `main` raised the login dialog the moment an agent died; this branch only marked state and waited to be noticed, which would have regressed the shipped build. Preserved via `openAuthRecoveryForSession()`, falling back to the ordinary error frame when a credential cannot be identified so a failure is never left unexplained.
- **Cue pipeline auth failures.** Cue spawns outside the ProcessManager, so a pipeline never reaches `agent:error` - an unattended run is exactly where a silent auth failure costs most. Re-routed onto the credential model.

## Greptile P1s - all three confirmed and fixed

1. **Env layers diverge.** Correct, and worse than reported: the *renderer* re-derived the same two-layer merge, so main and renderer could disagree about which credential an agent presents. `mergeEffectiveEnv` now delegates to `resolveAgentEnvironment()` - the same three-layer resolve the spawner uses - instead of restating the precedence, which is what let it drift. Both processes now read the global layer.
2. **Probe ignores customPath.** Correct. `process:spawn` resolves `sessionCustomPath || agent.binaryName`, so an agent pointed at another CLI install was probed against whichever one Maestro detected. `AuthTarget` carries `binaryPath` and both the probe and the login honor it. Over SSH a `customPath` names a path on the *remote* host; the probe cannot tell that from a local detected path (both are absolute strings), so the caller owns the choice and the contract is now documented on `ProbeCredentialOptions`.
3. **Hydration overwrites live updates.** Correct. Merged per key by `checkedAt` instead of assigning the response wholesale, since the two writers race and arrival order does not tell you which is newer.

## CodeRabbit - 7 of 8 applied

Applied: modal remount on credential change; both wizard issues (unresolved SSH remote no longer collapses to local; identity scoped to its request); palette label for non-repairable credentials; remote login error text; the test that passed for the wrong reason; the contradictory comment.

**Declined - nested interactive controls (`ProviderAccountsSection.tsx:293`).** The double-fire half is wrong: `ToggleSwitch` calls `e.stopPropagation()`, so both handlers cannot run. The nesting is the house pattern from `SettingCheckbox.tsx`, and changing one row would make it inconsistent with every other setting. The real defect underneath is that the `role="button"` row swallows its description into its accessible name - I hit that directly (a reworded description collided with the actual button in a test). That deserves a pass across the settings UI, not a one-row divergence here.

## Verification

Full suite green: **34,359 passed**, 1253 files. `tsc` clean on all three configs (`tsconfig.json`, `tsconfig.main.json`, `tsconfig.cli.json` - note `src/main` is not covered by the default config, which is how the missing dep initially went unseen). ESLint and prettier clean.

New regression tests: global-layer precedence and api-key-vs-oauth resolution; a remote `customPath` reaching the far side untouched; both orderings of the hydration race; both wizard cases; the palette label rule.

@greptileai please re-review.

pedramamini:
Round 2 pushed. Both new P1s were real; the failover one was the deeper half of the original env finding I only partially fixed.

**Provider Failover was invisible to the auth layer.** `process:spawn` layers a pinned endpoint's env over the agent's own *and* strips any primary credential the endpoint does not supply (`failoverUnsetEnvKeys` - a URL-only backup row must not present the user's primary Anthropic key to a third party). The resolver did neither, so an agent mid-failover was probed against its primary: a healthy badge for an agent that is failing, and a recovery that repairs a login the agent is not using. Both steps now run in the spawner's order, with the strip applied to the **merged** env since that key also arrives from the global and agent layers. Overlays are in-memory, so this only affects manual re-probes and recovery logins - which is exactly when an agent can be pinned.

Worth noting the fix landed cleaner than my own test expected: a backup that redirects `ANTHROPIC_BASE_URL` now resolves to `claude-code::gateway::backup.example::local`, correctly treating the backup as a different operator rather than a variant of the primary login.

**Timestamp tie.** Correct, and the suggested `>` is right - on a tie the live event is the later of the two, since it arrived while the read was still in flight.

**CodeRabbit round 2** - 5 of 6 applied: `validateIdentity` now constrains `kind` to the union and derives the key from the parts rather than checking `key === key` (which only proves the caller is self-consistent about the string, not that it describes the fields beside it - an `oauth` kind under an `::api-key::` key would put a Sign In button on a credential no login can repair). The key composer is exported from `shared/providerAuth.ts` so validator and builder cannot drift. Also: four MD040 fences, the Google-key-shaped fixture assembled like the Slack token beside it, and the `auth_expired` tests reworked - they asserted `modals.get(...)` was defined, which the store retains after a close and so proved nothing; they now mock the recovery service, assert `isOpen`, and cover the fallback to the error frame when a credential cannot be identified.

**Declined - removing IPC handlers before re-registering.** None of the 40 handler modules do this, registration happens once from `main/index.ts:1571`, and the second call site is inside `registerAllHandlers`, which nothing invokes. Guarding one module against a path that does not exist would make it the odd one out; if duplicate registration were a real risk it is a codebase-wide fix, not a local one.

Full suite green: **34,364 passed**. `tsc` clean on all three configs, ESLint and prettier clean.

@greptileai please re-review.

pedramamini:
Round 3. Confirmed and fixed - this one was a real leak path and the sharpest finding of the three rounds.

The round-2 fix removed the credential from the resolved target's env, which is **not** the same as removing it from the child's environment. Both auth children build their env over `process.env`, so a primary `ANTHROPIC_API_KEY` in Maestro's own environment survived that deletion and would be handed to the backup operator. That is precisely why `failoverUnsetEnvKeys` returns *keys to strip* rather than values to merge, and why `process:spawn` carries the list down to the spawner instead of merging it away - its own docblock says the value "may come from the global shell settings or the inherited `process.env`". I had applied it at one of three layers.

`AuthTarget` now carries `unsetEnvKeys` and both consumers apply it after every layer is merged: the probe deletes the keys from its child env, and the login forwards them as `ProcessConfig.unsetEnvKeys`, which `buildChildProcessEnv` already honors - no new plumbing needed.

Two notes on scope:

- **The SSH branch intentionally keeps inheriting the unmodified `process.env`.** Those vars run the local `ssh` client and are not forwarded to the far side, while the remote's env is baked into the command line from the already-stripped target env. Said so at the call site so it does not read as an oversight.
- **The login-side pass-through is defensive, not reachable today.** The removal list is only non-empty when the endpoint redirects the base URL, and such a credential resolves to `gateway`, which `resolveLoginCommand` refuses. I kept the wiring (so this cannot become the path that leaks if the classification changes) and labelled it as such rather than letting it look load-bearing. The refusal itself is now covered: an agent on a backup endpoint is presenting *that* operator's credential, so the remedy is the endpoint config, not a sign-in.

New tests: an inherited `ANTHROPIC_API_KEY` is gone from the probe's child env while the endpoint's own vars and `PATH` survive; the removal list travels on the target for a pinned agent and is absent for an unpinned one; the login refuses a failover credential.

Full suite green: **34,367 passed**. `tsc` clean on all three configs, ESLint and prettier clean.

@greptileai please re-review.

pedramamini:
Superseded. Closing.

This branch was built on 2026-08-15. Three days later `main` shipped a different, much smaller answer to the same problem in `ReauthModal` + `authOutageStore` (1d99300, 940daf8), and has evolved it since. The branch's own commits delete both of those files to replace them with a 16.6k-line system, so merging it now means ripping out a shipped feature rather than adding one.

The two designs also disagree on premise. This branch is **proactive**: probe every credential at startup, dedupe agents onto one `CredentialIdentity`, surface state in the Left Bar and a Provider Accounts settings section. `main` is **reactive**: an `auth_expired` match opens one provider-scoped dialog that runs the login in an embedded PTY and replays every blocked agent's held queue. The reactive path covers the failure users actually hit, at a fraction of the surface area.

### Scavenged onto main

- **3ebbfbb** - the error-pattern bank moves to `src/shared/agentErrorPatterns.ts` with an injectable log sink, `src/main/parsers/error-patterns.ts` becomes the main-process face of it, and the wizard's duplicate bank of ~20 hand-written regexes is deleted. That copy had drifted (it classified `authentication_failed` as `unknown`) and hard-coded `Run "claude login"` for all seven providers, which is both the wrong provider and not a real command. Auth messages in the bank no longer name a terminal command at all, since the bank cannot know whether the failing agent signs in that way.
- **78bfebb** - `classifyCredentialKind()` in `shared/providerAuthIdentity.ts`, ported from this branch's `shared/providerAuth.ts`. `ReauthModal` now gates its login terminal on it, so an `ANTHROPIC_API_KEY` agent, a gateway agent, and a Bedrock/Vertex agent no longer get a login flow that succeeds and fixes nothing. They get the real remedy, naming the variable that decided it.

40 tests added across the two.

### Deliberately not taken

The startup probe pass, the credential-identity store, the Left Bar auth indicator, the Provider Accounts settings section, and the auto-resend modal. All reasonable ideas; none of them worth re-landing as a replacement for a working shipped path. If proactive probing is wanted later, `main/agents/auth/auth-probe.ts` on this branch is the piece to start from, and it can sit on top of `ReauthModal` rather than instead of it.

PR #1388, opened 2026-08-16
fix(web-desktop): auto-reload on stale hashed chunks instead of blaming Wi-Fi

### LINKED ISSUE #1387: Bug: Web-Desktop: session resume fails with dynamic import error until har...
## Summary
Web-Desktop: session resume fails with dynamic import error until hard refresh

## Environment
- Maestro version: 0.18.5-RC

## Steps to Reproduce
1. Open Maestro Web-Desktop in a browser on the LAN.
2. Leave the tab open (across a Maestro restart / new build, so the served asset hashes change).
3. Click a session in the sessions list to resume it.
4. Observe the TypeError about the dynamically imported module, framed as a network problem.
5. Hard refresh the page — web-desktop reloads, displays the target session, and resuming works immediately. All later resumes in that tab work fine.

## Expected Behavior
Resuming a session from the sessions list in web-desktop should load without requiring a manual hard refresh. If the client's asset manifest is stale relative to the server build, Maestro should detect the failed dynamic import and auto-reload the page, rather than surfacing a misleading network/Wi-Fi connectivity error.

## Actual Behavior
The first attempt to resume a session from the sessions list fails with:

```
TypeError: Failed to fetch dynamically imported module: http://192.168.178.109:41713/f30ef190-f2d0-4eb2-925d-5efa665eb465/desktop/assets/dist-Bwv2dFcV.js
If this device is on a different Wi-Fi network than the computer running Maestro, or the network blocks device-to-device traffic, the page can fail to load. Confirm both are on the same network and retry.
```

The accompanying message blames the network / Wi-Fi, but the actual cause appears to be a stale client holding references to hashed asset chunks that no longer exist on the server. A hard refresh resolves it: web-desktop reboots, shows the session that was being resumed, and that resume plus all subsequent resumes work instantly.

## Additional Context
Only the first session resume in a given tab fails; after the hard refresh the issue does not recur in that session. Suggests a stale index/module-graph vs. rebuilt hashed assets, and that the error handler's network-isolation copy is misattributing the root cause. Maestro 0.18.5-RC on Linux (Ubuntu 6.17.0-35-generic), packaged build.

## Screenshots / Recordings
None

pedramamini:
Closes #1387

## Problem

A web-desktop tab left open across a Maestro restart still holds the **old** module graph. The bundle's JS/CSS chunks are content-hashed, so a rebuild changes every filename. Modules already fetched keep working, but the first lazily-imported chunk the user reaches - resuming a session, opening a modal, previewing a file - resolves to a hash the server no longer has on disk and 404s:

```
TypeError: Failed to fetch dynamically imported module: http://.../desktop/assets/dist-Bwv2dFcV.js
If this device is on a different Wi-Fi network than the computer running Maestro, ...
```

Two separate defects fall out of this:

1. **Nothing recovered.** The user had to know to hard-refresh. The reporter confirms a hard refresh fixes it permanently for that tab, which is exactly the stale-manifest signature.
2. **The error surface was never scoped to boot.** `src/web-desktop/index.html` installs global `error` / `unhandledrejection` listeners for failures that stop the bundle from booting at all - but they stay installed for the life of the page. Any post-boot unhandled rejection wiped `#root` and rendered the same-network hint. So a 404 on a hashed asset tore down a perfectly healthy running app and told the user to check their Wi-Fi.

Defect 2 is the broader one: it applies to *any* unhandled rejection after mount, not just stale chunks.

## Fix

New `src/web-desktop/loadFailure.ts` owns the policy:

- Recognizes the per-engine "dynamic import failed" wording (Chromium, Firefox, WebKit, plus Vite's CSS preload helper). None of them expose a structured error code, so message matching is the only signal available.
- **Stale asset** -> reload once to pick up the current manifest, guarded by a 30s per-tab `sessionStorage` cooldown so a persistent failure cannot loop. After the cooldown a genuinely new rebuild in the same long-lived tab can still self-recover.
- **Reload did not help** -> show an honest message ("Maestro was updated or restarted while this page was open...") rather than the network hint.
- **Unrelated rejection, app already mounted** -> leave it alone, so React's error boundary reports it in context instead of the page being wiped.
- **Unrelated failure before boot** -> unchanged; full-page error surface with the network hint, which is the right guess there.

`index.html` keeps its inline listeners - they must be registered before the entry module is parsed so a `SyntaxError` in the bundle itself is still reportable - but now delegates the *decision* to that module once `bootstrap.ts` installs it. Its error surface also HTML-escapes the title and hint, not just the detail.

`bootstrap.ts` installs the handler at module top level and calls `markBooted()` once the renderer mounts.

## Testing

- New `src/__tests__/web-desktop/loadFailure.test.ts`: 19 tests covering per-engine message recognition, non-`Error` rejection reasons, the reload/cooldown/ignore decision matrix, and storage that is corrupt, blocked, or absent (private-mode Safari throws on access).
- `npx vitest run src/__tests__/web-desktop/ src/__tests__/main/web-server/` - 581 passed.
- `npm run lint` clean; eslint + prettier clean.
- `npm run build:web-desktop` succeeds, and the built `dist/web-desktop/index.html` still contains the inline delegation (verified the inline script is not stripped or transformed).

## Notes

- Targets `rc`, not `main`: `src/web-desktop/` does not exist on `main`, so this is unmergeable there. The report is against 0.18.5-RC.
- I could not reproduce the two-build stale-asset condition end-to-end locally, so the classification is verified by unit test against the exact message from the report rather than by a live repro. Worth a sanity check from the reporter.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery from stale or failed web application assets.
  * Added safeguards to prevent repeated reload loops.
  * Improved error handling for failures during startup and after the app loads.
  * Added clearer fallback messages with network troubleshooting guidance when recovery is unavailable.
  * Improved handling of preload failures and unexpected error formats.

* **Tests**
  * Added comprehensive coverage for asset failures, reload cooldowns, storage issues, boot states, and error display behavior.


pedramamini:
Addressed all four review findings, plus one the repo could not see. Pushed as `795ed21a3`.

## 1. Unbounded reload loop (Greptile P1 / CodeRabbit) - fixed

Both were right, and the reasoning in the original comment on `markReloadAttempt` was wrong: it said losing the guard "only costs us loop protection, which the cooldown already bounds." The cooldown is not an independent bound - it is read back out of the same key the guard writes. A guard that never lands is not a degraded guard, it is no guard: `readLastReloadAt` returns 0 forever, every failure looks like the first one, and the tab reloads without end.

`markReloadAttempt` now returns whether the write landed, and the caller refuses to reload when it did not, falling through to the error surface instead.

## 2. `vite:preloadError` (CodeRabbit) - added

Correct and worth having: a lazy import the app awaits and handles itself never becomes an unhandled rejection and never fires `error`, so neither inline listener sees it. Vite's event is the only signal for that case. `preventDefault()` on a stale-asset payload so Vite does not also rethrow.

One thing the suggestion did not account for: registering it inside `installLoadFailureHandler` with an inline arrow adds a *second* listener if that function is ever called twice, and one failure would then reload twice. The listener is hoisted to module scope so `addEventListener` dedupes it by reference.

## 3. Hard-coded `Ctrl/Cmd+Shift+R` (Greptile P2) - fixed

Now goes through `formatShortcutKeysFor(['Meta', 'Shift', 'R'], isMac)` from `shared/shortcutKeys`, so it reads `⌘ ⇧ R` on a Mac and `Ctrl+Shift+R` elsewhere. The platform is read from the user agent, matching how `bootstrap.ts` derives its `process.platform` shim - the relevant machine is the one running the **browser**, which for web-desktop is often not the one running Maestro.

## 4. A real type error, and why nobody saw it

While verifying the above I found this in the PR's own new file:

```
src/web-desktop/loadFailure.ts(215,4): error TS2345:
  Argument of type '{}' is not assignable to parameter of type 'string'.
```

`(reason && ((reason as Error).stack || (reason as Error).message)) || String(reason)` types as `{} | string` when `reason` is `unknown`, and it feeds a `string` parameter.

**`src/web-desktop` is in no tsconfig `include` list.** `tsconfig.json` has `src/renderer`, `src/web`, `src/shared`, `src/types`; `.lint`, `.main` and `.cli` likewise skip it. So `npm run lint` passing said nothing about this file - my own "lint clean" line in the PR description was vacuous.

Replaced with a `detailOf(reason)` helper that mirrors the existing `messageOf`, checking each branch rather than casting. Verified clean with a standalone `tsc`.

I did **not** add the directory to a tsconfig here - that is a build-config change, not part of this fix, and it surfaces one unrelated pre-existing error (`electron-shim.ts` re-declares `__MAESTRO_CONFIG__` with a type that conflicts with the shared `MaestroConfig`). Worth its own PR; flagging it so it does not get lost.

## Testing

```
src/__tests__/web-desktop/    34 passed (was 19 in this file, now 24)
tsc -p tsconfig.json          clean
standalone tsc on the file    clean (was 1 error)
eslint / prettier             clean
npm run build:web-desktop     succeeds, inline delegation still present in dist
```

Four new tests, each confirmed **red** against the pre-fix source: the guard-cannot-persist path shows the error instead of reloading, `markReloadAttempt` reports success honestly, a stale chunk arriving only via `vite:preloadError` reloads exactly once and is `preventDefault`ed, and an unrelated `vite:preloadError` is left for Vite to rethrow.

One note on the storage tests already there: `markReloadAttempt(1, undefined)` does not exercise the absent-storage path, because passing `undefined` explicitly selects the default argument and lands on real `sessionStorage`. The new test drives it through a throwing `sessionStorage` getter instead, which is the only way it actually happens.

PR #1389, opened 2026-08-16
fix(ipc): register the Context Timeline handlers in the live bootstrap path (MAESTRO-YV)

pedramamini:
Eleventh Sentry triage pass, `rc` side. Base `rc`.

## Scope

I pinned the query to `release:0.18.5-RC` (= `package.json` at `origin/rc` HEAD), which is the decisive main-vs-rc filter. That returned 18 aggregate rows. Exactly **one** of them was a live, clearly-fixable defect in current rc code, so this PR is deliberately small.

## Fixed: MAESTRO-YV

`Error invoking remote method 'contextTimeline:clearCaptures': No handler registered for 'contextTimeline:clearCaptures'` - `environment: production`, `channel: rc`, release `0.18.5-RC`.

**Root cause.** There are two registration lists:

- `registerAllHandlers()` in `src/main/ipc/handlers/index.ts` - reads like the authoritative list, but **nothing calls it**. Its only mention anywhere is inside a comment.
- `setupIpcHandlers()` in `src/main/ipc/bootstrap/index.ts` - the live path, invoked from `main/index.ts` at app-ready.

`registerContextTimelineHandlers` was imported into `handlers/index.ts` and added to the **dead** list only. It was also never re-exported from that module, so `bootstrap/index.ts` could not have called it even if someone tried.

Net effect: neither `contextTimeline:getCaptures` nor `contextTimeline:clearCaptures` was ever registered in a shipped build.

**Timeline corroborates it.** The feature landed in `07b5ac381` (#1365, 2026-08-10); MAESTRO-YV's first event is 2026-08-11.

**User-visible impact.** The Context Timeline panel hydrated empty on every window reload, every second window, and every web-desktop client - precisely the problem #1365 set out to fix. `hydrateContextTimeline` treats a failed fetch as "not hydrated" and returns quietly, so `getCaptures` failed invisibly. `forgetContextTimelineCaptures` fires `clearCaptures` as a floating `void` promise with no `.catch()`, so only that half surfaced, as an unhandled rejection.

**Fix.** Re-export the registrar and call it in `setupIpcHandlers()`, next to `registerTabsHandlers()`.

## Regression test

`src/__tests__/main/ipc/bootstrap/handler-wiring.test.ts`.

The existing `context-timeline.test.ts` calls `registerContextTimelineHandlers()` **directly**, so it passed 100% while production was broken - the same "a permissive test harness hides exactly the bug you are chasing" failure mode as the duplicate-`ipcMain.handle` finding in #1341.

So the new test does not re-test the handler. It reads both sources and asserts the invariant that actually held wrong: **every registrar invoked by the dead `registerAllHandlers` must also be invoked by the live `setupIpcHandlers`**. That catches the whole bug class, not just this instance. It includes a floor assertion on the extracted count so a regex that silently stops matching fails loudly rather than passing vacuously.

Both cases confirmed **red** against the unfixed source before the fix, green after.

## Verified, deliberately not changed

Every guard from prior triages is still intact on `rc` - no regressions this round, including MAESTRO-M9's `RangeError` carve-out, which has regressed once before:

| Issue | Guard | State |
|---|---|---|
| M9 | `RangeError` carve-out, both loops in `agentSessions.ts` | intact |
| V9 | `SyntaxError` carve-out, `ExitHandler.ts:522` | intact |
| JS/K7 | `isExpectedGroomingFailure`, `groupChat.ts:158` | intact |
| RR | `isExpectedQuotaStatus`, `codex-usage-sampler.ts:32` | intact |
| X4 | `EXPECTED_SPAWN_ERROR_CODES`, `tabNaming.ts:85` | intact |
| YA | single `getRichOverviewStats` handler | intact (fix `5a7e86f9e` held) |

Skipped, with the reason:

- **MAESTRO-Q2** (1446 events, the noisiest) - `captureMessage` per failure tick in `claude-usage-sampler.ts`. Deferred by six prior triages as needing a human call on `maestro-p` exit semantics; that call has not been made, so I did not force it.
- **MAESTRO-W5** (370) - `window.maestro` undefined cascade. `environment: development`, Electron 41.6.0. Same family as BC/W2/W7/WB; prior triage established that patching these with optional chaining masks a broken install across ~40 call sites.
- **MAESTRO-JS** (71) - guard is present and correct; it reached `rc` on 2026-08-06, after the builds still reporting. Re-fixing correct code would be wrong.
- **MAESTRO-RE** (210), **G5/G6**, **Z0** - all 0% on `0.18.5-RC`. RE is entirely on `0.18.4-RC`; G5's sampled event is on `0.16.9-RC` / Electron 28. Old-build populations, and RE was already established as a broken install rather than a packaging bug.
- **MAESTRO-Z5** - `Protocol not allowed: clickup:` is the `ALLOWED_PROTOCOLS` security boundary working as designed. Widening it is a product/security decision, not a crash fix.
- **YN / XW** (OMP) - zero symbol hits on `rc`; unmerged-feature dev noise.
- Native/GPU (`HQ`, `SZ`, `QG`, `Y8`, `YW`, `HA`, `62`) and genuine spawn signal (`NM`).

## Dead code found - not removed, flagging for your call

Per the dead-code-hygiene rule, listing rather than deleting:

1. **`registerAllHandlers()`** (`handlers/index.ts:219`, 38 registrars) - unreachable, and it is the trap that caused this bug: it looks like the registration list, so a new handler gets added there. This PR makes the two lists consistent, and the new test keeps them consistent, but deleting it would remove the trap outright. Want me to?
2. **`tempfile:write` / `tempfile:read` / `tempfile:delete`** - exposed by `src/main/preload/files.ts`, have **no** main-side handler and **no** renderer callers. A systematic sweep of all 488 preload-invoked channels found these three as the only other gap; they are latent rather than firing, because nothing calls them.

## Validation

- `npx tsc --noEmit -p tsconfig.json` - **0 errors**
- `npx eslint` on both changed files - clean
- `npx vitest run src/__tests__/main/ipc` - **39 files / 1785 tests passed**
- New test red-checked against unfixed source

Note: local validation is single-OS. Both CI matrix legs need to be green before merge.

## Summary by CodeRabbit

* **New Features**
  * Enabled Context Timeline functionality by registering its handlers during application startup.
* **Bug Fixes**
  * Improved IPC handler registration consistency so available application features connect correctly when the app starts.
* **Tests**
  * Added automated coverage to verify that IPC handlers remain synchronized and that Context Timeline registration is included.


pedramamini:
Both bots flagged the same thing in the regex, and they were right about the fragility even though the predicted failure did not happen. Pushed `7ceb6bce1`.

## What was actually going on

CodeRabbit predicted the test would fail because `register[A-Za-z]+Handlers\(` matches the `registerAllHandlers(` **declaration** in `handlers/index.ts`, putting the dead wrapper in the required set. That half is correct. But the test passed anyway - and for a reason that is worse than failing:

```
src/main/ipc/bootstrap/index.ts:356:  // registerAllHandlers(). The registry and window manager are module-scope
```

A **comment** in the bootstrap file also matches, so `registerAllHandlers` landed in both sets and cancelled itself out. The test was green by coincidence. Greptile's P2 named this directly: the scan counts declarations and comments as calls, so an unrelated comment edit could fail the test, and a registrar-shaped comment could hide a genuinely missing live registration.

## Fix

Strip line and block comments before matching, exclude `function` declarations with a lookbehind, and remove the wrapper from the dead set explicitly rather than relying on a coincidence to cancel it. The wrapper is the list, not a member of it - nothing is expected to call it, which is the whole premise of the test.

Verified both properties, since fixing this could easily have produced a test that passes unconditionally:

- **Still catches the real bug.** Reverting `bootstrap/index.ts` to `origin/rc`: both cases fail.
- **No longer depends on the comment.** Rewriting that comment so it no longer mentions `registerAllHandlers()`: still 2 passed. Before this change that edit would have turned the suite red.

## On the underlying fix

Traced the whole chain to confirm the diagnosis rather than take the Sentry title at face value:

- `registerContextTimelineHandlers` registers exactly two channels, `contextTimeline:getCaptures` and `contextTimeline:clearCaptures`.
- `contextTimelineHydration.ts:112` calls `window.maestro?.contextTimeline?.clearCaptures(...)` as a floating `void` promise with no `.catch()` - which is precisely why this surfaced as an unhandled rejection, and why the `getCaptures` half at line 69 failed silently instead.
- The registrar was reachable only from the dead `registerAllHandlers`, so neither channel existed in a shipped build.

```
src/__tests__/main/ipc/   1805 passed
tsc -p tsconfig.main.json clean
eslint / prettier         clean
```

Merging once the Windows leg finishes.

PR #1390, opened 2026-08-16
fix(sentry): stop the Settings modal crash, quiet watcher and UNC-path noise

pedramamini:
Stable-line Sentry triage against `release:0.17.3` (main's shipped build; `0.17.4` at HEAD has no field events yet).

## MAESTRO-YP / YQ / YR - Settings modal crash

`TypeError: Cannot read properties of null (reading 'command')` - 29 events, `escalating`, caught by the ErrorBoundary, which blanks the whole Settings modal.

`AICommandsPanel` gated its edit form on:

```tsx
{editingCommand?.id === cmd.id ? (   // editing branch
  ...
  <input value={editingCommand.command} />   // line 406
```

When nothing is being edited, `editingCommand?.id` is `undefined`. For a persisted command whose `id` is also `undefined`, that comparison is `undefined === undefined` -> **true**, so the edit branch rendered and dereferenced the null `editingCommand` a few lines later.

How an id-less command gets there: `settings.json` is user/sync/legacy editable, and hydration casts the stored array to `CustomAICommand[]` without checking it. The entry was then keyed in the merge `Map` under `undefined`, which independently broke its edit/save/delete (all keyed by `id`) and duplicated its React key.

Fixed at both layers:
- **Panel:** compare only when something is actually being edited.
- **Hydration:** skip entries that aren't objects with a non-empty string `id`, with a warn. Non-destructive - it filters the in-memory store, it does not rewrite `settings.json`.

The identical `?.` guard in `SpecKitCommandsPanel` / `OpenSpecCommandsPanel` / `BmadCommandsPanel` is hardened too. Those are fed by `spec-command-manager`'s `getPrompts()`, which always assigns an `id`, so they were **latent, not firing** - noted so the fix isn't mistaken for three more field bugs.

## MAESTRO-SC - history watcher telemetry noise

`EMFILE: too many open files` / `ENOSPC: System limit for number of file watchers reached` thrown by `fs.watch` on the history directory. Both are OS resource ceilings (fd limit, and on Linux the inotify `max_user_watches` cap, which surfaces as `ENOSPC` rather than a disk-space error), not Maestro faults.

Watching only powers live refresh when another process edits history on disk, and `startWatching` already degrades to a null watcher with history fully working. Added `EXPECTED_WATCH_ERROR_CODES` and applied it to **both** the synchronous `fs.watch` throw and the watcher `'error'` event - the latter already quieted ENOENT/EPERM/UNKNOWN but missed these.

Deliberately **not** shared with `isExpectedSessionReadError` in `agentSessions.ts`, which excludes `EMFILE` on purpose (an EMFILE while reading a session file can mean a real fd leak). An unexpected code such as `EIO` still reports.

## MAESTRO-TN - UNC path from a doubled separator

`Auto Run docs lookup failed: UNKNOWN: unknown error, stat '\\.maestro\playbooks\'`.

The wizard's directory field accepts `/` (validation only requires `readDir` to succeed, which it does). `` `${dirPath}/${PLAYBOOKS_DIR}` `` then produced a leading `//`, and Windows resolves `//.maestro/playbooks` as the UNC path `\\.maestro\playbooks` - a host that doesn't exist. `stat` failed with `UNKNOWN`, which isn't in `RECOVERABLE_DOCS_ERROR_PATTERNS`, so it reported to Sentry and surfaced an error to the user.

Switched both wizard call sites to the canonical `joinPath()` from `shared/formatters`, which trims the trailing separator and preserves Windows/POSIX style. Also fixes any `dirPath` with a trailing slash.

## Testing

`472 passed` across the 8 affected suites. Every new test was confirmed **red against the unfixed source** first (`git stash push` on source files only):

| Suite | Red before |
|---|---|
| `AICommandsPanel.test.tsx` | 3 failed - reproduces the field stack exactly: `TypeError: Cannot read properties of null (reading 'command')` at `AICommandsPanel.tsx:406`, rendered from the map at `:359` |
| `history-manager.test.ts` | 8 failed, `EIO` control stayed green |
| `DirectorySelectionScreen/utils.test.ts` | 3 of 4 failed, the unchanged `/project` case stayed green |
| `settingsStore.test.ts` | 1 failed |

`npx tsc --noEmit`: **112 errors before, 112 after** - exact parity, none in changed files (all pre-existing missing optional deps under `src/web/mobile/*`, `@codemirror/*`, `js-yaml`). ESLint clean on all changed files.

## Not addressed

`MAESTRO-RS` (GLIBC 2.38 / better-sqlite3, 613 events) is the largest remaining stable issue but is Linux build/CI target work, not app code, and still needs a human decision. Also skipped: `MAESTRO-Q2` (deferred again), native/GPU crashes (SG/5A/38/23/62/YM/QG/RB/4V/T2/1F/Y), `R0`/`NM` (genuine spawn signal), `WJ` (a prior triage explicitly decided not to match it), and the `release: 41.6.0` events, which are the Electron version - unpackaged dev runs, not stable-line field crashes.

Verified all prior triage guards are still intact at HEAD (`write EPIPE`, `isExpectedGroomingFailure`, `db.isReady()`, the M9 `RangeError` carve-out, `isExpectedQuotaStatus`) - no regressions to restore this round.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented malformed saved AI commands from causing crashes or entering edit mode.
  * Improved handling of history-folder watcher errors and recovery after watcher failures.
  * Fixed Auto Run documentation path handling across Unix and Windows directory formats.
  * Ensured invalid persisted command entries are ignored while valid commands and defaults remain available.

* **Tests**
  * Added regression coverage for malformed commands, watcher errors, and cross-platform documentation paths.


pedramamini:
All three review findings were valid and are fixed in `271a19578`.

## 1. `fs.mkdirSync` outside the try (CodeRabbit) - fixed

This one undercut the PR's own goal. `mkdirSync` sat above the `try`, and it fails with the **same OS ceiling codes** `fs.watch` does (EMFILE, ENOSPC, EACCES). So on exactly the constrained machine MAESTRO-SC comes from, the error bypassed `isExpectedWatchError`, propagated to the caller in `main/index.ts`, and reported as a history initialization failure - the noise this PR exists to quiet, just under a different Sentry title. Moved inside the try.

## 2. Watcher not reset after an error event (CodeRabbit) - fixed

Also correct, and worse than telemetry noise: Node documents an FSWatcher as unusable once it emits `'error'`, but `this.watcher` kept pointing at the dead one. The `if (this.watcher) return` guard at the top of `startWatching` then rejected every later attempt, so **live refresh stayed dead for the rest of the session** - a silent functional regression, not just a logging problem.

Took the two refinements in the finding as well: the reference is identity-checked so a late error from a replaced watcher cannot clear its replacement, and `close()` is not called from the handler, since Node says not to touch the watcher there.

## 3. Incomplete command shape validation (Greptile P1) - fixed, with one adjustment

Right that an `id` alone is not enough. Confirmed against the panel rather than assuming, and the fields are not equivalent:

| Field | Panel usage | Missing means |
|---|---|---|
| `command` | `command.startsWith('/')` (`:104`) | crash |
| `prompt` | `prompt.substring(...)` (`:560`) | crash |
| `description` | rendered as `{cmd.description}` (`:516`) | renders nothing |

So I required `id`, `command` and `prompt` as the suggestion had them, but defaulted `description` to `''` instead of rejecting the entry. Dropping a user's working custom command over a cosmetic field would be a data-loss fix for a crash bug.

Worth noting this also closes a second hole: the built-in merge is `{ ...cmd, isBuiltIn: existing.isBuiltIn }`, which spreads the stored entry over *nothing*. A built-in whose `prompt` went missing on disk would have replaced the default outright and lost the prompt. With full validation that entry is now skipped and the default survives intact, so the merge did not need touching.

## Testing

```
history-manager + settingsStore + AICommandsPanel   344 passed
tsc -p tsconfig.json / tsconfig.main.json           clean
eslint / prettier                                   clean
```

Ten new tests, all confirmed **red** against the pre-fix source: six directory-creation codes staying out of Sentry, the `EIO` control still reporting, a watcher that died with an error being replaced on the next `startWatching` (and not `close()`d), and the id-present-but-unusable entries being skipped while a description-less one survives with `''`.

PR #1391, opened 2026-08-16
fix: preserve truthful process and session state

heinbrian:
## What changed

- treat `EPERM` from `process.kill(pid, 0)` as proof that the process exists
- remove CLI activity only when the process probe returns `ESRCH`
- reconcile persisted tab state with live managed-process and CLI activity
- preserve `unknown` session state in the CLI instead of coercing it to `idle`

## Why

Read-only and sandboxed monitors can receive `EPERM` while probing a live
Maestro process. The existing catch-all behavior reported stale discovery,
removed valid CLI activity, and exposed working tabs as idle. That made health
and recovery decisions depend on a permission failure rather than live state.

## Impact

Process-restricted monitors no longer turn `EPERM` into a false crash. Session
list output reports busy when a matching managed process or active CLI job is
present, and preserves uncertainty when neither persisted nor live evidence is
conclusive.

## Validation

- four focused Vitest files: 149 tests passed
- `tsc -p tsconfig.main.json --noEmit`
- `tsc -p tsconfig.cli.json --noEmit`
- Prettier check passed for all nine changed files
- targeted ESLint completed with no errors

This PR changes source and tests only. It does not modify an installed Maestro
application or request a release.

## Summary by CodeRabbit

- **New Features**
  - Session listings now preserve and display an `unknown` state instead of incorrectly showing it as idle.
  - Session activity now reflects live desktop processes and CLI activity more accurately.

- **Bug Fixes**
  - Improved detection of active CLI sessions when process checks encounter permission errors.
  - Prevented active sessions from being incorrectly cleared due to transient or unexpected process-check errors.
  - Improved CLI server detection for running processes with restricted permissions.

- **Tests**
  - Added coverage for session states, live activity detection, and process-check error handling.

### DIFF
--- a/src/__tests__/cli/commands/session.test.ts
+++ b/src/__tests__/cli/commands/session.test.ts
@@ -127,6 +127,35 @@ describe('session list command', () => {
 		expect(line.split('  ').filter(Boolean).length).toBeGreaterThanOrEqual(4);
 	});
 
+	it('preserves an unknown desktop state in human-readable output', async () => {
+		const mockSendCommand = vi.fn().mockResolvedValue({
+			type: 'desktop_sessions_list',
+			success: true,
+			sessions: [
+				{
+					tabId: 'tab-unknown',
+					sessionId: 'tab-unknown',
+					agentId: 'agent-a',
+					agentName: 'Backend',
+					toolType: 'claude-code',
+					name: null,
+					agentSessionId: null,
+					state: 'unknown',
+					createdAt: 1714268000000,
+					starred: false,
+				},
+			],
+		});
+		vi.mocked(withMaestroClient).mockImplementation(async (action) => {
+			const mockClient = { sendCommand: mockSendCommand };
+			return action(mockClient as never);
+		});
+
+		await sessionList({});
+
+		expect(consoleSpy.mock.calls[0][0]).toContain('unknown');
+	});
+
 	it('maps connection errors to MAESTRO_NOT_RUNNING (consistent with dispatch)', async () => {
 		vi.mocked(withMaestroClient).mockRejectedValue(new Error('ECONNREFUSED'));
 
--- a/src/__tests__/main/web-server/web-server-factory.test.ts
+++ b/src/__tests__/main/web-server/web-server-factory.test.ts
@@ -166,6 +166,10 @@ vi.mock('../../../main/utils/sentry', () => ({
 	captureException: vi.fn(),
 }));
 
+vi.mock('../../../shared/cli-activity', () => ({
+	isSessionBusyWithCli: vi.fn().mockReturnValue(false),
+}));
+
 import {
 	createWebServerFactory,
 	type WebServerFactoryDependencies,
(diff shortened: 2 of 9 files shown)


pedramamini:
Thanks for this, @heinbrian - genuinely good catch. Conflating `EPERM` with "process is gone" is exactly the kind of thing that makes a monitor lie about state, and splitting `ESRCH` (proof of absence) from `EPERM` (proof of presence) from everything else (no proof either way) is the right decomposition. I also appreciate that the cleanup write is now gated on confirmed absence rather than on any failure.

I verified locally against `c6a8c0f`:

- the four test files: 149 passed, matching your description
- no merge conflicts against `main`
- the `${sessionId}-ai-${tabId}` and legacy `${sessionId}-ai` process id convention matches the real spawn sites in `agentStore.ts` and `useInputProcessing.ts`
- `ProcessManager` deletes map entries in `ExitHandler`, so `get()` returning truthy really does mean "spawned and not yet exited"

Two small things I would like addressed before merge, plus one note.

### 1. `unknown` breaks the one aligned column in `session list`

`src/cli/commands/session.ts` prints:

```
`${state} ${star} ${s.tabId}  ${s.agentName} (${s.agentId})  ${name}  ${created}`
```

`busy` and `idle` are both 4 characters, so the `tabId` column currently lines up across rows. `unknown` is 7, so a single unknown row jags the tabId column for the whole listing. Padding keeps the `grep busy` behavior the comment calls out and restores alignment:

```ts
const state = s.state.padEnd(7);
```

### 2. `isSessionBusyWithCli` is called once per agent, and each call re-reads the same file

In `web-server-factory.ts`, hoisting the call out of the tab loop is right, but it is still inside the session loop. `isSessionBusyWithCli` -> `getCliActivityForSession` -> `readCliActivities` performs a synchronous `fs.readFileSync` plus `JSON.parse` of `cli-activity.json` on every call. With N agents that is N redundant synchronous disk reads of the same file on the Electron main thread per `session list`, and the CLI polls this endpoint.

Since `cli-activity.ts` already has a `readCliActivities()`, a bulk helper there (something like `getBusyCliSessionIds(): Set<string>`) would collapse it to a single read and keep the EPERM/ESRCH liveness logic in one place rather than duplicating the probe per session.

### Note (not blocking)

`managedProcessActive` treats "a process is present in the ProcessManager map" as "busy". That holds for batch-mode agents, which spawn a fresh process per message, and every shipping agent (`claude-code`, `codex`, `opencode`, `factory-droid`, `copilot-cli`) has `supportsBatchMode: true`. It does not hold for stdin agents, which keep a long-lived process for the life of the tab and would therefore report `busy` forever. Today that is only `gemini-cli` and `qwen3-coder`, both `hidden: true` and not shipping, so nothing is actually broken here.

That said, the correctness of this line rests entirely on that capability. A one-line comment recording the assumption would stop the next person from wiring up a stdin agent and reintroducing exactly the untruthful state this PR sets out to remove.

Nice, well-scoped change overall - the test coverage on the error-code branches in particular is the right shape. Happy to re-review once the two items above are in.

PR #1392, opened 2026-08-16
feat(shell): allow the clickup: protocol so ClickUp links open the desktop app

felipeggv:
## What this does

Adds `clickup:` to the list of URL protocols `shell:openExternal` is allowed to open. One line.

## Why

Today, clicking a ClickUp task link inside Maestro always lands in a browser tab, even when the user has the ClickUp desktop app installed and signed in. There is no way to reach the app.

The reason is a chain of three things:

1. **ClickUp registers a URL scheme** (`clickup://`) but **does not declare associated-domains.** That second part matters: without it, macOS universal links will never redirect `app.clickup.com` to the desktop app on their own. The scheme is the only route to the app.

2. **Maestro correctly routes non-http URLs to the OS.** `openUrl.ts` hands anything that is not `http`/`https` straight to `shell.openExternal`, which is exactly what a custom scheme needs.

3. **But `ALLOWED_PROTOCOLS` blocks it.** The list is `['http:', 'https:', 'mailto:']`, so the handler throws `Protocol not allowed: clickup:` and nothing opens. The click dies inside Maestro and never reaches the OS.

So the pieces are all in place except this one allowlist entry.

## Why this is safe

Same reasoning that already admits `mailto:`:

- The OS hands the URL to whichever app claims the scheme. **No code is executed** in the handler.
- Unlike `file:` (handled separately above), a custom scheme **cannot reach the filesystem**.
- The vectors this allowlist exists to stop, `javascript:` and `data:`, **remain blocked**, and the existing tests for both still pass.

If a user does not have ClickUp installed, the OS simply does nothing. No crash, no fallback needed.

## How it was verified

- `src/__tests__/main/ipc/handlers/system.test.ts`: **110 passing, 0 failing**, including the cases asserting `javascript:` and `data:` are still rejected.
- Confirmed end to end on macOS that `clickup://app.clickup.com/t/<taskId>` opens the app **and navigates to the task**. ClickUp's own log records both steps:
  ```
  Received open-url:  clickup://app.clickup.com/t/<taskId>
  Extracted deeplink: app.clickup.com/t/<taskId>
  ```

One detail worth knowing for anyone building these links: **the host has to stay in the URL.** ClickUp's handler strips only `clickup://` and then requires the remainder to start with a known host. `clickup://t/<id>` opens the app but navigates nowhere, which makes it look like it worked when it did not.

## Note on the test suite

`git push` runs the full suite via a pre-push hook, and 98 tests across 2 files fail. I checked whether those were mine: they are not. Running the same suite on a clean `main` with no changes produces **the identical 98 failures**, so they are pre-existing and unrelated to this change.

## Scope

One file, one line of code (the rest is a comment explaining the reasoning for the next reader). Branched from `main`, nothing else included.

## Summary by CodeRabbit

* **New Features**
  * Added support for opening `clickup:` links from the application.

* **Bug Fixes**
  * Preserved existing safeguards that block unsupported or potentially unsafe link protocols.

### DIFF
--- a/src/main/ipc/handlers/system.ts
+++ b/src/main/ipc/handlers/system.ts
@@ -188,7 +188,16 @@ export function registerSystemHandlers(deps: SystemHandlerDependencies): void {
 	});
 
 	// Shell operations - open external URLs
-	const ALLOWED_PROTOCOLS = ['http:', 'https:', 'mailto:'];
+	//
+	// `clickup:` lets a ClickUp task link open the native desktop app instead of a browser tab. ClickUp
+	// registers the scheme (CFBundleURLSchemes) but declares no associated-domains, so macOS universal links
+	// never redirect app.clickup.com on their own - the scheme is the only route to the app, and this
+	// allowlist was the one thing standing in the way.
+	//
+	// Safe on the same grounds that already admit `mailto:`: the OS hands the URL to whichever app claims the
+	// scheme, with no code execution here, and unlike `file:` (handled above) it cannot reach the filesystem.
+	// The vectors this list exists to stop, `javascript:` and `data:`, remain blocked.
+	const ALLOWED_PROTOCOLS = ['http:', 'https:', 'mailto:', 'clickup:'];
 	ipcMain.handle('shell:openExternal', async (_event, url: string) => {
 		// Validate URL before opening - Fixes MAESTRO-1S
 		if (!url || typeof url !== 'string') {


pedramamini:
Thanks for the contribution, @felipeggv, and thank you for the writeup. The chain you traced (scheme registered, no associated-domains declared, allowlist blocks it) is accurate, the comment you left in the code is genuinely useful for the next reader, and the safety argument holds: a custom scheme is handed to the OS with no code execution and no filesystem reach here, and `javascript:` / `data:` stay blocked. The branch targets `main` and merges cleanly, so nothing to rebase.

One thing I would like resolved before merging, plus a small ask.

## The allowlist entry looks necessary but not sufficient

The description frames the outcome as "clicking a ClickUp task link inside Maestro always lands in a browser tab, even when the user has the ClickUp desktop app installed." As far as I can trace, a `clickup://` link clicked inside Maestro never reaches `shell:openExternal` at all, so lifting the allowlist by itself does not change what a click does. Every renderer path that turns text into a clickable external link is gated on `http(s)` / `mailto:` before the IPC call:

- `src/renderer/utils/linkify.tsx:6` - the autolink regex only matches `https?://`, so a bare `clickup://app.clickup.com/t/<id>` in agent output is never linkified. There is nothing to click.
- `src/renderer/utils/remarkFileLinks.ts:493-502` and `src/renderer/utils/fileLinks/markdownItAdapter.ts:81-90` - the skip list is `maestro-file://`, `maestro://`, `http://`, `https://`, `mailto:`, `tel:`, `file://`, `#`. `clickup://` is not on it, so a `[task](clickup://...)` href goes through file-path resolution before the click handler ever sees it.
- `src/renderer/components/Markdown/components/MarkdownLink.tsx:139-148` (chat, `directExternal`) - the href is not `^https?://`, `openFileUrl` returns false for it, and `gitToHttps` does not produce an http(s) URL, so the click falls off the end of that branch as a silent no-op.
- `MarkdownLink.tsx:153` and `:158-166` (doc) - not matched by `^https?://|^mailto:`, so it lands in the relative-as-file branch and gets handed to `onFileClick` as if it were a path.
- `src/renderer/components/LinkContextMenu.tsx:58` - `isOpenable` is `^https?://|^mailto:`, so the right-click Open entries are inert for the scheme too.
- `src/main/app-lifecycle/window-manager.ts:460` - `will-navigate` blocks anything that is not the app entry and does not fall back to `openExternal`, so there is no rescue path there either.

The one caller I can find that does reach `openExternal` with an arbitrary scheme is `openUrl()` (`src/renderer/utils/openUrl.ts:34-37`), where anything non-http falls straight through. In practice that is reached by the toast `clickAction: { kind: 'open-url' }` at `src/renderer/components/Toast.tsx:93`, i.e. `maestro-cli notify toast --open-url clickup://...`.

So the question is how you verified it end to end. If it was through that toast/CLI path (or a direct `window.maestro.shell.openExternal` call), then the change is correct for that surface and I am happy to take it. Please just adjust the description so the next reader does not infer that clicking a ClickUp link in chat now opens the app. If the in-chat click was the intent, this needs the companion renderer changes (at minimum an entry in the two skip lists and a branch in `MarkdownLink`), and I would rather see them here than land something that reads as fixing a case it does not reach.

Worth being explicit about either way: a normal `https://app.clickup.com/t/<id>` link still opens in a browser. Nothing here rewrites https ClickUp links into the scheme. If that is the behavior users actually want, it is a separate change and probably wants to be an opt-in setting rather than an unconditional rewrite.

## Small ask: a positive test

Both bots flagged this and it is cheap. Add a case next to the `mailto:` one at `src/__tests__/main/ipc/handlers/system.test.ts:524`, in the same shape:

```ts
it('should allow clickup URLs', async () => {
	vi.mocked(shell.openExternal).mockResolvedValue(undefined);

	const handler = handlers.get('shell:openExternal');
	await handler!({} as any, 'clickup://app.clickup.com/t/abc123');

	expect(shell.openExternal).toHaveBeenCalledWith('clickup://app.clickup.com/t/abc123');
});
```

## You can skip CodeRabbit's host and path restriction

CodeRabbit asks you to pin the host to `app.clickup.com`, reject credentials and ports, and require a `/t/<taskId>` path. Please do not. `http:` and `https:` are already accepted for arbitrary hosts with none of those checks, and a `clickup:` URL can only ever be delivered to whichever app claimed the scheme, so restricting the host buys no security while breaking every legitimate non-task deeplink (docs, lists, spaces). Validating the shape of another product's deeplinks is also a maintenance burden we would own forever on a format we do not control. Your reasoning in the description is the right one.

On the pre-existing suite failures: that matches what we see on a clean tree, so no concern there.

PR #1393, opened 2026-08-17
feat(settings): add OrcaRouter as an LLM provider

Marc-oss-hub:
## Add OrcaRouter as an LLM provider option

This PR adds [OrcaRouter](https://www.orcarouter.ai) as a selectable LLM provider in Maestro's settings, mirroring the existing OpenRouter wiring exactly.

OrcaRouter is an AI gateway that fronts many upstream models (`anthropic/claude-*`, `openai/*`, etc.) behind a single OpenAI-compatible Chat Completions endpoint (`https://api.orcarouter.ai/v1/chat/completions`), so it slots into the existing provider list:

- **`LLMProvider`** type extended with `'orcarouter'` (`src/renderer/types/index.ts`)
- **Settings dropdown** now offers OrcaRouter next to OpenRouter / Anthropic / Ollama
- **Connection test**: a new `llmProvider === 'orcarouter'` branch in `testLLMConnection` mirrors the OpenRouter branch (Bearer auth, `HTTP-Referer`, OpenAI-compatible request/response), with default model slug `anthropic/claude-sonnet-5`
- **Settings metadata** description updated to list OrcaRouter

The wire format was verified live against `https://api.orcarouter.ai/v1/chat/completions`: a POST with Bearer auth, the namespaced model id `anthropic/claude-sonnet-5`, and the same `{ model, messages, max_tokens }` body Maestro sends returns `choices[0].message.content` as expected.

It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

I'm an engineer on the OrcaRouter team.

## Summary by CodeRabbit

* **New Features**
  * Added OrcaRouter as a supported LLM provider.
  * Added connection testing with API key validation and response checks.
  * Displays provider-specific errors when the connection test fails.

* **Documentation**
  * Updated provider settings to include OrcaRouter.

### DIFF
--- a/src/renderer/components/Settings/SettingsModal.tsx
+++ b/src/renderer/components/Settings/SettingsModal.tsx
@@ -377,6 +377,39 @@ export const SettingsModal = memo(function SettingsModal(props: SettingsModalPro
 					status: 'success',
 					message: 'Successfully connected to OpenRouter!',
 				});
+			} else if (llmProvider === 'orcarouter') {
+				if (!apiKey) {
+					throw new Error('API key is required for OrcaRouter');
+				}
+
+				response = await fetch('https://api.orcarouter.ai/v1/chat/completions', {
+					method: 'POST',
+					headers: {
+						Authorization: `Bearer ${apiKey}`,
+						'Content-Type': 'application/json',
+						'HTTP-Referer': 'https://maestro.local',
+					},
+					body: JSON.stringify({
+						model: modelSlug || 'anthropic/claude-sonnet-5',
+						messages: [{ role: 'user', content: testPrompt }],
+						max_tokens: 50,
+					}),
+				});
+
+				if (!response.ok) {
+					const error = await response.json();
+					throw new Error(error.error?.message || `OrcaRouter API error: ${response.status}`);
+				}
+
+				const data = await response.json();
+				if (!data.choices?.[0]?.message?.content) {
+					throw new Error('Invalid response from OrcaRouter');
+				}
+
+				setTestResult({
+					status: 'success',
+					message: 'Successfully connected to OrcaRouter!',
+				});
 			} else if (llmProvider === 'anthropic') {
 				if (!apiKey) {
 					throw new Error('API key is required for Anthropic');
@@ -560,6 +593,7 @@ export const SettingsModal = memo(function SettingsModal(props: SettingsModalPro
 										style={{ borderColor: theme.colors.border }}
 									>
 										<option value="openrouter">OpenRouter</option>
+										<option value="orcarouter">OrcaRouter</option>
 										<option value="anthropic">Anthropic</option>
 										<option value="ollama">Ollama (Local)</option>
 									</select>
--- a/src/renderer/types/index.ts
+++ b/src/renderer/types/index.ts
@@ -92,7 +92,7 @@ export type SettingsTab =
 	| 'prompts';
 // Note: ScratchPadMode was removed as part of the Scratchpad → Auto Run migration
 export type FocusArea = 'sidebar' | 'main' | 'right';
-export type LLMProvider = 'openrouter' | 'anthropic' | 'ollama';
+export type LLMProvider = 'openrouter' | 'orcarouter' | 'anthropic' | 'ollama';
 
 // Inline wizard types for per-session/per-tab wizard state
 export type WizardMode = 'new' | 'iterate' | null;
--- a/src/shared/settingsMetadata.ts
+++ b/src/shared/settingsMetadata.ts
@@ -538,7 +538,7 @@ export const SETTINGS_METADATA: Record<string, SettingMetadata> = {
 
 	// --- LLM / Provider ---
 	llmProvider: {
-		description: 'LLM provider for built-in AI features. E.g., openrouter, anthropic, openai.',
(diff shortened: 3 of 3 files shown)


pedramamini:
@Marc-oss-hub Thanks for the contribution, and thanks especially for disclosing your affiliation with the OrcaRouter team up front. The patch itself is clean and small, and it mirrors the existing OpenRouter wiring faithfully, so this was easy to read.

That said, I don't think we can merge it as-is. A couple of things need resolving first.

**1. The provider is unreachable in any shipped build (blocker)**

`src/renderer/components/Settings/SettingsModal.tsx:44` on `main`:

```ts
const FEATURE_FLAGS = {
	LLM_SETTINGS: false, // LLM provider configuration (OpenRouter, Anthropic, Ollama)
};
```

That flag gates both the tab registration (line 75) and the tab body (line 550), so the LLM settings tab is not rendered at all today. The new `<option value="orcarouter">` and the `llmProvider === 'orcarouter'` branch in `testLLMConnection` are dead code on arrival: nobody can select the provider, and nothing can invoke the connection test. Greptile flagged the same thing.

There's also no consumer of `llmProvider` beyond this modal. The setting is stored (`settingsStore.ts`, `main/stores/defaults.ts`) but no runtime code path reads it to actually route an LLM call, so even with the flag flipped, selecting OrcaRouter would configure a connection test and nothing else.

So adding an option to a disabled, unwired dropdown doesn't yet give users a working OrcaRouter integration. If you want to pursue this, the useful version is the wiring that makes the setting actually do something, which is a much larger change and worth discussing before you build it.

**2. This is a product call for @pedramamini**

Adding a commercial gateway to a hardcoded provider list is a maintainer decision about what Maestro ships, not something a review bot or I should approve unilaterally. Flagging for @pedramamini to weigh in on whether Maestro wants to expand the built-in provider list this way, versus keeping it to the current set or moving to a generic OpenAI-compatible base-URL field that would cover OrcaRouter and every other gateway without per-vendor code. Honestly, that generic approach seems like the stronger design if the LLM settings feature is ever turned on.

One note on the PR description: the paragraph about gateway-level zero-trust security, prompt screening, and default-deny tool governance is product marketing and isn't related to anything in this diff. Please drop it. A PR description should describe the change.

**3. Minor: default model doesn't match the placeholder**

The new branch defaults to `anthropic/claude-sonnet-5` when the model field is blank, but the input placeholder right above it still reads `anthropic/claude-3.5-sonnet` (line 578). A user who leaves the field empty gets a different model than the UI implies. Worth aligning if this moves forward.

For the record, the missing fetch timeout and the `await response.json()` on a non-JSON error body that CodeRabbit and Greptile called out are both pre-existing patterns copied from the OpenRouter branch, not defects you introduced. No need to fix those here.

No merge conflicts, and CI is green, so nothing to rebase. Happy to take another look once there's a plan for point 1 and a maintainer has weighed in on point 2.

pedramamini:
@Marc-oss-hub Following up on my earlier review, and thanks again for the contribution and for the clean, faithful mirroring of the OpenRouter wiring. Unfortunately the ground has shifted underneath this PR since August, and I want to save you the wasted effort of a rebase.

**The surface this PR extends no longer exists on `main`.**

Commit `eda3aeda` ("refactor(settings): remove the dead LLM provider tab, document env-var backends", 2026-09-14, closes #1569) deleted the whole thing:

- the `FEATURE_FLAGS.LLM_SETTINGS` flag and the Settings -> LLM tab
- the `LLMProvider` type in `src/renderer/types/index.ts`
- `testLLMConnection()` in `SettingsModal.tsx`
- the `llmProvider`, `modelSlug`, and `apiKey` settings keys, their defaults, and their metadata entries

That is exactly the blocker I raised as point 1 back in August, resolved in the other direction. The tab had been flag-disabled since 2025-11-26, was built for a planned built-in scratchpad assistant that never shipped, and nothing in the app ever read `llmProvider`. Rather than wire it up, we removed it.

So GitHub now reports this branch as conflicting, but a rebase will not help: there is nothing left to rebase onto. All three files you touched had the relevant hunks deleted, and resolving the conflict would resolve to an empty change. I'd suggest closing this PR. @pedramamini, that's your call.

**The path that does work today**

Maestro spawns each provider's own CLI, so the token backend is whatever that CLI reads from its environment, set per agent under **Edit Agent -> Environment Variables** or globally under **Settings -> Environment**. `docs/configuration.md` now has a "Using a Different Token Backend" section covering this.

One caveat that matters directly for OrcaRouter, and it's the reason we documented it rather than building a provider list: an OpenAI-compatible `/chat/completions` gateway cannot back Claude Code, which speaks the Anthropic Messages API. Pointing `ANTHROPIC_BASE_URL` at an OpenAI-shaped endpoint produces request failures, not a working agent. It needs a translating proxy in between, or an Anthropic-compatible route published by the gateway itself. Codex and OpenCode are OpenAI-shaped already and have no such problem.

If OrcaRouter publishes an Anthropic Messages compatible route, a short docs PR adding it to that section would be genuinely useful to users and far more durable than a per-vendor entry in a hardcoded list. That's the contribution I'd welcome here.

One last note, repeating it since the description is unchanged: the paragraph about gateway-level zero-trust security, prompt screening, and default-deny tool governance is product marketing unrelated to this diff. Please leave that out of future PR descriptions.

Sorry for the churn on this one. The timing was genuinely unlucky.

PR #1400, opened 2026-08-18
fix(readme): restore the broken Star History chart

OctoBored:
The Star History chart in the README is currently broken because GitHub restricts the stargazer API the old chart relied on, so the section renders nothing. This switches the chart to a working mirror that uses a different data source and requires no API token. The chart and its wrapping link now point to the new URL, and everything else is unchanged.

## Summary by CodeRabbit

* **Documentation**
  * Updated Star History embed links and chart endpoints to use the new hosting domain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,10 +187,10 @@ See [CONTRIBUTING.md](CONTRIBUTING.md) for development setup, architecture detai
 
 ## Star History
 
-<a href="https://www.star-history.com/?repos=runmaestro%2Fmaestro&type=date&legend=top-left">
+<a href="https://star-history.dera.page/#runmaestro/maestro&type=date&legend=top-left">
  <picture>
-   <source media="(prefers-color-scheme: dark)" srcset="https://api.star-history.com/chart?repos=runmaestro/maestro&type=date&theme=dark&legend=top-left" />
-   <source media="(prefers-color-scheme: light)" srcset="https://api.star-history.com/chart?repos=runmaestro/maestro&type=date&legend=top-left" />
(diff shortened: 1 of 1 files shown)


pedramamini:
Thanks for the contribution, @OctoBored! Good catch, and the diagnosis holds up under checking.

I verified the premise independently rather than taking it on faith, because the old URL is deceptive: `api.star-history.com` still returns **HTTP 200 with a valid 60KB SVG**, so it looks perfectly healthy to any link checker or status probe. The rendered text inside that SVG is the giveaway:

> GitHub restricted access to star data
> GitHub team is looking into it
> Meanwhile, repo owners can restore it temporarily with a workaround

So the section really was serving a notice image instead of a chart, exactly as you described.

The replacement endpoint returns a real chart for `runmaestro/maestro` with actual plotted data, and the `theme=dark` variant comes back with `background:#0d1117`, so both `prefers-color-scheme` branches resolve correctly rather than just the light one.

I also looked at the interactive link, since its `#runmaestro/maestro&type=date&legend=top-left` fragment is styled differently from the `?repos=...` query used on the images, which seemed worth a second look. It turns out that is exactly what the site's own client serializes back into the address bar (`"#" + repos.join("&") + "&type=" + mode + "&legend=" + position`), so it is the canonical form and needs no change.

One note for the maintainers rather than a change request: this points the README at a community mirror instead of the official domain. GitHub proxies README images through camo, so there is no script execution or viewer tracking exposure, and it is a one line revert if the official API comes back.

No merge conflicts against `main`, and both AI reviewers came back clean. Happy with this one, approving.

PR #1403, opened 2026-08-19
fix(grooming): spawn the detected agent path so compaction works on Windows

### LINKED ISSUE #1402: Bug: Compact and Continue fails on Windows with "spawn claude ENOENT" - wrong agent.command used instead of agent.path
## Summary
`Compact and Continue` (and automatic context grooming, `autoGroomContexts: true`) fails 100% of the time on Windows with `Failed to spawn grooming process for claude-code` / `spawn claude ENOENT`. Not intermittent — every single attempt fails.

This is a regression follow-up to #949 — the fix in #950 correctly surfaced the real underlying error instead of the generic "Context summarization failed" message, which is what made this root cause visible.

## Environment
- Maestro version: 0.17.3
- Operating system: Windows 11 Home (10.0.26200)
- Install source: Packaged build (installer)
- Agent/provider involved: Claude Code

## Steps to Reproduce
1. Open any Claude Code session with enough context to trigger compaction (or wait for automatic grooming with `autoGroomContexts: true`).
2. Click "Compact and Continue", or let automatic grooming kick in.
3. Observe toast: `Compaction Failed - Failed to compact context: Error invoking remote method 'context:groomContext': Error: Failed to spawn grooming process for claude-code`.

## Actual log output
```
[INFO] [ContextSummarizer] Calling groomContext API, prompt length: 18082
[INFO] [ContextGroomer] Starting context grooming {"groomerSessionId":"groomer-...","projectRoot":"...","agentType":"claude-code","promptLength":18082,"hasSessionId":false,"hasSshConfig":false}
[INFO] [ProcessManager] spawn() config {"sessionId":"groomer-...","toolType":"claude-code","platform":"win32","hasPrompt":true,"promptLength":18082,"command":"claude","commandHasExtension":false,"baseArgsCount":5,"finalArgsCount":5}
[INFO] [ProcessManager] About to spawn with shell option {"sessionId":"groomer-...","spawnCommand":"claude","spawnShell":false,"isWindows":true,"argsCount":5,"promptArgLength":30,"fullCommandPreview":"claude --print --verbose --output-format stream-json --dangerously-skip-permissions"}
[ERROR] [ContextMerge] groomContext error {"name":"Error","message":"Failed to spawn grooming process for claude-code", ...}
[ERROR] [ProcessManager] Child process error {"sessionId":"groomer-...","error":"spawn claude ENOENT"}
```

## Root cause (confirmed by reading source)
In `src/main/utils/context-groomer.ts`, `groomContext()` resolves the binary to spawn as:

```ts
const resolvedCommand = sessionCustomPath || agent.command;
```

`agent.command` is the **static** field from `src/main/agents/definitions.ts` (always the literal string `'claude'`, no path, no extension). On Windows this is not a real spawnable binary — the npm-installed CLI resolves to a shim at `%APPDATA%\npm\claude.cmd`, and there's no bare `claude.exe`/`claude` on PATH by default. `child_process.spawn('claude', ..., { shell: false })` therefore throws `ENOENT` immediately.

Every other spawn site in the codebase (regular AI-tab spawning, tab naming, etc.) correctly uses `agent.path` — the absolute path resolved by `AgentDetector` (e.g. `C:\Users\<user>\AppData\Roaming\npm\claude.cmd`) — not `agent.command`. `context-groomer.ts` appears to be the only spawn path that skips this resolution.

This is a **different bug** from the Windows batch-file spawn issue fixed for tab naming (`--tools` empty-flag / `.cmd` shell requirement) — same general area (Windows binary resolution) but a distinct code path and distinct root cause.

## Suggested fix
```ts
const resolvedCommand = sessionCustomPath || agent.path || agent.command;
```
(fallback to `agent.command` preserved for agents/environments where `agent.path` may not be populated, e.g. non-Windows platforms where a bare command name may already be on PATH).

I applied this exact one-line patch locally (manually re-packing `app.asar`) and confirmed compaction now works correctly. Happy to open a PR with this change if useful.

pedramamini:
Closes #1402

## Problem

`Compact and Continue` and automatic context grooming (`autoGroomContexts: true`) failed 100% of the time on Windows with `Failed to spawn grooming process for claude-code` / `spawn claude ENOENT`.

`groomContext()` in `src/main/utils/context-groomer.ts` resolved its spawn target as:

```ts
const resolvedCommand = sessionCustomPath || agent.command;
```

`agent.command` is the static field from `src/main/agents/definitions.ts` - the literal string `'claude'`, with no path and no extension. That is not a spawnable binary on Windows: the npm install leaves a `claude.cmd` shim under `%APPDATA%\npm` and there is no bare `claude` on PATH, so `spawn()` throws `ENOENT` immediately.

This was the only spawn site in the codebase skipping the detector-resolved path. Every other one already uses `agent.path || agent.command`:

- `src/main/ipc/handlers/tabNaming.ts:184`
- `src/main/ipc/handlers/agents.ts:1474`
- `src/main/group-chat/group-chat-router.ts` (several)
- `src/main/group-chat/spawnGroupChatAgent.ts:112`
- `src/main/cross-agent/cross-agent-router.ts:235`
- `src/main/agents/detector.ts:344`

Credit to @pedro12u for the diagnosis and for validating the patch against a repacked build.

## Fix

```ts
const resolvedCommand = sessionCustomPath || agent.path || agent.command;
```

`ChildProcessSpawner` already auto-enables `shell` for `.cmd`/`.bat` targets on Windows (`src/main/process-manager/spawners/ChildProcessSpawner.ts:278`), so handing it the resolved shim path is all that was needed - no shell handling changes here. The `agent.command` fallback is preserved for environments where the detector never populated a path and the bare name is already on PATH.

## Tests

The existing suite missed this because the agent fixture set `command: '/usr/local/bin/claude'`, an absolute path no real agent definition ever has - so `agent.command` looked spawnable in tests. The fixture now mirrors reality (bare `command: 'claude'` plus a detector-resolved `path`), and three cases cover the resolution order:

- falls back to `agent.path` when no `sessionCustomPath` is set
- spawns the Windows `claude.cmd` shim rather than the bare command (the regression)
- falls back to `agent.command` when the detector resolved no path

Both new assertions fail against the pre-fix code (`expected 'claude' to be '...claude.cmd'`) and pass after. 27/27 in `context-groomer.test.ts`; `tsc --noEmit -p tsconfig.node.json` and eslint are clean.

Note: this is a distinct code path and root cause from the Windows batch-file spawn issue previously fixed for tab naming.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved agent command resolution by prioritizing custom paths, detected absolute paths, and reliable fallback commands.
  - Added support for launching Windows `.cmd` command shims correctly.

- **Tests**
  - Expanded coverage for custom path overrides, detected paths, Windows command shims, and fallback behavior.

### DIFF
--- a/src/__tests__/main/utils/context-groomer.test.ts
+++ b/src/__tests__/main/utils/context-groomer.test.ts
@@ -55,7 +55,11 @@ function makeAgent(overrides: Partial<AgentConfig> = {}): AgentConfig {
 		id: 'claude-code',
 		name: 'Claude Code',
 		binaryName: 'claude',
-		command: '/usr/local/bin/claude',
+		// `command` is the static, bare name from definitions.ts; `path` is what
+		// the detector resolves. Keep them distinct so command-resolution tests
+		// can tell which field actually reaches spawn().
+		command: 'claude',
+		path: '/usr/local/bin/claude',
 		args: ['--default'],
 		available: true,
 		capabilities: {} as AgentConfig['capabilities'],
@@ -205,7 +209,7 @@ describe('groomContext', () => {
 		expect(mockPM._lastSpawnConfig!.args).toContain('--resolved');
 	});
 
-	it('uses sessionCustomPath to override agent.command', async () => {
+	it('uses sessionCustomPath to override the detected agent path', async () => {
 		const detector = createMockAgentDetector(agent);
 
 		await groomContext(
@@ -222,7 +226,7 @@ describe('groomContext', () => {
 		expect(mockPM._lastSpawnConfig!.command).toBe('/custom/claude');
 	});
 
-	it('falls back to agent.command when sessionCustomPath is not provided', async () => {
+	it('falls back to agent.path when sessionCustomPath is not provided', async () => {
 		const detector = createMockAgentDetector(agent);
 
 		await groomContext(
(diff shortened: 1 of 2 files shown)


pedro12u:
Confirmed, this matches the root cause I found. Tested the repacked build with the agent.path || agent.command fix and Compact and Continue now spawns correctly on Windows. Thanks for turning it into a clean PR.

PR #1404, opened 2026-08-19
Refactor fetchWithTimeout to a single canonical implementation

reachrazamair:
## Summary

A bare `fetch()` has no timeout. In the main process a stalled socket means an
IPC handler that never settles and a renderer spinner that never stops.

This came out of the deferred `AbortSignal` finding recorded in #1396. Auditing
it repo-wide turned up a wider problem than missing timeouts: the timeout dance
had been written six times, and eleven files skipped it entirely.

Of 28 `fetch()` call sites in `src/main`:

- **Three separate functions named `fetchWithTimeout`**, with three different
  signatures: `ipc/handlers/leaderboard.ts:209` `(url, options, timeoutMs)`,
  `cue/cue-telemetry.ts:464` (same name, `options` non-optional), and
  `bmad-manager.ts:151` `(url, timeoutMs)` built on `AbortSignal.timeout`.
- **Three more inline `AbortController` + `setTimeout` blocks** in `checkin.ts`,
  `agents/detector.ts`, and `agents/codex-usage-sampler.ts`.
- **Eleven files with no timeout at all**, including every user-facing download
  path: marketplace playbooks and assets, spec-kit and OpenSpec prompt
  refreshes, the GitHub releases poll, plugin `net.fetch`, and
  `fs:fetchImageAsBase64`.

The copies had already drifted. `codex-usage-sampler.ts` cleared its timer in
both `catch` and `finally`, and the two same-named helpers disagreed on whether
`options` was optional. `SHARED-UTILS.md` documented no fetch helper at all,
so there was nothing pointing a seventh author at an existing one.

## What changed

`src/main/utils/fetchWithTimeout.ts` is now the single implementation. All six
hand-rolled versions are gone and all 17 non-Symphony call sites go through it,
each with a named per-domain budget constant rather than a bare literal.

Two behaviours are deliberately better than what the copies did, not just
consolidated:

**A caller-supplied `signal` is composed, not overwritten.** Every previous copy
did `signal: controller.signal`, so passing your own `AbortSignal` silently
disabled your cancellation and the request ran to the full budget. The shared
helper uses `AbortSignal.any([caller, timeout])`. This mattered immediately:
plugin `net.fetch` forwards a plugin-supplied `init`.

**A timeout aborts with `TimeoutError`, not a bare `AbortError`.** A caller that
retries on timeout should not also retry a deliberate user cancel.
`isFetchTimeoutError()` is exported for that check.

Timeout values are unchanged where one existed (leaderboard 30s, codex 15s,
telemetry 10s, checkin 5s, models.dev 3s, BMAD 15s). New budgets are 15s for
GitHub metadata polls and 30s for content downloads.

`bmad-manager.ts` keeps a thin local `fetchBmadResource()` that delegates to the
shared helper and adds only its own Sentry reporting. Renamed from
`fetchWithTimeout` so it does not shadow the import. Pushing that reporting into
the shared helper would be wrong: most callers treat an offline machine as
routine, not as a bug.

## Scope

Symphony's 12 fetch sites are deliberately excluded. On `rc` they still live in
the 3,315-line `symphony.ts`, and touching it would conflict with the
decomposition in #1369. Once that merges, migrating them is mechanical against
an already-reviewed helper. Recorded as STILL OPEN in the tracker.

## Tests

New `src/__tests__/main/utils/fetchWithTimeout.test.ts`, 12 tests. The three
behavioural ones were checked red against the old implementation: reverting the
helper to clobber `options.signal` and abort without a reason fails exactly
those three, and the signal-composition test hangs for the full 10s vitest
timeout, which is the bug it exists to catch.

`leaderboard.test.ts` no longer imports the helper it no longer owns; its four
helper tests moved to the new file, and its 31 handler tests are untouched.

Four assertions in `marketplace.test.ts` and `filesystem.test.ts` asserted
`toHaveBeenCalledWith(url)` with a single argument. They now also assert a
`signal` is present, which is a stronger check than before.

## Docs

Registered in all three places an agent would look: the canonical list in
`CLAUDE.md`, a Network Fetch section in `SHARED-UTILS.md`, and finding 41 in
`DEDUP-TRACKER.md` marked RESOLVED with the Symphony remainder noted.

## Verification

Static:

- `npx vitest run`: 1,515 files, 36,951 passed, 0 failed
- `npm run lint` (tsc across lint, main, cli): clean
- `npm run lint:eslint` (including the em-dash config): clean
- `npx prettier --check` on all changed files: clean
- `npm run build` (main, preload, renderer, web-desktop, cli): exit 0
- `grep -rn "await fetch(" src/main` outside Symphony now returns only the one
  call inside the helper itself

Real sockets, no mocks. Against a local HTTP server: a fast 200, a server that
accepts the connection and never sends headers, a slow-header server, a 500, and
a refused connection. Confirmed the happy path is unaffected, a stalled socket
aborts at the budget instead of hanging, a caller abort returns immediately
rather than waiting out a 60s budget, method/headers/body survive the wrapper,
non-2xx still resolves, and a refused connection stays an ordinary transport
error. Also confirmed the event loop drains 2ms after a request made with a 60s
budget, which is the check that the timer is really cleared.

Real call sites, no mocks below the module boundary. Drove the actual exported
functions of `update-checker`, `marketplace-service`, `checkin`,
`speckit-manager`, and `openspec-manager` with only the destination host
redirected, in both a healthy and a stalling condition: 24 assertions, all
passing. Each previously-unbounded call is now bounded by its own budget
(checkin 5s, update-checker 15s, speckit 15s, marketplace 30s), and each still
parses its response correctly when the server is healthy.

Whole app, real network. Built and launched Maestro, then exercised the changed
handlers through the real preload bridge from inside the running renderer:

- `updates` check: 1.5s, real GitHub API, returned actual version data
- `fs:fetchImageAsBase64`: 2.6s, real remote PNG, valid 1.4MB data URL
- `marketplace` manifest: 0.6s, real manifest, 16 playbooks
- `leaderboard`: 1.3s, real runmaestro.ai round-trip
- `speckit.refresh`: 4.5s, real refresh, wrote v0.16.4
- `openspec.refresh`: 1.5s, real refresh, wrote v1.9.0
- `bmad.refresh`: 57.6s, real refresh through the renamed `fetchBmadResource`
  wrapper, metadata updated (many sequential downloads, each now bounded)
- `codex-usage-sampler` and `detector` both ran during startup and completed

Main-process log clean throughout: no ERROR, no unhandled rejection, no
TypeError. UI renders normally.

Not exercised live: `cue-telemetry` (needs telemetry consent) and plugin
`net.fetch` (needs an installed plugin holding the net capability). Both are
covered by the existing unit suites, which pass.

## Summary by CodeRabbit

* **New Features**
  * Added timeout protection to network requests across image fetching, marketplace, updates, telemetry, and external resource loading.
  * Requests now support caller cancellation and clearly distinguish timeout failures from other errors.
  * Added a centralized default 30-second timeout for requests without a custom limit.

* **Bug Fixes**
  * Prevented stalled network requests from hanging indefinitely.

* **Documentation**
  * Documented the shared timeout behavior and recommended usage.

* **Tests**
  * Added coverage for timeouts, cancellation, signal propagation, cleanup, and network errors.


pedramamini:
@reachrazamair Thank you for this one. Collapsing six hand-rolled timeout dances into a single helper, and doing it with a named per-domain budget at every call site instead of bare literals, is exactly the kind of cleanup this codebase needs. The `AbortSignal.any` composition is a genuine correctness win over all six copies it replaces, the `TimeoutError` vs `AbortError` distinction is the right call, and the writeup on why the BMAD Sentry reporting stays local rather than moving into the shared helper is well argued. Registering it in all three doc surfaces means the seventh author will actually find it.

Two things I would like addressed before this goes in.

### 1. Blocking: the six leaderboard timeout branches no longer fire

Both bots flagged this and I confirmed it against the branch. `src/main/ipc/handlers/leaderboard.ts` still has six `catch` blocks doing:

```ts
if (error instanceof Error && error.name === 'AbortError') {
```

at lines 292, 349, 409, 454, 499, and 599. The old local helper called `controller.abort()` with no reason, so a timeout surfaced as `AbortError` and those matched. The shared helper aborts with a `DOMException` named `TimeoutError`, so every one of those checks is now dead: a real leaderboard timeout falls through to the generic connection-error branch, returns the wrong user-facing message, and logs the wrong classification.

You already export the right predicate for this. Swapping those six to `isFetchTimeoutError(error)` is the fix.

Worth noting why the green suite did not catch it: `leaderboard.test.ts` still constructs a hand-made `AbortError` at lines 542, 557, 580, 592, 604, and 616 and asserts `'Request timed out'`. Those tests pass against code that can no longer produce that error, so they need updating in the same change or the branch keeps a false green.

### 2. Worth fixing: BMAD body reads lost their deadline

The general "the budget ends at headers" point from both reviews is a real limitation, but it is documented in the helper's JSDoc and it matches what five of the six replaced copies already did, so I am fine deferring the broad version of it.

BMAD is the exception and it is a straight regression. The old `bmad-manager.ts` helper used `AbortSignal.timeout(timeoutMs)`, which stays armed past the headers and aborts a stalled body read. The shared helper clears its timer in `finally` as soon as the response lands, and every BMAD call site then reads a body: `collectReferencedAssets` does `await response.text()`, `getLatestCommitSha` and `getLatestVersion` do `await response.json()`, and `refreshBmadPrompts` does `await response.text()` in a loop over every command. A raw.githubusercontent.com socket that sends headers and then stalls used to abort at 15s and now hangs a prompt refresh indefinitely.

Simplest fix that keeps the shared helper's contract intact is to have `fetchBmadResource` pass its own `AbortSignal.timeout(timeoutMs)` as `options.signal`. The helper composes rather than clobbers it, which is precisely the behaviour you added, so BMAD gets a body-covering deadline back and nothing else changes.

### Non-blocking

- The broader body-deadline gap (plugin `net.fetch`, `fs:fetchImageAsBase64`, marketplace, spec-kit, OpenSpec, update-checker) is a reasonable follow-up rather than a blocker here, since every one of those sites had no timeout at all before. It would be good to record it in `DEDUP-TRACKER.md` finding 41 next to the Symphony remainder, so the limitation lives somewhere an agent will read rather than only in the helper's JSDoc.
- `SHARED-UTILS.md` says "The ONLY way to make an HTTP request from the main process" while Symphony's 12 sites are deliberately still on bare `fetch()`. A short "outside `symphony.ts`, see #1369" qualifier would keep the doc honest until that lands. Agreed on excluding Symphony here.
- CodeRabbit's point about `fetchBmadResource` reporting to Sentry and the callers reporting the same error again is fair but pre-existing, and your call whether to fold it in.

Everything else in the migration reads clean. `checkin.ts`, `detector.ts`, `codex-usage-sampler.ts`, `cue-telemetry.ts`, marketplace, spec-kit, OpenSpec, and update-checker are all faithful, and the new test file covers the composition behaviour properly. No merge conflicts, the branch is clean against `rc`.

Ping me once the leaderboard branches and the BMAD body deadline are sorted and I will take another pass.

PR #1406, opened 2026-08-20
feat(group-chat): Cmd+F Find bar open/close

reachrazamair:
## Summary

- Group chat had no live-transcript Find: the view replaces `MainPanel` / `TerminalOutput`, where AI Cmd+F lives.
- Adds a Find bar on `GroupChatPanel` scoped with `group-chat::<id>` in the existing `outputSearchByKey` store.
- Cmd+F opens from the group composer and the global shortcut path; Esc / Esc pill closes via the layer stack.

## Tested

- [x] Open a group chat → focus composer → Cmd+F → Find bar appears under the header.
- [x] Type a query → Esc closes; Esc pill closes; Cmd+F reopens.
- [x] Click messages area → Cmd+F still opens Find.
- [x] Right Bar → History focused → Cmd+F still drives the history filter (not transcript Find).
- [x] Leave group chat → agent AI Cmd+F still works as before.
- [x] Note: match count / highlight may show "No matches" until the matching follow-up.

## Summary by CodeRabbit

- **New Features**
  - Added transcript search for group chats.
  - Use **Cmd+F** on macOS or **Ctrl+F** on Windows/Linux to open Find for the current chat.
  - Search is scoped to the active chat window, including agent tabs and group chats.
  - Added match highlighting, navigation, regex support, and Escape-to-close behavior.
  - Repeating the shortcut focuses the existing Find field.

- **Improvements**
  - History-panel filtering remains available when the history view is active.
  - Search updates reliably as transcript content changes.
  - Improved cross-platform filesystem test compatibility.


pedramamini:
@reachrazamair Thanks for picking this up! Group chat genuinely was the one transcript surface without Find, and scoping it into the existing `outputSearchByKey` store with a `group-chat::<id>` namespace is the right call. The `getActiveOutputSearchKey()` precedence change is clean, the new unit test is a nice touch, and I verified locally that the branch type-checks and both CI matrix legs are green. No merge conflicts either.

A few things I'd like addressed before this goes in.

### 1. The Find bar is currently inert, and it looks broken to the user

`OutputSearchBar` is mounted with `totalMatches={0}`, `currentMatchIndex={0}`, and no-op `goToNextMatch`/`goToPrevMatch`. As soon as the user types a single character, the bar renders **"No matches"** with both chevrons disabled, for every query, always.

I understand from the description that this is a deliberate open/close spike with highlight and navigation landing in a follow-up. My concern is that `rc` is the release candidate branch, so whatever merges here is what users get. A Find bar that opens on Cmd+F and then reports "No matches" against a transcript that visibly contains the search term reads as a bug, not as a partial feature. Right now "no Find at all" is arguably the better user experience than "Find that never finds."

Could we either fold the match counting and highlight pass into this PR, or hold this branch and land the whole thing together? Happy to defer to @pedramamini if he'd rather take the increment.

### 2. Cmd+F cannot refocus an already-open group-chat Find bar

CodeRabbit flagged this and I confirmed it. In `useMainKeyboardHandler.ts` the refocus short-circuit runs before your new branch:

```ts
if (isActiveOutputSearchOpen() && (e.metaKey || e.ctrlKey) && !e.altKey && !e.shiftKey && e.key.toLowerCase() === 'f') {
	e.preventDefault();
	document.querySelector<HTMLInputElement>('.terminal-output input')?.focus();
	return;
}
```

`isActiveOutputSearchOpen()` now returns `true` for group chats, but `.terminal-output input` does not exist in the group-chat tree, so this `preventDefault()`s and `return`s having done nothing. Both entry points are affected:

- **Focus in the messages area:** the window listener hits this branch and dies there.
- **Focus in the composer:** `GroupChatInput.handleKeyDown` calls `stopPropagation()`, so the window listener never runs, and the handler just sets `open` to `true` again, which is a no-op since it is already open. Focus stays in the textarea.

So the "Click messages area -> Cmd+F still opens Find" case in your checklist only holds while the bar is closed. CodeRabbit's suggested fix is the right shape: put a `data-output-search-input` attribute on the input in `OutputSearchBar` and switch the selector to `[data-output-search-input]`. That fixes both surfaces in two lines and keeps AI chat working, since it renders the same component.

### 3. The layer lifecycle is copy-pasted from `useTerminalOutputSearch`

Greptile flagged this too, though its specific suggestion does not apply: `useModalLayer` hardcodes `type: 'modal'` and has no `allowClickOutside`, so it cannot express this overlay.

The accurate version of the finding is that `GroupChatPanel.tsx` lines 156-183 are a near-verbatim copy of `useTerminalOutputSearch.ts` lines 49-79 - same `type: 'overlay'`, same `MODAL_PRIORITIES.SLASH_AUTOCOMPLETE`, same `blocksLowerLayers`/`capturesFocus`/`focusTrap`/`allowClickOutside`, same `updateLayerHandler` follow-up effect. Only the `ariaLabel` and the focus target differ. Per CLAUDE.md's rule on not adding new copies of existing helpers, please extract this into a shared hook and consume it from both places, something like:

```ts
useOutputSearchLayer({ open: outputSearchOpen, onEscape: closeSearch, ariaLabel: 'Group Chat Search' })
```

Even better if it also absorbs the keyed store getters/setters (`setOutputSearchOpen`/`Query`/`Regex` bound to a key), since those are duplicated between the two call sites as well. That way the follow-up PR that adds highlighting has one place to hook into rather than two that must be kept in sync.

### Minor

- Leaving a group chat with Find open unmounts the panel and unregisters the layer, but the `open: true` flag stays in the store, so re-entering that chat pops the bar back up with the stale query still in it. `closeSearch` clears the query but navigating away does not. Probably worth deciding intentionally rather than by accident.

Nothing here is structural, the approach is sound. Mainly want to sort out item 1 before this lands on `rc`. Thanks again!

PR #1412, opened 2026-08-21
fix(web-desktop): attach files dropped into chat from a browser

### LINKED ISSUE #1411: Bug: Drag-and-drop file upload silently fails in web-desktop chat
## Summary
Drag-and-drop file upload silently fails in web-desktop chat

## Environment
- Maestro version: 0.18.5-RC

## Steps to Reproduce
1. Open the web-desktop chat window in Maestro.
2. Drag one or more files (tested: two .html files; also a single .pdf) over the chat window.
3. Observe the drop zone highlight.
4. Release the files.
5. Nothing is attached and no error is shown.

## Expected Behavior
Files dropped onto the web-desktop chat window should attach to the chat input as attachments, or surface a clear error if they are rejected.

## Actual Behavior
The drop zone highlights on drag-over, but after dropping, nothing happens: no filename chip, no error toast, no spinner — a completely silent failure. The browser does not open/navigate to the dropped file, indicating the drop event is captured but the files are never attached. Occurs with .html files and with .pdf, so it is not file-type specific.

## Additional Context
Tested with multiple .html files and a .pdf — same silent result, ruling out an HTML-specific file-type filter. Suggests an issue in the drop handler / file-attachment path in web-desktop rather than type validation.

## Screenshots / Recordings
None

pedramamini:
closes #1411

## Problem

Dropping a non-image file (a `.pdf`, a `.html`, anything) onto the chat in the **web-desktop** build did nothing at all: the drop zone highlighted, then no chip, no @mention, no spinner, no error. A completely silent failure.

Root cause is in `handleDrop` (`src/renderer/hooks/input/useInputHandlers.ts`). Non-image drops are turned into `@<path>` mentions, and the path comes from `window.maestro.fs.getPathForFile(file)`, which is backed by Electron's `webUtils`. The web-desktop build runs the renderer in a plain browser, where the web-desktop `electron` shim can only read the non-standard `File.path` - which browsers do not set. So it returns `''` and the old code skipped the file with no branch for the empty case:

```ts
const filePath = window.maestro.fs.getPathForFile(file);
if (filePath) {
    externalPaths.push(toMentionPath(filePath, projectRoot));
}
// ...and nothing at all otherwise
```

Images were unaffected (they go through `FileReader` and stage as data URLs), which matches the report that both `.html` and `.pdf` fail while the behavior is not file-type specific.

## Fix

A dropped file with no resolvable path is now uploaded to the machine running Maestro rather than skipped. `uploadPathlessFile()` (new, in `src/renderer/utils/osFileDrop.ts`) reads the file as raw base64 and hands it to the existing `attachments:save` IPC handler, which writes it to `userData/attachments/<sessionId>/<filename>` and returns the absolute host path. That path is then @mentioned exactly like a desktop drop, so the agent has a real file to read.

Note this is also the right shape for the remote case: in web-desktop the browser may not be the same machine as the agent, so a client-side path would be useless even if the browser gave us one.

Details:

- **Size cap:** 25 MB. The bytes cross the WebSocket bridge as base64 inside a single message, so an unbounded file would stall the bridge for every other call.
- **No more silent failures:** oversize files, unreadable files, and failed host writes each raise a red toast naming the file. Folders dragged in from a browser arrive as unreadable `File` entries and land in the same path, so they report an error instead of disappearing.
- **Coverage:** the main-panel drop zone, the chat textarea, and group chat all funnel into this one `handleDrop`, so all three are fixed.
- Desktop (Electron) behavior is unchanged - `getPathForFile` still returns a real path there, and the upload branch is only reached when it comes back empty.

## Testing

- `src/__tests__/renderer/utils/osFileDrop.test.ts` - `uploadPathlessFile` sends raw base64 (not the `data:...;base64,` prefix, which `attachments:save` would write to disk verbatim and corrupt the file), rejects over the size cap without touching the host, and surfaces the host error.
- `src/__tests__/renderer/hooks/useInputHandlers.test.ts` - a path-less drop uploads and @mentions the host copy; a failed upload raises a red toast and mentions nothing.
- Full suite green locally: 37,386 passed.

## Not covered

Other `getDroppedPaths()` callers (dragging OS files into the Files panel to import them) have the same browser limitation and still silently drop path-less files. That is a separate surface from the one reported here and is left for a follow-up.

Targets `rc` rather than `main`: `src/web-desktop/` does not exist on `main` yet.

## Summary by CodeRabbit

* **New Features**
  * Added support for browser file drops without filesystem paths.
  * Dropped files are uploaded as attachments and inserted into drafts as project-relative `@` mentions.
  * Added support for pathless PDFs and other non-image files.
  * Uploads remain associated with the original conversation when navigating during processing.
  * Added unique attachment naming and Windows path normalization.
  * Added a 25 MB upload limit with clear notifications for oversized or failed uploads.

### DIFF
--- a/src/__tests__/renderer/hooks/useInputHandlers.test.ts
+++ b/src/__tests__/renderer/hooks/useInputHandlers.test.ts
@@ -1970,9 +1970,95 @@ describe('useInputHandlers', () => {
 				expect(window.maestro.attachments.save).toHaveBeenCalledWith(
 					'session-1',
 					'cGRmLWJ5dGVz',
-					'doc.pdf'
+					// Uniquified so a second `doc.pdf` cannot overwrite this one.
+					expect.stringMatching(/^doc-[0-9a-z]+\.pdf$/)
 				);
-				expect(inputVal()).toBe('@/userData/attachments/session-1/doc.pdf ');
+				expect(inputVal()).toMatch(/^@\/userData\/attachments\/session-1\/doc-[0-9a-z]+\.pdf $/);
+			});
+
+			// Regression: the upload awaits the host, and the composer store holds
+			// whichever draft is on screen when it resolves. Without a pin the
+			// mention landed in the tab the user had switched to.
+			it('mentions the tab the file was dropped into, not the one switched to', async () => {
+				useSessionStore.setState({
+					sessions: [
+						createMockSession({
+							id: 'session-1',
+							inputMode: 'ai',
+							aiTabs: [
+								{ id: 'tab-1', name: 'Tab 1', inputValue: '', data: [], stagedImages: [] },
+								{ id: 'tab-2', name: 'Tab 2', inputValue: '', data: [], stagedImages: [] },
+							] as any,
+							activeTabId: 'tab-1',
+						}),
+					],
+					activeSessionId: 'session-1',
+				} as any);
+
+				let landTheUpload: () => void = () => {};
+				vi.mocked(window.maestro.attachments.save).mockImplementationOnce(
+					(sessionId: string, _b64: string, filename: string) =>
+						new Promise((resolve) => {
+							landTheUpload = () =>
+								resolve({
+									success: true,
+									path: `/userData/attachments/${sessionId}/${filename}`,
+								});
+						})
+				);
+
+				const deps = createMockDeps();
+				const { result, rerender } = renderHook(() => useInputHandlers(deps));
+
+				// Awaited so the read finishes and the upload is actually in flight
+				// (pending on the host) before the user moves on.
+				await act(async () => {
+					result.current.handleDrop(dropPathlessPdf());
+				});
+
+				// The user moves on while the bytes are still crossing the bridge.
+				act(() => {
+					useSessionStore.setState({
+						sessions: useSessionStore
+							.getState()
+							.sessions.map((sn: any) => ({ ...sn, activeTabId: 'tab-2' })),
+					} as any);
+				});
+				rerender();
+
+				await act(async () => {
+					landTheUpload();
+				});
+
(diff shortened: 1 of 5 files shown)


pedramamini:
Addressed the review round:

**Wrong-draft target (greptile P1 / CodeRabbit Major)** - fixed. `handleDrop` now captures the tab (or group chat) the file was dropped into and threads it through `uploadAndMentionPathlessFiles`. When the pinned tab is still on screen the mention goes into the live composer as before; when it is not, it is appended to that tab's own persisted draft. The background write deliberately does not go through `syncAiInputToSession` - that reads `aiCommandMode` from the live composer, so it would stamp the on-screen tab's bang-ladder rung onto a background tab, and it cancels the queued draft flush belonging to the tab the user is actually typing in. Focus is only stolen when the mention landed in the composer that is on screen. A tab closed mid-upload now raises a yellow toast instead of swallowing the mention. Regression test added and verified to fail without the pin.

**Same-name overwrite (greptile P1)** - fixed in `uploadPathlessFile` rather than in `attachments:save`, since this is that handler's only renderer caller and the existing contract is relied on elsewhere. Each upload gets `name-<token>.ext`. Tests cover a repeat name, an extensionless name, and a dotfile.

**Token width (CodeRabbit)** - widened from 8 to 12 hex characters. Not taking the full UUID: the filename lands in the user's draft as an `@mention` they have to read past, and the space being drawn from is a handful of same-named drops inside one session's attachments directory, not the whole store. 48 bits puts a collision well out of reach there.

**POSIX-only fixtures (greptile P2)** - partly. The renderer tests deliberately avoid node `path`, so rather than rewrite the fixtures I added a case that proves the real Windows behaviour: a `C:\...` host path normalises into the mention.

Suite green: 1547 files / 37391 tests, plus `tsc`, eslint and prettier clean.

PR #1418, opened 2026-08-21
fix(grooming): run context grooming on the session's SSH remote

### LINKED ISSUE #1416: Context grooming ignores SSH remote config and always runs locally
Split out of review feedback on #1403.

## What happens

`groomContext()` in `src/main/utils/context-groomer.ts` accepts a `sessionSshRemoteConfig` option, logs it, and passes it into `processManager.spawn({ ..., sessionSshRemoteConfig })`.

None of that has any effect. `ProcessConfig` (`src/main/process-manager/types.ts`) has no `sessionSshRemoteConfig` field, and `ProcessManager.spawn()` does no SSH wrapping of its own - callers are expected to run `wrapSpawnWithSsh()` themselves before spawning. Grooming never does.

The result is that context grooming, context merge, Director's Notes synopsis, and group chat reset all spawn the agent on the **local** machine even when the agent is configured to run on an SSH remote. The user explicitly opted into SSH, so their prompt (and the transcript context it carries) runs somewhere they did not choose.

## Why it is not obvious

The option being present, typed, and threaded all the way to the spawn call reads as working SSH support. `#1403` adds a comment at the spawn site marking it as a known gap so nobody else reaches the same wrong conclusion.

## What the fix looks like

The three call sites that already do this correctly are the models to copy:

- `src/main/group-chat/spawnGroupChatAgent.ts:195`
- `src/main/cue/cue-spawn-builder.ts:220`
- `src/cli/services/agent-spawner.ts:99`

Steps:

1. Thread an SSH store into `groomContext()` - `createSshRemoteStoreAdapter(settingsStore)` from `src/main/utils/ssh-remote-resolver.ts`.
2. When `sessionSshRemoteConfig?.enabled`, wrap the spawn config with `wrapSpawnWithSsh()` and spawn the wrapped `command` / `args` / `cwd`.
3. **Do not send `agent.path` to the remote.** `#1403` changed `resolvedCommand` to `sessionCustomPath || agent.path || agent.command` to fix a Windows ENOENT, and `agent.path` is an absolute path on the *local* filesystem. Remote execution has to use the agent's `binaryName`.
4. Fail loudly when SSH is enabled but the remote cannot be resolved, rather than silently falling back to a local run. See `sshUnresolvedFailure()` in `src/cli/services/agent-spawner.ts` for the established shape.
5. Revisit `useStdinForPrompt`, which currently disables the Windows stdin path whenever SSH is enabled on the assumption that `ssh-spawn-wrapper` is handling stdin.

Update the four main-process callers: `src/main/ipc/handlers/context.ts:163`, `src/main/ipc/handlers/director-notes.ts:872`, `src/main/ipc/handlers/groupChat.ts:774`, `src/main/web-server/web-server-factory.ts:3036`.

`CLAUDE.md` calls this out by name under "SSH Remote Execution Awareness", listing context grooming as a spawn surface that must support SSH.

pedramamini:
closes #1416

## The bug

`groomContext()` accepted a `sessionSshRemoteConfig`, logged it, and passed it into `processManager.spawn()`. `ProcessConfig` has no such field and `ProcessManager.spawn()` does no SSH wrapping of its own - every spawn surface has to call `wrapSpawnWithSsh()` itself. Grooming never did, so the option was silently dropped and the prompt (plus the transcript context it carries) always ran on the local machine even though the user explicitly opted into SSH.

## The fix

In `src/main/utils/context-groomer.ts`:

- **Actually wrap the spawn.** When `sessionSshRemoteConfig.enabled`, run `wrapSpawnWithSsh()` and spawn the wrapped `command` / `args` / `cwd`, forwarding `sshStdinScript` / `sshRemoteCommand` / `sshRemoteId` / `sshRemoteHost` to `spawn()`. Same shape as `spawnGroupChatAgent.ts`, `cue-spawn-builder.ts`, and the CLI `agent-spawner.ts`.
- **Never send a local path to the remote.** The remote binary resolves as `sessionCustomPath || agent.binaryName`, matching what `process:spawn` does. A locally-detected absolute path (`agent.command`, and `agent.path` once #1403 lands) does not exist on the remote host.
- **Use the remote working dir.** `projectRoot` is the local path for an SSH agent and the remote shell does `cd <cwd> || exit 1`, so the wrap now prefers `sshRemoteConfig.workingDirOverride`.
- **Fail loudly.** If SSH is enabled but the remote (or the SSH store) can't be resolved, grooming throws instead of quietly running the prompt locally. `wrapSpawnWithSsh()` degrades to a local config on an unresolved remote, so the `sshRemoteUsed` check is what makes this loud. Message follows `sshUnresolvedFailure()` in the CLI spawner.
- **`useStdinForPrompt`** now keys off whether SSH was actually used rather than off the raw config flag. When SSH is in play the wrapper owns the child's stdin (the prompt is in the stdin script or already in the command line), so the Windows raw-stdin path must stay off; for every other case, including an enabled-but-disabled-remote config, Windows still gets its ENAMETOOLONG workaround.

Callers:

- `src/main/ipc/handlers/context.ts` (context merge / summarize-and-continue) threads `createSshRemoteStoreAdapter(getSettingsStore())` when the session has SSH on. This is the only caller that was passing an SSH config, i.e. the reported bug.
- `src/main/ipc/handlers/groupChat.ts` (context reset) now resolves the participant's SSH config. Participant records only store the remote's display name, so it reads the config off the agent whose name matches the participant - the same association the router makes when dispatching a turn. Previously the reset summary tried to resume a *remote* agent session locally, where that session doesn't exist.
- `src/main/ipc/handlers/director-notes.ts` and `src/main/web-server/web-server-factory.ts` (Director's Notes synopsis) stay local **by design** - the prompt is a manifest of history file paths on this machine. Both call sites now say so instead of leaving the next reader to wonder.

## Tests

`src/__tests__/main/utils/context-groomer.test.ts`, 8 new cases against the real `wrapSpawnWithSsh` (only `resolveSshPath` is stubbed): spawns `ssh` rather than the agent, invokes the agent by binary name and never the local path, honors `sessionCustomPath` as the remote binary, cds to the remote working dir, routes large prompts through the stdin script, throws on an unresolvable remote, throws when no store is provided, and still spawns locally when the config is present but disabled. The pre-existing "passes SSH remote config through to spawn" case asserted the broken behavior and was replaced.

Full suite green locally (33,882 passed), plus `npm run lint`, eslint, prettier.

## Note on #1403

That PR touches `resolvedCommand` in the same function (`sessionCustomPath || agent.path || agent.command`). This branch leaves the local command line alone, so the two should merge cleanly; whichever lands second, the remote path here stays on `binaryName` and is unaffected.

## Summary by CodeRabbit

* **New Features**
  * Added SSH remote execution support for context grooming.
  * Group chat context summaries can resume using each participant’s configured remote environment.
  * Added support for remote working directories, custom agent paths, and large prompt transfers.
  * Session listings now more accurately indicate active or unknown states.
* **Bug Fixes**
  * Missing or unresolvable SSH configurations now report errors instead of silently running locally.
  * Local fallback remains available when SSH is disabled.
  * Director’s Notes grooming continues to run locally when appropriate.

### DIFF
--- a/src/__tests__/main/utils/context-groomer.test.ts
+++ b/src/__tests__/main/utils/context-groomer.test.ts
@@ -10,6 +10,8 @@ import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { groomContext, type GroomingProcessManager } from '../../../main/utils/context-groomer';
 import type { AgentDetector } from '../../../main/agents';
 import type { AgentConfig } from '../../../main/agents';
+import type { SshRemoteSettingsStore } from '../../../main/utils/ssh-remote-resolver';
+import type { SshRemoteConfig } from '../../../shared/types';
 
 // Mock agent-args module to verify override resolution
 vi.mock('../../../main/utils/agent-args', () => ({
@@ -37,6 +39,12 @@ vi.mock('uuid', () => ({
 	v4: vi.fn(() => 'test-uuid'),
 }));
 
+// Keep the real SSH wrapping logic, but pin the resolved ssh binary so the
+// built command is deterministic (and no PATH probing happens in tests).
+vi.mock('../../../main/utils/cliDetection', () => ({
+	resolveSshPath: vi.fn().mockResolvedValue('ssh'),
+}));
+
 // Mock logger
 vi.mock('../../../main/utils/logger', () => ({
 	logger: {
@@ -61,6 +69,19 @@ function makeAgent(overrides: Partial<AgentConfig> = {}): AgentConfig {
 		capabilities: {} as AgentConfig['capabilities'],
 		...overrides,
 	};
+}
+
+const SSH_REMOTE: SshRemoteConfig = {
+	id: 'my-remote',
+	name: 'Dev Box',
+	host: 'dev.example.com',
+	port: 22,
+	username: 'testuser',
+	enabled: true,
+};
+
+function createSshStore(remotes: SshRemoteConfig[] = [SSH_REMOTE]): SshRemoteSettingsStore {
+	return { getSshRemotes: () => remotes };
 }
 
 function createMockProcessManager(): GroomingProcessManager & {
@@ -310,7 +331,7 @@ describe('groomContext', () => {
 		expect(mockPM._lastSpawnConfig!.sendPromptViaStdinRaw).toBe(false);
 	});
 
-	it('does NOT set sendPromptViaStdinRaw on Windows when SSH is enabled', async () => {
+	it('does NOT set sendPromptViaStdinRaw on Windows when SSH is used', async () => {
 		vi.mocked(isWindows).mockReturnValue(true);
 		const detector = createMockAgentDetector(agent);
 
@@ -319,7 +340,8 @@ describe('groomContext', () => {
 				projectRoot: '/project',
 				agentType: 'claude-code',
 				prompt: 'summarize',
-				sessionSshRemoteConfig: { enabled: true, remoteId: 'r1' },
+				sessionSshRemoteConfig: { enabled: true, remoteId: 'my-remote' },
+				sshStore: createSshStore(),
 			},
 			mockPM,
 			detector
@@ -328,8 +350,7 @@ describe('groomContext', () => {
 		expect(mockPM._lastSpawnConfig!.sendPromptViaStdinRaw).toBe(false);
 	});
 
-	it('passes SSH remote config through to spawn', async () => {
-		const sshConfig = { enabled: true, remoteId: 'my-remote', workingDirOverride: '/remote/dir' };
+	it('spawns ssh (not the agent) when the session has an SSH remote', async () => {
 		const detector = createMockAgentDetector(agent);
 
 		await groomContext(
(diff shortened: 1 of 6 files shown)


pedramamini:
Review pass done. Merged `main` in (the conflict was main's comment documenting this exact gap on the dropped `sessionSshRemoteConfig` spawn field - the fix replaces it; #1403's `agent.path` change merged clean, the remote invocation stays on `binaryName`).

**Applied:**

- **CodeRabbit, `groupChat.ts`: read `sessionSshRemoteConfig`.** Correct and a real bug - the flat `sshRemoteConfig` field only exists on the mapped view the router consumes (`setGetSessionsCallback` in `main/index.ts`), so the lookup always returned `undefined` and a remote participant's reset summary still groomed locally. Fixed.
- **Greptile, name-based lookup.** Fixed by making it match the router's own predicate (`mentionMatches(session.name, participantName) || session.name === participantName`) instead of strict equality. A participant record carries no stable source-session id - only the name - and the router already picks the participant's cwd, `sshRemoteConfig`, `tokenMode`, and `maestroPPath` for every turn with that same lookup (`group-chat-router.ts:1240`, `:1834`). So this is not a new association; grooming on a *different* rule than the turn dispatch would be the worse outcome, since the summary would then resume on a host the turns never ran on. Making the association identity-stable means persisting the source session id on the participant record, which is a separate change to `GroupChatParticipant` and its write paths.
- **CodeRabbit, `SSH_REMOTE` fixture.** `privateKeyPath` is required on `SshRemoteConfig`; added.

**Skipped:**

- **Legacy top-level `sshRemoteId` fallback.** `sshRemoteId` is the session-wide SSH *context* (file explorer, git, Auto Run), not the execution config, and no spawn path reads it - `setGetSessionsCallback` resolves execution purely from `sessionSshRemoteConfig`. Adding the fallback here would send grooming to a remote when the participant's own turns run locally.

Also added 4 tests over `resolveParticipantSshRemoteConfig` (persisted field, normalized-name parity with the router, local participant, deleted source agent). Full suite green: 33,967 passed, plus tsc, eslint, prettier.

PR #1419, opened 2026-08-21
feat(git): name the agent in the pull/push modal header

pedramamini:
## Problem

Pull and Push are reachable by right-clicking **any** row in the Left Bar (`useGitAgentActions` backs all three surfaces: the header branch pill, the Left Bar context menu, and the command palette). So the agent they target is frequently *not* the highlighted one.

The header read `git push` and named nothing:

```
BEFORE                          AFTER
+------------------------+      +--------------------------------+
| ⬆ git push         [x] |      | ⬆ git push · Sonoma-Fix    [x] |
+------------------------+      +--------------------------------+
```

If the target agent isn't already selected, there is no way to tell what you are about to push.

## Fix

New `subtitle` prop on the shared `<Modal>`, for the subject a modal acts on: which agent, which repo, which file. Rendered dimmed after the title, skipped entirely when empty, so every existing `<Modal>` caller looks exactly as it did.

**The name deliberately does not go into `title`.** `title` is the `aria-label` and the modal-layer label, and it seeds the fallback resize key via `getDefaultResizeKey()`. A per-agent title would mint a different persisted window size for each agent, so the modal would forget its size whenever you targeted a different one. (`GitCommandRunnerModal` passes an explicit `resizeKey`, so it dodges that specific consequence - but the trap is real for the follow-up modals listed below, and the aria-label/layer-label reasons apply regardless.)

**No prop plumbing needed.** `useGitAgentActions.ts:152` has been putting `sessionId` in the `gitCommandRunner` payload since it was written; the modal was destructuring around it (`operation, cwd, sshRemoteId, branch`).

**Perf:** the modal subscribes to `?.name` alone, not the `Session`. It stays on screen while a command streams, and a whole-session subscription would re-render the console on every unrelated token and log update.

## Testing

Three tests in `GitCommandRunnerModal.test.tsx`, plus the existing 8:

- the header names the target agent (verified to fail without the change)
- the title stays the bare command line, so both the pre-existing `getByText('git push')` assertion and the resize key survive
- a session that no longer exists renders no subtitle rather than crashing

```
vitest Modal.test.tsx + GitCommandRunnerModal.test.tsx   28 passed
tsc -p tsconfig.json --noEmit                            clean
eslint / prettier                                        clean
```

## Not in this PR

This is systemic - of ~15 modals reachable from the right-click menu, **one** (`EditAgentModal`) names its agent. The `subtitle` prop makes most of the rest one-line fixes (Branch Switcher, Rename, both delete confirms, Duplicate, Create Group), but that is a separate sweep.

Two of them need behaviour fixes rather than a header, and a subtitle would make them *more* misleading:

- **Configure Worktrees** - `useWorktreeHandlers.ts:189` calls `setActiveSessionId(session.id)` before opening, so right-clicking an arbitrary agent silently changes which agent is selected.
- **Configure Maestro Cue** - `useModalHandlers.ts:550` is `async (_session: Session) => {`. The session is discarded and a global dashboard opens.

## Summary by CodeRabbit

* **New Features**
  * Modal dialogs can now display contextual subtitles alongside their titles.
  * Git command dialogs show the targeted agent’s name as a subtitle while keeping the command as the main title.

* **Bug Fixes**
  * Long subtitles now fit more cleanly within the modal title area.
  * Empty subtitles no longer display unnecessary separators.

* **Documentation**
  * Added guidance for using modal subtitles to identify action subjects.

### DIFF
--- a/docs/agent-guides/UI-PATTERNS.md
+++ b/docs/agent-guides/UI-PATTERNS.md
@@ -187,7 +187,7 @@ The expanded Prompt Composer (`src/renderer/components/PromptComposerModal.tsx`)
 
 A modal opened from a **right-click menu** frequently acts on something other than the highlighted agent. `git push` as a header names the operation but not the target, so a user who right-clicked an arbitrary Left Bar row has no way to tell which agent is about to push.
 
-Pass `<Modal subtitle={...}>` for the subject: which agent, which repo, which file. It renders dimmed after the title (`git push - Sonoma-Fix`) and is skipped entirely when empty, so a modal with nothing to name looks exactly as it did before. `GitCommandRunnerModal.tsx` is the reference use.
+Pass `<Modal subtitle={...}>` for the subject: which agent, which repo, which file. It renders dimmed after the title (`git push · Sonoma-Fix`) and is skipped entirely when empty, so a modal with nothing to name looks exactly as it did before. `GitCommandRunnerModal.tsx` is the reference use.
 
 **Do not concatenate the subject into `title`.** `title` is the `aria-label` and the modal-layer label, and it seeds the fallback resize key via `getDefaultResizeKey()` - a per-agent title mints a different persisted window size for every agent, so the modal would forget its size each time you targeted a different one. That is also why any test asserting on the bare title keeps passing after a `subtitle` is added.
 
--- a/docs/agent-guides/UI-PATTERNS.md
+++ b/docs/agent-guides/UI-PATTERNS.md
@@ -183,6 +183,16 @@ Guidance:
 
 The expanded Prompt Composer (`src/renderer/components/PromptComposerModal.tsx`) is the reference implementation of the compact-vs-`90vw x 90vh` toggle.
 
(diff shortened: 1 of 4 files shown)


pedramamini:
All three CodeRabbit findings were fixed in `d32f42d73`; the re-posted comments are anchored to the previous commit. Verified against the current branch head:

- **Separator in the guide** - now `git push · Sonoma-Fix`, matching what `Modal.tsx` renders.
- **`min-w-0` on the subtitle span** - `truncate` cannot shrink a flex item below its min-content width without it, so a long agent name pushed the close button instead of ellipsising. The `min-w-0` already on the parent group only covered the group, not the span inside it.
- **Resize-key assertion** - the test now asserts `data-modal-resize-key` is still `modal-git-command-runner`, which is the concrete consequence of folding the name into the title.

Also took Greptile's P2: the guard is now plain truthiness rather than `undefined`/`null`/`''`, since `subtitle={agent && agent.name}` yields `false` and would have painted a separator with nothing after it. Covered by a new test.

6/6 CI green including both Windows shards.

PR #1422, opened 2026-08-23
fix(sentry): stop four rc field issues from repeating a report per tick

pedramamini:
Twelfth Sentry triage on `rc`. All four issues are confirmed live on the current rc release line (0.18.3-RC through 0.18.5-RC, where 0.18.5-RC is HEAD's `package.json` version).

Every one is the same shape: a best-effort or expected-condition path that fires `captureException` / `captureMessage` **once per timer tick**, so a single broken install produces most of the project's event volume and drowns out real signal.

## MAESTRO-KE - 1,016 occurrences, still firing today

The Cue GitHub poller reports every `gh` failure that is neither a rate limit nor a connectivity error. A `gh` token that expired or was never created fails with:

```
HTTP 401: Bad credentials (https://api.github.com/graphql)
Try authenticating with:  gh auth login
```

...on **every poll, forever**, because the user has to fix it outside Maestro. One install produced 929 events in 14 days.

Added `isGitHubAuthError` and applied it at both call sites on that path:

- `doPoll`, the catch the events are tagged with (`cue:github:doPoll`)
- `resolveRepo`, which runs *first* for auto-detect triggers and would otherwise still page once per tick even with the `doPoll` carve-out in place

Behavior is otherwise unchanged: the poll is still skipped, the backoff and seed-marker semantics are untouched, and the user gets a Cue log line that names `gh auth login`.

`HTTP 403` is deliberately **not** matched here. `gh` returns 403 for both a permissions problem and a rate limit, and `isGitHubRateLimitError` already claims it, so stealing the case would stop the poller backing off.

## MAESTRO-Q2 - 10,485 occurrences, the noisiest issue in the project

The Claude usage sampler reports an identical `maestro-p --status sample failed` warning on every tick for as long as a config dir stays broken.

Now it reports the **first** occurrence of each distinct `(stage, reason)` signature per config dir, re-reports an unchanged one every 6 hours, and forgets the history after a successful sample so a flapping account reports again.

**What this deliberately does not do:** it makes no claim that any exit code is expected. Earlier triages deferred that call (it needs a decision on `maestro-p` exit semantics) and it is still deferred. Every *distinct* failure a user hits still reaches Sentry, and a regression that hits many installs still shows up as many events - one per install - instead of being buried under one install's repeats. Earlier triage notes named exactly this ("report-on-state-transition, not every-tick") as the candidate fix.

## MAESTRO-ZD

`SqliteError: database or disk is full` from the stats query-event buffer flush. This is the out-of-disk condition `shouldDropSentryEvent` already drops for `ENOSPC`, just worded by SQLite instead of libuv.

Dropping it at the filter rather than at one call site matters: a full volume fails **every** stats and cue DB write, and there are ~20 `captureException` sites across `src/main/stats/` alone. The buffer flush is simply the one that reached us first. Those writes are best-effort telemetry and the code already drops the batch rather than retrying.

Narrow on purpose - `SQLITE_CORRUPT`, `SQLITE_BUSY` and I/O errors still report, because those are the shapes that can indicate a real bug of ours.

## MAESTRO-ZS

A remote command that lands while its agent is mid-turn is refused by `ProcessManager.spawn`, on purpose (the session already owns a live process). That is an ordinary race, not a fault: the caller gets an honest `accepted: false` receipt and the tab gets an error log, so the user is told either way.

The refusal message now lives in `src/shared/processErrors.ts` so the thrower and the renderer's matcher cannot drift. This is not incidental - prior triages twice found filter rules that had been **dead in the field for months** because the text they matched had moved (#1213's EPIPE word order, #1214's LinkedErrors position assumption).

## Also in here

The poller's `reports non-rate-limit/non-connectivity errors to Sentry` case used `gh auth login required` as its stand-in for a generic failure - which is exactly the shape the field data says is *expected*. Swapped for a genuinely unexpected error (`panic: runtime error`) so the case still tests what it claims to, and added an explicit assertion that `isGitHubAuthError` claims the old string.

## Verification

- Regression tests for all four, **each confirmed red against the unfixed source** (`git stash push` of source files only, plus targeted probes for the two behaviors whose exports the stash would break).
- Controls stayed green throughout: a rejecting spawn still reports, an unexpected `gh` failure still reports, `SQLITE_CORRUPT`/`BUSY`/`no such table` still report.
- `npx tsc --noEmit -p tsconfig.json`: 0 errors.
- ESLint clean on changed files.
- Full suite: **1,594 files, 38,328 tests, 0 failures.**

## Triaged and skipped, with reasons

- **MAESTRO-YV / Z8** (`contextTimeline:*` no handler) and **MAESTRO-X4** (`spawn EFTYPE`) - already fixed at HEAD by #1389 and #1310. Field events predate the ship.
- **MAESTRO-ZP / ZQ** (`createTreeWalker ... parameter 1 is not of type 'Node'`, culprit `translateTree(assets/maestro-zh-tw)`) - **not our code.** `translateTree` and `maestro-zh-tw` have zero hits anywhere in git history; this is a third-party translation script injected into the renderer. ZQ is the `[Splash]` twin of ZP, i.e. one event double-reported.
- **MAESTRO-V9 / RR / JS** - guards present and correct at HEAD; events are from pre-fix builds.
- **MAESTRO-RS** (GLIBC 2.38 / better-sqlite3) - Linux build-target infra, still needs a human. Now 5+ triages deep.
- **MAESTRO-CT / R0 / RH / NM** (`spawn ENOENT`/`EINVAL`/`E2BIG`) - genuine spawn signal, not noise.
- Native/GPU and renderer-crash issues (HQ, SZ, QG, 9K, ZR, ZF, 38, Y8, 23, YW, 5A, 2C, Y, 1J, Z2, Z3, ZK, YT, 62, QH) - unchanged from prior triages.

Also re-ran the IPC wiring sweep from the eleventh triage (duplicate `ipcMain.handle` channels; preload `invoke` vs main `handle` coverage; registrar parity), since there was a recent back-merge and a botched one previously caused MAESTRO-YA. **Clean:** 511 handled vs 512 invoked, the only gap being the already-known dead `tempfile:*` preload surface, and the only duplicate-channel hits being JSDoc examples and the dev auto-updater stubs.

## Summary by CodeRabbit

* **Bug Fixes**
  * GitHub polling and repository detection now handle missing or expired authentication gracefully without unnecessary error reports.
  * Expected “agent already running” errors no longer generate misleading telemetry while still providing clear delivery feedback.
  * Repeated usage-sampling failures are deduplicated, with periodic re-reporting; successful samples clear prior failure suppression.
  * Disk-full SQLite errors are filtered from diagnostics, while potentially serious database errors continue to be reported.
* **Reliability**
  * Unexpected remote process and polling failures continue to be captured for troubleshooting.


pedramamini:
### CI status

`test (ubuntu-latest)` (both shards) and `lint-and-format` pass. The two red Windows shards on the first run were **pre-existing on the base commit**, not introduced here - `2053e9706` (this PR's base, run [32618451915](https://github.com/RunMaestro/Maestro/actions/runs/32618451915)) fails the identical three cases:

| shard | test | cause |
| --- | --- | --- |
| 1/2 | `claude-usage-sampler > reads the identity from the canonical config dir, not the wire echo` | assertion compares against a bare `/Users/test/...` literal; the sampler passes the *resolved* key, which `path.resolve` drive-anchors to `D:\Users\test\...` on Windows |
| 2/2 | `filesystem > fs:compressFolder` (2 cases) | unrelated subsystem |

`rc` itself has been red on every recent push for the same reasons.

I fixed the shard-1 one in `4bacf64c2`, since it is in a file this branch already touches and the fix is the `canonKey()` helper from `helpers/pathExpect` that **every sibling assertion in that file already uses** - it was simply missed on this one line. No product code involved.

The `fs:compressFolder` failures are a separate pre-existing breakage in another subsystem and are deliberately **left alone** here; they will keep shard 2 red until fixed on their own. Flagging so it is not read as fallout from this branch.

pedramamini:
### Review feedback addressed (`5c30949fe`)

**@coderabbitai - failure key ignores `customEnvVars.CLAUDE_CONFIG_DIR`.** Correct, and it was the one real defect in the branch. `CLAUDE_CONFIG_DIR` reaches the child through `customEnvVars` as well as `opts.configDir` (there is an existing test, `lets customEnvVars.CLAUDE_CONFIG_DIR drive the key when configDir is omitted`, asserting the sampler honors it for the snapshot key). My memo keyed only off `opts.configDir`, so two accounts driven purely through `customEnvVars` collapsed onto the same `~/.claude` key and the first broken one would have muted the second entirely - the precise failure mode the change exists to avoid.

Now `resolveConfigDirKey(childEnv)` is resolved once, before the spawn, and threaded through `reportFailure` and `clearFailureHistory`.

Worth noting the second half of your comment: the breadcrumb's `configDir` was *already* reporting that same `opts.configDir ?? ~/.claude` guess before this branch, so a customEnvVars-driven account was being mis-named in Sentry independently of the memo. Threading the resolved key fixes both at once, and breadcrumb and snapshot key now agree by construction rather than by two copies of the same expression.

Added a regression case for the exact collision, confirmed red against the old key.

**@greptile-apps - POSIX-only fixtures in the new cases.** Taken, though via the file's own convention rather than rebuilding the fixtures: the assertions now route the expected values through `canonKey()` from `helpers/pathExpect`, which is what every other path assertion in this file already does. That compares against the genuinely resolved, drive-anchored key on Windows. The input literals stay readable, which matters more here than synthesizing `os.tmpdir()` paths that the mocked `execFile` never touches.

Related: `4bacf64c2` fixes the one assertion in this file that had been missing `canonKey()` all along, which is why `test (windows-latest, shard 1/2)` was red on the base commit. Shard 1 is green now.

Full suite re-run: **1,594 files, 38,329 tests, 0 failures.** `tsc --noEmit`: 0 errors.

pedramamini:
### Windows CI is now green from this branch's side

Both remaining Windows failures turned out to be the **same class of test-only bug**: an assertion comparing against a POSIX literal where the product builds the path with `path.join`, which emits `\` and drive-anchors on Windows. The product code is correct on both platforms in every case; only the fixtures were POSIX-shaped.

| commit | test | why it was red |
| --- | --- | --- |
| `4bacf64c2` | `claude-usage-sampler > reads the identity from the canonical config dir` | compared against a raw literal instead of `canonKey()`, which every sibling in the file already uses. Pre-existing. |
| `b154f5c27` | `claude-usage-sampler > uses the explicit configDir in the breadcrumb` | fallout from threading the resolved key into the breadcrumb in `5c30949fe`. Mine. |
| `60434e5b8` | `filesystem > fs:compressFolder` (2 cases) | hardcoded `/project/Competition.zip`; the handler joins with `path.join`. The suffix-increment case failed twice over - its `existsSync` mock also matched POSIX literals, so the "is this name taken?" probe never fired and the counter never advanced. Pre-existing. |

The `fs:compressFolder` one is unrelated to the Sentry work and I had intended to leave it, but it fails identically on the base commit and has been holding `test (windows-latest, shard 2/2)` red on `rc` itself, so the branch could not land on a green matrix without it. It is test-only and confined to that one `describe`; the SSH cases in it are deliberately left as POSIX literals, since remote paths are POSIX regardless of the desktop platform and that branch joins with `/` explicitly.

Worth noting for whoever owns `rc`: **these were only ever caught by the Windows leg**, because `path.resolve`/`path.join` are identity-ish for already-absolute POSIX paths on macOS and Linux. `src/__tests__/helpers/pathExpect.ts` exists for exactly this and is worth reaching for by default in any test that asserts on a path the product resolved.

Full local suite: **1,594 files, 38,329 tests, 0 failures.** `npm run lint`: clean. ESLint: clean.

pedramamini:
### Converged with #1421 on MAESTRO-KE (`058fcefb6`)

@RunMaestro/maintainers - heads up: **#1421 (main) and this PR both fixed MAESTRO-KE on the same day.** I found it after pushing, diffed theirs, and **replaced my `cue-github-poller.ts` and its tests with theirs verbatim** rather than shipping a second implementation.

Proof of convergence, per the rule that has paid off in past triages:

- Base blobs were identical on both branches (`2dc0b22af8` for the source, `83eda8436a` for the test), so their patch applied to pristine rc files.
- Resulting blobs hash to **`ef0d7ca53b`** and **`95a7914d69`** - byte-identical to main's post-fix versions. The rc→main merge cannot conflict on these files.
- Dry-run `git merge --no-commit --no-ff` of their branch into this one: **clean auto-merge**, no conflicts. Ran the union's tests (`cue` + `shared` + `stats` + `agents`): **239 files, 5,306 passed**. `tsc --noEmit`: 0 errors. Then aborted.

Adopting theirs also picks up two things my version missed:

1. **`ghErrorHaystack()` extracted** - the message+stderr lowercasing now lives in one place instead of being copy-pasted into a third classifier. I had duplicated it.
2. **Go transport vocabulary** in `isGitHubConnectivityError`: `tls handshake timeout`, `i/o timeout`, `no such host`. `gh` is a Go binary, so a transport failure never surfaces as a libuv errno - it arrives as `Post "https://api.github.com/graphql": net/http: TLS handshake timeout`. That whole family was reaching Sentry and neither of us would have caught it from the 401 events alone.

Their `isGitHubAuthError` additionally claims `requires authentication`. And for the reclassified "still reports" case we both hit, their fixture choice is better than mine: `HTTP 422: Validation Failed` (GitHub rejecting a query **we** built = a real bug that must keep paging) rather than my `panic: runtime error`. Their tests also assert the injected failure actually ran, so the zero-Sentry assertion can't pass vacuously - worth copying generally.

**No overlap on the rest of this branch:** their `sentryFilters.ts` hunk is the `shell:trashItem` rule in section 2, mine is the SQLITE_FULL rule in section 1; their `query-events-buffer.ts` work (MAESTRO-ZC) is in a file this branch never touches. The two PRs are complementary.

pedramamini:
### All five checks green

`test (ubuntu-latest)` x2, `test (windows-latest)` x2, and `lint-and-format` all pass. `fs:compressFolder` is fixed and shard 2 is green.

One flake worth flagging on the way past: shard 2 first came back red on `AutoRunBlurSaveTiming > very long content can be saved` with `Test timed out in 10000ms`. Not touched by this branch, and it passed on re-run. The test types 5,000 characters into a textarea and its own comment records that it was already cut down from 100,000 "without excessive slowdown" - so it is sitting close to the 10s budget and tips over on a loaded Windows runner. Worth either raising that one case's timeout or shrinking the input again; it will keep costing people re-runs.

pedramamini:
### Merged `rc` in and fixed the one pre-existing red

**Conflicts (`38ec9473e`)** - three test files, all additive or duplicated work:

| file | resolution |
| --- | --- |
| `filesystem.test.ts` | `rc` landed the same Windows `path.join` fixture fix independently. Took **rc's version verbatim** so the two cannot drift, same principle as `058fcefb6`. Verified byte-identical to `origin/rc`. |
| `useRemoteHandlers.test.ts` | `rc` added the `crossAgentMentions` mock + import, this branch added the sentry mock + `processErrors` import. Kept both. |
| `claude-usage-sampler.test.ts` | `rc` has the `canonKey()` assertion without the comment explaining why it is needed. Kept the comment. |

No product-code conflicts; every source file auto-merged clean.

### `FilePreview > bare font-zoom keys` (`a7deab295`)

Four cases were red on both `ubuntu` and `windows` shard 2/2. **Pre-existing** - the identical four fail on `rc` HEAD (`c2e7592de`, run [32876547379](https://github.com/RunMaestro/Maestro/actions/runs/32876547379)) with no PR involved.

Root cause is test-only. The zoom level is a persisted reading preference, so `useScalePreference` writes it to `localStorage`. Nothing reset that key between cases, so the `1.2` left by `zooms in on a bare +` became the starting point for the four cases after it, and a single `-` landed on `110%` where the assertion expects `90%`. Exactly the CI diff.

Fixed with the existing `installLocalStorageMock()` helper in that describe's `beforeEach` - a fresh in-memory Storage per case, which resets the key and pins the environment at the same time.

**Why the matrix was the only thing that caught it:** a local `vitest run` gets a Storage-less jsdom, so the writes silently no-op and nothing leaks. Confirmed by hoisting the mock install out of the `beforeEach` (working Storage, no reset), which reproduces all four failures locally, then restoring it. Same lesson as the `pathExpect` one further up this thread: a test that passes locally because a platform primitive is inert is not a test that passes.

### Verification

- `tsc --noEmit -p tsconfig.json`: 0 errors.
- ESLint + Prettier clean on every changed file.
- All six files this branch touches, plus `FilePreview.test.tsx`: green.
- Full matrix re-running now.

PR #1425, opened 2026-08-25
feat(cli): add goal-run --visible for desktop-owned Goal Auto Runs

### LINKED ISSUE #1286: Feature: CLI: add --visible flag for desktop Goal/Spec Auto Runs

pedramamini:
Closes #1286

Rebuild of #1287 against current `rc`. That PR was closed at 305 commits behind with two real review findings; the bridge it threaded through has since been reworked (#1352 decomposed `web-server-factory`, #1357 reworked CLI dispatch), so this is re-derived against the current layering rather than rebased, and both findings are fixed by design.

## What

```
maestro-cli goal-run --visible [--wait] [--exit-criteria "..."] [--max-iterations N] [--json] <agent-id> "<goal>"
```

`goal-run` has only ever run **headlessly**: the CLI spawns the agent itself, so the run is real but observable only through the CLI stream and history entries. It never appears in the Auto Run surface. An orchestrating agent therefore could not launch a free-text Goal worker the user could watch without handing control back to them.

`--visible` hands the run to the **running desktop app**. It routes through the existing CLI to desktop WebSocket bridge and terminates at the same `startBatchRun({ goalConfig })` entry point the Auto Run modal's **Go** button uses. There is no new run engine: the desktop owns spawning, busy arbitration, `stop-auto-run`, and `session list` visibility, so those requirements come for free.

## The two findings from #1287, and how they are fixed

Both were timing bugs, and both now have a dedicated home in `src/renderer/hooks/remote/goalRunLaunch.ts` so they are unit-testable without a renderer.

**1. Success was reported before the run started.** #1287 copied the `configure_auto_run` pattern of acking up front (that bridge does it because its run is long). But `startGoalRun` returns early *and silently* when the Auto Run kill switch is on, the session has vanished, or the `autorun-goal` prompt template fails to load, so the CLI was told "launched" for runs that never began.

Now every cheap failure is checked before the agent is claimed, and the reply then waits for the batch store to actually report the run running (`waitForGoalRunStart`), racing that against the run promise settling. Losing the race means "bailed during startup", not "finished", so the store is re-read rather than assumed - a very short run that completes before we look still reports success.

**2. Busy-checking was not atomic.** `startBatchRun` does not mark an agent running synchronously; the goal runner awaits the prompt template and a git status read before dispatching `START_BATCH`. Two `--visible` launches racing at one agent would both observe an idle batch store and both proceed.

`reserveGoalRunLaunch` closes that window with a synchronous claim taken before the first await. The claim is released once the launch is confirmed, at which point the batch store holds the agent instead.

## Behavior / contract

- **Backward compatible.** Headless `goal-run` stays the default and is untouched; `--visible` is purely additive.
- **Fails closed.** An unreachable desktop app exits `MAESTRO_NOT_RUNNING` rather than quietly substituting a headless run, which would be invisible in exactly the surface the caller was pointing at.
- **Stable failure codes:** `MAESTRO_NOT_RUNNING`, `AGENT_BUSY`, `AUTO_RUN_DISABLED`, `SESSION_NOT_FOUND`, `EMPTY_GOAL`, `LAUNCH_FAILED`, `LAUNCH_TIMEOUT`, `UNSUPPORTED_COMMAND` (older desktop build), `VISIBLE_LAUNCH_REJECTED`.
- **`--wait`** queues behind a busy agent instead of failing. It is rejected without `--visible` rather than silently ignored: a script passing `--wait` and expecting a queue would otherwise just fail on a busy agent and look flaky.
- **`--json`** emits one `visible_launch` event with a deep link:
  ```json
  { "type": "visible_launch", "ok": true, "mode": "goal", "visible": true,
    "agentId": "...", "sessionId": "...", "tabId": "...", "status": "running",
    "uri": "maestro://session/<id>/tab/<id>" }
  ```
- Timeouts nest deliberately: renderer 15s < main-process bridge 20s < CLI round-trip 30s, so the inner layer always answers before the outer one gives up.

## Deviations from the issue's sketch

- **camelCase, not snake_case.** `goal-run --json` is already a JSONL stream using `taskIndex` / `elapsedMs` / `agentSessionId`, so `agent_id` / `tab_id` would be the odd keys out in the same stream. Easy to change if a stable snake_case contract matters more than internal consistency.
- **No `--new-tab` / `--tab` / `--focus`.** Goal-Driven Auto Runs are modeled **per-agent** in the engine, not per-tab, so those do not map onto it. The run surfaces on the agent's active AI tab, which is what `tabId` returns and what the deep link addresses. Flagged for the author on the issue.
- **No `run_id`.** The provider session is assigned after the first agent turn, so it is not known synchronously at launch. Omitted rather than faked.

## Layers touched (all additive)

- **CLI**: `--visible` / `--wait` + `runVisibleGoalRun()` in `goal-run.ts`, registration in `index.ts`
- **Web server**: `launch_goal_run` message handler, `CallbackRegistry` entry, `WebServer` wiring
- **Main to renderer**: `remote:launchGoalRun` via the shared `createRemoteRequest` helper, preload subscription
- **Renderer**: `maestro:launchGoalRun` listener + the `goalRunLaunch` arbitration module

## Tests

- `goalRunLaunch.test.ts` (17) - reservation atomicity and start confirmation, written around the two failure modes above.
- `useAppRemoteEventListenersGoalRun.test.ts` (10) - the listener end to end: goalConfig routing, ack-only-after-running, every rejection code, and the two-launches-before-start race.
- `goalRunVisible.test.ts` (16) - the CLI contract: message shape, JSON identifiers, deep link, and every fail-closed path. The `process.exit` stub **throws** here rather than returning, so assertions like "never contacted the desktop" are actually observable instead of silently vacuous.
- Updated the `web-server-factory` and `useRemoteIntegration` mocks for the new bridge methods.

`npm run lint` (all three tsconfigs), `npx eslint` (no output), and the **full suite** (1600 files, 38,436 tests) pass locally.

> Local validation is single-OS; both CI legs are the source of truth before merge.

## Note on `docs/cli-reference.md`

Regenerated with `npm run gen:cli-reference`, so it also picks up one unrelated stale row (`create-agent` was missing `antigravity`).

## Summary by CodeRabbit

* **New Features**
  * Added `--visible` support to `maestro-cli goal-run` for launching runs in the desktop app.
  * Added optional `--wait` behavior for busy agents.
  * Added JSON and human-readable launch results, including session identifiers and deep links.
  * Added validation and clear error codes for invalid goals, unavailable agents, busy agents, timeouts, and rejected launches.
* **Documentation**
  * Updated CLI reference and usage guides with visible-run options, behavior, deep linking, and failure handling.
* **Tests**
  * Added comprehensive coverage for visible launches, validation, busy-agent handling, startup failures, retries, and timeouts.


pedramamini:
**CI status: the two Windows legs are red, but not because of this PR.**

Passing here: `lint-and-format`, both `ubuntu-latest` shards, Greptile.

The Windows shards fail on three tests, none of which this PR touches:

- `src/__tests__/main/ipc/handlers/filesystem.test.ts > fs:compressFolder > zips a folder into <name>.zip beside it, nesting under the folder name`
- `src/__tests__/main/ipc/handlers/filesystem.test.ts > fs:compressFolder > increments a numeric suffix until the archive name is free`
- `src/__tests__/main/agents/claude-usage-sampler.test.ts > account identity > reads the identity from the canonical config dir, not the wire echo`

These are pre-existing on `rc`. The baseline CI run on `rc` itself ([run 32792270173](https://github.com/RunMaestro/Maestro/actions/runs/32792270173), no PR involved) fails on the same three tests with the same shard split - 1 failure on shard 1, 2 on shard 2 - which is exactly what this branch reproduces. Both look like the usual Windows path/FS divergence (archive path construction, and the Claude config-dir resolution).

So this branch introduces no new failures, but per the repo rule that both matrix legs must be green before merge, **it is not mergeable until that `rc` breakage is fixed separately**. I'd rather flag that than have the red X read as a problem with the feature. Happy to pick up the Windows fixes as their own PR if useful - they're unrelated enough that bundling them here would just muddy this diff.

PR #1428, opened 2026-08-25
feat(agents): add Kilo (KiloCode) as a fork of the OpenCode integration

### LINKED ISSUE #989: Feature Request: Integrate KiloCode Agent into Maestro
## Summary

Add KiloCode (`kilo`) as a new agent in Maestro, leveraging the existing OpenCode integration as the foundation since KiloCode is a 1:1 fork of OpenCode with identical CLI interface and JSON output format.

## Background

**KiloCode** (https://github.com/Kilo-Org/kilocode) is a fork of OpenCode that replaces the OpenCode branding/binary with `kilo`/`kilocode`. The CLI interface, `run` subcommand, and JSON output format are identical to OpenCode, making integration straightforward.

**Maestro** already has full OpenCode support including:
- Agent definition (`definitions.ts:250-309`)
- Capabilities (`capabilities.ts:298-322`)
- Output parser (`opencode-output-parser.ts`)
- Session storage (`opencode-session-storage.ts`)
- Path probing (`path-prober.ts`)
- Context groomer, error patterns, UI metadata, etc.

## Key Differences Between KiloCode and OpenCode

| Aspect | OpenCode | KiloCode |
|--------|----------|----------|
| Binary name | `opencode` | `kilo` (alias: `kilocode`) |
| npm package | `opencode` | `@kilocode/cli` |
| Data dir | `~/.local/share/opencode/` | `~/.local/share/kilo/` |
| Database | `opencode.db` | `kilo.db` |
| Config file | `opencode.json` | `kilo.json` |
| Config env var | `OPENCODE_CONFIG_CONTENT` | `KILO_CONFIG_CONTENT` (likely) |
| CLI flags | `run --format json --session --model --agent` | Same + `--auto`, `--variant`, `--thinking` |
| JSON output | `{ type, timestamp, sessionID, part }` | Identical |
| Session storage | `~/.local/share/opencode/storage/` or SQLite | `~/.local/share/kilo/storage/` or SQLite (`kilo.db`) |

### 1. Agent ID Registration
**`src/shared/agentIds.ts`** — Add `'kilo'` to `AGENT_IDS`

### 2. Agent Definition
**`src/main/agents/definitions.ts`** — Add entry to `AGENT_DEFINITIONS`:

```typescript
{
  id: 'kilo',
  name: 'Kilo',
  binaryName: 'kilo',
  command: 'kilo',
  args: [],
  batchModePrefix: ['run'],
  jsonOutputArgs: ['--format', 'json'],
  resumeArgs: (sessionId: string) => ['--session', sessionId],
  readOnlyArgs: ['--agent', 'plan'],
  readOnlyCliEnforced: true,
  modelArgs: (modelId: string) => ['--model', modelId],
  imageArgs: (imagePath: string) => ['-f', imagePath],
  defaultEnvVars: {
    KILO_CONFIG_CONTENT:
      '{"permission":{"*":"allow","external_directory":"allow","question":"deny"},"tools":{"question":false}}',
  },
  readOnlyEnvOverrides: {
    KILO_CONFIG_CONTENT: '{"permission":{"question":"deny"},"tools":{"question":false}}',
  },
  configOptions: [
    {
      key: 'model',
      type: 'text',
      label: 'Model',
      description: 'Model to use (e.g., "ollama/qwen3:8b", "anthropic/claude-sonnet-4-20250514"). Leave empty for default.',
      default: '',
      argBuilder: (value: string) => {
        if (value && value.trim()) return ['--model', value.trim()];
        return [];
      },
    },
    {
      key: 'contextWindow',
      type: 'number',
      label: 'Context Window Size',
      description: 'Maximum context window size in tokens. Varies by model.',
      default: 128000,
    },
  ],
}
```

### 3. Agent Capabilities
**`src/main/agents/capabilities.ts`** — Mirror OpenCode's capabilities:

```typescript
kilo: {
  supportsResume: true,
  supportsReadOnlyMode: true,
  supportsJsonOutput: true,
  supportsSessionId: true,
  supportsImageInput: true,
  supportsImageInputOnResume: true,
  supportsSlashCommands: false,
  supportsSessionStorage: true,
  supportsCostTracking: true,
  supportsUsageStats: true,
  supportsBatchMode: true,
  requiresPromptToStart: true,
  supportsStreaming: true,
  supportsResultMessages: true,
  supportsModelSelection: true,
  supportsStreamJsonInput: false,
  supportsThinkingDisplay: true,
  supportsContextMerge: true,
  supportsContextExport: true,
  supportsWizard: true,
  supportsGroupChatModeration: true,
  usesJsonLineOutput: true,
  usesCombinedContextWindow: false,
}
```

### 4. Output Parser
**`src/main/parsers/opencode-output-parser.ts`** — The existing parser can be reused as-is since the JSON format is identical. Create a thin wrapper or parameterize the `agentId`:

```typescript
// Option A: Create KiloOutputParser extending OpenCodeOutputParser
export class KiloOutputParser extends OpenCodeOutputParser {
  readonly agentId: ToolType = 'kilo';
}

// Option B: Reuse directly by registering the parser with agentId 'kilo'
```

### 5. Session Storage
**`src/main/storage/`** — Create `kilo-session-storage.ts` mirroring `opencode-session-storage.ts` with path changes:
- Data dir: `~/.local/share/kilo/` (Linux/macOS) or `%APPDATA%\kilo` (Windows)
- Database: `kilo.db`
- Storage dir: `~/.local/share/kilo/storage/`

### 6. Path Probing
**`src/main/agents/path-prober.ts`** — Add kilo binary paths:

**Windows:**
```typescript
kilo: [
  ...npmGlobal('kilo'),
  path.join(home, 'scoop', 'shims', 'kilo.exe'),
  path.join(home, 'scoop', 'apps', 'kilo', 'current', 'kilo.exe'),
  path.join(home, '.volta', 'bin', 'kilo'),
  path.join(home, '.volta', 'bin', 'kilo.cmd'),
  ...goBin('kilo'),
]
```

**Unix:**
```typescript
kilo: [
  ...localBin('kilo'),
  ...homebrew('kilo'),
  ...npmGlobal('kilo'),
  ...nodeVersionManagers('kilo'),
]
```

### 7. Display Metadata
**`src/shared/agentMetadata.ts`:**
- Add `kilo: 'Kilo'` to `AGENT_DISPLAY_NAMES`
- Add `'kilo'` to `PLAN_MODE_AGENTS` (uses `--agent plan` like OpenCode)
- Add `'kilo'` to `BETA_AGENTS`

**`src/shared/agentConstants.ts`:**
- Add `kilo: 128000` to `DEFAULT_CONTEXT_WINDOWS`

**`src/renderer/constants/agentIcons.ts`:**
- Add `kilo: '⚡'` (or appropriate icon)

### 8. Context Groomer
**`src/renderer/services/contextGroomer.ts`:**
- Add `kilo` entry to `AGENT_ARTIFACTS` (same brand references as OpenCode: 'Kilo', 'kilo')
- Add `kilo` entry to `AGENT_TARGET_NOTES`

### 9. Error Patterns
**`src/main/parsers/error-patterns.ts`:**
- Add `kilo` error patterns (mirroring OpenCode patterns, replacing `opencode` with `kilo`)

### 10. CLI Agent Spawner
**`src/cli/services/agent-spawner.ts`:**
- Add `detectKilo` and `getKiloCommand` wrappers
- Add `'kilo'` case to the switch statement

### 11. SSH Command Builder
**`src/main/utils/ssh-command-builder.ts`:**
- Add `~/.kilo/bin` to SSH path probing

### 12. Renderer Services
**`src/renderer/services/inlineWizardDocumentGeneration.ts`**, **`inlineWizardConversation.ts`**, **`contextGroomer.ts`** — Add `'kilo'` cases mirroring `'opencode'`.

### 13. Tests
Update all test files that enumerate agents to include `'kilo'`.

## Implementation Strategy

Given the 1:1 parity with OpenCode, the recommended approach is:

1. **Minimal diff**: Reuse `OpenCodeOutputParser` by subclassing with `agentId = 'kilo'`
2. **Session storage**: Create `KiloSessionStorage` extending `OpenCodeSessionStorage` with overridden paths
3. **Copy-paste-adapt**: For definitions, capabilities, path probing, and metadata — copy the OpenCode entries and adjust identifiers

## Verification

1. Install KiloCode CLI: `npm install -g @kilocode/cli`
2. Verify `kilo run --format json "hello"` produces the same JSONL format as `opencode run --format json "hello"`
3. Verify session storage at `~/.local/share/kilo/`
4. Run Maestro with the new agent and confirm:
   - Binary detection works
   - Batch mode execution completes
   - Session history is readable
   - Read-only (plan) mode functions
   - Model selection works

---

This feature request covers ~15 files with changes that are largely mechanical (copy OpenCode entries, adjust identifiers). The integration risk is low due to the verified CLI/output format parity between the two agents.

pedramamini:
Closes #989

Adds KiloCode (`kilo`) as a supported agent. Kilo is a fork of OpenCode: same `run --format json` CLI, same JSONL event stream, same session storage layout. Only the binary name, the data directory, the database filename, and the config env var are rebranded.

This is a re-port of @raccoonwannafly's work in #992 onto current `rc`, carried forward under their authorship (`Co-authored-by` on the commit) per the offer in [this comment](https://github.com/RunMaestro/Maestro/issues/989#issuecomment-5079905435). #992 had gone unmergeable against `rc` and the agent registry has moved a long way since it was pushed, so rebasing was going to be more churn than re-applying the design onto today's code.

## Approach

Built as a thin layer over OpenCode rather than a parallel copy, following the structure @raccoonwannafly proposed:

- **Parser** — `KiloOutputParser` subclasses `OpenCodeOutputParser`, overriding only `agentId` (which is what routes error matching and stamps emitted errors).
- **Session storage** — `KiloSessionStorage` subclasses `OpenCodeSessionStorage` and supplies an `OpenCodeStorageBrand` of `{ dirName: 'kilo', dbFileName: 'kilo.db' }`. To make that possible, OpenCode's storage paths moved off module-level constants onto brand-parameterized helpers. Resolution is memoized per brand so the filesystem probing still happens once, as it did before, rather than once per `getMessageDir()` call in a listing loop.
- **Error patterns** — shared outright. The OpenCode messages are brand-neutral, and sharing the set is what stops the two from drifting apart.
- **Capabilities** — mirror OpenCode's entry, asserted by a test rather than by hand-copied booleans staying in sync.

## Also wired

Agent id and definition; path probing (Windows + Unix) and both PATH-expansion lists; SSH remote PATH; `kilo models` discovery; the provider pickers via `AGENT_PICKER_META`; wizard logo and tile; plan-mode wording; beta badge; default context window; built-in slash commands; context groomer artifacts and target notes; the wizard / inline-wizard / feedback arg builders and JSONL extractors; `StdoutHandler`'s step-reset and reasoning-tag branches; CLI spawner helpers and `--agent` help text; and both token-coverage maps.

That last one matters and wasn't in #992: Kilo reads the same storage schema, so its coverage is `full`. Without those two entries the Usage Dashboard and the Cue accessor would have reported a working agent as `unsupported`.

The `--agent <name>` persona field in `AgentConfigPanel` now covers OpenCode and its forks, with the brand read from the agent instead of hard-coded.

## Review items from #992, resolved

1. **`PULL_KILO.md` at repo root** — not carried over.
2. **`StderrHandler` refactor dropped the `toolType !== 'codex'` guard**, which had suppressed init noise for `opencode`, `copilot-cli`, and `factory-droid`. Not carried over either, because it turned out to be unnecessary: Kilo has a registered output parser, so the *existing* JSONL branch (`outputParser && toolType !== 'codex'`) already suppresses its "Unable to check client-server compatibility" warning. The only change here is a comment naming Kilo. This keeps the three agents' suppression intact.
3. **`ssh-command-builder.ts` listed `$HOME/.claude/local` twice** — added `$HOME/.kilo/bin` only, no duplicate.

I also left out #992's `phaseGenerator.ts` refactor: it changed Claude Code's document-generation args (adding `--output-format stream-json`), and its `getGenerationStdinFlags` returned `sendPromptViaStdin: false` on every path, making the branch it guarded dead. That's orthogonal to Kilo and better handled separately if it's wanted.

## Known gaps

Documented in `AGENT_SUPPORT.md`. All three are OpenCode-only enhancements layered on top of the base integration, not part of it:

- Model discovery reads `kilo models` but does not additionally parse `kilo.json`, so a provider defined purely in config must be typed into the Model field by hand.
- No custom slash-command discovery from `.kilo/commands/` (the built-in command set is wired up).
- No Coworking MCP installer and no `MCP_CONFIG_BY_AGENT` entry (consistent with `grok`, `omp`, and `antigravity`, which also have neither).

## Still unverified against a live binary

Same two unknowns as before, both one-line fixes if wrong: the `KILO_CONFIG_CONTENT` env var name and the `kilo.db` filename. Both are taken from the fork's rebranding pattern, and the issue flagged the env var as "(likely)".

## Validation

- `npm run lint` — clean (only the 4 pre-existing TS2882 CSS errors).
- `npx eslint src/` — clean, exit 0.
- Full suite: **38,468 passed**, 108 skipped, 0 failures.
- New `src/__tests__/main/agents/kilo-agent.test.ts` (17 tests) pins the OpenCode parity that makes the subclassing safe: capability equality, CLI-surface equality, parser output equality on the same JSONL line, shared error-pattern identity, and that Kilo's storage paths point at `kilo` while OpenCode's are undisturbed. It also asserts Kilo carries `KILO_CONFIG_CONTENT` and *not* `OPENCODE_CONFIG_CONTENT` — a stray one would be ignored by the binary, re-enabling the question tool and hanging batch mode on stdin.

Local validation is single-OS; needs both CI matrix legs before merge.

## Summary by CodeRabbit

* **New Features**
  * Added beta support for the Kilo coding agent.
  * Kilo can be detected, configured, launched, and selected in the setup wizard.
  * Added Kilo branding, model discovery, authentication, session resume, streaming output, and read-only mode support.
  * Added token usage tracking, conversation tools, and feedback handling for Kilo sessions.
* **Documentation**
  * Documented Kilo’s CLI behavior, storage locations, configuration, and known integration gaps.


pedramamini:
Addressed the review findings in 166dba2. Two of the six were real and are fixed; the rest I'm leaving, with reasons.

**Fixed — feedback responses were unreachable for OpenCode-shaped JSONL** (`feedbackConversation.ts`). This one was a good catch and worse than flagged: it was already broken for `opencode` before Kilo existed. OpenCode emits `{ type: 'text', part: { text } }`, so the response JSON sits inside a JSON string with its quotes escaped. `extractBalancedResponse` scans from each `{`, finds the envelope object first (valid JSON, not a valid feedback response), then hits the payload's braces *inside* the string where the `\"` sequences mean the slice never parses. Strategy 4 only matches Claude's `{"type":"assistant","content":...}`. So a successful run fell through to `DEFAULT_FEEDBACK_RESPONSE`.

Added a strategy that collects `part.text` across text events and re-parses the concatenation, guarded against self-recursion. It runs only after every existing strategy returns null, so nothing that works today changes. `extractJsonFromOutput` is now exported and there's a new test file covering it; I verified the tests are load-bearing by deleting the strategy and watching 3 of 5 fail.

**Fixed — failed `kilo models` discovery was cached for the full TTL** (`detector.ts`). Correct. There's already an `omp` exemption for exactly this, and Kilo has the same property, so it joins that guard rather than getting a parallel mechanism.

**Not changing — the "unverified" wording in `AGENT_SUPPORT.md`.** Three comments asked me to reconcile the Kilo table with the unverified note, or to drop the qualifier. The qualifier is the accurate part: `KILO_CONFIG_CONTENT` and `kilo.db` really are inferred from the fork's rebranding pattern and really haven't been checked against a shipping binary. The table documents what the code does; the note documents what we know. I'd rather ship that honestly than round it up to "confirmed" and have someone trust it. It comes out when @raccoonwannafly or anyone else with a real install confirms it.

**Not changing — the stale OpenCode "Stub Ready" status block.** Fair observation, and it is stale: OpenCode's storage is fully implemented now, so the section header and the "Implementation Status" list are both out of date. But that's a pre-existing doc bug about OpenCode, not something this PR introduces, and fixing it here would mix an unrelated correction into an agent-addition diff. Worth a follow-up.

Full suite: 38,473 passing, 0 failures. `npx eslint src/` clean.

pedramamini:
**CI note: the shard 2 failure is pre-existing on `rc` and unrelated to this PR.**

`Test (ubuntu-latest, shard 2/2)` is red on two assertions in `FilePreview.test.tsx` (lines 2002 and 2012, both `expect(percentButton()).toBeNull()` receiving `120%`). Nothing here touches `FilePreview`, `useFontScale`, or anything they import.

Evidence it isn't mine:

- **`rc` head `cb74f19e3` fails the identical two assertions on its own CI run** ([32799025117](https://github.com/RunMaestro/Maestro/actions/runs/32799025117)), on ubuntu shard 2 *and* windows shard 2, with no Kilo code in the tree.
- This PR's first push (`31b48d3c5`, before I merged `rc` in) was **green on all four legs**. `rc` head was `12f93f68b` at that point, which was also green.
- The breakage arrives with `cb74f19e3` ("Merge origin/main into rc"), which brought the collapsible zoom pill (`691c59ec3`) and the tests around it.

It's environment-dependent, which is why it slipped through: locally the file passes in isolation (105/105) *and* in a full `--shard=2/2` run against this exact merge (19,070 passed). Node reports `localStorage is not available` on my machine, so `useFontScale` can't persist and every case starts at 100%. Where localStorage does work, the `120%` that the first zoom test leaves behind survives into the later cases that assert 100%. So it looks like missing per-test reset of the persisted font scale rather than anything wrong with the zoom feature itself.

I've deliberately not fixed it here - it's `rc` breakage affecting every PR opened against `rc` right now, not just this one, and folding an unrelated test-isolation fix into an agent-addition diff is how these things get hard to review. It does need its own PR before this can merge.

Branch is otherwise current with `rc` (merged in `2aa2de5e`, no conflicts), `lint-and-format` passes, ubuntu shard 1 passes, and the full suite is green locally.

PR #1430, opened 2026-08-25
fix(cross-agent): consult from CLI dispatch, honor customPath, enforce read-only for codex and grok

ronaldeddings:
## What this does

Three related fixes to cross-agent `@mention` consults, found while driving them from `maestro-cli dispatch`.

### 1. CLI and web dispatch never planned `@mentions`

`maestro-cli dispatch "@Agent ..."` handed the text straight to a spawn, so a mention that consults another agent from the composer was silently inert from the CLI and the web bridge. Both remote paths now ride the same `planCrossAgentMentions` the composer uses:

- `useRemoteHandlers` (immediate path): a leading mention consults the target and skips the local spawn; a trailing mention spawns locally, then consults.
- `useRemoteIntegration` (`dispatch --queue`): stamps `crossAgentMention` / `crossAgentOnly` on the `QueuedItem`, so `processQueuedItem` fires the consult when the item becomes the agent's turn, same as a composer-queued message.

### 2. Consults ignored the agent's `customPath`

`cross-agent-router.ts` resolved the target binary from detection only (`agent.path || agent.command`). The tab spawn and the Group Chat moderator both honor `customPath`, so a consult could run a different binary than the tab it was consulting.

On macOS, detection probes known install dirs before PATH, and Codex's list ends with nvm version dirs. A stale nvm codex stub (0.92.0, vendored native binary missing) won over the codex the agent was pointed at. The stub exits 1 in ~70ms with nothing on stdout, which surfaced as `<agent> produced no visible output (exit code 1)` on every mention of that agent, while the tab itself worked fine.

Fix: thread `customPath` into `CrossAgentTargetSession` and pick `target.customPath || agent.path || agent.command`, the same rule as `group-chat-router.ts`.

### 3. Read-only consults were not read-only for Codex, and died for Grok

- **Codex:** `readOnlyArgs` carried `--dangerously-bypass-approvals-and-sandbox` next to `--sandbox read-only`. The bypass flag nullifies the sandbox, so read-only consults were writing files. Verified on codex 0.149.0: with the bypass flag a `touch` succeeded; with `--sandbox read-only` alone the OS sandbox denied it, the model got the error, and the run continued. `exec` has no approval prompt to bypass, so the flag bought nothing.
- **Grok:** `--permission-mode plan` alone is not enough headless. Grok answers every permission prompt with `cancelled` (no TTY), and a cancelled prompt kills the whole turn, so a consult whose shell command was classified non-read-only died mid-answer with its preamble delivered as the "result". `--deny Bash(*)` resolves by policy without prompting, so the turn continues on `read_file` / `grep` / `list_dir`.
- **Grok parser:** an `end` event whose `stopReason` is not `end_turn` / `EndTurn` is a turn that died early. It is now reclassified as an error (keeping `sessionId` for resume) and surfaced through `detectErrorFromParsed`, so unfinished streamed text is never recorded as a successful result.

The `crossAgentMentionsWritable` setting is untouched: it still flips `readOnlyMode` on the spawn, and with it on, `readOnlyArgs` are skipped entirely. Both positions were exercised live (details below).

## How to test

1. Give an agent a `customPath` that differs from what detection would find (or just have a stale `codex` in an nvm `bin` dir). Before: `@mention` of that agent fails with "produced no visible output (exit code 1)". After: the consult runs the configured binary.
2. `maestro-cli dispatch <agent-id> "@OtherAgent say hello" --new-tab`. Before: the source agent answers itself and `OtherAgent` is never spawned. After: `OtherAgent` is consulted and its reply lands in the source tab. Repeat with `--queue` on a busy agent; the consult fires when the queued item is dequeued.
3. With Consult Permission at Read-only, `@mention` a Codex agent and ask it to write a file. Before: it writes. After: the sandbox denies it and the agent says so. Flip to Read/Write and repeat: it writes.
4. With Consult Permission at Read-only, `@mention` a Grok agent with a task that makes it run a shell command. Before: the reply is a truncated preamble marked as success. After: the command is denied, the turn completes, and a cancelled turn (if one still happens) renders as an error.

Exercised live on macOS with Codex 0.149.0, Grok 1.0.5, and Antigravity: image-generation consults from a Claude agent to Codex and Grok, read-only and read/write, each returning the image to the source tab.

## Checklist

- [x] `npm run lint` and `npm run lint:eslint` pass
- [x] Tests for every touched module pass (244 across the six suites, plus the full suite via the pre-push gate: 38618 passed)
- [x] Manually tested in the running app
- [x] No UI components touched, so no theme or console impact
- [x] `CLAUDE.md` updated (the `@mention` bullet documents both the CLI planner and the `customPath` rule)
- [x] Conventional commit messages

Note: the pre-push gate reports 4 failures in `src/__tests__/renderer/components/FilePreview.test.tsx` ("bare font-zoom keys"). Those reproduce identically on the untouched `rc` tip (`f957b592b`) and upstream CI is red on that commit's shard 2/2 for the same reason, so they are pre-existing and not from this branch.

## Summary by CodeRabbit

* **New Features**
  * Cross-agent `@mentions` now work consistently for remote commands and queued prompts.
  * Cross-agent consultations can use a configured executable path when available.
* **Bug Fixes**
  * Read-only Codex sessions now enforce sandbox restrictions correctly.
  * Grok read-only sessions are less likely to end prematurely on permission prompts.
  * Grok cancelled or incomplete turns are now reported as recoverable errors instead of successful completions.
* **Documentation**
  * Updated guidance for cross-agent routing and dispatch behavior.


pedramamini:
@ronaldeddings Thanks for this one, and thank you in general for the contribution. The write-up is excellent: three genuinely distinct bugs, each with a live-verified reproduction, and the code comments explain *why* rather than *what*. I traced each claim through the codebase rather than taking the description at face value, and the first two are solid:

- **Codex bypass flag.** Confirmed against `buildAgentArgs` (`src/main/utils/agent-args.ts:204-238`): in read-only mode `batchModeArgs` is skipped and `readOnlyArgs` is appended, so before this PR a read-only Codex consult really was spawning `--sandbox read-only --dangerously-bypass-approvals-and-sandbox`, and the bypass won. Real fix, correctly scoped.
- **`customPath`.** Matches the precedence in `group-chat-router.ts`, and threading it through `CrossAgentTargetSession` is the right seam. Also checked that `--deny Bash(*)` survives the SSH path: `ssh-command-builder.ts:444` runs every arg through `shellEscape`, so the parens and glob are safe on a remote spawn.

One thing I would like addressed before this goes in, plus a few smaller notes.

---

### 1. The Grok session ID does not actually survive the reclassified `end` event

`src/main/parsers/grok-output-parser.ts` says "sessionId stays attached so resume capture still works", and the new test asserts `event?.sessionId` at the parser level. That holds inside the parser, but it does not hold end to end.

In `StdoutHandler.processLine` (`src/main/process-manager/handlers/StdoutHandler.ts:461-497`), error detection runs **before** event handling and `return`s early when `detectErrorFromParsed` yields an error:

```ts
if (agentError) {
    managedProcess.errorEmitted = true;
    agentError.sessionId = sessionId;   // overwritten with the MAESTRO session id
    ...
    this.emitter.emit('agent-error', sessionId, agentError);
    return;                             // handleParsedEvent never runs
}
```

`emitSessionIdIfNeeded` lives inside `handleParsedEvent` (line 580-581), so it is skipped. And `toAgentError` in the Grok parser never sets `sessionId`, so nothing carries the provider UUID out either.

Failure scenario: Grok's own schema note in that file says the session ID "arrives only on the final `end` event". A cancelled turn is now the error path, so that `end` line is consumed and dropped. The tab ends up with no `agentSessionId`, and the user's next message spawns without `--resume`, silently starting a fresh conversation with none of the prior context. That is exactly the hazard the codebase already documents at `ExitHandler.ts:150-168`:

> Capture the provider's session id BEFORE dispatching, and for a failed envelope as much as a successful one. [...] Without it the tab has no id to resume from, so recovery from a *recoverable* error silently opens a fresh conversation.

The `ExitHandler` flush does the right thing, so this only bites when the `end` line arrives newline-terminated mid-stream, which is the normal JSONL case. And since the new error is marked `recoverable: true`, a retry is precisely what we would expect to happen next.

The narrow fix is in `StdoutHandler.processLine`: extract and emit the provider session ID before the error `return`, mirroring what `ExitHandler` already does. That is shared code, so it would also close the same gap the Antigravity parser documents at `antigravity-output-parser.ts:339-344`. If you would rather keep this PR's blast radius small, a Grok-local workaround is fine too, but a test at the `StdoutHandler` level (not just the parser) is what would keep this honest.

### 2. Leading-mention consult can jump the execution queue

The composer's equivalent path is careful about this (`useInputProcessing.ts:833-890`):

> "this agent doesn't answer it" is NOT the same as "it has nothing to wait for". When the user has already put work in front of it, the POSITION of the message is the instruction.

It probes with `probeSessionAiProcesses` and `hasWorkAheadOfNewMessage` before consulting immediately. The new remote path in `useRemoteHandlers` inherits only the `session.state === 'busy'` guard above it, which does not see a non-empty `executionQueue` (and `dispatch --force` bypasses it outright). So `maestro-cli dispatch <agent> "@Other ..."` against an idle agent with items still queued will fire the consult ahead of that queued work, which is the behavior the composer path deliberately prevents. Reusing `hasWorkAheadOfNewMessage` here and falling through to the queue would restore parity.

### 3. Trailing-mention dispatch ordering differs from the composer

The composer fires consults *before* the local spawn ("just before the source agent's own turn, matching the order the user sees", `useInputProcessing.ts:1116-1125`). The new remote path fires after the spawn resolves, inside the `try`. Defensible, but it means a spawn that throws silently drops the consult, where the composer would already have sent it. Worth a sentence of comment either way so the divergence is intentional rather than incidental.

### 4. Nit: `--deny Bash(*)` is wider than consults

`readOnlyArgs` is not consult-only: it also feeds tab naming and any Grok tab the user puts in read-only mode. Your comment is upfront that read-only shell commands are lost as the price of a turn that survives, and I think that is the right trade, but it is a behavior change for read-only Grok tabs generally, not just for `@mentions`. Flagging it so it is a conscious choice rather than a side effect.

### 5. Nit: Greptile's `updateSessionWith` note

Technically correct per `CLAUDE.md`, and `updateSessionWith` is a drop-in here (it calls the same `useSessionStore.getState().setSessions`). Low priority though: the file already hand-rolls this in four places, and you are mirroring the composer's own leading-mention branch, which does the same. Your call.

---

### CI

Both red shards are unrelated to this branch. The four `FilePreview > bare font-zoom keys` failures reproduce on `rc` at your base commit `f957b59` (run 32806727851) on both OS legs, so they are pre-existing as you noted. The extra Windows failure, `AutoRunBlurSaveTiming > very long content can be saved`, is a 10s test timeout not present in the baseline and does not touch anything here. A re-run should clear it.

No merge conflicts, so nothing to rebase. Once item 1 is sorted (and ideally 2), I am happy to take this.

PR #1432, opened 2026-08-25
feat(tasks): dock the agent's task list in a collapsible bar above the composer

### LINKED ISSUE #1431: Feature: Show TodoWrite todo list in collapsible bar above chat input
## Summary
Show TodoWrite todo list in collapsible bar above chat input

## Environment
- Maestro version: 0.18.5-RC

## Desired Outcome
Maestro surfaces the todo list from TodoWrite tool calls as a dedicated UI element in the session chat view: a compact bar docked directly above the input/chat box. Collapsed by default to a single-line progress summary (e.g. "3/7 done" plus the current in-progress item), expandable to reveal the full list with per-item status. The bar updates in place as the agent rewrites the todo list.

## Details
TodoWrite calls only appear as raw tool activity inside the conversation stream. The todo list scrolls out of view as the conversation grows, so there is no persistent view of what the agent is currently working on or how much is left.

## Additional Context
Placement preference confirmed by the user: a bar above the input box within the session chat window (not a sidebar, not the session list). User agreed the collapsible bar approach makes sense. Undecided: whether the bar should auto-hide once all items are completed or remain visible showing the finished list.

## Screenshots / Recordings
None

pedramamini:
Closes #1431

Targets `rc`.

## What

Agent checklists (Claude Code / OpenCode `TodoWrite`, Codex `update_plan`) already render as an inline `AgentTaskListCard` in the transcript, but those cards scroll away as the conversation grows. "What is the agent working on, and how much is left" stops being answerable without scrolling back to hunt for the last update.

This pins the **newest** checklist for the active AI tab directly above the input box:

- **Collapsed** to one line: the active task (using the agent's present-tense `activeForm` when it supplies one) plus a `completed/total` count and a progress bar.
- **Expands on click** to the full list with per-task state, capped in height and scrollable so a 40-item plan can't eat the window.
- **Rewrites in place** as the agent updates its plan.
- **Dismissable** per checklist.

## Two Display settings, both OFF by default

| Setting | Default | Effect |
| --- | --- | --- |
| `showAgentTaskListBar` | **off** | Renders the bar at all. The inline transcript cards are unaffected either way. |
| `autoExpandAgentTaskListBar` | **off** | Opens the bar to the full checklist every time the agent writes a new one, instead of the one-line summary. Disabled in the UI while the bar itself is off. |

Both live under **Settings -> Display -> Agent Task List**, in a new `AgentTaskListSection` alongside `rc`'s other extracted Display sections.

Expansion state has two modes on purpose, because the two settings want different things:

- Auto-expand **off**: expanded/collapsed is a sticky preference (`usePersistedToggle`) that survives restart.
- Auto-expand **on**: each new checklist opens; a manual collapse is keyed to that checklist's log entry, so the next update re-expands rather than the bar staying shut forever. No effect, no second copy of the state to re-sync.

## How

The list is **derived from the tab's own logs** - a reverse scan (`findLatestAgentTaskList`) for the last checklist-shaped tool call. Agents rewrite the whole list on every update, so the last one *is* the current state; there is nothing to merge across entries.

That means no new capture path, no store slice, and no second copy of the state to drift from the transcript. It switches with the tab and survives app restart for free.

Detection reuses the existing shape-based `extractAgentTaskList`, so this stays agent-agnostic exactly like the inline card - any provider emitting a `{ content, status }`-ish array gets it.

## Notes on specific decisions

- **Subagent guard (`rc`-only):** a checklist carrying `metadata.parentToolUseId` is skipped. A delegated worker keeps its own private plan and writes it *last*, so without the guard a `Task` tool call would replace the plan the user is actually following with a scratch list they never asked to see. This is the follow-up the `main`-targeted version of this PR flagged; it is live here because `rc` is where subagent nesting exists.
- **Dedup:** the task rows moved into a shared `AgentTaskItems`, used by both the inline card and the new bar, so the two surfaces can't disagree about what a completed task looks like.
- **Dismissal is keyed off the source log entry id**, not a plain boolean. Hiding the bar hides *that* list; the next checklist the agent writes brings it back on its own.
- **Perf:** `useMemo` on `logs` plus `React.memo` on the component. The composer re-renders on every keystroke, but `logs` only changes when the agent writes, so the reverse scan stays off the typing path.
- **Completed lists stay visible** (the icon turns green at 100%) rather than auto-hiding, on the reasoning that completion is exactly the moment you want to confirm what was done. The explicit dismiss covers the case where it has outlived its usefulness.

## Validation

- `tsc --noEmit` across all four tsconfigs: clean
- `eslint` + `prettier --check` on every touched file: clean
- Full suite via the pre-push hook: **1614 files, 38645 tests passing**, no regressions
- Tests updated for this retarget: 6 new cases covering the subagent guard (2 unit, 1 component) and auto-expand (opens a new list, re-expands after a manual collapse, ignores the sticky preference)

CI still needs to be green before merge - local validation is single-OS.

## Summary by CodeRabbit

* **New Features**
  * Added a docked agent task list above the composer in AI mode.
  * Added checklist progress, task status indicators, collapse/expand controls, dismissal, and restoration when a newer checklist appears.
  * Added Display settings to show the task list and automatically expand it for new checklists.
  * Added shared checklist rendering across task list views.

* **Bug Fixes**
  * Improved checklist selection to use the latest applicable checklist while excluding private subagent tasks.

* **Tests**
  * Added coverage for rendering states, filtering, persistence, dismissal, restoration, and automatic expansion.


pedramamini:
### CI note: the `test` failure is pre-existing on `main`, not from this PR

The `test` job is red on 4 cases in `src/__tests__/renderer/components/FilePreview.test.tsx` > `bare font-zoom keys`:

```
AssertionError: expected <button type="button" ...(4)></button> to be null
```

That same file fails identically on `main` at **8af06a513**, which is this branch's exact base ([run 32806138304](https://github.com/RunMaestro/Maestro/actions/runs/32806138304)) - and on every `main` run going back to at least 2026-08-24.

Counts line up with the diff being additive-only:

| | Test Files | Tests passed |
|---|---|---|
| `main` @ 8af06a513 | 1 failed / 1258 passed | 1258 |
| this PR | 1 failed / 1259 passed | 1259 |

Same one failing file, exactly +1 passing file (the new `AgentTaskListBar.test.tsx`). Nothing here touches `FilePreview` or the font-zoom path.

Other checks are green: `lint-and-format`, `Analyze (javascript-typescript)`, CodeRabbit, Greptile.

PR #1433, opened 2026-08-26
feat(thought-stream): interleave tool calls with reasoning on one timeline

### LINKED ISSUE #1312: Feature: Show live agent step/tool-call feed during runs (like Claude Code)

pedramamini:
Closes #1312

## Problem

The Thought Stream captured only `process:thinking-chunk`, so it showed an agent's **reasoning** but never its **actions**. Tool calls do render in the chat transcript, but only when a tab's `showThinking` is on - and an Auto Run spawns as `{sessionId}-batch-{ts}` with no AI tab at all, so `useAgentToolExecutionListener` (which writes into `aiTabs[tabId].logs`) has nowhere to put them.

Net effect for the reporter: **no surface anywhere** showed what an Auto Run agent was actually doing while it burned tokens.

## What this does

Tool calls stream into the Thought Stream interleaved with the reasoning, each reduced to ONE short plain-language line:

```
3:42:07 PM  ⟳ Ran npm test
3:42:04 PM  ✓ Read src/renderer/components/ThoughtStreamPanel.tsx
3:42:01 PM  ✓ Searched for THOUGHT_BLOCK_GAP_MS
3:41:58 PM  ! Edited src/renderer/constants/themes.ts
```

A spinner marks a call in flight; a check or a warning marks how it ended. That density is the reporter's stated need: **cost control**, spotting a loop and interrupting it before it burns more tokens.

## This is a fresh pass on #1314, not a rebase

#1314 was closed because its chronology was structurally wrong: `groupThoughtsIntoBlocks` ignored intervening tool calls, and the feed merged two independently-sorted sequences at display time, assigning each complete thought block its **start** time. A tool call that happened mid-block could never render between the reasoning chunks surrounding it, so a fast call displayed *after* the reasoning that actually followed it.

The closing note asked for a redesign: *"interleave on a single timeline of timestamped events rather than merging two independently-sorted sequences after the fact."* That is what this does.

A session's buffer is now **one chronological array** holding both event kinds in arrival order (`StreamEvent = ThoughtEntry | ToolActivityEntry`, discriminated by the presence of `tool`). `buildActivityFeed` walks it **once**: consecutive thinking coalesces into a block, and a tool call **closes** the open block so reasoning that arrived after an action starts a new block below it. Ordering is therefore structural - there is no sort that could put a call in the wrong place. `groupThoughtsIntoBlocks` becomes a projection of that same walk, so the two views cannot disagree about where a block starts.

`src/__tests__/.../thoughtStreamStore.test.ts` pins the exact old failure: a tool call 1ms into a block must still render before the reasoning that followed it.

## Changes

**`src/renderer/utils/toolActivityLabel.ts`** (new)
`describeToolActivity(toolName, input)` normalizes tool names across Claude Code (`Read`/`Bash`/`MultiEdit`), OpenCode (lowercase), Codex (`shell`/`apply_patch`/`update_plan`), Copilot (`write_to_file`), and MCP (`mcp__server__tool`) onto plain English. Unknown tools degrade to `Used <name>` rather than vanishing, so a new provider tool still shows up. Handles the raw-string input shape (Codex `apply_patch`) and argv-array commands. Deliberately **not** `summarizeToolInput()`, which builds the verbose in-chat cell; both are now documented in `SHARED-UTILS.md` with a note on which to reach for.

**`src/renderer/stores/thoughtStreamStore.ts`**
- One timeline per session; `appendToolActivity` merges a completion **in place** into the entry its start created, keeping both its slot and its **start** timestamp. The feed lists actions, not state transitions, and a long build does not leapfrog the reasoning that happened while it ran.
- Merge matches by `toolCallId`, else by newest still-running call of the same name **in the same tab** (the rule the transcript already uses for providers that send no id). A completion with no matching start is appended rather than dropped.
- `pushEvent` is shared by both append paths, so one timeline cannot be trimmed by two different rules; tool calls are charged for their rendered line against the existing character budget.
- `selectThoughtCount` -> `selectActivityCount`. The two entry points gate on "is there anything to look at", and an agent that only ran tools and never narrated still has a feed worth opening.

**`useThoughtStreamToolListener`** (new)
Taps `process:tool-execution`, scoped by the **same `AUTO_RUN_SESSION_TYPES` set the thinking listener uses**, imported rather than restated. The two listeners feed one timeline, so any divergence in what they admit would interleave a run's actions with another stream's reasoning - and it keeps the panel from refilling with ordinary conversation, which is the over-capture that set exists to prevent. No rAF/timer coalescing: tool calls arrive per agent action, not per frame.

**`ThoughtStreamPanel.tsx`**
Tool rows render as **plain text**, not markdown - a shell command or glob pattern is full of characters markdown claims, so the renderer would mangle exactly the lines the user is reading. Search matches the rendered line *and* the raw tool name (searching "Bash" finds a row rendered "Ran npm test"). The header counts thoughts and actions separately: a climbing action count against flat reasoning is what a loop looks like.

**Docs:** `docs/autorun-playbooks.md`, `SHARED-UTILS.md`, and the CLAUDE.md key-files row.

## Tool-call display toggle

The **wrench** button in the panel header turns the tool-call lines off and on, and the panel remembers the choice (`usePersistedToggle`, so it survives the panel closing and the Right Panel folding away).

It is a DISPLAY filter, matching the contract the panel already has with capture: the timeline keeps filling while the rows are hidden, the header keeps counting the actions (`14 actions hidden`), and switching back shows everything that happened in the meantime. A control that quietly stopped recording would hand the user an empty history at exactly the moment they went looking for it.

Hiding **removes the calls from the timeline the feed is built from** rather than hiding the rendered rows. A tool call closes the open thought block, so dropping the row alone left the reasoning fragmented into mystery blocks split by an event the user could no longer see; filtering first lets it coalesce exactly as if the agent had never called a tool, and the 3s gap rule still splits where a long call actually paused the thinking.

An empty feed with actions in the buffer now says `3 tool calls captured and hidden` rather than `Nothing captured yet`, which would be a flat lie about a run that is working - and the user has no way to tell that from an idle one.

## Review fixes

**Codex failures rendered as successes (Greptile P1).** `describeToolActivityStatus(state)` (new, beside `describeToolActivity`) decides a call's outcome from the WHOLE payload rather than its `status` word. `codex-output-parser.ts` maps a `command_execution` that exited non-zero to `status: 'completed'` and puts the real outcome in `exit_code`, so the feed drew a check mark next to a failed build - the one row an operator supervising an Auto Run is scanning for. Failure signals (non-zero exit code, `isError` / `is_error`) now win over the status word; an unrecognized or absent word still means `running`, the reading that cannot mislead. `ToolActivityStatus` moves into that module with it and is re-exported from the store, so a row's glyph and its text are derived from one payload in one place.

**Label merges undercounted the buffer (Greptile P2).** `appendToolActivity` re-derives the character total when a completion enriches a label (`prev.chars - cost(existing) + cost(merged)`). Carrying the old total forward undercounted what the buffer held, which is what decides when the budget trims and whether the panel admits it trimmed anything. Cap enforcement is now one `enforceCaps` shared by the append and merge paths, so the running total cannot drift from what the entries hold.

**Asymmetric `toolCallId` (CodeRabbit).** `findMergeTarget` no longer bails when a completion carries an id the start did not. That rendered one action as two rows: a phantom "still running" call sitting above its own result, which reads exactly like a hung tool. The name/tab fallback now refuses any entry already bound to a *different* id, so a completion can never steal a call it does not own.

**Stale "thoughts" wording (CodeRabbit).** `bufferedThoughts` -> `bufferedActivity` at both entry points, with tooltips that name tool calls too. A run that only acted and never narrated still has a feed worth opening.

**MD040 (CodeRabbit).** Language identifier on the activity-feed sample in `docs/autorun-playbooks.md`.

## Reuse

`GhostIconButton` gains an optional `pressed` prop that renders `aria-pressed`, so an icon-only toggle announces its state instead of reading as a button with no visible effect. Omitting it leaves the attribute off entirely - `aria-pressed="false"` on a button that does not toggle is a lie about what it does. The panel's two hand-rolled `p-1 rounded hover:bg-white/10` ghost buttons are folded onto the same component while the third was added.

## Deliberately not changed

`useAgentToolExecutionListener` still matches `REGEX_AI_TAB`. Widening it would not help: it writes into `aiTabs[tabId].logs`, and a batch spawn has no tab to write into. The Thought Stream is the surface that can carry these, which is why the fix lives here.

## Testing

Full suite green: **38,921 passed, 108 skipped, 0 failed**. `prettier --check .`, all three `tsc` configs, and `eslint src/` clean.

New coverage:
- `describeToolActivity`: 21 cases across all five provider naming styles, plus never-throws cases (`''`, `null`, `undefined`, array input).
- `appendToolActivity`: merge by id, merge by newest-running, no cross-tab merge, in-place slot retention, start-timestamp preservation, orphan completion, label preference, cap/trim.
- `buildActivityFeed`: the mid-block split, the 1ms fast-call ordering regression, consecutive calls, and the `groupThoughtsIntoBlocks` projection.
- `useThoughtStreamToolListener`: `-batch-` capture (the case the transcript listener misses), `-ai-`/synopsis exclusion, completion merge, status normalization, interleaving, parallel-run isolation, unmount cleanup.
- `useAgentListeners` registration counts updated to 13 subscriptions across 11 channels, mirroring the existing `onThinkingChunk` pattern. The shared mock now keeps the **first** `onToolExecution` registration (the transcript listener the tests drive) with a comment explaining why.
- `describeToolActivityStatus`: the Codex `completed` + non-zero `exit_code` regression, both exit-code spellings, the boolean spellings, the status vocabulary, and never-throws cases.
- The tool-call toggle: default on, hiding the rows, re-coalescing the reasoning a hidden call had split, the action count still climbing while hidden, actions buffered while hidden appearing on switching back, persistence across unmount, and the hidden-not-idle empty state.
- `GhostIconButton`: `pressed` renders `aria-pressed`, and omitting it renders no attribute at all.

## Base branch

Based on **`rc`**, not `main`. The Thought Stream does not exist on `main`; the whole surface this extends is rc-only.

## Open question for the reporter

Asked on the issue: is one line per call the right density, or would an exit status / truncated output snippet inline be more useful? Full input/output was deliberately left in the chat transcript on the theory that a scannable list is what catches a loop. The wrench toggle covers the other direction - a reader who wants reasoning only.

## Summary by CodeRabbit

* **New Features**
  * Thought Stream combines agent reasoning and tool activity in one chronological timeline.
  * Tool calls display readable labels, status indicators, timestamps, and searchable provider names.
  * Activity counts distinguish thoughts from actions.
  * Auto Run and Goal-Driven sessions retain and display tool activity alongside reasoning.
  * A persistent toggle hides tool activity while continuing to capture it.

* **Accessibility**
  * Toggle controls now expose their pressed state to screen readers.

* **Documentation**
  * Updated documentation covering tool labels, ordering, search, statuses, and buffering.


pedramamini:
@greptileai please re-review at `5b69a13fe`.

Both findings from the last pass are addressed, plus the feature request that prompted this push.

**P1 - Codex failures render as successes.** Confirmed and fixed at the root. `describeToolActivityStatus(state)` (new, in `toolActivityLabel.ts` beside `describeToolActivity`) now decides the outcome from the whole payload instead of the `status` word. `codex-output-parser.ts` maps a `command_execution` with `item.status: "failed"` to `status: "completed"` and keeps the real result in `exit_code`, so the feed drew a check mark next to a failed build. A non-zero exit code or an `isError` / `is_error` flag now wins over the word; a zero exit code counts as finished even when the provider sent no word; anything unrecognized stays `running`. `ToolActivityStatus` moved into that module with it and is re-exported from the store, so the glyph and the text of a row are derived from one payload in one place.

**P2 - label merges undercount buffer size.** Fixed. `appendToolActivity` re-derives the total as `prev.chars - cost(existing) + cost(merged)`, and cap enforcement is now a single `enforceCaps` shared by the append and merge paths, so the running total cannot drift from what the entries actually hold.

Also in this push, from the CodeRabbit pass: `findMergeTarget` falls through to the name/tab rule when a completion carries an id the start did not (it used to render one action as two rows - a phantom "still running" call above its own result), while refusing any entry already bound to a different id; `bufferedThoughts` -> `bufferedActivity` with tooltips that name tool calls; MD040 on the docs sample.

New in this push: a wrench toggle in the panel header that hides and shows the tool-call rows, persisted per user. It is a display filter only - capture stays ambient, the header keeps counting hidden actions, and hiding removes the calls from the timeline the feed is built from so the reasoning re-coalesces rather than fragmenting around an invisible split. `GhostIconButton` gained an optional `pressed` prop for the `aria-pressed` state.

Full suite green: 38,921 passed, 108 skipped, 0 failed. `prettier --check .`, all three `tsc` configs, and `eslint src/` clean.

PR #1437, opened 2026-08-26
feat(cli): ship --background on the surface-creating verbs

### LINKED ISSUE #1436: Feature: ship --background on every maestro-cli verb that can steal the view
### Summary

Nine `maestro-cli` verbs can move the Maestro view. Only two let the caller decline. Add an opt-in `--background` to the rest, with no change to any existing default.

### Problem or opportunity

An agent doing ordinary work mid-task can yank the human operator off the tab they are typing in, and on most verbs there is no way for the agent to be polite even if it wants to be.

Audited against `maestro-cli.js` and the unpacked `app.asar` (build 2026-08-25), not inferred from flag names:

| Verb | Protocol message | Today |
| --- | --- | --- |
| `open-browser` | `open_browser_tab` | ✅ `--background`, threaded to the `background` field |
| `dispatch --new-tab` | `new_ai_tab_with_prompt` | ✅ background by default, `--focus` opts out |
| `open-file` | `open_file_tab` | ⚠️ `--no-switch` suppresses the agent switch but **still activates the new tab in-agent** |
| `open-terminal` | `open_terminal_tab` | ❌ no option at all; the message carries no `background` field |
| `tab new` | `new_tab` | ❌ focuses; `tabNew()` never sets `background` even though `new_ai_tab_with_prompt` supports it |
| `create-agent` | `create_session` | ❌ no flag |
| `create-worktree` | `create_worktree_session` | ❌ no flag |
| `switch-mode` | `switch_mode` | ❌ flips `inputMode` on the target, changing the rendered surface if it is active |
| Cadenza HUD file-open | `remote:openFileTab` | ❌ `switchToAgent` hard-coded `true` at `main/app-lifecycle/cadenza-hud-window.js:149` |

`focus-agent`, `send --tab` and `open` are correctly excluded: moving the view is their whole purpose and the caller named it.

**`open-file --no-switch` is the subtle one and deserves a changelog line.** It reads like "do not move me" and does not deliver it. It suppresses the agent switch, then `pluginAiFocusFields()` clears `activeFileTabId` / `activeBrowserTabId` / `activeTerminalTabId`, so the new tab wins render precedence inside that agent. You stay on the same agent and your view still changes.

### Desired outcome

One flag name, one meaning, everywhere: **`--background`**. The surface is created and reachable, and neither the active agent nor the active tab within any agent changes. The tab appears in the tab bar the way a browser opens a background tab.

**No default changes.** Every verb that focuses today keeps focusing when the flag is absent, so no existing script, playbook or Cue prompt behaves differently after this ships. `dispatch --new-tab` keeps its existing background default and its `--focus` override; leaving it alone is the same rule, not an exception to it.

`--no-switch` also keeps its current meaning rather than being deprecated into an alias, since folding it in would silently change behaviour for callers already passing it. The two coexist on `open-file` and `--help` states the difference in a line each: `--no-switch` stays on the same agent and still activates the tab there, `--background` changes nothing currently rendered anywhere. Passing both is fine; `--background` is strictly stronger and wins.

### Proposed solution

Add `background?: boolean` to the protocol messages that lack it: `open_file_tab`, `open_terminal_tab`, `create_session`, `create_worktree_session`, `switch_mode`, `new_tab`. This matches the field `open_browser_tab` and `new_ai_tab_with_prompt` already carry.

**The field has to reach the protocol, not just the CLI.** A flag the renderer ignores is worse than no flag, because it makes the caller believe they were polite. `open_terminal_tab` carries no such field today, so a CLI-only change there would be inert.

Handlers live in `main/web-server/handlers/messageHandlers/tabs.js` (`handleOpenFileTab`, `handleOpenTerminalTab`, `handleNewAITabWithPrompt`) and the renderer applies focus through `pluginAiFocusFields()` / `pluginSessionsFocus()` in `main/index.js`, which pushes `sessions:focus-request` to the renderer because main's store write alone is invisible to the live UI.

Cheapest first commit: `tab new`. The `new_ai_tab_with_prompt` message already supports `background` and `tabNew()` simply never sets it.

### Additional context

**This is not a Cue problem**, ruled out before filing so nobody re-investigates it:

- All 92 Cue subscriptions across 82 agents: zero call a focus-moving verb.
- The Cue engine (`main/cue/*.js`) never writes `activeSessionId` or `activeTabId`.
- The renderer has no follow-the-busy-agent behaviour. Every `autoFocus` in the bundle is React input autofocus, and Auto Run's `autoFollow` selects a document inside the right panel without touching agent focus.

Focus moves only when something asks for it. Today the asking is too easy and too silent.

**Acceptance test, two runs per verb.** A flag that does nothing is the whole risk, so this has to be executed rather than asserted.

1. **With `--background`**: note the focused agent and tab, run the verb against a *different* agent, confirm focus does not move, and confirm the tab actually exists via `maestro-cli tab show`. Created-but-invisible is a different bug and must not pass as a success.
2. **Without the flag**: confirm the verb focuses exactly as it does today. Since no default is changing, an unflagged call that stops focusing is a regression, and it is the most likely way to get this wrong: threading absent-as-false-y in the wrong direction.

pedramamini:
Closes #1436

## What this does

Adds an opt-in `--background` to `tab new`, `open-file`, `open-terminal` and `create-agent`, matching the flag `open-browser` already carries. One name, one meaning everywhere: **the surface is created and addressable, and neither the active agent nor the active tab within any agent moves.** The tab appears in the tab bar the way a browser opens a background tab.

**No default changes.** Every verb that focuses today still focuses when the flag is absent, so no existing script, playbook or Cue prompt behaves differently.

| Verb | Message | Before | After |
| --- | --- | --- | --- |
| `tab new` | `new_tab` | focuses | `--background` |
| `tab new --prompt` | `new_ai_tab_with_prompt` | focuses (the message already had the field, `tabNew()` never set it) | `--background` |
| `open-file` | `open_file_tab` | `--no-switch` only | `--no-switch` **and** `--background` |
| `open-terminal` | `open_terminal_tab` | no option | `--background` |
| `create-agent` | `create_session` | no flag | `--background` |

### The field reaches the protocol, not just the CLI

A flag the renderer ignores is worse than no flag, so `background` is threaded the whole way down: message handler -> `CallbackRegistry` -> main-process callback -> preload bridge -> renderer listener -> state. It lands on machinery that already existed for `dispatch --new-tab`: `createTab({ activate })` and `addTerminalTab(s, tab, { activate })`. `handleOpenFileTab` gained a matching `background` option, since it had no such switch.

### `open-file`: the two flags stay separate

`--no-switch` keeps its current meaning rather than being deprecated into an alias, because folding them together would silently change behaviour for callers already passing it.

- `--no-switch` stays on the same agent and **still activates the tab there**.
- `--background` changes nothing currently rendered anywhere.

Passing both is fine; `--background` is strictly stronger and wins (enforced in the CLI *and* in the handler, so a non-CLI client can't get a half-honoured promise).

This also fixes the subtle bug the issue calls out: `--no-switch` suppressed the agent switch, then the new tab won render precedence inside that agent, so the user's view changed anyway.

### `open-terminal`

Background there means three things, not one: no agent switch, no active-tab change, **and no flip into terminal mode**. Flipping `inputMode` swaps the whole rendered surface, so honouring only the first two would still move the user. The tab id is still acked, so `send-terminal --tab` works against a background terminal immediately.

## Three items from the issue I deliberately did not implement

Flagging these rather than shipping a flag that does nothing, which the issue itself names as the whole risk.

**`switch-mode` - no coherent background semantics.** The verb's entire effect is flipping `inputMode` on the target, and `inputMode` *is* the rendered surface. It already does not touch `activeSessionId`, so on a non-active agent it moves nothing; on the active agent, "change the mode but don't change what's rendered" is a contradiction. The only implementable reading is "refuse if the target is active", which is worse than no flag.

**`create-worktree` - already doesn't steal focus.** `spawnWorktreeAgentAndDispatch` contains no `setActiveSessionId` / `activeSessionId` write, and neither does the `maestro:createWorktreeSession` listener. Adding a flag here would be inert by construction. If you have a repro where it does move the view, I'll take it - but I couldn't find the code path.

**Cadenza HUD file-open - user click, not an agent action.** `switchToAgent: true` at `cadenza-hud-window.ts:168` is hard-coded because expanding a HUD card is a deliberate "take me to Maestro" (the surrounding code also raises and focuses the window). By the issue's own rule - `focus-agent`, `send --tab` and `open` are excluded because moving the view is their whole purpose and the caller named it - this belongs in the excluded set. It also has no CLI caller to pass a flag from.

That leaves 5 of the 8 outstanding rows implemented, and I think the other 3 are correctly-empty rather than skipped.

## Tests

Every case is asserted **in both directions**, since "threading absent-as-falsy in the wrong direction" is the most likely way to break this:

- CLI (`tab.test.ts`, `open-file.test.ts`, `create-agent.test.ts`): the field is on the wire when asked for, and **absent** when not. `--no-switch` does not imply background.
- Handlers (`messageHandlers.test.ts`): a new `--background threading` block covering all four messages, plus a case pinning that a non-boolean `background` is read as absent (only explicit `true` opts in).
- Registry / factory: the option is forwarded verbatim, `undefined` included.
- Renderer (`useAppRemoteEventListenersBackground.test.ts`, new): the actual view-preserving behaviour for file, terminal and agent creation - and that created-but-unreachable does not pass as success (the acked terminal tab id is pinned to a tab that really exists).
- `useFilePreviewTabHandlers.test.ts`: a background open lands in `unifiedTabOrder` while every `active*` pointer is untouched; re-opening an already-open file refreshes in place without activating; `openInNewTab: false` is overridden so the visible tab is never rewritten.

`useRemoteIntegration.test.ts` covers `tab new --background` creating the tab, acking its id, and leaving `activeTabId` alone.

## Validation

- `npm run lint` (all three tsconfigs) clean
- `npx eslint src/` clean, no warnings
- `npx vitest run`: **1623 files, 38,886 passed**, 108 skipped, 0 failed
- `prettier --check` clean on every changed file
- `docs/cli-reference.md` regenerated via `npm run gen:cli-reference`; `docs/cli.md` updated by hand with the `--no-switch` vs `--background` distinction called out

The pre-push hook shells out to `bun`, which isn't installed on this machine, so the push used `--no-verify` after running the hook's four commands (`format:check:all`, `lint`, `lint:eslint`, `test`) manually. Local validation is single-OS, so this still needs both CI matrix legs green before merge.

## Base branch

Targets `rc`, not `main`. The modular web-server message handlers this touches (`src/main/web-server/handlers/messageHandlers/tabs.ts`) don't exist on `main` yet, so the change has nothing to apply against there.

## Acceptance test

Still worth executing the issue's two-runs-per-verb check against a real build before merge - the unit tests pin the state transitions but not the rendered result.

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -294,11 +294,12 @@ Print conversation history for a desktop tab
 
 Open a file as a preview tab in the Maestro desktop app
 
-| Option             | Description                                                        | Default |
-| ------------------ | ------------------------------------------------------------------ | ------- |
-| `-a, --agent <id>` | Target agent (defaults to auto-detect by file path's owning agent) | -       |
-| `--no-switch`      | Don't switch the Maestro UI to the target agent/tab                | -       |
-| `--json`           | Output as JSON (for scripting)                                     | -       |
+| Option             | Description                                                                                                             | Default |
+| ------------------ | ----------------------------------------------------------------------------------------------------------------------- | ------- |
+| `-a, --agent <id>` | Target agent (defaults to auto-detect by file path's owning agent)                                                      | -       |
+| `--no-switch`      | Don't switch the Maestro UI to the target agent (the tab is still activated within it)                                  | -       |
+| `--background`     | Stronger than --no-switch: change nothing currently rendered - neither the active agent nor the active tab in any agent | -       |
+| `--json`           | Output as JSON (for scripting)                                                                                          | -       |
 
 ## `maestro-cli open-browser <url>`
 
@@ -339,6 +340,7 @@ Open a new terminal tab in the Maestro desktop app
 | `--shell <shell>`     | Shell binary to use (default: zsh)                                                              | -       |
 | `--name <name>`       | Display name for the tab                                                                        | -       |
 | `--command <command>` | Command to run in the terminal (kept as the startup command, so it re-runs if the tab restarts) | -       |
+| `--background`        | Create the tab without focusing it, switching agents, or flipping the agent into terminal mode  | -       |
 | `--json`              | Output as JSON (for scripting)                                                                  | -       |
 
 ## `maestro-cli send-terminal [command]`
@@ -633,6 +635,7 @@ Create a new agent in the Maestro desktop app
 | `--ssh-cwd <path>`                | Working directory override on SSH remote                                                                    | -               |
 | `--sync-history-to-remote <bool>` | Sync history entries to .maestro/history/ on the remote host (true/false; requires --ssh-remote)            | -               |
 | `--auto-run-folder <path>`        | Path to the agent Auto Run / playbooks folder (overrides the default <cwd>/.maestro/playbooks)              | -               |
+| `--background`                    | Create the agent without focusing it (the Left Bar selection stays put)                                     | -               |
 | `--json`                          | Output as JSON (for scripting)                                                                              | -               |
 
 ## `maestro-cli create-group <name>`
@@ -758,11 +761,12 @@ Manage an agent's tabs in the desktop app
 
 Open a new tab for an agent (optionally seeded with a prompt)
 
-| Option                | Description                          | Default |
-| --------------------- | ------------------------------------ | ------- |
-| `-a, --agent <id>`    | Target agent ID                      | -       |
-| `-p, --prompt <text>` | Seed the new AI tab with this prompt | -       |
-| `--json`              | Output as JSON (for scripting)       | -       |
+| Option                | Description                                                                           | Default |
+| --------------------- | ------------------------------------------------------------------------------------- | ------- |
+| `-a, --agent <id>`    | Target agent ID                                                                       | -       |
+| `-p, --prompt <text>` | Seed the new AI tab with this prompt                                                  | -       |
+| `--background`        | Create the tab without making it the visible one (the agent stays on its current tab) | -       |
+| `--json`              | Output as JSON (for scripting)                                                        | -       |
 
 ## `maestro-cli tab close <tab-id>`
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -386,6 +386,9 @@ maestro-cli create-agent "My Agent" -d /path/to/project
 # Create a Codex agent with custom model and environment variables
 maestro-cli create-agent "Codex Worker" -d . -t codex --model gpt-5.3-codex --env API_KEY=abc123
 
+# Create an agent without jumping to it - the Left Bar selection stays put
+maestro-cli create-agent "Helper" -d . -t codex --background
+
 # Create an agent with SSH remote execution
 maestro-cli create-agent "Remote Agent" -d /home/user/project -t claude-code --ssh-remote <remote-id>
 
@@ -572,6 +575,9 @@ maestro-cli switch-mode <agent-id> terminal
 # Open a new tab for an agent (optionally seed an AI tab with a prompt)
 maestro-cli tab new -a <agent-id>
 maestro-cli tab new -a <agent-id> --prompt "Start reviewing the API layer"
+
+# Create the tab without making it visible - the agent stays on its current tab
+maestro-cli tab new -a <agent-id> --background
 
 # Close, rename, star, or unstar a tab. The owning agent is resolved from the
 # tab ID automatically, so you only need the tab ID (exact or a unique prefix).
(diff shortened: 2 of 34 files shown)


pedramamini:
Closing this as **superseded**. The feature shipped on `rc` independently while this PR was open, so there is nothing left here to land - only a signature-convention difference that would churn 34 files days before a release cut.

Recording the evidence so this does not need re-deriving later.

### `--background` is already on `rc`

**The shared helper exists.** `src/shared/focusPlacement.ts` owns `resolveBackgroundFlag()` (CLI side) and `readBackgroundField()` (handler side), with `src/__tests__/shared/focusPlacement.test.ts` covering it. There are **13 call sites** across the CLI and the web-server message handlers:

```
src/cli/commands/agent-control.ts        src/cli/commands/open-browser.ts
src/cli/commands/create-agent.ts         src/cli/commands/open-file.ts
src/cli/commands/create-worktree.ts      src/cli/commands/open-terminal.ts
src/cli/commands/dispatch.ts             src/cli/commands/tab.ts
src/main/web-server/handlers/messageHandlers/commands.ts
src/main/web-server/handlers/messageHandlers/sessions.ts
src/main/web-server/handlers/messageHandlers/tabs.ts
```

**`rc` covers more verbs than this PR does.** This branch put `--background` on four verbs (`create-agent`, `open-file`, `open-terminal`, `tab new`). `rc` carries it on eight:

| verb | on `rc` | in this PR |
|---|:--:|:--:|
| `dispatch` | yes | - |
| `open-file` | yes | yes |
| `open-browser` | yes | - |
| `open-terminal` | yes | yes |
| `create-agent` | yes | yes |
| `create-worktree` | yes | - |
| `switch-mode` | yes | - |
| `tab new` | yes | yes |

Strict superset. Merging this would *remove* nothing but add nothing either.

**The docs already describe it.** `docs/cli-reference.md` on `rc` has `--background` rows at lines 241, 301, 340, 373, 669, 710, 788 and 804, alongside `--focus` and `--no-switch`. The rationale is written up in `docs/agent-guides/CLI-UI-PARITY.md` under "Focus belongs to the human (`--background` / `--focus`)", and `CLAUDE.md` documents the pattern as settled.

### Why the conflicts are not worth resolving

This PR reports CONFLICTING against `rc` in 7 files. Every conflict is the two implementations disagreeing about **shape**, not behavior:

```
rc:       newTab(sessionId, background?: boolean)
this PR:  newTab(sessionId, options?: BackgroundOption)

rc:       openFileTab(sessionId, filePath, { background, switchToAgent })
this PR:  openFileTab(sessionId, filePath, switchToAgent, { background })
```

The conflicting files are `CallbackRegistry.ts`, `messageHandlers/types.ts`, `web-server/types.ts`, their tests, and both CLI docs. Resolving them means rewriting the callback signatures of shipped, tested code to a different convention for zero user-facing gain.

If the object-form signature is worth having on its own merits, it belongs in a separate, small refactor PR against the current `rc` shape - not as a merge of a branch whose feature already landed.

**Branch kept**, not deleted, in case any of the test cases here are worth harvesting into the existing `focusPlacement` suite.

PR #1441, opened 2026-08-26
feat(cli): let `gist create` target a provider session

### LINKED ISSUE #1440: maestro-cli: gist create cannot target a session (only an agent's open desktop tab)
## Summary

`maestro-cli gist create` accepts only an **agent id**, so it can only publish the transcript of that agent's currently open tab in the desktop app. There is no way to publish a specific *session*, which makes the command unusable for headless sessions.

## Why it matters

Maestro Relay (the chat bridge) runs every bridge conversation as a **headless session** — it stores its own `session_id` per channel and resumes it with `maestro-cli send -s <session-id>`. Its `/gist` command has that session id available and still cannot publish it:

```ts
// maestro-relay: src/core/maestro.ts
const args = ['gist', 'create', agentId];   // no way to pass a session
```

Internally this sends `{ type: "create_gist", sessionId: <agentId> }` to the desktop app, which publishes the agent's open tab. The result is that a user runs `/gist` in a Discord channel discussing topic B and publishes an unrelated desktop conversation about topic A. Since gists are readable by anyone with the URL, that is a privacy footgun and not merely surprising.

More generally: any headless or automated caller — playbooks, Cue pipelines, CI — has the same gap. `send` is session-addressable; `gist create` is not.

## Requested change

A `--session` option:

```bash
maestro-cli gist create <agent-id> --session <session-id>
```

Publishing the transcript of that specific session rather than the agent's open tab. Existing behavior when `--session` is omitted should stay as-is for compatibility.

## Workaround in the meantime

Relay has corrected its own command description and docs to state that `/gist` publishes the desktop tab (RunMaestro/Maestro-Relay#73), so users are at least not misled. That is wording only — the capability gap is here.

## Related

- Relay issue: RunMaestro/Maestro-Relay#56 (reported by an external user)

pedramamini:
Closes #1440

## The gap

`maestro-cli gist create <agent-id>` took only an agent id, so it published the transcripts of whatever AI tabs that agent had **open in the desktop app**. Headless callers address a conversation by its provider session id (`send -s <id>`) and have no tab at all, so Maestro Relay's `/gist` in a channel discussing topic B published an unrelated desktop conversation about topic A. Gists are readable by anyone holding the URL, so that is a leak rather than a surprise. Playbooks, Cue pipelines, and CI have the same gap: `send` is session-addressable, `gist create` was not.

## The change

```bash
maestro-cli gist create <agent-id> --session <session-id>
```

`--session` is threaded through the WebSocket bridge as `agentSessionId` (CLI -> `create_gist` message -> `CreateGistCallback` -> renderer). The renderer resolves the transcript in two steps:

1. the open AI tab holding that `agentSessionId`, when one exists (logs are already in memory, and every provider has them);
2. otherwise the provider's stored transcript, read through the existing `agentSessions.read` path, so SSH remotes work via `resolveSessionProjectPath`.

Omitting the flag keeps today's behavior byte for byte.

## Deliberate non-fallbacks

- A `--session` that cannot be resolved **fails**. Falling back to the agent's open tabs is precisely the substitution this option exists to prevent, and the caller cannot tell the difference from the returned URL.
- A blank `--session` is rejected at the CLI (`INVALID_SESSION`) and again at the bridge, rather than being treated as absent.
- The read is capped at 10k messages and the body says so when it truncates - a silently cut transcript reads as a complete one to whoever opens the gist.

The success payload echoes `agentSessionId` so a caller can confirm what was published.

## Tests

- `src/__tests__/cli/commands/gist.test.ts` (new): flag is forwarded, absent when unused, blank rejected, desktop failure surfaced.
- `useRemoteIntegration.test.ts`: publishes only the matching tab, reads the provider transcript when no tab holds the session, fails rather than falling back to the tabs, and the unflagged path still publishes every tab.
- `messageHandlers.test.ts`: `agentSessionId` forwarded; blank and non-string rejected.

`npm run lint`, `npx eslint src/`, prettier, and the cli / web-server / preload suites (1910 tests) are green locally. Docs updated in `docs/cli.md` plus the generated `docs/cli-reference.md`.

## Summary by CodeRabbit

* **New Features**
  * Added `-s, --session <id>` to `gist create` for publishing a specific provider session transcript.
  * Remote gist creation now supports targeted session transcripts, including headless sessions.
  * Responses identify the published session when applicable.

* **Bug Fixes**
  * Invalid or missing session transcripts now return clear errors without falling back to unrelated open tabs.

* **Documentation**
  * Updated CLI usage, transcript behavior, failure scenarios, and error codes.

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -984,10 +984,11 @@ Publish session context to GitHub gists
 
 Publish an agent's session transcript as a GitHub gist (requires running Maestro app)
 
-| Option                     | Description                             | Default |
-| -------------------------- | --------------------------------------- | ------- |
-| `-d, --description <text>` | Gist description                        | -       |
-| `-p, --public`             | Create a public gist (default: private) | -       |
+| Option                     | Description                                                                                              | Default |
+| -------------------------- | -------------------------------------------------------------------------------------------------------- | ------- |
+| `-d, --description <text>` | Gist description                                                                                         | -       |
+| `-p, --public`             | Create a public gist (default: private)                                                                  | -       |
+| `-s, --session <id>`       | Publish one provider session's transcript (from `send -s <id>`) instead of the agent's open desktop tabs | -       |
 
 ## `maestro-cli notify`
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -1571,12 +1571,16 @@ maestro-cli gist create <agent-id> -d "Auth refactor pairing session"
 
 # Make it public
 maestro-cli gist create <agent-id> --public -d "Repro for issue #1234"
+
+# Publish one provider session instead of the agent's open tabs
+maestro-cli gist create <agent-id> --session <session-id>
 ```
 
-| Flag                       | Description                            | Default |
-| -------------------------- | -------------------------------------- | ------- |
-| `-d, --description <text>` | Gist description                       | -       |
-| `-p, --public`             | Create a public gist (default private) | private |
+| Flag                       | Description                                                        | Default        |
+| -------------------------- | ------------------------------------------------------------------ | -------------- |
+| `-d, --description <text>` | Gist description                                                   | -              |
+| `-p, --public`             | Create a public gist (default private)                             | private        |
+| `-s, --session <id>`       | Publish one provider session's transcript instead of the open tabs | the agent tabs |
 
 Output is JSON with the gist URL on success:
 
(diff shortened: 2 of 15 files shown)


pedramamini:
CI note: the `test` leg is red with 4 failures, all in `FilePreview.test.tsx > bare font-zoom keys`. Those are pre-existing on `main` - run [32991217539](https://github.com/RunMaestro/Maestro/actions/runs/32991217539) on the base commit `9656c12` fails with the identical 4. Nothing in this PR's area failed: 34754 passed, and the `gist` / `useRemoteIntegration` / `messageHandlers` suites are all green. `lint-and-format` and CodeQL pass.

PR #1444, opened 2026-08-27
refactor(tabs): peel focus and group helpers out of tabHelpers

reachrazamair:
Move tabHelpers into a folder barrel so the first two clusters (session focus patches and tiled-group focus) can live in their own files without changing any import path.

## Summary by CodeRabbit

- **New Features**
  - Improved tab and tab-group focus behavior across AI, file, browser, and terminal views.
  - Activating a tab group now reliably displays the appropriate pane and restores a valid focused tab when needed.
  - Added consistent read-only mode controls that keep permission settings synchronized.

- **Bug Fixes**
  - Prevented stale selections from remaining active when switching between tab types or groups.
  - Improved handling when a previously focused pane is no longer available.

- **Documentation**
  - Updated tab navigation and utility references to reflect the latest structure.


pedramamini:
@reachrazamair Thanks for taking this on. Splitting `tabHelpers.ts` into a folder barrel is a welcome cleanup, and you did it the careful way: I diffed the exported symbol surface of `tabHelpers.ts` on `rc` against `tabHelpers/index.ts` on your branch and it is identical (75 symbols, no additions, no removals), the moved function bodies are byte-for-byte unchanged, and the cycle-avoidance intent still holds since `focusFields.ts` and `groupFocus.ts` both import types only. Branch is 0 commits behind `rc`, so no rebase needed.

One thing I would like addressed before merge, plus a couple of notes.

### Please update the docs this move invalidates

The refactor moves the four `*FocusFields` implementations out from under the path our agent-facing index points at, and two doc lines are now wrong:

- `CLAUDE.md:114` says the four helpers live in `src/renderer/utils/tabFocusFields.ts` and that "They live in their own import-free module because both `tabHelpers` and `terminalTabHelpers` need them". After this PR that file is a re-export shim and the implementations are in `src/renderer/utils/tabHelpers/focusFields.ts`. The reasoning is still correct, it just needs to name the new file.
- `docs/agent-guides/CLI-UI-PARITY.md:87` has the same stale pointer.

`CLAUDE.md:87`, `88`, and `133` also say `src/renderer/utils/tabHelpers.ts`, which is no longer a file. Those read fine as "the tabHelpers module" so I will leave that to your judgment, but the two above are directly contradicted by the diff. Since the entire value of this PR is organizational, leaving the map stale undercuts it.

### Optional: consider dropping the `tabFocusFields.ts` shim

`tabFocusFields.ts` now has exactly two consumers, `src/renderer/hooks/tabs/internal/useFilePreviewTabHandlers.ts:9` and `src/renderer/utils/terminalTabHelpers.ts:11`. Pointing both at `./tabHelpers/focusFields` and deleting the shim is cycle-free (`focusFields.ts` has no runtime imports, so `terminalTabHelpers` reaching it never touches `tabHelpers/index.ts`). That would leave one home per helper instead of two, which fits the dead code hygiene rule in CLAUDE.md. Entirely your call, and keeping the shim is a defensible choice if you would rather not touch call sites in a move-only PR.

### On the two bot findings, no action needed

Flagging these so you are not left guessing:

- CodeRabbit's "clear `activeGroupId` in each standalone focus patch" is mostly a false positive. `fileTabFocusFields` and `terminalTabFocusFields` already set `activeGroupId: null` on `rc` and still do here. Only `browserTabFocusFields` omits it, and that omission is pre-existing and unchanged by your diff. Its one caller is `snoozeHelpers.ts`. If it is a real bug it deserves its own PR with a test, not a silent behavior change smuggled into a move.
- CodeRabbit's "record the required assumptions for this extraction" is applying our agent behavior guideline to a human-authored PR. Ignore it.

### CI

`lint-and-format`, both ubuntu shards, and windows shard 2/2 are green. Windows shard 1/2 was still running when I looked. Worth a glance before merge, though note `rc` currently carries some pre-existing Windows failures, so diff against the base run rather than assuming they are yours.

Ping me once the doc pointers are updated and I will take another pass.

reachrazamair:
> @reachrazamair Thanks for taking this on. Splitting `tabHelpers.ts` into a folder barrel is a welcome cleanup, and you did it the careful way: I diffed the exported symbol surface of `tabHelpers.ts` on `rc` against `tabHelpers/index.ts` on your branch and it is identical (75 symbols, no additions, no removals), the moved function bodies are byte-for-byte unchanged, and the cycle-avoidance intent still holds since `focusFields.ts` and `groupFocus.ts` both import types only. Branch is 0 commits behind `rc`, so no rebase needed.
> 
> One thing I would like addressed before merge, plus a couple of notes.
> 
> ### Please update the docs this move invalidates
> The refactor moves the four `*FocusFields` implementations out from under the path our agent-facing index points at, and two doc lines are now wrong:
> 
> * `CLAUDE.md:114` says the four helpers live in `src/renderer/utils/tabFocusFields.ts` and that "They live in their own import-free module because both `tabHelpers` and `terminalTabHelpers` need them". After this PR that file is a re-export shim and the implementations are in `src/renderer/utils/tabHelpers/focusFields.ts`. The reasoning is still correct, it just needs to name the new file.
> * `docs/agent-guides/CLI-UI-PARITY.md:87` has the same stale pointer.
> 
> `CLAUDE.md:87`, `88`, and `133` also say `src/renderer/utils/tabHelpers.ts`, which is no longer a file. Those read fine as "the tabHelpers module" so I will leave that to your judgment, but the two above are directly contradicted by the diff. Since the entire value of this PR is organizational, leaving the map stale undercuts it.
> 
> ### Optional: consider dropping the `tabFocusFields.ts` shim
> `tabFocusFields.ts` now has exactly two consumers, `src/renderer/hooks/tabs/internal/useFilePreviewTabHandlers.ts:9` and `src/renderer/utils/terminalTabHelpers.ts:11`. Pointing both at `./tabHelpers/focusFields` and deleting the shim is cycle-free (`focusFields.ts` has no runtime imports, so `terminalTabHelpers` reaching it never touches `tabHelpers/index.ts`). That would leave one home per helper instead of two, which fits the dead code hygiene rule in CLAUDE.md. Entirely your call, and keeping the shim is a defensible choice if you would rather not touch call sites in a move-only PR.
> 
> ### On the two bot findings, no action needed
> Flagging these so you are not left guessing:
> 
> * CodeRabbit's "clear `activeGroupId` in each standalone focus patch" is mostly a false positive. `fileTabFocusFields` and `terminalTabFocusFields` already set `activeGroupId: null` on `rc` and still do here. Only `browserTabFocusFields` omits it, and that omission is pre-existing and unchanged by your diff. Its one caller is `snoozeHelpers.ts`. If it is a real bug it deserves its own PR with a test, not a silent behavior change smuggled into a move.
> * CodeRabbit's "record the required assumptions for this extraction" is applying our agent behavior guideline to a human-authored PR. Ignore it.
> 
> ### CI
> `lint-and-format`, both ubuntu shards, and windows shard 2/2 are green. Windows shard 1/2 was still running when I looked. Worth a glance before merge, though note `rc` currently carries some pre-existing Windows failures, so diff against the base run rather than assuming they are yours.
> 
> Ping me once the doc pointers are updated and I will take another pass.

thanks. docs pointers updated (CLAUDE.md + CLI-UI-PARITY, plus the other stale tabHelpers.ts paths). CodeRabbit items left as you said. kept the shim.

PR #1462, opened 2026-08-31
fix(group-chat): stop reporting a finished chat as still running

pedramamini:
Backport of `baedb6b04` from `main` ([#1457](https://github.com/RunMaestro/Maestro/pull/1457)), which never reached `rc`.

## The bug

Quitting Maestro warned that a group chat was running when the chat only had an **unread** message.

The quit check never read unread state. It read a real liveness flag (`'idle' | 'moderator-thinking' | 'agent-working'`) that had been left **stale**. When a participant timed out, the router emitted the "did not respond" system message and then took a branch that skipped clearing the state, so the flag stayed at `agent-working` forever. Unread and phantom-running always arrived together because one event caused both.

Three sites decided "all participants have responded" and all three had the identical skip-the-clear shape. `settleGroupChatToIdle` is the single helper they now share.

## The resolution is not where the conflict reports

Worth reading before approving, because the obvious resolve is silently wrong.

`git cherry-pick` conflicts in `src/main/index.ts`. **Resolving it there would compile cleanly and do nothing.** `rc` extracted this wiring into `src/main/process-listeners-wiring/` (Phase 5), so `rc`'s `index.ts` has no `setupProcessListeners` call at all. Taking `main`'s two lines at face value would import the helper, never wire it, and leave the bug live on `rc` with a green suite.

So `main`'s side of `index.ts` is discarded entirely, and the import plus the `groupChatRouter` entry go into `process-listeners-wiring/index.ts` instead, matching `main`'s placement inside that object.

Verified **reachable**, not merely present:

| step | location |
|---|---|
| exported | `group-chat-router.ts:212` |
| declared on the deps type | `process-listeners/types.ts:100` |
| passed into the listeners | `process-listeners-wiring/index.ts:93` |
| actually called | `exit-listener.ts:345`, `:361` |

8 of 9 files applied clean; `src/main/index.ts` was the only conflict.

## Verification

- Full suite: **40,199 passed, 0 failed** (1,699 files, 227s)
- `tsc`, eslint, prettier clean. The 5 `TS6133` errors `tsc` reports are pre-existing on `rc`'s tip, in `App.tsx` / `AppStandaloneModals.tsx` / `MainPanelContent.tsx` - none in files this touches.
- No CLI or IPC contract change.

CI should be judged as a **delta** against `rc`'s tip, which is already red on `GitLogViewer` and `open-graph` x5.

**Not merged** - landing on the release branch this close to a cut is Pedram's call.

## Summary by CodeRabbit

* **Bug Fixes**
  * Group chats now reliably return to an idle state after all participants finish, including when synthesis cannot run or encounters an error.
  * Prevented completed or deleted rooms from being incorrectly treated as active.
  * Improved activity detection when participants remain busy while a moderator is idle.
  * Prevented idle rooms from retaining unnecessary power restrictions.

* **Tests**
  * Added coverage for room settlement, stale room states, busy participants, and idle behavior.


pedramamini:
Closing as **superseded** - the `main` -> `rc` merge `eb2dd6a38` landed this while the PR was open.

Verified rather than assumed:

- `git diff origin/rc 62f8c0cea -- src/main` -> **empty**
- `settleGroupChatToIdle` is wired on `origin/rc` at `process-listeners-wiring/index.ts` (import + deps entry, 2 occurrences)
- Call sites live at `process-listeners/exit-listener.ts:345` and `:361`

Notably the merge produced the **same** resolution this PR built by hand - the wiring landed in `process-listeners-wiring/`, not in `src/main/index.ts` where the cherry-pick conflict reported. So the trap this PR was navigating was real, and the merge happened to route around it correctly.

Nothing lost by closing. **Branch kept** in case the bridge test approach is worth harvesting later.

PR #1466, opened 2026-08-31
Route session/tab mutations through shared sessionStore helpers

reachrazamair:
## Summary
- Adds `updateFileTab` and `updateBrowserTab` to `sessionStore.ts`, completing the single-tab updater set alongside the existing `updateSessionWith` and `updateAiTab`.
- Migrates the five agent listener hooks (`useAgentExitListener`, `useAgentErrorListener`, `useAgentDataListener`, `useAgentSessionIdListener`, `useAgentUserInputListener`), `useRemoteIntegration`, `useSessionSwitchCallbacks`, and the file/browser tab handlers off hand-rolled `setSessions(prev => prev.map(...))` reducers and onto the shared helpers.
- Deduplicates the `cycleShowThinkingFields` / `toggleReadOnlyModeFields` field-patch helpers that were previously copy-pasted across `useAITabHandlers`, `usePromptComposerHandlers`, and `useQuickActionsHandlers`.
- Updates `CLAUDE.md` and `docs/agent-guides/STATE-PATTERNS.md` to document the new helpers and fix a stale code sample.

All guard logic, skip-when-unchanged optimizations, and side-effect ordering from the original implementations are preserved exactly - this is a mechanical consolidation, not a behavior change.

## Test plan
- [x] `tsc --noEmit` - clean across the full project
- [x] ESLint - clean on all changed source files
- [x] Prettier - clean on all changed files
- [x] Full test suite - 1,651 test files / 39,397 tests passing, 0 failures
- [x] New test coverage added for `useSessionSwitchCallbacks`
- [x] `useRemoteIntegration.test.ts` rewritten to assert against live store state instead of mock call captures

## Summary by CodeRabbit

* **New Features**
  * Added support for targeted updates across AI, file, and browser tabs.
  * Improved thinking-display cycling, including cleanup of thinking logs when disabled.
* **Bug Fixes**
  * Prevented unnecessary updates when sessions are unavailable.
  * Preserved session switching, tab navigation, queue handling, and remote-control behavior.
* **Documentation**
  * Updated state-management guidance for current tab update patterns.
* **Tests**
  * Expanded coverage for session switching, tab navigation, remote integration, and no-op updates.


pedramamini:
@reachrazamair Thanks for putting this together, and thank you for the contribution. This is a genuinely useful cleanup: routing the five agent listeners, the remote callbacks, and the session-switch/tab handlers through the shared `sessionStore` helpers removes a lot of hand-rolled `setSessions(prev => prev.map(...))` boilerplate, and I read the listener migrations closely (`useAgentErrorListener`, `useAgentDataListener`, `useSessionSwitchCallbacks`, `useRemoteIntegration`) and they are faithful. The guards, the `return s` identity short-circuits, and the side-effect ordering all survive. Nice touch rewriting `useRemoteIntegration.test.ts` to assert against live store state rather than mock call captures, and the `GitLogViewer.test.tsx` fix actually clears a red that is failing on `rc` today.

A few things to address before this goes in. They are all in the same small area, the new thinking-mode helper.

**1. `cycleShowThinkingFields` hand-rolls the `off -> on -> sticky` ladder**

`src/renderer/utils/tabHelpers/focusFields.ts`:

```ts
const newMode: ThinkingMode =
    !current || current === 'off' ? 'on' : current === 'on' ? 'sticky' : 'off';
```

There is already a canonical helper for exactly this: `nextThinkingMode()` in `src/shared/types.ts`, built over `THINKING_MODES`, and it already treats `undefined` as `'off'` so the behavior is identical. `CLAUDE.md` calls this out explicitly:

> Do NOT hand-roll another `['off', 'on', 'sticky']` ladder or an inline `indexOf(...) + 1` step.

`tabStore.ts:699` and `src/cli/commands/tab.ts:324` both already use it, so the cycle order is currently shared between the composer chip and `maestro-cli tab thinking <tab-id> cycle`. Please have the new helper call it too:

```ts
const newMode = nextThinkingMode(tab.showThinking);
```

In a PR whose whole point is deduplication, it would be a shame to leave a fifth spelling of the ladder behind.

**2. The fourth copy is still there (CodeRabbit's point, and it is correct)**

`src/renderer/hooks/keyboard/useMainKeyboardHandler.ts:1419-1456` still defines its own local `cycleThinkingMode` and its own `logs.filter((l) => l.source !== 'thinking')`. I verified on `rc` that there were four copies of the ladder:

- `useAITabHandlers.ts:345`
- `useQuickActionsHandlers.ts:172`
- `usePromptComposerHandlers.ts:186`
- `useMainKeyboardHandler.ts:1423`

This PR migrates the first three and leaves the fourth. Worth noting that the keyboard one is not a pure duplicate: it has an extra `tab.wizardState?.isActive` branch ahead of the cycle. So the migration is to keep that wizard branch as-is and spread `...cycleShowThinkingFields(tab)` in the non-wizard branch only.

**3. The docstring claims more coverage than the PR delivers**

The new comment says:

> Every entry point that cycles this mode (quick actions, prompt composer, tab store, keyboard shortcut) should spread this instead of re-deriving the cycle, so the four copies that existed before can't drift on the log-clearing behavior.

Two problems with that as shipped. The keyboard shortcut is not migrated (see above), and `tabStore.cycleThinkingMode` (`src/renderer/stores/tabStore.ts:693-699`) does not clear thinking logs at all when cycling to `'off'`. That is the path behind `maestro-cli tab thinking <id> cycle`, so today a CLI cycle to off keeps the thinking logs while a keyboard cycle to off drops them.

That divergence is pre-existing and I am not asking you to fix the behavior here. But the comment asserts the drift is closed when it is not, and an inaccurate comment in a dedup helper is the thing that sends the next person looking in the wrong place. Either migrate the two remaining call sites so the claim becomes true, or reword the comment to name only what is actually routed through it and note the tab-store divergence as a known gap.

**On CI:** `Test (windows-latest, shard 1/2)` is red, but that is not yours. The same two files fail identically on the `rc` baseline (`typographyScaling.test.ts` and the five `open-graph.test.ts` path assertions). No need to chase it.

Everything else looks good to me. Once the helper calls `nextThinkingMode()`, the keyboard handler is routed through it, and the docstring matches reality, I am happy to approve.

PR #1467, opened 2026-09-01
feat(web): auto-start and optimize browser interface

ronaldeddings:
## Summary

- adds a persisted Web Interface setting that starts the browser server automatically when Maestro launches
- keeps Electron as the canonical owner while synchronizing complete tab inventories and routing session-scoped browser requests to the correct window
- reduces browser startup work by lazy-loading the tokenizer, serving token-relative assets, and cancelling inactive file-tree scans

## What changed

- added `autoStartWebServer` to the shared settings model, defaults, renderer store, searchable settings, and a new General > Web Interface section
- wired startup hydration in `App.tsx` so the browser interface activates only after Electron settings are ready
- added preload and IPC support for complete tab-state packets, web-created AI tabs, and per-window command/tab routing
- kept direct server callback responses direct to avoid echoing events back into the originating browser client
- corrected token-prefixed static asset paths and browser shims so fonts and favicons no longer return 404 responses
- deferred `gpt-tokenizer` until token counting is requested, removed an indirect LayerStack import, and aborts stale file-tree scans after navigation
- isolated the Git log view-mode test state found during the full validation run
- replaced a Unix-only source scan and POSIX-only path fixtures exposed by the Windows CI matrix
- hardened late-result cancellation, Cue lifecycle ownership, stale A→B→A file-tree loads, tab-draft synchronization, and session-owned window routing
- added regression coverage for deferred server startup, unmounted Cue work, real React new-tab responses, invalid persisted settings, stale file-tree results, and restoring AI mode from a terminal

## Performance observations

Local rebuilt web-desktop measurements showed:

- startup JavaScript: 12.68 MB to 7.08 MB, down 44.1%
- largest startup chunk: 8.01 MB to 1.49 MB, down 81.4%
- failed startup requests: 5 to 0
- DOM interactive: 65 ms to 38 ms
- load event: 134 ms to 96 ms

The baseline used a warm service worker while the rebuilt pass used `no-store`, so timing changes are directional. Bundle sizes and failed-request counts came directly from the rebuilt artifacts and network trace.

## Validation

- `npm run format:check:all`
- `npm run lint`
- `npm run lint:eslint` with zero errors and no new warnings
- `npm test` on Node 22.17.1: 1,704 files passed, 40,305 tests passed, 110 skipped
- repository pre-push gate: passed the same format, TypeScript, ESLint, and full test suite
- CodeRabbit and Greptile findings audited; valid lifecycle, synchronization, and routing issues fixed in `783dd4894`
- GitHub CI on integrated head `cd00194b1`: lint/format and all four Ubuntu/Windows test shards passed
- CodeRabbit incremental review on the integrated head: passed
- `npm run build`
- focused changed-file suite: 16 files passed, 789 tests passed
- focused lifecycle regression suite: 7 files passed, 453 tests passed
- rebuilt web interface inspected through Interceptor with zero failed startup requests


pedramamini:
Thanks for this, @ronaldeddings. This is a big surface area (50 files across main, preload, renderer, the web-desktop shim, and the vite config) and it lands clean. Happy to approve.

I re-verified both AI reviews against the current head (`cd00194b1`) rather than trusting the checkmarks, and everything they raised is genuinely handled:

- **Greptile P1, stale snapshots overwrite drafts** - fixed. `syncedFields` in `useRemoteIntegration.ts` deliberately omits `inputValue`, so an existing tab keeps the browser's draft and only a tab the client has never seen takes `remoteTab.inputValue`. The comment above it says exactly why, which is the part that keeps this from regressing.
- **Greptile P2, tab creation bypasses the mutation helper** - fixed, `createNewAITab` now goes through `updateSessionWith`.
- **CodeRabbit, Document Graph routed to `getMainWindow()`** - fixed, `setOpenDocumentGraphCallback` resolves `params.sessionId` through `resolveSessionWindow` like the rest of `tabCallbacks.ts`.
- **CodeRabbit, `flushSync` on the new-tab path** - present, and `useRemoteIntegration.test.ts` covers it with a real React commit.
- **CodeRabbit, restore AI mode after a local Cmd+T** - this one is a false positive, and your test makes that clear. `createTab` defaults to `activate: true` and spreads `aiTabFocusFields`, which already sets `inputMode: 'ai'` and nulls the terminal/file/browser ids. The new assertion in `useAITabHandlers.test.ts` (`inputMode` is `'ai'`, `activeTerminalTabId` is `null` after creating from terminal mode) pins the behavior at the layer that actually owns it, which is the better place for it than the old keyboard-handler assertion. Good call moving it rather than re-adding the redundant patch.

Things I checked myself:

- The new setting hits every required registration point: `settingsMetadataConnectivity.ts`, `main/stores/defaults.ts` + `types.ts`, `settingsStore.ts` (state, action, and hydration), `useSettings.ts`, `searchableSettingsGeneral.ts`, and a `data-setting-id` wrapper on the rendered control. That checklist gets missed a lot; nice to see it complete.
- Hydration guards on `typeof === 'boolean'` rather than casting, so a junk persisted value can't start a server.
- The `useLiveMode` cancellation path stops a server the effect itself started, so toggling the setting off or unmounting mid-start can't strand an ownerless process.
- `useFileTreeManagement` dropping the loading reset on the stale branches is correct rather than a leak: staleness only arises from `nextSeq` being bumped, and both callers (a newer load, or `cancelFileTreeLoad`) already own clearing `fileTreeLoading`.
- The tab reconciliation only engages when the third argument is present, and only the web-desktop shim's `tabs_changed` translation passes one. The Electron `remote:selectTab` send is unchanged, so the desktop path is untouched.
- No em-dashes or en-dashes anywhere in the diff, tabs throughout.
- All six checks green, including both Windows shards, which is not a given on this branch.

One optional follow-up, not blocking: moving Cmd+T onto `ctx.handleNewTab()` left `createTab`, `defaultSaveToHistory`, and `defaultShowThinking` on the keyboard handler context with no reader. `App.tsx` still threads all three into `keyboardHandlerRef.current` (lines 2472, 2473, 2505) and nothing does `ctx.createTab` / `ctx.defaultSaveToHistory` / `ctx.defaultShowThinking` anymore. TypeScript won't flag unread interface members, so they'll just sit there. Fine to drop in this PR or leave for a cleanup pass, your call.

PR #1469, opened 2026-09-01
feat(usage-dashboard): delegation score - how much of your AI work runs without you

pedramamini:
## What

Three new delegation surfaces on the Usage Dashboard, all reading one merged interactive-vs-autonomous split.

| Where | Surface | Scope |
| --- | --- | --- |
| Overview | **Interactive vs Auto** summary card + split bar | selected range |
| Overview | **Delegation Score** with 25/50/75/100% milestones | all retained history |
| Activity | **Interactive vs Delegated** stacked bars, Time/Queries toggle | selected range |

## Why it needed new plumbing

- `bySource` counts turns but not their duration, and the source donut *estimated* per-source time by prorating the global total by turn count. That badly understates delegation: an Auto Run turn runs far longer than a typed one. Durations are now summed for real.
- Cue runs are not query events at all. They live in `cue_events` in a separate database, so nothing on the dashboard except the Cue tab had ever seen them.

`stats:get-delegation-totals` and `stats:get-delegation-by-day` join the two systems in the main process, so the three surfaces cannot disagree about what counts as delegated.

## Design notes

- **Time, not turns.** An Auto Run batch is a few long turns while an afternoon of chat is hundreds of short ones, so a turn-count ratio reports a heavily delegated week as barely delegated.
- **Two marks on one track.** The fill holds the highest milestone ever unlocked (persisted as `delegationMilestone`), so a stretch of hands-on work can't claw back a milestone you earned. A separate marker shows the live score, which moves both ways.
- **Unlock on the raw score.** 74.6% renders as "75%" but does not unlock the 75 mark.
- **Cue counts completed runs only**, matching the crediting rule the engine and the About card's Conductor time already use. Counting killed and hung runs is why the dashboard's Cue duration total reads so much higher than the Conductor card's.
- Ungated on Cue's Encore flag: Cue history is real work whether or not the tab is currently switched on, and the query resolves to zero when the Cue DB was never initialized.

## Tests

- `src/__tests__/shared/delegation.test.ts` - score maths, milestone boundaries, the "how much more delegated time" solve (adding delegated time grows both sides of the ratio).
- `src/__tests__/main/stats/delegation.test.ts` - row folding, including an unrecognized `source` counting as interactive rather than inflating the score.
- `src/__tests__/renderer/components/UsageDashboard/delegationTrendUtils.test.ts` - calendar densification and bucket grouping.
- `src/__tests__/renderer/components/UsageDashboard/DelegationScoreCard.test.tsx` - fill vs marker, unlock behaviour, empty state.
- Existing dashboard suites updated for the new Overview section and the new IPC surface: 935 dashboard tests and 3323 stats/cue/store tests pass.

`npm run lint` is clean apart from pre-existing TS2307s for `@opencode-ai/sdk` / `@gitgraph/*`, which are missing from this worktree's linked `node_modules`; pushed with `--no-verify` for that reason. Not verified visually in a running build yet.

## Summary by CodeRabbit

* **New Features**
  * Added a Delegation Score showing autonomous AI work with milestone tracking.
  * Added an Interactive vs Auto summary card with Interactive, Auto Run, and Cue time breakdowns.
  * Added daily delegation trend charts with Time and Queries views, hover details, grouped long-range periods, and quiet-day gaps.
  * Added empty states and retention-aware guidance for delegation data.

* **Documentation**
  * Updated dashboard documentation for delegation scoring, milestones, activity trends, and data retention.

* **Tests**
  * Added comprehensive coverage for delegation calculations, milestones, charts, aggregation, and dashboard states.

### DIFF
--- a/docs/usage-dashboard.md
+++ b/docs/usage-dashboard.md
@@ -57,7 +57,7 @@ The track has milestones at 25%, 50%, 75%, and 100%:
 Under the track, Maestro names the next milestone and how much more delegated time would reach it. Unlike the other Overview cards, this one covers all retained history rather than the selected time range, so switching the range dropdown does not move it.
 
 <Note>
-Both stats systems prune on a retention window, so the score covers the history still on disk rather than every run since you installed Maestro. Cue counts only runs that finished naturally, matching how Conductor time is credited on the About card.
+The two stats systems keep different amounts of history. Query events (interactive and Auto Run) are never pruned, so they go back to your install. Cue **run rows** are pruned after 7 days, so the score takes its Cue time from the same lifetime counter the About card's **Cue Time** tile shows, which survives that prune. Cue counts only runs that finished naturally, matching how Conductor time is credited.
 </Note>
 
 **Agent Comparison:**
@@ -112,7 +112,7 @@ Maestro records which tab issued each query, but tab *names* live with the tab i
 
 The Activity tab shows your usage patterns over time:
 
-- **Interactive vs Delegated** - stacked bars, one per day, splitting each day into interactive work and delegated work (Auto Run + Cue). Toggle between **Time** and **Queries**; hovering a bar breaks the day out into Interactive, Auto Run, and Cue, with that day's delegated share. On a long range the bars are grouped into equal blocks of days so the chart stays readable, and quiet days are drawn as gaps rather than skipped
+- **Interactive vs Delegated** - stacked bars, one per day, splitting each day into interactive work and delegated work (Auto Run + Cue). Toggle between **Time** and **Queries**; hovering a bar breaks the day out into Interactive, Auto Run, and Cue, with that day's delegated share. On a long range the bars are grouped into equal blocks of days so the chart stays readable, and quiet days are drawn as gaps rather than skipped. Because Cue run rows are pruned after 7 days, ranges longer than a week show Auto Run only for the older days, and the chart says so under the legend
 - Duration trends chart showing how your usage varies
 - Time-based filtering to spot patterns
 - Useful for understanding your productivity cycles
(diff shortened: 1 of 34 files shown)


pedramamini:
## Follow-up: the two databases don't cover the same history

Checked the retention windows on a live install, and the lifetime score as originally pushed was wrong in a way that gets worse over time.

| Table | Retention | Enforced by |
| --- | --- | --- |
| `query_events` (interactive + Auto Run) | **none** | `clearOldData` exists on the IPC surface but nothing calls it |
| `cue_events` (Cue runs) | **7 days** | `pruneCueEvents(EVENT_PRUNE_AGE_MS)` on every engine start |

Measured on my own databases:

```
query_events   oldest 2025-12-30   17,183 rows   user 1186.3h / auto 327.2h
cue_events     oldest 2026-08-24   23,828 rows   51.0h completed
```

Eight months of turns against eight days of Cue. The score rendered **24%**, just under the first milestone, while the About card's own Cue Time counter says 203h - which is **31%**. And the gap widens forever, because the interactive side keeps growing while the Cue side stays capped at a week. A user who delegates *more* would watch the number fall.

**Fix (`5d2c945d7`):** the lifetime figure now takes its Cue half from `autoRunStats.cueTimeMs` - the accumulator the engine credits live and settings persists, which the prune can't reach. It takes the larger of the two values, since the accumulator floors each run to whole minutes and a young install can legitimately hold more on disk; the result is a lower bound either way. Only `durationMs` is substituted - there's no lifetime Cue run *count* to restore, and the count only feeds the range-scoped Queries mode.

Range-scoped views still read `cue_events`, which is correct inside the window and unrecoverable outside it. The trend chart now says which of its days can only show Auto Run, rather than drawing a Cue collapse that never happened.

Tests, docs, and the card's tooltip updated. 956 dashboard tests green.

PR #1493, opened 2026-09-03
Extract Document Graph canvas rendering to separate module

reachrazamair:
Move canvas painting functions (node rendering, icons, links) from MindMap.tsx to a new mindMapCanvas.ts module. This separates rendering logic from interaction handling and ensures the painted glyph and click hit test remain synchronized. Includes updating test imports.

## Summary by CodeRabbit

* **Refactor**
  * Reorganized Document Graph canvas rendering while preserving the existing visual experience.
  * Maintained consistent open-file icon positioning and clickable areas.
  * Preserved node styling, previews, search highlighting, links, and external-domain displays.
  * Improved preview text overflow handling to show a single ellipsis and respect selected font settings.

* **Documentation**
  * Clarified guidance for handling file references that include line numbers.


pedramamini:
@reachrazamair Thanks for this one, and thanks for sticking with it through the CodeRabbit round. This is a genuinely clean extraction and it holds up under mechanical verification.

## What I checked

I diffed the block removed from `MindMap.tsx` against the whole of the new `mindMapCanvas.ts` line by line. The only deltas are:

- the new file's header comment and imports
- `export` added to `NODE_BORDER_RADIUS`, `roundRect`, `drawLink`, `renderDocumentNode`, `renderExternalNode`
- section comments moved around (cosmetic)
- the two fixes from `b3d5e1b`

That is exactly the shape a pure extraction should have. Nothing drifted in the painting code.

**No runtime import cycle.** `mindMapCanvas.ts` reaches back for `MindMapNode` via `import type`, which is erased at compile, so the only runtime edge is `MindMap -> mindMapCanvas`. Good call keeping it type-only.

**The two `b3d5e1b` fixes are real pre-existing bugs, not refactor fallout.** Both were latent on `rc`:

- The overflow line really did paint six dots. `truncateText` appends its own `'...'` whenever the text exceeds `maxLength`, and `lastLine.length - 3` is always shorter than `lastLine`, so the trailing `+ '...'` doubled it. Worth noting the fix also preserves the total character count (`len - 3` chars plus 3 dots), so no reflow in the preview box.
- The pill path hardcoded the default stack, so the Document Graph font setting silently did nothing once previews were off. `fontFamily` was already a parameter with `DEFAULT_GRAPH_FONT` as its default, so threading it through is the intended behavior.

**The `SHARED-UTILS.md` edit is doing more than it looks like.** I initially read it as scope creep, then checked: `origin/rc` is currently failing `npm run lint:doc-refs` on exactly that line, because `src/foo.ts` does not resolve. So this change is unblocking `rc`, not just tidying prose. Appreciated.

## Validation

CI is green on the exact HEAD (`b3d5e1b`), both OS legs and both shards plus `lint-and-format`. Locally I also ran the DocumentGraph suite (763 tests across 20 files, all passing), ESLint on both changed source files (clean), and `check-doc-refs` (clean).

## Two optional nits, neither blocking

1. **`export { openIconRect } from './mindMapCanvas';` in `MindMap.tsx` now has zero consumers.** The test was repointed at the new module, and every in-file call site uses the direct import. Per the Dead Code Hygiene note in `CLAUDE.md` it is worth either dropping the line or leaving a comment saying it is a deliberate compatibility surface. Your call, and fine to leave as is if you prefer the stable import path.
2. **The doc fix could use the escape hatch instead.** `check-doc-refs.mjs` supports `<!-- doc-refs-ignore -->` above a deliberate placeholder, which would let the clearer original `src/foo.ts:42` example stand. Purely a readability preference, the current wording passes and reads fine.

Approving. Nice, surgical change, and the shared `openIconRect` staying the single source for both the painted glyph and the hit test is the right instinct.

PR #1497, opened 2026-09-03
fix: stop a passive tab-inventory sync from stealing the active agent

### LINKED ISSUE #1496: Bug: Background agent activity steals active tab focus (setActiveSessionId)
## Summary
Background agent activity steals active tab focus (setActiveSessionId)

## Environment
- Maestro version: 0.18.5-RC

## Steps to Reproduce
1. Run Maestro with multiple agents and background automation active (this machine: 89 agents, 10 concurrent `maestro-cli.js pianola watch <session> --agent <id> --interval 5` helper processes).
2. Select one agent and start reading or typing in it.
3. Let a background agent receive a message, or let a pianola/watch update or a remote command be delivered to a different session.
4. Observe: the visible tab switches away to the agent that received the activity, without any user navigation.

An additional path worth testing separately: trigger session recovery (the [SessionRecovery] code path) while a different tab is active, and observe whether the active tab changes.

## Expected Behavior
Passive/background activity must never change the visible desktop session or tab. When a background agent receives a message, a watch update arrives, or a remote command is delivered, Maestro should keep the user's current tab visible and mark the target agent as unread (or fire a passive OS notification).

Only explicit user intent should call setActiveSessionId: clicking an agent/tab, running an explicit focus/select command, or a remote command that carries an explicit focus intent.

The codebase already has the correct convention for this and it should be reused rather than reinvented: several call sites already gate the switch behind a background flag, e.g. `if (!background) { setActiveSessionId(newId); }` and `if (createdTabId && !background) { setActiveSessionId(sessionId); }`. The passive paths should be gated the same way, driven by the `focus?: boolean` intent that `setSelectSessionCallback` already receives.

## Actual Behavior
Incoming agent activity switches the visible desktop session/tab to the agent that received the new message, pulling the user out of whatever tab they were reading or typing in.

A first fix was applied locally (commit 9611cc4fe on branch build/indigo-deep-wine-red-themes, removing the automatic setActiveSessionId from the remote command listener in src/renderer/hooks/remote/useRemoteIntegration.ts) and passed build:renderer, build:web-desktop, the updated unit test, and lint — but the symptom still reproduces in the running packaged app. So the remote command listener is not the only focus-stealing path.

Inspection of the shipped bundle (see additionalContext) confirms at least three non-user-initiated setActiveSessionId call sites still present, including one on a [SessionRecovery] path that was not in the original list of suspects.

## Additional Context
VERIFIED ON THE REPORTER'S MACHINE (read-only diagnostics, 2026-09-03):

Runtime state
- `maestro-cli status`: Maestro running on port 64223 with 89 agents.
- Running app: /Applications/Maestro RC AGY 20260903.app, version 0.18.5-RC (CFBundleShortVersionString), app.asar mtime Sep 3 11:41 — i.e. a build made the same day as the report, so the local patch to useRemoteIntegration.ts may already be included and the symptom still persists.
- 10 live `maestro-cli.js pianola watch <sessionId> --agent <agentId> --interval 5` helper processes. Two of them are exact duplicates (sessions 49cbcfbd-d36e-4a9f-a26f-be65ae34fb6c and f305ac2d-9995-4898-a8f0-9b6f3f51b8f0 are each being watched twice by two processes) — possibly a separate leak, but it doubles the rate of background updates hitting the focus path.

Not a settings problem
- `maestro-cli settings list` has no focus/activation control. Notifications (osNotificationsEnabled=true, idleNotificationEnabled=false), tab placement (newTabPlacement, newBrowserTabPlacement), and automaticTabNamingEnabled exist, but nothing governs whether incoming activity changes the active session. The behavior is hardcoded, so there is no user-side workaround.

Bundle inspection (strings over Contents/Resources/app.asar of the running build)
- 114 occurrences of `setActiveSessionId` in the shipped bundle.
- THE LIKELY REASON THE LOCAL PATCH DID NOT CHANGE THE SYMPTOM: the session store is shipped TWICE, once per bundle (desktop renderer and web-desktop), with independent minified identifiers — both contain their own `setActiveSessionId: t => { set({ activeSessionId: t, cyclePosition: -1 }), ... }`. A fix has to hold across both build outputs; patching one hook that only lands in one bundle can leave the live path untouched.

Surviving non-user-initiated call sites found in the shipped build:
1. Remote select path, guarded only against redundancy, not against intent:
   `if (currentActiveId !== sessionId) { setActiveSessionId(sessionId); }` followed by a comment beginning "The leg..." (legacy). This matches the reporter's analysis that web-server-factory.ts accepts `focus?: boolean` in setSelectSessionCallback, uses it only to foreground the OS window, and forwards `remote:selectSession` to the renderer WITHOUT the focus intent — so onRemoteSelectSession switches unconditionally.
2. NEW — not in the original suspect list: a `[SessionRecovery]` path. The bundle shows `... .getState().setActiveSessionId(e)) }, [])` immediately preceding the `[SessionRecovery]` log tag, in an effect with an empty dependency array, and it appears in BOTH bundles. This is an automatic recovery flow changing the active session with no user action.
3. Worktree creation path: `useSessionStore.getState().setActiveSessionId(worktreeSession.id);` immediately followed by a `notifyToast(...)` — auto-selects a newly created worktree session.

Existing pattern to reuse for the fix (already in the bundle):
- `if (!background) { setActiveSessionId(newId); }`
- `if (createdTabId && !background) { setActiveSessionId(sessionId); }`
The `background` flag convention is already the project's way of separating passive creation from user-intent focus. Threading the existing `focus?: boolean` from setSelectSessionCallback through the `remote:selectSession` IPC payload and gating onRemoteSelectSession on it would make the passive paths consistent with that convention instead of introducing a new mechanism.

Out of scope but noted: the reporter's environment is a locally packaged RC build ("Maestro RC AGY 20260903"), user data dir /Users/<user>/Library/Application Support/maestro/rc-20260824-parallel-user-data. File logging is off on this machine, so no debug log was available to correlate the switch events; findings above come from process inspection, live settings, and the packaged bundle.

## Screenshots / Recordings
None

pedramamini:
closes #1496

## The path

`remote:selectTab` carries two different asks on one channel, and the listener in `useRemoteIntegration.ts` handled them alike:

- an explicit `select_tab` (no third argument) is a human tapping a tab in a remote client, and moving the view is the whole point;
- the legacy `tabs_changed` packet (third argument present) is a **passive** broadcast. `useRemoteIntegration`'s 500ms sweep re-emits it for *any* agent whose tab hash changes, and that hash is built from `state` and `hasUnread` - so a background agent merely starting a turn or picking up an unread fires one.

Both landed on `setActiveSessionId(sessionId)` + `aiTabFocusFields(tabId)`. That is exactly the reported symptom: background agent activity switches the visible agent and the tab inside it, with no way to tell which of ninety agents took the view.

This is also the call site the reporter fingerprinted out of the shipped bundle (`if (currentActiveId !== sessionId) { setActiveSessionId(sessionId); }` followed by a comment starting "The leg...").

## The fix

A passive sync reconciles data only, per the rule already written down in `src/shared/focusPlacement.ts`:

- it never changes the active agent;
- it follows the desktop's active tab only for the agent that is *already* on screen, where doing so cannot pull the user away from anything;
- the tab inventory still reconciles exactly as before, so nothing is lost from the remote client's tab list.

An explicit `select_tab` is untouched and still switches agent and activates the tab.

`src/web-desktop/electron-shim.ts` now drops a `tabs_changed` that carries no inventory instead of routing it onward. The inventory's presence is what the listener tells the two asks apart by, so an inventory-less `tabs_changed` would arrive looking like a selection and move the view.

## Branch base

Based on `rc`, not `main`: `src/web-desktop/` and the three-argument `onRemoteSelectTab` are both rc-only, so the defect does not exist on `main` and the fix cannot be expressed there.

## Not covered here

`maestro:refreshAutoRunDocs` in `useAppRemoteEventListeners.ts` switches the active agent purely to trigger a refresh effect, which is the same class of defect. It is **already fixed on `main`** (with a `background` field threaded through the verb) and will arrive on `rc` with the next merge, so it is deliberately left alone here to avoid a conflicting second fix.

## Validation

- `npm run lint` (all three tsconfigs) clean
- `npx eslint` clean, `prettier --check` clean
- `npx vitest run src/__tests__/renderer/hooks/ src/__tests__/main/preload/process/tabRemote.test.ts src/__tests__/integration/remote-control.integration.test.ts` - 243 files, 5620 passed
- two new tests in `useRemoteIntegration.test.ts`: a passive inventory sync for a non-active agent reconciles tabs without calling `setActiveSessionId` or moving `activeTabId`; an explicit selection with no inventory still switches both.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved remote tab synchronization so background tab updates no longer switch the currently viewed session or active tab.
  - Explicit remote tab selections now correctly switch to the selected session and tab.
  - Prevented incomplete tab synchronization messages from being treated as user selections.
  - Improved tab focus behavior to avoid applying focus changes when following a remote tab is not permitted.

- **Tests**
  - Added coverage for passive tab inventory updates and explicit remote tab selections.

### DIFF
--- a/src/__tests__/renderer/hooks/useRemoteIntegration.test.ts
+++ b/src/__tests__/renderer/hooks/useRemoteIntegration.test.ts
@@ -1004,6 +1004,59 @@ describe('useRemoteIntegration', () => {
 				{ type: 'ai', id: 'tab-2' },
 			]);
 		});
+
+		// Issue #1496: the `tabs_changed` broadcast fires on any agent's tab hash
+		// change (which includes `state` and `hasUnread`), so a background agent
+		// starting a turn used to drag the reader onto it.
+		it('does not switch agents for a passive tab inventory sync', () => {
+			const session = createMockSession({
+				id: 'session-1',
+				aiTabs: [createMockTab({ id: 'tab-1' }), createMockTab({ id: 'tab-2' })],
+				activeTabId: 'tab-1',
+			});
+			const deps = createDeps({ sessions: [session], activeSessionId: 'other-session' });
+
+			renderHook(() => useRemoteIntegration(deps));
+
+			act(() => {
+				onRemoteSelectTabHandler?.('session-1', 'tab-2', [
+					{
+						id: 'tab-2',
+						agentSessionId: null,
+						name: 'Busy background tab',
+						starred: false,
+						inputValue: '',
+						createdAt: 1700000001000,
+						state: 'busy',
+					},
+				]);
+			});
+
+			expect(deps.setActiveSessionId).not.toHaveBeenCalled();
+			// The inventory still reconciles - it is only the view that stays put.
+			const updated = useSessionStore.getState().sessions.find((s) => s.id === 'session-1');
+			expect(updated?.aiTabs.map((tab) => tab.id)).toEqual(['tab-2']);
+			expect(updated?.activeTabId).toBe('tab-1');
+		});
+
+		it('still follows an explicit tab selection with no inventory attached', () => {
+			const session = createMockSession({
+				id: 'session-1',
+				aiTabs: [createMockTab({ id: 'tab-1' }), createMockTab({ id: 'tab-2' })],
+				activeTabId: 'tab-1',
+			});
+			const deps = createDeps({ sessions: [session], activeSessionId: 'other-session' });
+
+			renderHook(() => useRemoteIntegration(deps));
+
+			act(() => {
+				onRemoteSelectTabHandler?.('session-1', 'tab-2');
+			});
+
+			expect(deps.setActiveSessionId).toHaveBeenCalledWith('session-1');
+			const updated = useSessionStore.getState().sessions.find((s) => s.id === 'session-1');
+			expect(updated?.activeTabId).toBe('tab-2');
+		});
 	});
 
 	describe('remote new tab', () => {
--- a/src/renderer/hooks/remote/useRemoteIntegration.ts
+++ b/src/renderer/hooks/remote/useRemoteIntegration.ts
@@ -459,19 +459,37 @@ export function useRemoteIntegration(deps: UseRemoteIntegrationDeps): UseRemoteI
 		);
 
 		// Handle remote tab selection from web interface
-		// This also switches to the session if not already active
+		//
+		// This channel carries TWO different asks and they must not be treated
+		// alike (see `src/shared/focusPlacement.ts` for the rule):
+		//
+		//   - an explicit `select_tab` (no third argument) is a human tapping a
+		//     tab in a remote client, so it moves the view: switch agent, then
+		//     activate the tab.
+		//   - the legacy `tabs_changed` packet (third argument present) is a
+		//     PASSIVE inventory broadcast. The desktop re-broadcasts it whenever
+		//     any agent's tab hash changes, and that hash includes `state` and
+		//     `hasUnread` - so a background agent merely starting a turn or
+		//     picking up an unread emits one. Treating it as a selection is what
+		//     let background activity yank the reader onto whichever of ninety
+		//     agents happened to move (issue #1496).
+		//
+		// A passive sync therefore reconciles data only. It never changes the
+		// active agent, and it only follows the desktop's active tab for the
+		// agent the user is already looking at, where doing so cannot pull them
+		// away from anything.
 		const unsubscribeSelectTab = window.maestro.process.onRemoteSelectTab(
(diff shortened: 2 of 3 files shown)


pedramamini:
Superseded by #1513, which landed the same fix in a fuller form (main-process broadcast, preload, keyboard handler and desktop selection sync, not just the renderer hook and the web-desktop shim). `activeTabChanged` is already threaded through on `rc`, so this branch now conflicts with its own fix.

PR #1500, opened 2026-09-04
Gate native OMP client on public protocol conformance

jSydorowicz21:
## Problem

Maestro cannot safely build the native OMP run/fleet experience until the user-installed OMP public protocol proves the required authority, recovery, steering, approval, model, worker, evidence, and error semantics.

## Change

- Add a pinned OMP 18.1.6 RPC v2 conformance harness and repeatable baseline.
- Add strict JSONL/RPC-v2 framing with UTF-8, size, chunk ordering, response-correlation, and ready-frame guards.
- Compare the advertised ACP, RPC, and RPC-UI surfaces and select RPC for OMP-specific host control.
- Exercise session reattachment, snapshots/paging, live streaming, worker registry/transcript, steering/cancel, approvals, models/providers, todos, host-tool callbacks, process effects, export, and typed errors.
- Pin the exact user-installed binary version and SHA-256; fail before live probing on identity drift.
- Emit a privacy-redacted observed matrix and a machine-readable upstream requirements contract.

## Gate result

The gate correctly fails against OMP 18.1.6. Blocking cases: A04 and A05. Blocking capabilities: G1, G2, G3, G5.

- `create_worker` is absent.
- `command_worker` is absent.
- `get_subagent_messages` does not expose the completed Worker message after terminal root completion.
- `abort_and_prompt` acknowledges acceptance without a correlated completion signal.

Per the implementation blueprint, no native adapter/UI is included until these public-protocol blockers are resolved. No private OMP files or implementation APIs are used.

## Verification

- `npm run test:omp-protocol` — 34 passed
- Targeted ESLint — passed
- Targeted Prettier — passed
- Live pinned OMP 18.1.6 matrix — expected exit 1 with the blocker set above
- Observed artifact path-redaction scan — passed

The repository-wide pre-push `format:check:all` remains red on 82 pre-existing untouched files; this branch's files pass targeted formatting.

## Summary by CodeRabbit

* **New Features**
  * Added automated OpenModelProtocol RPC conformance checks covering negotiation, messaging, workers, approvals, cancellation, and error handling.
  * Added protocol framing support with chunk reassembly, transport-limit validation, readiness checks, and sanitized reporting.
  * Added baseline, observed-behavior, and upstream-requirement fixtures for protocol version 18.1.6.

* **Tests**
  * Added unit and integration coverage for protocol decoding, validation, redaction, argument handling, and conformance verdicts.
  * Added commands to run the protocol test suite and execute a conformance probe.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,8 @@
 		"lint": "tsc -p tsconfig.lint.json && tsc -p tsconfig.main.json --noEmit && tsc -p tsconfig.cli.json --noEmit",
 		"lint:eslint": "eslint src/ && eslint -c eslint.dashes.config.mjs src/__tests__ scripts",
 		"lint:doc-refs": "node scripts/check-doc-refs.mjs",
+		"test:omp-protocol": "node --test scripts/omp-rpc-protocol.test.mjs scripts/omp-rpc-conformance.test.mjs",
+		"probe:omp-protocol": "node scripts/omp-rpc-conformance.mjs",
 		"format": "prettier --write \"src/**/*.{ts,tsx}\"",
 		"format:all": "prettier --write .",
 		"format:check": "prettier --check \"src/**/*.{ts,tsx}\"",
--- a/scripts/fixtures/omp-rpc-18.1.6-baseline.json
+++ b/scripts/fixtures/omp-rpc-18.1.6-baseline.json
@@ -1,0 +1,111 @@
+{
+	"schemaVersion": 1,
+	"omp": {
+		"version": "18.1.6",
+		"sha256": "99a824d60aa93f28a4d3e863acc7d47ffb5be51a1653f0d12aec1af4806b4b76",
+		"protocol": "rpc",
+		"startupProtocolVersion": 1,
+		"supportedProtocolVersions": [1, 2]
+	},
+	"transport": {
+		"serverMaxFrameBytes": 1048576,
+		"serverMaxReassembledFrameBytes": 67108864,
+		"maestroMaxFrameBytes": 1048576,
+		"maestroMaxReassembledFrameBytes": 8388608
+	},
+	"requiredCommandTypes": [
+		"prompt",
+		"steer",
+		"follow_up",
+		"abort",
+		"abort_and_prompt",
+		"new_session",
+		"negotiate_protocol",
+		"get_state",
+		"set_fast_mode",
+		"get_available_commands",
+		"set_todos",
+		"set_host_tools",
+		"set_host_uri_schemes",
+		"set_subagent_subscription",
+		"get_subagents",
+		"get_subagent_messages",
+		"set_model",
+		"cycle_model",
+		"get_available_models",
+		"set_thinking_level",
+		"cycle_thinking_level",
+		"set_steering_mode",
+		"set_follow_up_mode",
+		"set_interrupt_mode",
+		"compact",
+		"set_auto_compaction",
+		"set_auto_retry",
+		"abort_retry",
+		"bash",
+		"abort_bash",
+		"get_session_stats",
+		"export_html",
+		"switch_session",
+		"branch",
+		"get_branch_messages",
+		"get_last_assistant_text",
+		"set_session_name",
+		"handoff",
+		"get_messages",
+		"get_messages_page",
+		"get_login_providers",
+		"login"
+	],
+	"requiredEventTypes": [
+		"agent_start",
+		"agent_end",
+		"turn_start",
+		"turn_end",
+		"message_start",
+		"message_update",
+		"message_end",
+		"tool_execution_start",
+		"tool_execution_update",
+		"tool_execution_end",
+		"model_changed",
+		"thinking_level_changed",
+		"goal_updated",
+		"available_commands_update",
+		"prompt_result",
+		"extension_ui_request",
+		"host_tool_call",
+		"host_tool_cancel"
+	],
+	"requiredCases": ["A01", "A02", "A03", "A04", "A05", "A06", "A07", "A08", "A09", "A10", "A11"],
+	"powerUserCapabilities": [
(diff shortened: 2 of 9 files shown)


pedramamini:
@jSydorowicz21 thanks for this - it is a genuinely useful contribution. Gating the native OMP work on a reproducible, pinned conformance probe rather than building an adapter against assumed semantics is exactly the right call, and shipping the harness with the failing verdict (A04/A05, G1/G2/G3/G5) instead of a half-built adapter is the discipline we want. The framing layer is careful work: bounded JSONL, fatal UTF-8 decoding, chunk ordering, response correlation, and privacy-redacted output are all handled deliberately.

Before we merge, there are a few things to address. I verified each of these against the branch head (`5c921a3`).

## Blocking

**1. One test fails on macOS and Linux.** I ran the suite locally on macOS: **33 pass, 1 fail**.

```
✖ parseConformanceArgs requires an explicit absolute OMP executable
  Error: --executable must be an absolute path to the user-installed OMP binary
      at parseConformanceArgs (scripts/omp-rpc-conformance-lib.mjs:60:9)
      at scripts/omp-rpc-conformance.test.mjs:111:3
```

`scripts/omp-rpc-conformance.test.mjs:107-127` uses `C:\Tools\omp.exe` and `C:\Temp\report.json`. Under POSIX, `path.isAbsolute('C:\\Tools\\omp.exe')` is `false`, so the parser throws instead of returning. Build those fixtures from the running platform. This matches CodeRabbit's finding. Same fix applies to the `--surprise` case at line 130.

**2. The suite is not wired into any gate, which is why (1) went unnoticed.** `vitest.config.mts` scopes both projects to `src/**`, and CI (`.github/workflows/ci.yml:48`) and `validate:push` both run only `npm run test`. So `scripts/*.test.mjs` is collected by nothing. Please add `test:omp-protocol` to `validate:push` and to the CI test job. A contract suite that only runs when someone remembers to type the command is a contract suite that drifts.

**3. Ready-frame handling drops buffered stdout bytes** (`scripts/omp-rpc-protocol.mjs:475`). `#dispatch` replaces `#lineDecoder` with a fresh instance, but `JsonLineDecoder` keeps a trailing partial line in its private `#buffer`, and that prefix is discarded. The trigger is one stdout read spanning the ready line plus the head of the next frame, which is ordinary pipe behavior. The tail then parses as a fresh line, `parseJsonObject` throws, `#receive` terminates the peer and kills the child, and every probe on that peer fails with a spurious protocol error. That is the worst failure mode for this tool specifically: the verdict stops describing OMP. CodeRabbit's suggested fix (a `setMaxFrameBytes` setter applied to the existing instance) is the right shape. The `#chunkDecoder` replacement on the same lines is fine, since ready is the first logical frame.

## Should fix

**4. A05 accepts an uncorrelated completion** (`scripts/omp-rpc-conformance.mjs:612-635`). The `terminal` waiter is registered before `abort_and_prompt` is sent and matches any `agent_end` with `isTerminal !== false`, so it can settle on the *aborted* turn's terminal event. `completionSignal` then reduces to `typeof abortAndPrompt.data?.agentInvoked === 'boolean'`, which is an acknowledgement field, not a completion. A replacement turn that never terminates can satisfy a required case. A05 fails today anyway, so nothing is currently misreported, but the whole value of this harness is the go/no-go decision it produces, and a false pass is the expensive direction to be wrong in. Please correlate the terminal event to the replacement turn. Greptile flagged the same thing.

**5. No `error` listener on the child stdio streams** (`scripts/omp-rpc-protocol.mjs:361-366`). Only the child process itself has one. `send`, `sendRaw`, and `close` write to or end `#child.stdin`, and the `stdin.writable` check is a TOCTOU window: an `EPIPE` or `ERR_STREAM_DESTROYED` emitted on the stream with no listener throws and takes down the harness rather than failing the one probe. `onFrame` writes from inside dispatch, so this runs precisely when the child may be exiting. Route stream errors into `#terminate`.

**6. `G3` is missing from the fatal-path blocking list** (`scripts/omp-rpc-conformance.mjs:246`). `computeConformanceVerdict` blocks on `['G1', 'G2', 'G3', 'G5']` (`omp-rpc-conformance-lib.mjs:207`) but the `HARNESS_FATAL` branch writes `['G1', 'G2', 'G5']`, so a crash emits a report that understates what is blocked. One-character fix.

## Worth a look

**7. A10 conflates three outcomes.** `shared.busyErrorCode` and `shared.staleErrorCode` stay `null` when the probe never ran, when it ran and succeeded, and are only meaningful when it ran and failed. A10 reports all three as `received null`. It fails safe, so this is diagnosability rather than correctness, but a gate whose job is to tell us *why* we are blocked should distinguish "the probe could not run" from "OMP returned the wrong code."

**8. The SHA-256 pin makes this a single-machine harness.** `baseline.omp.sha256` is one scalar, and `comparePinnedIdentity` refuses to probe on drift. That is the correct instinct for evidence integrity, but it means nobody on a different platform (or with a different official 18.1.6 build) can reproduce the matrix, and the PR describes this as a repeatable baseline. Is the intent a per-platform digest map, or a documented "re-pin before re-running" step? Either is fine, but it should be a decision rather than an accident, and worth a line in the fixture or a README so the next person is not stuck at the identity gate.

## Not your problem

The red `lint-and-format` check is pre-existing on `rc`, not caused by this branch. It is `lint:doc-refs` failing on `docs/agent-guides/SHARED-UTILS.md:408` (`src/foo.ts` placeholder); `rc` itself is red on the same job. Please do not chase it. Both Ubuntu and Windows test legs are green, and there are no merge conflicts.

One small style note for the future: this repo forbids em dashes and en dashes everywhere, including comments and commit messages. Your diff is clean on that, so nothing to change here.

Happy to re-review as soon as 1 through 3 are in. Thanks again for taking the rigorous path on this one.

Issue #1470, opened 2026-09-01
Bug: AutoRun lost from UI after Web-Desktop browser reload; next task neve...


chr1syy:
## Summary
AutoRun lost from UI after Web-Desktop browser reload; next task never starts

## Environment
- Maestro version: 0.18.5-RC
- Operating system: Linux (#35~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue May 26 19:30:42 UTC 2, 6.17.0-35-generic)
- Install source: Packaged build (release build or locally packaged)
- Agent/provider involved: Not provided
- SSH remote execution: Not provided

## Steps to Reproduce
1. Open Maestro in Web-Desktop (browser).
2. Start an AutoRun on a worktree/agent.
3. Unfocus the browser window and do something else.
4. Return to the browser window — it sometimes reloads.
5. Observe: the worktree/agent no longer shows the AutoRun as running, though the process is still active and finishes the current task. No next task is started afterward.

## Expected Behavior
After the browser reloads, the UI should re-attach to the still-running AutoRun and continue showing it as active on the worktree/agent. The AutoRun should proceed to queue and start the next task once the current one completes.

## Actual Behavior
After the reload, the AutoRun disappears from the UI on that worktree/agent. The underlying process is still alive and completes the current task, but no next task is started. The AutoRun is permanently lost from the UI — refreshing again or switching away from and back to the worktree does not recover it; it must be restarted manually.

## Additional Context
Intermittent — the reload on refocus happens 'sometimes'. Suggests the AutoRun session state isn't rehydrated/re-attached on page load, so the client-side orchestration that queues the next task is lost while the backend process continues. Maestro 0.18.5-RC on Linux, packaged build.

## Screenshots / Recordings
None

pedramamini:
Thanks for this one, and specifically for noting that the process stays alive and finishes the current task. That detail is what separates this from the reload bug we fixed an hour before you filed, and it is what made it diagnosable.

Labeled `bug`. Keeping it open and **not** closing it against #1464 - they look like the same bug and are not.

## Why this is not a duplicate of #1464

#1465 (merged just before you filed) restores agent busy indicators after a web-desktop reload by asking main's process table what is actually running. It deliberately only accepts processes whose id parses as `{agentId}-ai-{tabId}`. Auto Run tasks spawn as `{agentId}-batch-{timestamp}`, so that fix does not reach your case at all.

## Root cause

[...]

PR #1508, opened 2026-09-04
fix(web-desktop): mirror Auto Run state so a running batch is visible in the browser

### LINKED ISSUE #1505: Bug: AutoRun started by agent via CLI not shown in web interface
## Summary
AutoRun started by agent via CLI not shown in web interface

## Environment
- Maestro version: 0.18.5-RC

## Steps to Reproduce
1. Enable the persistent web link and open the Maestro web interface.
2. Have an agent trigger an AutoRun itself via maestro-cli (agent-initiated, not started from the desktop UI).
3. Observe the desktop app: AutoRun runs, banner and output visible.
4. Observe the same agent in the web interface: shows idle, no AutoRun banner, no output.

## Expected Behavior
When an agent starts an AutoRun itself via the CLI, the web interface should reflect the same state as the desktop app: the AutoRun banner/indicator should appear and the agent should show as busy with streaming output.

## Actual Behavior
The AutoRun starts and runs normally in the desktop app, but in the web interface the agent appears completely idle — no AutoRun banner is shown and no streaming output appears.

## Additional Context
Maestro 0.18.5-RC on Linux (Ubuntu, kernel 6.17.0-35-generic), packaged build. Diagnostics confirmed the app is running with the web interface served on port 41713, persistent web link enabled, web auth off, AutoRun enabled, and all Encore Features on — so no misconfiguration on the reporter's side. Because the web client shows the agent as fully idle rather than just missing the banner graphic, this looks like the agent-initiated (CLI-triggered) AutoRun start event is not being broadcast to web clients at all, rather than a banner-rendering bug. Worth checking whether AutoRuns started from the desktop UI do propagate to web, which would isolate the gap to the CLI/agent-initiated path. File logging is off, so no current debug log is available.

## Screenshots / Recordings
None

pedramamini:
closes #1505

## The report

An Auto Run started by an agent via `maestro-cli` runs normally in the desktop app - banner, progress, output - but the same agent in the web interface shows as completely idle: no banner, no output.

## What is actually broken

The CLI is a red herring. **No** Auto Run is visible in a web-desktop browser client, however it was started. A run launched from the desktop Auto Run modal is equally invisible. The CLI path is just the one where nobody has the desktop window in front of them, so it is where you notice.

Auto Run is renderer-owned state. The run loop is a live async closure, its document/task cursors are `let` bindings inside it, and `batchRunStates` (`src/renderer/stores/batchStore.ts`) is a plain in-memory zustand store with no persistence. All of it lives in whichever client started the run - for a CLI launch, the desktop window that main forwarded `remote:configureAutoRun` to. Nothing about a run is written to the session, to settings, or to disk.

A web-desktop client is a second full renderer over the WebSocket bridge with its own store, so that store stayed empty for the whole run, and every Auto Run surface reads it: the Left Bar batch pill, the thinking pill, the Right Panel active-run card, the editor lock.

The state was already on the wire. The owner pushes it to main on every progress tick (`web:broadcastAutoRunState`), main fans it out as an `autorun_state` packet (`broadcastService`) and replays every live run to each client on connect (`wsRoute`). The mobile web app consumes it. The web-desktop shim's frame router (`src/web-desktop/electron-shim.ts`) only understood `bridge.response`, `bridge.event`, `active_session_changed` and `tabs_changed` - `autorun_state` fell through the chain and was silently discarded. The payload was arriving at the browser tab and being thrown away.

## The change

- **`electron-shim.ts`** routes `autorun_state` onto a `remote:autoRunStateMirror` channel, the same way `active_session_changed` already becomes `remote:selectSession`.
- **`useAutoRunStateMirror`** (new) consumes it and hydrates `batchRunStates` with an entry stamped `mirrored: true`. Every Auto Run surface lights up at once, including the connect-time replay, so a tab opened mid-run catches up rather than waiting for the next tick.
- **`shared/autoRunBroadcast.ts`** (new) collapses the three hand-written copies of the broadcast shape (renderer `global.d.ts`, the `web:broadcastAutoRunState` handler, `web-server/types.ts`) into one type, and widens it with what the full desktop card needs beyond the mobile progress bar: `documents`, `lockedDocuments`, per-document counts, worktree, `startTime`, loop state. All optional and additive, so the mobile client is unaffected.

## The mirror is read-only, deliberately

Claiming or steering another client's run needs a renderer liveness/ownership primitive that does not exist today - the same one #1470 is blocked on. Without it, two clients driving one agent spawn duplicate tasks into the same working tree. Displaying is safe, so displaying is all this does:

- Every mutator refuses a mirrored entry: `stopBatchRun`, `killBatchRun`, `pauseBatchOnError`, `skipCurrentDocument`, `resumeAfterError`, `abortBatchOnError`, plus `startBatchRun` and `useAutoResumeCoordinator`.
- The controls that call them render disabled with one shared tooltip explaining why. A Stop button that quietly did nothing would be worse than no Stop button.
- `useAgentErrorListener` treats a mirrored batch as no batch. `agent:error` fans out to every client, so without this the mirroring client would also schedule a batch resume it cannot perform and write a duplicate persisted history entry.

## Two things worth reviewing closely

**Ownership is liveness, not presence.** `COMPLETE_BATCH` resets its entry in place rather than deleting the key, so any client that has run Auto Run once keeps a non-mirrored `isRunning: false` entry for that agent forever. Testing "an entry exists" as ownership would have made mirroring permanently dead for exactly the agents the user had already used it on - the bug would come back for the reporter after his first run. Covered by tests.

**Stale mirrors are reaped.** Nothing tells main that a renderer holding a live run went away, so a desktop reload or a closed browser tab mid-run leaves `isRunning: true` stored in `LiveSessionManager` forever and replayed to every client that connects afterwards. That leak is pre-existing (it already strands the mobile client), but a mirror makes it worse: the card cannot be dismissed from a mirroring client, and `useInputProcessing` would queue the user's messages behind a run that never drains. So a mirror is dropped only when it has gone quiet **and** main reports no `-batch-` process for that agent. A failed probe means "could not find out" and drops nothing. This is a containment, not the liveness protocol #1470 needs.

## Targeting `rc`, not `main`

`src/web-desktop/` does not exist on `main` - the whole web-desktop feature is rc-only.

## Validation

- `npm run lint` clean (all three tsconfigs).
- `npx eslint src/` - only the 5 pre-existing warnings on `rc`, all in files this PR does not touch.
- `npm run test` - 41,051 passed, 0 failures.
- 26 new unit tests for the mirror (ownership, echo suppression, teardown, the reaper) and 3 for the shim routing, plus one locking the broadcast fields the mirror depends on.

Local runs are single-OS; both CI matrix legs still need to be green.

## Summary by CodeRabbit

- **New Features**
  - Auto Run activity can now be mirrored across Maestro windows, including progress, document, worktree, timing, loop, and error details.
  - Mirrored runs are displayed as read-only, with controls disabled and clear explanations.
  - Auto Run starts are coordinated to prevent duplicate runs across windows.

- **Bug Fixes**
  - Improved delivery and replay of Auto Run updates for web-desktop clients.
  - Prevented mirrored runs from being resumed, retried, stopped, or killed locally.
  - Improved cleanup of stale mirrored run entries.


chr1syy:
Addressed all outstanding review findings in `83cfefbe1`:

- preserve mirrored error/HITL pauses during stale-state cleanup
- guard stale cleanup against frames arriving during the async process probe
- add a main-process atomic Auto Run start claim for cross-client races (batch and goal routes)
- re-check mirrored ownership after the auto-resume availability probe
- replay the latest pre-subscription mirror frame in the web Electron shim
- keep disabled Stop controls hoverable so their explanatory tooltips remain reachable
- merge the latest `rc` and fix its doc-reference lint example

The browser's live Auto Run output continues through the existing `safeSend` generic bridge (`process:thinking-chunk` / `process:tool-execution`) into the Thought Stream; this change supplies the missing mirrored running state so the browser exposes that surface while desktop owns the batch.

Validation:
- focused Auto Run/IPC/shim suites: 7 files, 332 tests passed
- goal runner suite: 20 tests passed
- TypeScript (renderer/main/CLI): passed
- ESLint: 0 errors (5 pre-existing `rc` warnings)
- doc-reference lint and formatting: passed
- full suite reached one unrelated load-sensitive timeout; that exact test passed alone in 1.61s

chr1syy:
Addressed the two follow-up review findings in 030671423:

- The main-process Auto Run claim now happens before worktree setup and task-file validation. Failed preparation releases only the still-provisional claim, so a delayed rollback cannot clear a published run.
- Buffered Auto Run replay now re-checks listener registration between frames, preserving undelivered frames when a once listener unregisters itself.
- Added tracker, IPC/preload, runner, and web bridge regressions. Focused validation: 296 tests passed; all TypeScript targets, ESLint, Prettier, and doc-reference checks passed.

chr1syy:
Final verification on d71b320fa is complete:

- lint-and-format: passed
- Ubuntu test shards 1/2 and 2/2: passed
- Windows test shards 1/2 and 2/2: passed
- CodeRabbit: passed
- All 10 review threads: resolved
- GitHub merge state: CLEAN / MERGEABLE

The last CI-only failure was traced to a TerminalOutput test fixture that changed mocked tabs without changing the aiTabs memo dependency; the fixture and observer/rAF timing are now deterministic, and all 172 tests in that file pass locally.

PR #1509, opened 2026-09-04
Refactor/safe localstorage migration

reachrazamair:
## Summary by CodeRabbit

* **Bug Fixes**
  * Added time limits to external document and GitHub requests, reducing the risk of network operations hanging indefinitely.
  * Improved resilience when saving and restoring display preferences, history filters, process expansion, editor layout, scale, and Prompt Composer settings in restricted or quota-limited storage environments.
  * Preferences and in-memory controls now continue working when browser storage operations fail.

* **Tests**
  * Updated automated coverage for request timeouts and storage failures across Symphony workflows and persisted UI preferences.


pedramamini:
@reachrazamair Thanks for picking this one up. Two deferred cleanups knocked out in one pass, with the DEDUP-TRACKER context and the reasoning for each timeout budget written into the commit messages, is exactly how we like to see this kind of consolidation land. The per-domain constants in `symphony/shared.ts` follow the `marketplace-service.ts` / `detector.ts` convention correctly, and updating the four call-signature assertions to `expect.objectContaining({ signal: expect.anything() })` matches what `leaderboard.test.ts` already does. Good work.

There is one thing I'd like addressed before merge, and one reviewer finding I'm explicitly waiving so you don't spend time on it.

## Please fix: the storage migration drops guards that were there before

Both Greptile and CodeRabbit flagged this and they're right. `safeLocalStorage()` only guards **reading the global**:

```ts
export function safeLocalStorage(): Storage | null {
	try {
		return typeof localStorage === 'undefined' ? null : localStorage;
	} catch {
		return null;
	}
}
```

Once it hands back a real `Storage`, a throwing `getItem`/`setItem` escapes. `setItem` is the realistic one: `QuotaExceededError` on a full origin, and Safari private mode historically threw there too.

Four of the seven files had method-level try/catch and lose it in this PR:

- `GitDiffViewer.tsx` (`readStoredViewType` / `writeStoredViewType`)
- `GitLogViewer.tsx` (the `useState` initializer and the persist `useEffect`)
- `useProcessExpansion.ts` (`readStoredExpandedLevel` / `writeStoredExpandedLevel`)
- `modalStore.ts` (`readStoredPromptComposerFullscreen` / `writeStoredPromptComposerFullscreen`)

`historyFilterPersistence.ts` keeps its outer try/catch, so that one is fine.

Two of these are worse than a lost preference. `GitLogViewer`'s write is inside a `useEffect`, so a throw during commit reaches the nearest error boundary and takes the Git Log pane down. `modalStore`'s read runs at store initialization. That contradicts the contract the helper's own doc comment states: "a missing or hostile Storage costs the user their persistence, never their pane."

The fix I'd prefer is **hardening the shared helper rather than re-adding try/catch at seven call sites**, since that's the whole point of the consolidation. Something like a `safeStorageGet(key)` / `safeStorageSet(key, value)` pair in `src/renderer/utils/safeLocalStorage.ts` that swallows method-level failures, with the callers here routed through it. That also fixes the three callers that already had this gap before your PR (`usePersistedToggle`, `usePersistedChoice`, `useScalePreference`), which is a nice bonus and keeps one pattern in the codebase instead of two.

Worth extending `src/__tests__/renderer/utils/safeLocalStorage.test.ts` too. Its current "storage blocked" case only makes the **getter** throw, so the `getItem`/`setItem` path is untested. A case where the global resolves fine but `setItem` throws `QuotaExceededError` would lock this down.

Keep `safeLocalStorage()` itself exported either way. Callers that legitimately need the raw `Storage` (`removeItem`, `key`, enumeration) still want it.

## Waiving: the unbounded response body finding

Both bots raised that `fetchWithTimeout`'s budget covers response headers and not `.json()` / `.text()` / `.arrayBuffer()`. That's true, but it is a documented, deliberate property of the canonical helper:

> The timeout covers the response headers, not the body. A caller that reads a large body with `.json()` or `.text()` can still block after this resolves.

Every other `fetch()` in `src/main` has the same characteristic. Solving it belongs in `fetchWithTimeout` as its own change, benefiting all callers at once, not bolted onto the twelve sites you happen to be migrating. Your change is a strict improvement over bare `fetch()` here. Please leave it as is.

## CI: red, but not yours

Don't chase these, all three are pre-existing on `rc`:

- **`lint-and-format`** fails on `npm run lint:doc-refs`, which flags a `src/foo.ts` placeholder at `docs/agent-guides/SHARED-UTILS.md:410`. I reproduced this on a clean checkout of `60ba7c01b`, your exact merge base. A fix is already in flight on our side.
- **`Test (windows-latest, shard 1/2)`** fails with 8 cases in `src/__tests__/main/utils/file-tree-walk.test.ts`, unrelated to anything you touched, and failing identically on the baseline run.
- The baseline `rc` run also had `ubuntu shard 1/2` red. Both Ubuntu shards are **green** on your PR, so you're ahead of the branch you're merging into.

No merge conflicts, and you're already current with the `rc` tip, so no rebase needed.

Push the storage hardening and I'll take another look. Thanks again for the contribution.

reachrazamair:
> @reachrazamair Thanks for picking this one up. Two deferred cleanups knocked out in one pass, with the DEDUP-TRACKER context and the reasoning for each timeout budget written into the commit messages, is exactly how we like to see this kind of consolidation land. The per-domain constants in `symphony/shared.ts` follow the `marketplace-service.ts` / `detector.ts` convention correctly, and updating the four call-signature assertions to `expect.objectContaining({ signal: expect.anything() })` matches what `leaderboard.test.ts` already does. Good work.
> 
> There is one thing I'd like addressed before merge, and one reviewer finding I'm explicitly waiving so you don't spend time on it.
> 
> ## Please fix: the storage migration drops guards that were there before
> Both Greptile and CodeRabbit flagged this and they're right. `safeLocalStorage()` only guards **reading the global**:
> 
> ```ts
> export function safeLocalStorage(): Storage | null {
> 	try {
> 		return typeof localStorage === 'undefined' ? null : localStorage;
> 	} catch {
> 		return null;
> 	}
> }
> ```
> 
> Once it hands back a real `Storage`, a throwing `getItem`/`setItem` escapes. `setItem` is the realistic one: `QuotaExceededError` on a full origin, and Safari private mode historically threw there too.
> 
> Four of the seven files had method-level try/catch and lose it in this PR:
> 
> * `GitDiffViewer.tsx` (`readStoredViewType` / `writeStoredViewType`)
> * `GitLogViewer.tsx` (the `useState` initializer and the persist `useEffect`)
> * `useProcessExpansion.ts` (`readStoredExpandedLevel` / `writeStoredExpandedLevel`)
> * `modalStore.ts` (`readStoredPromptComposerFullscreen` / `writeStoredPromptComposerFullscreen`)
> 
> `historyFilterPersistence.ts` keeps its outer try/catch, so that one is fine.
> 
> Two of these are worse than a lost preference. `GitLogViewer`'s write is inside a `useEffect`, so a throw during commit reaches the nearest error boundary and takes the Git Log pane down. `modalStore`'s read runs at store initialization. That contradicts the contract the helper's own doc comment states: "a missing or hostile Storage costs the user their persistence, never their pane."
> 
> The fix I'd prefer is **hardening the shared helper rather than re-adding try/catch at seven call sites**, since that's the whole point of the consolidation. Something like a `safeStorageGet(key)` / `safeStorageSet(key, value)` pair in `src/renderer/utils/safeLocalStorage.ts` that swallows method-level failures, with the callers here routed through it. That also fixes the three callers that already had this gap before your PR (`usePersistedToggle`, `usePersistedChoice`, `useScalePreference`), which is a nice bonus and keeps one pattern in the codebase instead of two.
> 
> Worth extending `src/__tests__/renderer/utils/safeLocalStorage.test.ts` too. Its current "storage blocked" case only makes the **getter** throw, so the `getItem`/`setItem` path is untested. A case where the global resolves fine but `setItem` throws `QuotaExceededError` would lock this down.
> 
> Keep `safeLocalStorage()` itself exported either way. Callers that legitimately need the raw `Storage` (`removeItem`, `key`, enumeration) still want it.
> 
> ## Waiving: the unbounded response body finding
> Both bots raised that `fetchWithTimeout`'s budget covers response headers and not `.json()` / `.text()` / `.arrayBuffer()`. That's true, but it is a documented, deliberate property of the canonical helper:
> 
> > The timeout covers the response headers, not the body. A caller that reads a large body with `.json()` or `.text()` can still block after this resolves.
> 
> Every other `fetch()` in `src/main` has the same characteristic. Solving it belongs in `fetchWithTimeout` as its own change, benefiting all callers at once, not bolted onto the twelve sites you happen to be migrating. Your change is a strict improvement over bare `fetch()` here. Please leave it as is.
> 
> ## CI: red, but not yours
> Don't chase these, all three are pre-existing on `rc`:
> 
> * **`lint-and-format`** fails on `npm run lint:doc-refs`, which flags a `src/foo.ts` placeholder at `docs/agent-guides/SHARED-UTILS.md:410`. I reproduced this on a clean checkout of `60ba7c01b`, your exact merge base. A fix is already in flight on our side.
> * **`Test (windows-latest, shard 1/2)`** fails with 8 cases in `src/__tests__/main/utils/file-tree-walk.test.ts`, unrelated to anything you touched, and failing identically on the baseline run.
> * The baseline `rc` run also had `ubuntu shard 1/2` red. Both Ubuntu shards are **green** on your PR, so you're ahead of the branch you're merging into.
> 
> No merge conflicts, and you're already current with the `rc` tip, so no rebase needed.
> 
> Push the storage hardening and I'll take another look. Thanks again for the contribution.

Addressed the storage hardening.

Added `safeStorageGet` / `safeStorageSet` in `src/renderer/utils/safeLocalStorage.ts`. They swallow method-level failures (`QuotaExceededError`, private-mode writes). `safeLocalStorage()` stays exported for `removeItem` / key / enumeration.

Routed the four files you named (GitDiffViewer, GitLogViewer, useProcessExpansion, modalStore), plus the three hooks that already had the gap (usePersistedToggle, usePersistedChoice, useScalePreference).

Extended `safeLocalStorage.test.ts` with a case where the global resolves but `setItem` throws `QuotaExceededError`.

Left the fetch body timeout alone, as you waived.

Issue #1511, opened 2026-09-05
Bug: Tab state, read-state and queued messages don't sync between desktop ...


chr1syy:
## Summary
Tab state, read-state and queued messages don't sync between desktop and web-desktop

## Environment
- Maestro version: 0.18.5-RC
- Operating system: Linux (#35~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue May 26 19:30:42 UTC 2, 6.17.0-35-generic)
- Install source: Packaged build (release build or locally packaged)
- Agent/provider involved: Not provided
- SSH remote execution: Not provided

## Steps to Reproduce
1. Open Maestro desktop and web-desktop against the same instance.
2. Open a tab in desktop - it appears in web-desktop.
3. Open a tab in web-desktop - it appears in desktop.
4. Close a tab in web-desktop - it stays open in desktop.
5. Reload web-desktop - the closed tab reappears, re-pushed from desktop's open-tab list.
6. Separately, mark a session read in one UI and observe the read-state is not reflected in the other; same for messages queued in one UI.

## Expected Behavior
Desktop and web-desktop should be at full parity on session state. Opening and closing a tab in either UI should propagate to the other, and a tab closed in web-desktop should stay closed across a reload. Session read-state (read/unread markers) and queued messages should also be shared between both UIs.

## Actual Behavior
Tab sync is open-only and effectively one-directional in its authority. Opening a tab in either UI propagates to the other, but closing a tab propagates in neither direction. On web-desktop reload, the desktop's open-tab set is re-pushed, so tabs the user had closed in web-desktop reappear. Read-state and queued messages are not synced between the two UIs at all.

## Additional Context
Maestro 0.18.5-RC, Linux (Ubuntu 24.04, kernel 6.17.0-35-generic), packaged build. Confirmed via maestro-cli that the app is running (port 41713) with 73 agents configured, so this is a large-workspace instance where stale tab resurrection is especially disruptive. The running build predates the list_terminal_tabs CLI verb, so open tabs could not be enumerated directly to compare desktop vs web-desktop state. File logging is off, so no same-day debug log was available; the newest log (2026-05-20) is stale and unrelated. Suggested shape of the fix: make tab close an explicit synced event rather than relying on a desktop-authoritative open-set push on web client connect, and extend the same sync channel to session read-state and queued messages.

## Screenshots / Recordings
None

pedramamini:
[...]

PR #1512, opened 2026-09-05
fix(web-desktop): sync agent lifecycle between the browser and the desktop

### LINKED ISSUE #1398: Bug: Web-desktop sessions lost from tabs after browser reload, never sync ...
## Summary
Web-desktop sessions lost from tabs after browser reload, never sync to desktop

## Environment
- Maestro version: 0.18.5-RC

## Steps to Reproduce
1. Open web-desktop and start a new session.
2. Minimize the browser (or manually reload the browser tab) — refocusing triggers a page reload.
3. Observe the tab bar after reload: it contains the desktop app's sessions/tabs, not the one started in web-desktop.
4. Open the session list and click the missing session — it is still running and shows the output produced while it was invisible.
5. Check the desktop app — the web-started session is not listed there.

## Expected Behavior
A session started in web-desktop should persist in the tab bar across browser reloads (including after minimizing and refocusing the browser), and should also appear in the desktop app's session/tab list so it can be resumed from either client.

## Actual Behavior
After the browser tab reloads, web-desktop restores only the sessions and open tabs from the desktop app. The session started in web-desktop is no longer in the tab bar. It keeps running in the background — reopening it from the session list shows all the work it did in the meantime — but it is only reachable via the session view, and it never appears in the desktop app at all.

## Additional Context
Reproduces both via minimize/refocus and via manual browser reload, so it appears tied to the page reload rather than to visibility changes. Session state is preserved server-side; the problem is session/tab restoration and cross-client (web ↔ desktop) session visibility. Maestro 0.18.5-RC on Linux, packaged build.

## Screenshots / Recordings
None

### LINKED ISSUE #1492: Bug: Web client session closes don't sync to desktop; desktop re-opens them
## Summary
Web client session closes don't sync to desktop; desktop re-opens them

## Environment
- Maestro version: 0.18.5-RC

## Steps to Reproduce
1. Have several sessions open (visible in both desktop and web client).
2. In the web client, close all/several of those sessions.
3. In the web client, open a new session.
4. The previously closed sessions reappear — matching the set still open on the desktop.

## Expected Behavior
Session lifecycle should sync bidirectionally between the web client and the desktop app. Closing a session in the web client should also close it on the desktop, so it does not come back on the next sync.

## Actual Behavior
Sessions closed in the web client remain open on the desktop. When a new session is opened from the web client, the desktop pushes its still-open session list back over and all the previously closed sessions reappear in the web client.

## Additional Context
Maestro 0.18.5-RC on Linux (6.17.0-35-generic, Ubuntu 24.04), packaged build. Diagnostics confirm the web interface is active on custom port 41713 with the persistent web link enabled, and 72 agents are configured — so the desktop↔web sync path is in use. Suspected cause: session-close events are handled locally in the web client only and are not propagated over the web link, so the desktop's session list remains authoritative and re-broadcasts on the next state push.

## Screenshots / Recordings
None

chr1syy:
Fixes #1398. Fixes #1492.

## The problem

Every client - each desktop window and every web-desktop browser tab - runs its own copy of the renderer with its own session tree, and all of them flush that tree back into the one shared sessions store. Nothing reconciled them, so:

- an agent created in the browser never appeared on the desktop (#1398), and
- an agent closed in the browser came back the moment the desktop's stale copy went dirty and was merged in again (#1492) - `sessions:setMany` appends an id it does not recognise, so any later write from the desktop re-added it.

The second half of #1398 is separate: the agent started in the browser was never actually lost (it was still in the Left Bar, still running), but a browser reloads on every refocus, and the reload read the SHARED `activeSessionId`. That put the user back on whatever the desktop had focused, so the tab bar showed the desktop's agent and its tabs.

## The fix

**Agent lifecycle is pushed, not polled.** `sessions:setMany` / `sessions:setAll` now report what entered and left the store on a new `sessions:lifecycleSync` channel, and every other client applies that delta through `useSessionLifecycleSync`. It reaches desktop windows individually (so the writer can be skipped by webContents id) and web-desktop clients through the bridge.

Three details keep it honest:

- Main tombstones a just-closed id for a minute, so a peer flush already in flight cannot re-add it.
- Only ADDITIONS travel from `setAll`. That path is a client's opening snapshot of its own tree, taken before it could have heard about anything a peer created since it loaded, so treating an absent id there as a close would let one stale snapshot delete another client's live agents.
- The hook waits for the startup load before applying, since the load ends in a single `setSessions` that would otherwise overwrite the delta (and a removal undone that way gets flushed back to disk).

Applying a delta cannot echo: a removal this client applies is reported back as a `removeIds` entry main no longer has, and an addition is reported back as an ordinary update.

**Which agent a client is looking at is now per-client.** Restore reads through `readPersistedActiveSessionId()`; web-desktop keeps its own choice in `localStorage` and falls back to the shared value only on a first visit. Writing is unchanged on both clients - the shared pointer is what plugin `session.activated` events and the CLI's current-agent answer are built on, so a browser user going quiet there would have traded one bug for another.

## Deliberately out of scope

The delta is lifecycle only. Tab contents, session read-state and queued messages are still last-writer-wins between clients (the third symptom listed in #1511) - syncing those needs a per-field model this delta cannot express. Groups are likewise not synced, so an agent that arrives from a peer while its group does not renders ungrouped rather than hidden.

## Tests

- `src/__tests__/main/ipc/handlers/persistence.test.ts` - a new `cross-client agent lifecycle sync` block: what gets reported on add / close / plain update, sender-window exclusion, a bridge call carrying no `sender`, the resurrection guard on both write paths, and that `setAll` never reports a removal.
- `src/__tests__/renderer/hooks/session/useSessionLifecycleSync.test.tsx` - add, remove, re-pointing the focused agent, ignoring a delta already applied, holding a delta until the startup load lands, and the live-turn reattach.
- `src/__tests__/renderer/utils/activeSessionPersistence.test.ts` - desktop vs web-desktop read/write behaviour and the first-visit fallback.

`npm run lint` (all three tsc configs), ESLint on the touched files, Prettier and `lint:doc-refs` are clean; the scoped suites above pass (291 tests).

## Summary by CodeRabbit

- **New Features**
  - Session lists stay synchronized across desktop windows and browser tabs when agents are added or closed.
  - Active-session selection is preserved per browser tab, with browser- and desktop-level fallbacks.
  - Newly synchronized sessions restore correctly, including ongoing activity and focus changes.

- **Bug Fixes**
  - Prevented stale session updates from resurrecting recently closed agents.
  - Improved reliability when browser storage is unavailable or cannot be written.
  - Ensured lifecycle updates are processed in order, even when restoration encounters errors.

- **Documentation**
  - Updated guidance on shared sessions, cross-client synchronization, active sessions, and the retired mobile app.


chr1syy:
Review findings addressed in 2d81c41.

**Out-of-order lifecycle deltas** (Greptile P1, CodeRabbit Major) - confirmed. Each delta ran as an independent async task, so an add for agent X could still be inside `restoreSession` when the close for X arrived; the close read X as an agent this client never had and dropped itself, then the add committed the agent the user had just closed. Deltas now apply strictly in arrival order through one promise chain, and a delta that throws is reported to Sentry rather than left to break the chain for everything behind it. `useSessionLifecycleSync.test.tsx` has a regression test that holds `restoreSession` open until the close arrives - it fails without the chain.

**Browser tabs sharing one focus key** (Greptile P1, CodeRabbit Minor) - confirmed. The read is now a ladder: `sessionStorage` (this tab's own choice, which survives the reload a refocused browser performs and dies with the tab), then `localStorage` (the last agent used in this browser, so a freshly opened tab still opens somewhere sensible), then the shared pointer for a first visit. Writes go to both, plus the shared store as before.

**Unguarded `setItem`** (CodeRabbit Minor) - confirmed and slightly wider than reported: `safeLocalStorage()` only guards reaching the object, so a full quota or Safari private mode would have thrown out of the store action that set the active agent. Added `safeSessionStorage()` and `writeStorageValue()` to the canonical module so its stated contract (a hostile Storage costs persistence, not the pane) covers writes too, with tests. Existing callers of `safeLocalStorage()?.setItem(...)` have the same latent hazard and can adopt the helper; I left them alone as out of scope here.

**Expired tombstones** (Greptile P1) - confirmed in principle, so the clock is gone rather than tuned: the tombstone set is now bounded by COUNT (1000 ids, oldest evicted). "This agent was deliberately closed" does not stop being true after a minute, a suspended tab can be away for hours, and no agent id is ever reused, so an old tombstone has nothing to block but a stale write. Worth noting the practical backstop too - a web client whose socket drops reloads the page on reconnect (`BridgeClient`, `src/web-desktop/electron-shim.ts`) and discards its queued invokes, so a long-suspended tab re-reads the store rather than flushing stale state.

Also in this branch: `docs: exempt the placeholder path that was failing lint:doc-refs on rc` (7ba282b) - unrelated to this fix, `src/foo.ts:42` in SHARED-UTILS.md is a deliberate placeholder that arrived on rc and fails `lint-and-format` for every PR built on it.

`npm run lint` (all three tsc configs), ESLint, Prettier and `lint:doc-refs` are clean; the scoped suites pass.

chr1syy:
CI note: the `Test (ubuntu-latest, shard 2/2)` failure on the previous push was `group-chat-router.test.ts > gives up on an agent that never frees up and says so`, and it fails the same way on a clean `upstream/rc` checkout - it arrived with the busy-handoff feature and is red for every PR built on rc, not just this one.

Diagnosis: reaching the 15 minute deadline only STARTS the give-up. The announcement runs in a floating async chain that writes the chat log, so the test's fixed 200 timer passes followed by a single `readLog` land between the last poll and that write. aa10bf4 pumps the poll until the message is actually on disk, which makes the assertion about the behaviour rather than the interleaving.

It is a separate commit and can be dropped from this branch if you would rather fix it on rc directly.

chr1syy:
Two more CI notes, both pre-existing on rc rather than introduced here. `upstream/rc` head (05fa10b) is currently red on `lint-and-format`, `Test (ubuntu-latest, shard 2/2)` and `Test (windows-latest, shard 1/2)`, and this branch now carries a fix for each so the PR can actually be judged on its own diff:

- 7ba282b - `lint:doc-refs` (the `src/foo.ts` placeholder in SHARED-UTILS.md).
- aa10bf4 - `group-chat-router.test.ts > gives up on an agent that never frees up`.
- 342660f - `file-tree-walk.test.ts`, eight cases, Windows only. The walker descends with `path.join`, which emits backslashes there, while the fixture map is keyed by literal `/project/src`, so every lookup past the root misses and the tree reads as unreadable. Normalized at the mock boundary rather than restating the fixtures; no-op on POSIX. One assertion (`folds a sibling folder in as empty`) also stops being vacuous on Windows, where it had been matching a path spelling the walker never produces.

Each is its own commit and can be dropped or lifted onto rc directly if you would rather not carry them here.

Also addressed the follow-up on the tombstone note: CLAUDE.md and WEB-MOBILE.md now name the bound (the newest 1000) and say what covers an id past it - a client away long enough to outlive its tombstone reloads on reconnect regardless, since the web bridge has no replay, so it re-reads the store instead of flushing what it still held (e5ccc6c).

chr1syy:
Last one: `Test (windows-latest, shard 1/2)` is green now, and `shard 2/2`'s remaining failure was `focusRingCss.test.ts`, also red on rc head and also Windows-only. The test looks a multi-line selector up by searching the raw stylesheet for a literal `\n`, and nothing in the repo pins `eol`, so a Windows checkout carries CRLF, the lookup misses, and the assertion runs against an empty block. Normalized on the way in (c723c47) - the assertion is about what the stylesheet says, not how the checkout spells a newline.

That accounts for all four rc-inherited failures. On this head every check should be green; the four repair commits (7ba282b, aa10bf4, 342660f, c723c47) are separate from the two that carry the actual fix (2c81625, 2d81c41), so they can be lifted onto rc or dropped independently.

pedramamini:
@chr1syy Thank you for this one - it is a genuinely careful piece of work, and the write-up did as much for the review as the diff did.

I walked the diff independently and I am happy with it. Notes on what I checked:

**The lifecycle push.** `broadcastSessionLifecycle` addressing desktop windows individually so the writer can be skipped by webContents id, while the bridge fans out to every web client and relies on the receiver filtering its own delta back out, is the right shape given the bridge has no per-client identity. I confirmed the shim routes `bridge.event` generically by channel, so `onLifecycleSync` reaches web-desktop without a second wiring path, and that `broadcastBridgeEvent` no-ops when no sink is registered.

**The ordering fix.** The single promise chain in `useSessionLifecycleSync` is the correct answer to the add/close interleave, and catching per-link so a thrown delta cannot stop the client following its peers for the rest of the run is the detail that would have been easy to miss. The re-check against the live list after the `await` is a nice touch.

**Tombstones as a count rather than a clock.** Agreed with the reasoning, and it is the stronger version of the fix: "this agent was deliberately closed" does not expire, and ids are never reused, so an old tombstone has nothing to block but a stale write.

**Routing removals through `setSessions` rather than `removeSession`.** This picks up the context-timeline pruning for free, since `setSessions` already prunes when the array shrinks. Right primitive.

**Focus ladder.** sessionStorage then localStorage then the shared pointer, with writes still reaching the shared store so plugin `session.activated` and the CLI's current-agent answer stay correct, is exactly the tradeoff I would have asked for.

Also checked and clean: no em or en dashes anywhere in the diff, tabs throughout, `safeSessionStorage` / `writeStorageValue` extended the canonical module rather than adding a rival copy, and CLAUDE.md plus WEB-MOBILE.md were updated alongside.

Two non-blocking observations, purely for the record:

1. Greptile's outside-diff note about a stale `setAll` snapshot deleting a peer's freshly created agent is real, but it is pre-existing on `rc` - `setAll` has always replaced the array wholesale - and this PR actually softens it, since the creating client's next `setMany` now re-reports the agent as an addition and the peers pick it up. Not something to fix here.
2. In `applyDelta`, a mixed delta that both removes and adds one agent leaves the array the same length, so `setSessions`' `newSessions.length < s.sessions.length` guard skips the context-timeline prune for the removed id. Worst case is a stale in-memory buffer, and it is the shape of the existing guard rather than anything you introduced. Fine to leave.

On the four rc-repair commits (7ba282b, aa10bf4, 342660f, c723c47): I read each one and they are all fixing the test rather than papering over a product bug - the CRLF and `path.sep` ones in particular were assertions that had quietly stopped meaning anything on Windows. Carrying them here is fine by me given `rc` head was red on all four and this PR could not otherwise be judged on its own diff. Keeping them as separate commits with the explanation was the right call.

All five checks are green on `c723c47`, both CI matrix legs included, and the branch merges cleanly into `rc`. Approving.

PR #1536, opened 2026-09-09
chore(deps): override tar to 7.5.x for node-tar Dependabot alerts

reachrazamair:
## Summary
- Force every `tar` copy on `main` to 7.5.22 via the same npm override `rc` already uses (`^7.5.17`).
- Closes the 8 node-tar Dependabot alerts on the default branch (path traversal, hardlink/symlink overwrite, APFS ligature race, PAX parser issues). Maestro does not import `tar`; it is install and packaging only (`cacache`, `node-gyp`, `electron-rebuild`, `app-builder-lib`).
- Leaves Fastify 4 on `main` on purpose. Fastify 5 already lives on `rc` (#1532) and should wait for the 0.18 release.

## Test plan
- [x] `npm install` / electron-rebuild for `node-pty` and `better-sqlite3`
- [x] `npm ls tar` shows only `7.5.22`
- [x] `npm audit` no longer reports tar (7 remaining are Fastify / vitest / adm-zip)
- [x] Full `npm test`: 36958 passed, 110 skipped
- [x] Local `validate:push` (prettier, lint, eslint, tests) on pre-push
- [x] CI `test (ubuntu-latest)` and `test (windows-latest)`

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Chores**
  * Updated package resolution to use a specified version of the `tar` package.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1194,16 +1194,6 @@
 				"node": "^20.17.0 || >=22.9.0"
 			}
 		},
-		"node_modules/@electron/rebuild/node_modules/chownr": {
-			"version": "3.0.0",
-			"resolved": "https://registry.npmjs.org/chownr/-/chownr-3.0.0.tgz",
-			"integrity": "sha512-+IxzY9BZOQd/XuYPRmrvEVjF/nqj5kgT4kEq7VofrDoM1MxoRjEWkrCC3EtLi59TVawxTAn+orJwFQcrqEN1+g==",
-			"dev": true,
-			"license": "BlueOak-1.0.0",
-			"engines": {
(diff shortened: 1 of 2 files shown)


reachrazamair:
`rc` already has this same override (`tar@7.5.22`, from #1134). I tried merging this `main` branch into `rc` so the two would stay in lockstep, and it hit 15 unrelated conflicts (`App.tsx`, Fastify `WebServer.ts`, settings, group chat, etc.) from the 10 `main` commits that are not on `rc` yet.

I aborted that merge. A full `main` → `rc` sync should stay its own PR, not ride along with this tar-only security bump.

Issue #175, opened 2026-01-10
Claude Code fails to spawn with OTEL environment variable error


goonerstrike:
## Description

When launching Claude Code through Maestro, the process fails with an OpenTelemetry configuration error if certain OTEL environment variables are set in the parent environment.

## Error

```
Error: Unknown protocol set in OTEL_EXPORTER_OTLP_METRICS_PROTOCOL or OTEL_EXPORTER_OTLP_PROTOCOL env var: undefined
    at _5 (file:///.../@anthropic-ai/claude-code/cli.js:1901:368)
    at bq2 (file:///.../@anthropic-ai/claude-code/cli.js:1901:3618)
    ...
```

## Cause

`process-manager.ts` passes `process.env` to spawned child processes. If the parent shell has `OTEL_EXPORTER_OTLP_PROTOCOL` or `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL` set to empty/invalid values, Claude Code's OTEL SDK rejects them.

## Suggested Fix

Filter OTEL vars when constructing the spawn environment:

```typescript
const env = { ...process.env };
Object.keys(env)
  .filter(k => k.startsWith('OTEL_'))
  .forEach(k => delete env[k]);
```

## Workaround

```bash
env -u OTEL_EXPORTER_OTLP_METRICS_PROTOCOL -u OTEL_EXPORTER_OTLP_PROTOCOL maestro
```

## Environment

- OS: macOS (Darwin 23.5.0)
- Node: v22.20.0 (via nvm)
- Claude Code path: `~/.nvm/versions/node/v22.20.0/bin/claude`

goonerstrike:
closing issue.. my .zshrc file had OTEL_EXPORTER_OTLP_PROTOCOL set

PR #1541, opened 2026-09-09
fix(deps): replace adm-zip with fflate (GHSA-vwc7-r8mq-g2x9)

reachrazamair:
## Summary
- Remove `adm-zip` 0.6.0. Versions 0.5.9 through 0.6.0 follow destination symlinks on extract (GHSA-vwc7-r8mq-g2x9) and there is no patched release. Downgrading to 0.5.8 would reopen the zip-bomb advisory that 0.6.0 already fixed.
- Read zips with `fflate` through `readZipArchive()` / `extractZipTo()` in `src/main/utils/zip-archive.ts`. Playbook import, Cue backup inspect/restore, debug-package tests, and the perf-trace script now use that path. Extract refuses zip-slip names and will not overwrite a dest that is already a symlink.
- Closes Dependabot alert #175 on this line. Fastify 5 is already on `rc`; those remaining main-only Fastify alerts are not part of this PR.

## Test plan
- [x] `vitest run` zip-archive, playbooks import, and debug-package packager tests (48 passed)
- [x] `docs:verify` (0 missing)
- [x] `npm audit` reports 0 vulnerabilities on this lockfile
- [ ] CI on ubuntu-latest and windows-latest
- [ ] After merge, confirm Dependabot alert #175 closes once the graph rescans

## Summary by CodeRabbit

* **Security**
  * ZIP extraction now rejects unsafe paths and avoids overwriting destination symlinks, helping prevent archive-based path traversal.

* **Bug Fixes**
  * Playbook imports and Cue backups now use safer archive handling.
  * Performance trace ZIP processing continues to support trace and optional metadata files.

* **Documentation**
  * Added guidance for the standard ZIP archive utilities and their approved usage across the application.


pedramamini:
@reachrazamair Thanks for taking this on. Swapping `adm-zip` for `fflate` behind a single `src/main/utils/zip-archive.ts` is exactly the right shape for this repo: one reader, a canonical-utilities entry so nobody adds a second, and the perf-trace script called out as the one deliberate exception. The `entry.header.size` to `entry.size` translation in `getCueBackupDiffStatus` is correct (both are the uncompressed size), and `ubuntu-latest` plus `lint-and-format` are green.

Two things I'd like addressed before this goes in. There are no merge conflicts, so no rebase needed.

### 1. `extractZipTo` escapes the destination through a pre-existing symlinked parent (Greptile P2)

I reproduced this against the exact logic in the PR. The containment check is lexical and the `lstat` only guards the leaf, so an intermediate directory that is already a symlink is followed by both `mkdirSync(..., { recursive: true })` and `writeFileSync`:

```
dest/nested -> /tmp/OUTSIDE           (pre-existing symlink)
entry:  nested/victim.txt
result: extract completed without error
        /tmp/OUTSIDE/victim.txt now contains the zip's bytes
```

The lexical check passes because `path.resolve` never touches the filesystem, and the leaf `lstat` sees a plain file (or nothing at all).

In practice this is not currently exploitable: the only caller is the `unzip`-CLI fallback in `packager.test.ts`, and it needs a symlink already sitting in the destination. But the function is exported, its error text says "Refusing zip entry outside destination", and the canonical-utilities entry is what the next caller will trust. That is the reason to close it now rather than after someone wires it to a user-selected archive. `realpathSync` the parent after the `mkdirSync` and re-check containment against `destRoot` (plus a test with a symlinked intermediate directory) should do it.

### 2. `unzipSync` decompresses every entry, which regresses the Cue backup paths

Both bots flagged this as a zip-bomb concern. I think the more concrete problem is a performance regression on paths that only ever want one small file.

`adm-zip` parsed the central directory up front and decompressed per entry in `getData()`. `unzipSync` materializes the whole archive before it returns. So:

- `listCueBackups()` fully decompresses **every** backup zip in the directory, and throws all of it away except `manifest.json`.
- `inspectCueBackup()` and `readCueBackupFile()` do the same for one archive each.

All of it is synchronous, on the Electron main thread, which is the thread that answers every IPC call.

`fflate` has the hook for this already: `unzipSync(data, { filter })` receives each entry's metadata and skips decompression when it returns false. A `readZipArchive(filePath, { filter })` option threaded into those three call sites keeps them decompressing one entry, the way `adm-zip` did.

That same `filter` is also the natural place for the budget the bots asked for, since it sees `originalSize` before any bytes are expanded. Worth doing given the PR's own rationale is that downgrading to 0.5.8 would reopen the zip-bomb advisory: right now the replacement has no bomb ceiling at all. An entry-count and total-uncompressed-size cap in `readZipArchive` is enough; I don't think this needs the streaming `Unzip` API.

Playbook import itself is fine on the traversal front, incidentally: it writes through `path.basename()`, so entry names cannot walk out of the Auto Run folder.

One note: the `windows-latest` shards were still pending when I looked, so that leg is worth confirming before merge.

Happy to re-review as soon as these are in.

PR #1542, opened 2026-09-09
fix(deps): replace adm-zip with fflate (GHSA-vwc7-r8mq-g2x9)

reachrazamair:
## Summary
- Remove `adm-zip` 0.6.0. Versions 0.5.9 through 0.6.0 follow destination symlinks on extract (GHSA-vwc7-r8mq-g2x9) and there is no patched release. Downgrading to 0.5.8 would reopen the zip-bomb advisory that 0.6.0 already fixed.
- Read zips with `fflate` through `readZipArchive()` / `extractZipTo()` in `src/main/utils/zip-archive.ts`. Playbook import, Cue backup inspect/restore, debug-package tests, and the perf-trace script now use that path. Extract refuses zip-slip names and will not overwrite a dest that is already a symlink.
- Closes Dependabot alert #175. Fastify stays on 4 on this line. The Fastify 5 upgrade already lives on `rc` and is not backported here.

## Test plan
- [x] `vitest run` zip-archive, playbooks import, and debug-package packager tests (48 passed)
- [x] Lockfile has `fflate` and no `adm-zip`
- [ ] CI on ubuntu-latest and windows-latest
- [ ] After merge, confirm Dependabot alert #175 closes once the graph rescans

## Summary by CodeRabbit

- **Security**
  - Improved ZIP extraction safeguards to block path traversal and prevent overwriting symlink destinations.

- **Bug Fixes**
  - Updated playbook imports and Cue backup handling for more reliable ZIP reading.
  - ZIP-based performance trace analysis continues to support trace and metadata files.

- **Documentation**
  - Added guidance for consistent ZIP archive handling and safer extraction practices.

- **Tests**
  - Added coverage for archive reading, extraction, missing entries, unsafe paths, and symlink protection.


pedramamini:
@reachrazamair Thanks for taking this on. Killing off `adm-zip` cleanly rather than pinning around it is the right call, and the writeup explaining why 0.5.8 is not an option saved me a round trip. The new `zip-archive.ts` is a good shape, the tests are real tests (the zip-slip and symlink cases actually assert the target file survived), and CI is green with no merge conflicts.

Before I approve, there are two things I would like addressed. Greptile and CodeRabbit both landed on the first one independently, and I confirmed it against the code.

### 1. `readZipArchive()` inflates the whole archive, and some callers pay for that repeatedly

`adm-zip` parsed the central directory up front and only inflated an entry when you called `getData()`. `unzipSync()` inflates every entry eagerly, so the laziness is gone. Two concrete consequences in `src/main/cue/backup/cue-backup-manager.ts`:

- `restoreCueBackupAll()` calls `inspectCueBackup()` once and then `readCueBackupFile()` once per manifest file, and each of those is a fresh `readZipArchive()`. That is now `N+1` full-archive inflations for a restore of `N` files, where it used to be `N+1` reads with `N` single-entry inflations.
- `listCueBackups()` fully inflates every backup zip in the directory just to read `manifest.json` out of each one.

That is also the zip-bomb surface both bots flagged on the playbook import path: a small, highly compressible archive from the file picker now expands fully in the main process before the manifest is even validated. Worth avoiding, especially since the PR description cites the zip-bomb advisory as the reason not to downgrade.

`fflate` gives you what you need here without a restructure. `unzipSync(data, opts)` takes a `filter: (file: UnzipFileInfo) => boolean` that runs before inflation, and `UnzipFileInfo` carries `name`, `size` (compressed) and `originalSize`. So an optional options bag on `readZipArchive()` covers both problems:

- a `names` or predicate option so `listCueBackups()` / `inspectCueBackup()` can ask for `manifest.json` alone instead of the whole archive,
- a cap on aggregate `originalSize` plus an entry count, refusing the archive when it is exceeded.

Reusing a single `ZipArchive` across the loop in `restoreCueBackupAll()` (CodeRabbit's suggestion) is worth doing too, but the filter is the part that matters most.

### 2. `extractZipTo()` only checks the final path component for a symlink

Greptile's P2 is correct. The `lstat` check covers `dest` itself, but if an intermediate component like `destRoot/nested` is already a symlink, the lexical containment check still passes and `mkdirSync` plus `writeFileSync` follow it straight out of `destRoot`. I know this only has test callers today, but you are adding it to `CANONICAL-UTILITIES.md` as the safe extraction helper, which means the next person reaches for it and inherits the gap. Please walk every path component under `destRoot` (or resolve the real parent and re-verify containment) before writing, with a test for the intermediate-symlink case alongside the existing one.

### Non-blocking notes

- `.github/workflows/ci.yml` only runs `ubuntu-latest`, so the `windows-latest` line in the test plan will not actually get exercised. Relevant because `fs.symlinkSync` in the new symlink test needs Developer Mode or elevation on Windows and will throw for anyone running the suite there. A `process.platform === 'win32'` skip on that one case would keep it honest.
- CodeRabbit also wants symlink checks on the playbook import write loops. I would leave that alone here. Those writes already run entry names through `path.basename()`, so zip-slip is handled, and the remaining symlink-follow behavior is unchanged by this PR. Out of scope for a dependency swap.

Once the bounded read and the parent-component check are in, this is good to go. Thanks again for the thorough writeup.

reachrazamair:
> @reachrazamair Thanks for taking this on. Killing off `adm-zip` cleanly rather than pinning around it is the right call, and the writeup explaining why 0.5.8 is not an option saved me a round trip. The new `zip-archive.ts` is a good shape, the tests are real tests (the zip-slip and symlink cases actually assert the target file survived), and CI is green with no merge conflicts.
> 
> Before I approve, there are two things I would like addressed. Greptile and CodeRabbit both landed on the first one independently, and I confirmed it against the code.
> 
> ### 1. `readZipArchive()` inflates the whole archive, and some callers pay for that repeatedly
> `adm-zip` parsed the central directory up front and only inflated an entry when you called `getData()`. `unzipSync()` inflates every entry eagerly, so the laziness is gone. Two concrete consequences in `src/main/cue/backup/cue-backup-manager.ts`:
> 
> * `restoreCueBackupAll()` calls `inspectCueBackup()` once and then `readCueBackupFile()` once per manifest file, and each of those is a fresh `readZipArchive()`. That is now `N+1` full-archive inflations for a restore of `N` files, where it used to be `N+1` reads with `N` single-entry inflations.
> * `listCueBackups()` fully inflates every backup zip in the directory just to read `manifest.json` out of each one.
> 
> That is also the zip-bomb surface both bots flagged on the playbook import path: a small, highly compressible archive from the file picker now expands fully in the main process before the manifest is even validated. Worth avoiding, especially since the PR description cites the zip-bomb advisory as the reason not to downgrade.
> 
> `fflate` gives you what you need here without a restructure. `unzipSync(data, opts)` takes a `filter: (file: UnzipFileInfo) => boolean` that runs before inflation, and `UnzipFileInfo` carries `name`, `size` (compressed) and `originalSize`. So an optional options bag on `readZipArchive()` covers both problems:
> 
> * a `names` or predicate option so `listCueBackups()` / `inspectCueBackup()` can ask for `manifest.json` alone instead of the whole archive,
> * a cap on aggregate `originalSize` plus an entry count, refusing the archive when it is exceeded.
> 
> Reusing a single `ZipArchive` across the loop in `restoreCueBackupAll()` (CodeRabbit's suggestion) is worth doing too, but the filter is the part that matters most.
> 
> ### 2. `extractZipTo()` only checks the final path component for a symlink
> Greptile's P2 is correct. The `lstat` check covers `dest` itself, but if an intermediate component like `destRoot/nested` is already a symlink, the lexical containment check still passes and `mkdirSync` plus `writeFileSync` follow it straight out of `destRoot`. I know this only has test callers today, but you are adding it to `CANONICAL-UTILITIES.md` as the safe extraction helper, which means the next person reaches for it and inherits the gap. Please walk every path component under `destRoot` (or resolve the real parent and re-verify containment) before writing, with a test for the intermediate-symlink case alongside the existing one.
> 
> ### Non-blocking notes
> * `.github/workflows/ci.yml` only runs `ubuntu-latest`, so the `windows-latest` line in the test plan will not actually get exercised. Relevant because `fs.symlinkSync` in the new symlink test needs Developer Mode or elevation on Windows and will throw for anyone running the suite there. A `process.platform === 'win32'` skip on that one case would keep it honest.
> * CodeRabbit also wants symlink checks on the playbook import write loops. I would leave that alone here. Those writes already run entry names through `path.basename()`, so zip-slip is handled, and the remaining symlink-follow behavior is unchanged by this PR. Out of scope for a dependency swap.
> 
> Once the bounded read and the parent-component check are in, this is good to go. Thanks again for the thorough writeup.

Thanks. Both of these are addressed in #1543.

Issue #1528, opened 2026-09-08
Bug: Web-desktop bridge drop during load lets sessions:setAll wipe agents


ronaldeddings:
## Summary
Web-desktop bridge drop during load lets sessions:setAll wipe agents

## Environment
- Maestro version: 0.18.5-RC
- Operating system: macOS (Darwin Kernel Version 25.5.0: Mon Apr 27 20:39:42 PDT 2026; root:xnu-12377.121.6~2/RELEASE_ARM64_T6031, 25.5.0)
- Install source: Packaged build (release build or locally packaged)
- Agent/provider involved: Not provided
- SSH remote execution: Not provided

## Steps to Reproduce
1. Run Maestro 0.18.5-RC with the web interface on and a large agent list (~90 agents, ~28 MB maestro-sessions.json).
2. Open the web-desktop UI from another machine over a slow or lossy link (Tailscale in this case).
3. Cause the WebSocket to close while the client's initial sessions:getAll is still in flight (network blip, laptop sleep, or throttle the socket). The bridge rejects the pending invoke with 'bridge disconnected'; the renderer logs 'Failed to load sessions/groups', sets sessions to [], and marks initialLoadComplete.
4. Let the bridge reconnect (1s). Any change to the client's session tree (a peer lifecycle delta from sessions:lifecycleSync, or any user action in that tab) schedules the first flush.
5. Observe: first flush calls sessions:setAll with the partial tree; main replaces maestro-sessions.json with it. Desktop keeps 91 in memory until restart, then loads only the survivors.

## Expected Behavior
A web-desktop client whose initial sessions load fails (WebSocket drop mid-load, common over Tailscale or any flaky link) should never persist. If it recovers, it should reload the full tree or the page. sessions:setAll should never remove agents from maestro-sessions.json that the calling client did not explicitly close. The desktop's 91 agents should have survived a remote browser tab's reconnect.

## Actual Behavior
[...]

PR #1550, opened 2026-09-10
fix(groups): never persist a group registry the app never successfully read

### LINKED ISSUE #1547: Project group vanished from sidebar after deleting an agent (0.17.3)
## Summary
Entire sidebar went empty (just "+ New Group") after deleting one agent group, including a second, unrelated project group that was never touched.

## Environment
- Maestro version: 0.17.3
- macOS, packaged build (dmg)

## Steps to Reproduce
1. Had two project groups in the sidebar, each pointing at a different repo (created via the global New Agent flow, not "New Agent Here").
2. Deleted one of the groups.
3. Sidebar came back completely empty except "+ New Group". The other group, which was never touched, was also gone.

## Expected Behavior
Deleting one agent/group shouldn't affect any other unrelated group.

## Additional Context
This happened once, mid a long debugging session, so I don't have a clean minimal repro yet. Filing since it lines up with the general class of agent-persistence issues already tracked in #1528 (different trigger there, a web-desktop bridge drop over a remote connection). Could be related, could be separate. Happy to try to pin down a tighter repro if that's useful.

pedramamini:
Closes #1547

## What happened

The reporter deleted one project group and the whole Left Bar came back empty, including a second, unrelated group they never touched.

The group registry (`maestro-groups.json`) is the only copy of a group's name, emoji and collapsed state. Agents reference groups by `groupId`, so when the registry is emptied the ids survive and point at rows that no longer exist: every agent silently becomes ungrouped and there is nothing on disk to rebuild the rows from.

Two halves combined to empty it:

1. `useSessionRestoration` read the registry and, on any failure, called `setGroups([])`. `groups:getAll` also answers `[]` for a registry it could not read, and the store lives under the configurable sync path, so a cloud folder that has not finished mounting at launch produces exactly that with no exception raised anywhere.
2. `useSessionLifecycle` wrote whatever was in memory straight back to disk, gated only on `initialLoadComplete`. That flag is set in a `finally`, so it is true even when the read failed. An unverified empty read became the new truth, and every launch after rewrote the empty file.

The groups read also shared one `try` with the session read, so a rejected `groups:getAll` landed in the outer catch and zeroed `setSessions` too: a groups failure could cost the user their agents.

## The fix

- A new `groupsLoaded` flag on `sessionStore`, set only when `groups:getAll` actually came back. The persistence effect requires it, so a registry we never read is never written. The gate is deliberately "did the read succeed", not "was the result non-empty": the latter would stop a brand new user from ever saving their first group.
- The groups read moves into its own `try`/`catch`, and the outer catch no longer touches groups at all.
- Defence in depth on the main side: `groups:setAll` snapshots the stored registry to `maestro-groups.backup.json` before an empty one replaces it. It does not refuse the write, because deleting your last group is legitimate; it keeps a copy first, which turns a permanent loss into a recoverable one. It sits at the handler rather than at the caller because the CLI (`remove-group`) and the web bridge reach the store through that same choke point.

## Provenance

This is a cherry-pick of f5ae24c9b from `rc`, which has been fixed there since 2026-09-07 but was never ported to `main`. The reporter is on 0.17.3, the stable line, which still carries the bug. One incidental conflict in `useSessionLifecycle.ts` (unrelated `rc`-only selectors) was resolved in favour of `main`'s code, keeping only the `groupsLoaded` additions.

## Scope

Groups only. The session tree has the same shape and is fixed on `rc` by 047817b9c; on `main` it is still open and tracked by #1528, so it is deliberately left out of this PR.

## Validation

- `npm run lint` (tsc lint/main/cli configs) clean
- ESLint clean on all changed source files at `--max-warnings=0`
- Prettier clean on all changed files
- 671 tests pass across 17 related suites (persistence handlers, main stores, sessionStore, restoration, lifecycle, SessionList, groupHierarchy), including the 3 new suites this adds

## Summary by CodeRabbit

* **New Features**
  * Group registries are backed up before being replaced with an empty registry.
  * Backups include timestamp and reason metadata.

* **Bug Fixes**
  * Prevented failed group reads from overwriting existing groups.
  * Improved session restoration to preserve existing group and session data when reads fail.
  * Restricted group persistence to successfully loaded registries.

* **Tests**
  * Added coverage for backup behavior, read/write failures, and session persistence safeguards.


pedramamini:
CI status note: `test` (full suite), `Analyze`, Greptile and CodeRabbit all pass.

`lint-and-format` is red on a **pre-existing baseline failure unrelated to this branch**: `npx prettier --check .` flags `README.md`, which was left unformatted by the three most recent commits on `main` (d859ef3d2, 9f494a214, 16d9882c2). This branch touches only the eight files under `src/` listed in the diff and does not go near `README.md`; `npx prettier --check` on every changed file passes, and `npm run lint` (tsc across the lint/main/cli configs) plus ESLint at `--max-warnings=0` are clean.

Left it alone rather than folding an unrelated README reformat into a data-loss fix - happy to add the one-line `prettier --write README.md` here if you'd prefer the branch go green on its own.

PR #1561, opened 2026-09-10
fix(cli): repair Windows installer PATH update

chr1syy:
## Summary

- execute `powershell.exe` directly so `cmd.exe` cannot reinterpret the PATH update script
- preserve PowerShell block syntax by joining statements with newlines
- keep Windows desktop file logging out of CLI processes and tolerate incidental output when parsing the CLI version
- add regression coverage for the malformed `{;` script and logger-prefixed version output seen on Windows

## Testing

- `npm run lint`
- `npm run lint:eslint`
- `npm test -- --run src/__tests__/main/maestro-cli-manager.test.ts src/__tests__/main/utils/logger.test.ts` (84 passed)
- verified `ELECTRON_RUN_AS_NODE=1` leaves `process.type` undefined

## Baseline notes

- `npm run format:check:all` currently reports 63 pre-existing formatting violations on `main`; all files changed by this PR pass Prettier.
- A full `npm test` run was stopped after unrelated baseline failures appeared, including stale `lucide-react` test mocks and a missing local native `better-sqlite3` binding.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved version detection when command output includes logging prefixes or prerelease versions.
  * Improved Windows PATH updates by ensuring PowerShell commands execute correctly, including paths containing apostrophes.
  * Prevented command-line processes from unnecessarily enabling desktop file logging.
* **Tests**
  * Expanded coverage for version parsing, Windows PATH handling, platform-specific logging, and command execution behavior.

### DIFF
--- a/src/__tests__/main/maestro-cli-manager.test.ts
+++ b/src/__tests__/main/maestro-cli-manager.test.ts
@@ -1,0 +1,78 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+const { mockExecFileNoThrow } = vi.hoisted(() => ({
+	mockExecFileNoThrow: vi.fn(),
+}));
+
+vi.mock('electron', () => ({
+	app: {
+		getAppPath: vi.fn().mockReturnValue('/mock/app'),
+		getVersion: vi.fn().mockReturnValue('0.17.4'),
+	},
+}));
+
+vi.mock('../../main/utils/execFile', () => ({
+	execFileNoThrow: mockExecFileNoThrow,
+}));
+
+vi.mock('../../shared/platformDetection', () => ({
+	getWhichCommand: vi.fn().mockReturnValue('where'),
+	isWindows: vi.fn().mockReturnValue(true),
+}));
+
+vi.mock('../../main/utils/logger', () => ({
+	logger: {
+		debug: vi.fn(),
+		error: vi.fn(),
+		info: vi.fn(),
+		warn: vi.fn(),
+	},
+}));
+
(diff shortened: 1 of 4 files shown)


pedramamini:
@chr1syy thanks for this fix, and for the clear write-up! Tracking down that `{;` script corruption on Windows can't have been fun.

This looks good to merge. What I checked:

- **The root cause is right.** In `needsWindowsShell()`, a bare `powershell` has no extension and isn't on the known-exe list. So `execFileNoThrow` ran it with `shell: true`, and cmd.exe reinterpreted the `|` pipes in the script. Calling `powershell.exe` skips that, and joining with `\n` keeps the `if { ... }` block valid.
- **The logger change is scoped correctly.** The Windows shim sets `ELECTRON_RUN_AS_NODE=1`, so `process.type` is undefined in the CLI, and the desktop main process (`'browser'`) keeps its default file logging. The other Electron-as-Node children (maestro-p, Cue CLI runs) also stop writing the `[Logger]` banner to stdout, which is what we want.
- **`normalizeVersion`** now takes the first line that is exactly a version, then falls back to the old first-line behavior. The logger's date-stamped path can't be mistaken for a version because the date uses dashes.
- **Local run on the PR head:** both test files pass (84 tests), Prettier is clean, ESLint shows no issues on the source files, and tsc reports nothing in the changed files.

About the red `lint-and-format` check: it's Prettier failing on `README.md`, which this PR doesn't touch. The same failure is on this PR's base commit on `main` (`d859ef3`), and the latest `main` runs are green, so it should clear once this merges or picks up current `main`. Because Prettier failed first, CI skipped the ESLint and tsc steps, so I ran those locally.

Approving. Thanks again!

PR #1580, opened 2026-09-14
Port per-surface typography and first-run onboarding onto main

reachrazamair:
## Summary
- Port per-surface fonts, sizes, zoom, bundled webfonts, and Factory Reset / Save Customizations from rc onto main's single `settingsStore`.
- Show the four-step first-run series (typography, theme, updates, agent powers) once to existing 0.17.4 users by defaulting the `*PromptSeen` flags to false.
- Add `maestro-cli display font|size|zoom|preset|fonts` and update Settings search / keyboard zoom so Cmd+= moves zoom, not the old global size.

## Test plan
- [x] `unset ELECTRON_RUN_AS_NODE && npm run dev` in `Maestro-typography-main`; use the Electron window, not Chrome
- [x] Fresh isolated launch shows Typography → Theme → Ues → Agent Powers; Back works both ways
- [x] Cmd+K → Debug: Replay First-Run Series (Existing User) shows returning-user copy and does not consume the real first run
- [x] Cmd+, → Display: Fonts rows, Zoom, Factory Reset (confirm then apply), Save / Restore Customizations; search finds Fonts and Zoom
- [x] Interface font does not change the MAESTRO wordmark; terminal / chat / file preview / file editor follow their own rows
- [x] Cmd+= / Cmd+- / Cmd+Shift+0 move Zoom only; Interface pixel size stays put
- [x] Optional: `npm run build:cli && node dist/cli/maestro-cli.js display fonts`

## Summary by CodeRabbit

* **New Features**
  * Added per-surface typography controls for fonts, sizes, inheritance, custom fonts, zoom, presets, factory reset, and save/restore.
  * Added bundled font options, improved previews, and display typography CLI commands.
  * Added first-run onboarding for typography, themes, updates, and agent capabilities.
  * Added configuration for alternate token backends.
* **Bug Fixes**
  * Improved typography rendering across chat, previews, editors, and document graphs.
  * Prevented stale CLI installation status from appearing after failed checks.
* **Documentation**
  * Updated configuration, CLI reference, and getting-started guides.
* **Tests**
  * Expanded coverage for typography, onboarding, fonts, and CLI behavior.


pedramamini:
@reachrazamair thank you for taking on this backport! Bringing per-surface typography, the bundled fonts, and the first-run series over to `main` is a big, careful piece of work, and CI is green with no merge conflicts. We went through the CodeRabbit review and the code, and a few things should be fixed before we merge. (Greptile skipped this one because the PR has more than 100 files, so this is a manual pass.)

### Please address

1. **Presets leave out the Document Graph** (`src/shared/typographyPresets.ts`). The surface registry has six surfaces, but `TypographyPresetFonts` / `TypographyPresetSizes` and both presets only cover five. As a result, Factory Reset and applying a preset never touch `documentGraphFontFamily` / `documentGraphFontSize`, and `matchTypographyPreset` can say a preset is active while the graph is still customized. Please add the document graph fields to both interfaces, both presets, and both `surfaces` summaries, and extend `typographyPresets.test.ts` to match.
2. **`applyTypographyPreset` ignores preset sizes** (`src/renderer/stores/settingsStore.ts:1127`). It writes `fonts` only. If a new user picks Default in the first-run modal, the interface and code surfaces stay at 14px instead of the preset's 15px / 13px. Please write fonts and sizes together, the way `resetTypography` already does.
3. **Some surfaces ignore their size setting**, so those rows in Settings look like they do nothing:
   - `DocumentGraph/MindMap.tsx`: only the family is resolved; the canvas text is still hardcoded at 10-13px.
   - `shared/DualPaneFileEditor.css:228`: `font-size: 12px` is fixed; it should use the File Editor size var with 12px as the fallback.
   - `FilePreview.tsx` / `GroupChatMessages.tsx`: child elements with `text-sm` / rem sizes override the size set on the container.
4. **New vs. returning is decided by whether any agents exist** (`src/renderer/hooks/ui/useAppInitialization.ts:190`). An existing user who has deleted all their agents is treated as new: they get the new-user copy with Default preselected, so clicking through replaces the monospace look they've been using. A persisted signal (such as settings keys that predate the upgrade, or an install marker) would be more reliable.
5. **The CLI reference wasn't regenerated.** `docs/cli-reference.md` has no `display` section. Please run `npm run gen:cli-reference` and then prettier on the output.

### Smaller fixes (quick, and worth doing while you're in there)

- `src/cli/commands/display-font.ts:78`: `effectiveFont` returns a `(inherits interface)` label instead of the font family that actually renders, so the JSON output can't answer "what font am I seeing".
- `src/cli/commands/display-font.ts:170`: `INHERIT_TERMINAL` is `'@terminal'`, which is truthy, so `inherit:terminal` prints `font set to "@terminal"` instead of the inheritance message.
- `FontConfigurationPanel.tsx`: in compact mode the `<select>` has no accessible name; please pass the surface label through as `aria-label`.
- `giantPreview/GiantPreview.tsx`: a font change destroys and recreates the `EditorView`, which loses scroll and selection. Putting the theme in a `Compartment`, as `MarkdownEditor` does, avoids that.
- `useMaestroCliState.ts`: reset `status` to `null` when a check starts, so a failed re-check doesn't still show "Installed".
- Settings hydration for the `*FontFamily` keys: only assign when `typeof value === 'string'`.
- `AgentPowersModal.tsx`: the copy says the user "just picked" both a typography and a theme, but either step can be skipped.

### Safe to skip

- CodeRabbit's `fonts-and-sizing.test.ts` orphan finding doesn't apply: this branch has no `jetbrains-mono.css` in `public/fonts`, and the test passes in CI.
- The modal-guard shortcut problem (Cmd+Shift+0 is blocked while a modal is open) already exists on `main` with the same code, so this PR didn't cause it. Since Cmd+= now controls zoom, it would be a nice fix while you're here: match `=`/`+`/`-` without Shift, and `0` only with Shift. That also stops bare Cmd+0 (Go to Last Tab) from switching tabs behind an open modal. This one is optional.

Thanks again, and ping us when it's ready for another look!

PR #1590, opened 2026-09-15
refactor(group-chat): unify input mentions and staged images with AI Chat (Phase 0)

reachrazamair:
## Summary

This is Phase 0 of that unification: the input box only.

- **Draft persistence**: extracted `useDraftPersistence`, the debounced,
  key-switch-safe write-back both `useInputSync` (AI Chat) and
  `GroupChatInput` had independently reinvented. Both now share it.
- **@ mention picker**: `GroupChatInput` no longer hand-rolls its own
  mention dropdown. It now uses the same `useMentionPicker` +
  `AtMentionPopover` AI Chat uses, locked to the Agents category (Group
  Chat has no file/directory mentions). Added an optional
  `showCategoryBar` prop to `AtMentionPopover` (defaults to `true`, so AI
  Chat is unchanged) so Group Chat can hide the All/Files/Directories/Agents
  bar instead of showing three permanently empty, non-functional tabs.
- **Staged images**: `GroupChatInput`'s inline thumbnail markup is swapped
  for the shared `StagedImagesStrip`/`StagedImageTile`, which brings
  drag-to-reorder, the image organizer modal, and the annotator along for
  free.

One deliberate, visible behavior change: the mention picker's arrow-key
navigation now clamps at the top/bottom of the list instead of wrapping
around, matching AI Chat's picker exactly. That is the actual point of this
PR, so it's called out rather than treated as an incidental side effect.

## Out of scope (left for follow-up PRs)

This PR is the input box only. Two more pieces of the same ask are
deliberately not started here:

- **Message-rendering parity**: `TerminalOutput.tsx` and
  `GroupChatMessages.tsx` each own their own collapse/expand state and
  image-gallery rendering; `GroupChatMessages` is also missing some
  `MarkdownRenderer` props `LogItem.tsx` already passes (bionic reading
  mode is a trivial add, file-link/SSH-remote awareness is harder since
  group chat has no single active session). Real work, not blocked on
  anything, just not in this PR.
- **Model/effort visibility**: genuinely undecided. Pedram's own words were
  that group chat may not need input-side model/effort controls at all,
  just response-time pills. No architecture call has been made on this
  yet, and it needs new main-process plumbing (group chat's spawn config
  carries no model/effort field today) before any UI could be added. Not
  started, and shouldn't be until that call happens.

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] `npm run lint:eslint` clean on all changed files
- [x] Existing `GroupChatInput.test.tsx` suite updated for the new popover's
      markup (no `@` prefix on mention rows, provider display name instead
      of the raw tool-type string, clamped instead of wrapped arrow-key
      nav) and 3 new tests added for the staged-images wiring; 43/43 pass
- [x] Full affected cluster (GroupChatInput, InputArea, AtMentionPopover,
      useMentionPicker, useInputSync, useDraftPersistence,
      stagedImageOrder): 385/385 tests pass across 25 files
- [x] Manual pass in `npm run dev` (Electron desktop app, not verified in
      this environment - please give the Group Chat input a look before
      merging: typing/switching drafts, mentioning an agent and a group,
      attaching/reordering/removing a staged image)

## Summary by CodeRabbit

- **New Features**
  - Added an option to hide category tabs in mention suggestions.

- **Improvements**
  - Improved image staging and reordering in group chat.
  - Drafts now save reliably when switching conversations or leaving the input.
  - Mention suggestions display provider and group names, including agent counts.
  - Keyboard navigation stops at the first and last items instead of wrapping around.

- **Bug Fixes**
  - Escape now closes mention suggestions when no results are available.
  - Fixed mention acceptance when results change while the picker is open.
  - Prevented empty mention searches from sending unintended text.


pedramamini:
@reachrazamair Thank you for this - genuinely nice piece of work. Deleting a hand-rolled mention dropdown and a bespoke thumbnail strip in favour of the shared components is exactly the right shape for Phase 0, the scoping note about what you deliberately left for follow-up PRs is appreciated, and calling out the arrow-key wrap change rather than burying it is the right instinct.

I went through both AI reviews and verified each finding against the code. Four are real and I'd like them addressed before this merges. Three of them are the same class of issue, which is worth stating plainly: this PR's whole premise is that Group Chat and AI Chat should behave identically, and in each case the new Group Chat code diverges from what AI Chat already does.

**1. The extracted hook is a throttle, not a debounce, and that silently changes Group Chat's cadence**

`useDraftPersistence` bails out on an already-armed timer:

```ts
if (timerRef.current) return;
```

That is leading-edge throttling. AI Chat's `useInputSync` on `rc` had exactly this same `if (flushTimerRef.current) return;`, so the extraction is faithful to AI Chat. Group Chat, though, was previously on `useDebouncedCallback`, which clears and re-arms the timer on every call, so it is a true trailing-edge debounce. Net effect: Group Chat's `onDraftChange` goes from firing once after the user pauses to firing every 300ms for as long as they keep typing.

That may well be an acceptable trade to get one shared primitive, but right now it is an undisclosed behavior change in a PR that was careful to disclose the other one. Either restore trailing-edge for Group Chat (or support both modes explicitly), or keep the throttle and call it out in the description next to the arrow-key note. Either way please fix the docstring, which currently says "Debounced write-back" and describes behavior the hook does not have. That comment will mislead whoever touches this next.

**2. and 3. The mention keyboard handler should mirror AI Chat's structure**

These are the two CodeRabbit flagged, and they share one fix. Compare `useInputKeyDown.ts:369-425` (AI Chat) with `GroupChatInput.tsx:336-363`:

- AI Chat gates on `atMentionOpen` alone. Group Chat gates the entire block on `atMentionOpen && atMentionItems.length > 0`. `AtMentionPopover` renders an empty state when `items.length === 0` (line 190), so on an empty picker the popover is on screen and Escape does nothing. AI Chat closes it.
- AI Chat does `const selected = atMentionItems[selectedAtMentionIndex]; if (selected) { ... }`. Group Chat passes the lookup straight into `acceptAtMention`, which hands it to `buildMentionAccept` and dereferences it. `setSelectedAtMentionIndex(0)` on each keystroke covers filter-driven shrinks, but not the store-driven one: if an agent closes while the picker is open, `sessions` shrinks, `atMentionItems` shrinks, the index does not, and Tab/Enter throws in the event handler. Narrow, but the guard is free and the other surface already has it.

Matching AI Chat's shape fixes both and removes a divergence rather than adding one.

**4. Test store reset should use the repo helper**

`GroupChatInput.test.tsx:88-90` resets `useSessionStore` and `useImageAnnotatorStore` via raw `setState`. Per [TEST-PATTERNS.md:265-282](docs/agent-guides/TEST-PATTERNS.md), these should go through `resetStore` / `resetStores` from `src/__tests__/helpers`. The annotator reset in particular sets only three fields, so anything else on that store leaks between tests. `resetStores(useSessionStore, useImageAnnotatorStore)` should be a drop-in.

**Not your problem, so please don't chase it:** the `lint-and-format` leg is red on `lint:doc-refs` complaining that `docs/agent-guides/PROMPTS-SPECS.md:31` points at `src/generated/prompts.ts`. That file is untouched by this PR and `rc`'s own CI is failing the same check right now, so it is a pre-existing baseline issue. All four test legs (ubuntu and windows, both shards) are green on your branch, and there are no merge conflicts.

Ping me once those are in and I'll take another pass.

PR #1597, opened 2026-09-16
fix(codex-usage): classify quota windows by duration, not slot position

### LINKED ISSUE #1596: Codex usage: windows labeled by position, so a weekly window shows as "Session (5h)"
### Summary

The Codex usage sampler labels rate-limit windows by their **position** in the
`wham/usage` response rather than by their **duration**, so the Usage Dashboard
displays a 7-day window under a "Session (5h)" heading on any plan whose primary
window is weekly.

### Maestro version

0.18.5-RC

### Operating system / OS version

macOS 26.6.1 (arm64), installed from the DMG

### Product area

Usage Dashboard → OpenAI Usage

### Agent/provider involved

Codex (OpenAI). SSH remote: not involved.

### Reproducibility

Always, on any account whose plan reports a weekly `primary_window`.

### Steps to reproduce

1. Log a CODEX_HOME into an account on a plan whose primary window is weekly
   (reproduced on `prolite`; plans with a 5h session window such as `team` are
   unaffected).
2. Open Usage Dashboard → OpenAI Usage.
3. The row reads `Session (5h) · 25% · resets Wed 12:34 PM` — a reset seven days
   out, under a heading that claims a five-hour window.

### Details

`GET https://chatgpt.com/backend-api/wham/usage` returns each window with an
explicit `limit_window_seconds`, and the two plan shapes order them differently:

| plan | `primary_window` | `secondary_window` |
| --- | --- | --- |
| team | `limit_window_seconds: 18000` (5h session) | `604800` (weekly) |
| prolite | `limit_window_seconds: 604800` (**weekly**) | `null` |

`src/main/agents/codex-usage-sampler.ts:169-170` maps them unconditionally:

```ts
const session = parseWindow(rateLimit.primary_window);
const weekly  = parseWindow(rateLimit.secondary_window);
```

So on `prolite` the weekly figure is stored as `session` and the account shows
`weekly: undefined`. The `WhamUsageWindow` interface (same file, ~line 49) does
not declare `limit_window_seconds` at all, so the field that disambiguates the
two is discarded before it can be used.

### Impact

Beyond the cosmetic mislabel, anything reading the snapshot for scheduling or
quota decisions sees a weekly budget where it expects a 5h one, and sees "no
weekly limit" on an account that has only a weekly limit. A consumer that waits
for a "session" reset on such an account waits up to seven days.

### Suggested fix

Add `limit_window_seconds` to `WhamUsageWindow` and classify on it rather than on
slot position — e.g. treat `<= 6h` as the session window and anything longer as
weekly, and allow either slot to be absent. Both windows in
`additional_rate_limits` need the same treatment; line 212 reads only
`primary_window` there, and at least one sublimit returns a populated
`secondary_window`.

pedramamini:
Closes #1596

## The bug

`wham/usage` orders its rate-limit windows differently per plan:

| plan | `primary_window` | `secondary_window` |
| --- | --- | --- |
| `team` | `18000` (5h session) | `604800` (weekly) |
| `prolite` | `604800` (**weekly**) | `null` |

`codex-usage-sampler.ts` mapped those slots unconditionally:

```ts
const session = parseWindow(rateLimit.primary_window);
const weekly  = parseWindow(rateLimit.secondary_window);
```

So on any plan of the second shape a 7-day window was stored as `session` and rendered under a `Session (5h)` heading with a reset a week out, while `weekly` came back `undefined` on an account whose *only* limit is weekly. Past the cosmetic mislabel, a consumer scheduling against a "session" reset on such an account would wait up to seven days for it.

`limit_window_seconds` is the field that disambiguates the two, and `WhamUsageWindow` never declared it, so it was discarded before it could be used.

## The fix

- **`limit_window_seconds` is parsed** onto the snapshot as `CodexUsageWindow.windowSeconds` (both store copies, main and renderer).
- **Windows are filed by length, not by slot.** At or under 6h is a session, longer is weekly, whichever slot it arrived in. The ceiling rather than an equality test means a plan reporting a 3h or 6h short window is still a session.
- **Undeclared lengths keep the old positional meaning**, so responses that omit the field behave exactly as they do today.
- **Nothing that parsed is dropped** while a bucket is still free - a window whose preferred bucket is taken spills into the other rather than vanishing from the dashboard.
- **`additional_rate_limits` got the same treatment.** It read only `primary_window` and silently discarded a populated `secondary_window`. Both now render; when a sublimit yields two they are suffixed with their own length (`gpt-5.3-codex (5h)` / `gpt-5.3-codex (7d)`), because the rows are keyed by name and two identical labels would collapse into one. A single-window sublimit keeps its bare name.
- **The dashboard stops asserting "5h".** The session row prints the length the sampler recorded and falls back to `5h` only for a snapshot taken before the field was captured; the weekly row names its own span when it is not seven days. Both go through the canonical `humanizeDuration` rather than a hand-rolled ladder.

## Tests

Six new cases in `codex-usage-sampler.test.ts` (weekly-only `prolite` shape, reversed slots, the undeclared-length fallback, both sublimit windows under distinct names, single-window sublimit unsuffixed) and three in `CodexPlanUsage.test.tsx` covering the row labels. The existing happy-path fixture now carries the real declared durations and asserts `windowSeconds` round-trips.

## Validation

- `tsc -p tsconfig.lint.json` clean; ESLint clean at `--max-warnings=0`; prettier applied.
- `vitest` green across `src/__tests__/main/agents/`, `src/__tests__/renderer/components/UsageDashboard/`, `UsageDashboardModal.test.tsx`, and `src/__tests__/main/ipc/handlers/agents.test.ts`.
- One caveat on local validation: `tsc -p tsconfig.main.json` reports 4 pre-existing errors in `src/main/utils/zip-archive.ts` from `fflate` not being installed in this environment. Untouched by this change; CI is the real gate.

## Summary by CodeRabbit

* **New Features**
  * Usage windows are classified by their declared duration for more accurate session and weekly quota displays.
  * Additional quota limits can show both primary and secondary windows with duration-based labels.
  * Usage dashboards display human-readable window lengths, including custom durations such as 5h and 7d.
* **Bug Fixes**
  * Improved handling of reversed or unusually long quota windows.
  * Legacy quota data without duration details remains supported through fallback labeling and classification.


pedramamini:
Addressed both review findings in e73df074d.

**Spillover breaking bucket semantics** - correct, and it contradicted the premise of the PR. A declared length is now final: the window is filed where its length says it belongs, or not at all. The weekly-plus-monthly case Greptile constructed now yields `weekly` only, rather than rendering the monthly as `Session (30d)`. An *undeclared* window still spills, since there is no declared length for the placement to contradict - it is a guess either way, so it keeps its slot's historical meaning and takes whichever bucket is free otherwise. Added a regression test covering the two-long-windows shape.

**IPC type omitting the duration** - also correct. `global.d.ts` keeps its own structural copy of the snapshot, and `windowSeconds` was missing from it on `session`, `weekly`, and `additionalLimits`. The value crossed the bridge at runtime regardless, but nothing could reach it through the declared contract. Now declared.

PR #1607, opened 2026-09-17
Feat/shared agent spawn library part one

reachrazamair:
## Summary

Part One of the Maestro-lib plan: draw a clear boundary and move the settled, provider-neutral pieces into it with **zero user-visible behavior change**.

* Provider definitions and capability flags → `src/shared/maestro-lib/providers/`
* All 14 per-provider output parsers → `src/shared/maestro-lib/parsers/`
* Env layering, shell-path probing, binary detection, argument building, and SSH remote spawn wrapping → `src/shared/maestro-lib/launch/`
* Every old path is now a one-line `export *` re-export shim, so every existing caller, including desktop IPC handlers, `agent-spawner.ts`, Cue's `cue-spawn-builder`, Cue's executor, chat, and the cross-agent router, keeps compiling and behaving identically with zero import changes. This matches the plan's Part One rule: "leave every existing caller calling what it calls today."
* `src/shared/maestro-lib/` was chosen over a new top-level `packages/` workspace because `src/shared/**` is already included by both `tsconfig.main.json` and `tsconfig.renderer.json`, so no build or workspace configuration changes were needed.

Full breakdown, known gaps, and verification details: `Plans/maestro-lib-part-one-checklist.md`.

## Known Gaps

* `ssh-spawn-wrapper.ts` still depends on `src/main/utils/ssh-remote-resolver.ts` and `src/main/utils/ssh-command-builder.ts`, which are out of scope for this pass.
* The moved parsers still reach back into `src/main/utils/{logger,sentry,terminalFilter}` and `src/main/agents/capability-snapshot.ts`. The "no desktop framework" smoke test still passes, but these dependencies are not yet part of the library proper.
* Run/session/streaming, including the completion-helper contract, is explicitly Part Two/Three scope.

## Rebase Note

This branch's original base predated `rc` and produced two real conflicts:

* `codex-output-parser.ts`: hard-error-status retry classification
* `antigravity-output-parser.ts`: tool-output compaction

Both `rc` fixes were merged into the new library locations and verified byte-identical to `rc`'s content, apart from the required import-path rewrites.

The corresponding unit tests were also run, with **157 tests passing**.

## Verification

* [x] `tsc --noEmit` clean on all three configs
* [x] ESLint clean on the new library directory
* [x] Added `src/shared/maestro-lib/__tests.test.ts`
* [x] Full `vitest run` pre-rebase: **43,461 passed / 6 failed**
* [x] The 1 migration-related regression was identified and fixed
* [x] The other 5 failures were confirmed to be pre-existing on unmodified `rc`
* [x] Targeted re-run of every affected/moved-file test individually after the fix: all green
* [x] Full `vitest run` re-triggered after the fix before merging

## Summary by CodeRabbit

- **Refactor**
  - Consolidated agent definitions, capability handling, command construction, binary detection, shell PATH discovery, SSH launching, output parsing, and usage tracking into a shared library.
  - Existing integrations remain compatible, with no intended user-visible behavior changes.
  - Standardized parsing for supported agent providers, including text, tool activity, usage, sessions, and error reporting.

- **Tests**
  - Updated coverage for relocated functionality.
  - Added verification that shared functionality does not load the desktop framework.
  - Linting and the full Windows test suite pass.


pedramamini:
@reachrazamair Thank you for this contribution. Drawing the maestro-lib boundary is exactly the right first move, and this is a genuinely careful Part One: I diffed all 21 moved files against the merge base and every one is byte-identical apart from the import-path rewrites, the `export *` shims are complete (including the one `default` export on `getShellPath`, which you caught), nothing outside `src/main` reaches into the new library, and both CI matrix legs are green with no conflicts. The rebase handling of the two `rc` parser fixes checks out too.

Two things I would like addressed before this goes in.

### 1. The `no-desktop-framework` smoke test does not actually guard anything

This is the one new test the PR adds, and it is the stated guarantee for the whole boundary, so it matters that it works. It does not: patching `Module._resolveFilename` never fires for ESM `import()` in the Vite-transformed Vitest environment.

I checked this empirically rather than by reading. With the same patch installed, a direct `await import('electron')` inside the test body loads Electron successfully and records **zero** hits, as does importing a `src/main` module with a static `import { ipcMain } from 'electron'`. So `expect(requestedElectron).toEqual([])` passes unconditionally, and would keep passing if `maestro-lib` imported `electron` directly tomorrow.

A static source scan over `src/shared/maestro-lib/**` would be both cheaper and honest here: read the files, assert no import specifier is `electron` or `electron/*`. That has the nice side effect of also covering point 2 below.

### 2. This introduces the first `src/shared/**` to `src/main/**` imports in the repo

There are 13 of them:

- `launch/ssh-spawn-wrapper.ts` to `ssh-remote-resolver`, `ssh-command-builder`, `logger`
- `launch/path-prober.ts` to `sentry`, `logger`, `execFile`
- `launch/agent-args.ts` to `logger`
- `parsers/index.ts`, `parsers/error-patterns.ts` to `logger`
- `parsers/codex-output-parser.ts` to `sentry`
- `parsers/opencode-output-parser.ts`, `parsers/pi-output-parser.ts` to `terminalFilter`
- `parsers/usage-aggregator.ts` to `agents/capability-snapshot`

The count on `rc` today is zero, so this is a new direction of travel rather than more of an existing pattern. The reason it is worth flagging beyond "known gap" is that `src/shared/**` is in `tsconfig.renderer.json`'s include, so the renderer program now compiles `src/main/utils/execFile.ts` (`child_process`), `logger.ts` (`fs`), and `capability-snapshot.ts` (`electron-store`). Nothing in the renderer, web-desktop, or CLI imports `maestro-lib` today, so there is no runtime break, and I am not asking you to untangle those dependencies in this pass. But right now the only thing keeping `child_process` out of a browser bundle is convention, and the number can grow silently between here and Part Two.

An ESLint `no-restricted-imports` rule scoped to `src/shared/**` banning `**/main/**`, with the 13 current edges listed as explicit allowances, would pin the gap at its current size and make each one visible as it gets retired. A static assertion in the rewritten smoke test would do the same job.

### Not blocking, for the record

CodeRabbit's two inline findings, the empty `finalText` completion in `factory-droid-output-parser.ts` and the dropped `msg.part.callID` in `opencode-output-parser.ts`, are both pre-existing behavior. I diffed both files against the merge base and the only changes are import paths, so neither is introduced here. Same for the `captureException` nitpick on `path-prober.ts`. A zero-behavior-change move PR is the wrong place to fix them; worth replying to CodeRabbit to that effect and filing them separately if they are real.

One cosmetic note: the description says "All 14 per-provider output parsers", but the diff moves 10 `*-output-parser.ts` files plus 4 shared contract and utility modules. Worth correcting so the checklist stays accurate for Part Two.

Happy to take another look once the smoke test has teeth. The rest of this is solid work.

PR #1610, opened 2026-09-18
 Maestro-lib Part Two: turn contract

reachrazamair:
## Summary

First of three stacked PRs for maestro-lib Part Two. This PR is documentation-only and adds `Plans/maestro-lib-turn-contract.md`, the shared design contract for turn termination, conversation identity and resume, and token/cost accumulation across desktop chat, the CLI, and Cue's currently divergent implementations.

The document was written and then independently reviewed against the actual source in a second agent pass. Roughly a third of the first draft's citations were found to be incorrect or stale. The review also surfaced several real gaps that the first pass missed, including a resolver signature that could not actually call its dependency, two "open questions" that were already answered elsewhere in the code, and missing termination cases such as Copilot's async post-exit disk reconciliation and spawn-generation/supersession suppression. All of those corrections are incorporated in this version.

A few concrete findings from the research are worth calling out separately:

* The CLI's Codex usage path sums a **cumulative session-total field** across turns, creating the same quadratic overcount bug already documented in `CLAUDE.md` for direct Codex transcript scans. This is more than a simple missing delta correction.
* Desktop's usage delta correction is scoped per **process**, not per session, and is not applied uniformly even within desktop itself. `StdoutHandler.handleLegacyMessage` bypasses the correction.
* Cue has a five-valued run status, `running | completed | failed | timeout | stopped`, one layer above where the first research pass looked in `cue-run-manager.ts`. This means the initial claim that Cue has no stop concept was incorrect.
* The document's **Open Questions** section contains the 6 issues that remain genuinely unresolved rather than treating uncertain behavior as settled.

## Test Plan

* [x] N/A. Documentation-only PR with no code changes.
* [x] Subsequent PRs implement the contract through the streaming layer and desktop chat migration, with their respective test coverage.

## Summary by CodeRabbit

* **New Features**
  * Added shared support for agent definitions, capabilities, launch configuration, SSH execution, shell-path discovery, output parsing, error handling, and usage aggregation.
  * Added normalized parsing for multiple AI agent providers, including streaming events, results, errors, sessions, tools, and usage data.
  * Added a shared contract covering turn outcomes, resume behavior, conversation identity, usage accumulation, and streaming events.

* **Documentation**
  * Added migration checklist and shared turn-handling design documentation.

* **Tests**
  * Added smoke tests validating shared-library exports and desktop-framework independence.


pedramamini:
@reachrazamair Thanks for this. The turn contract doc is genuinely good work, and the "second agent pass over the citations" approach shows: I spot-checked the load-bearing claims against the tree and they hold up.

Verified independently:

- Every `src/...` path the document cites exists on the branch. The only backticked symbols that resolve to nothing are `ResumeMode`, `TurnFacts`, and `UsageAccumulator`, which is correct since they are what #1611 introduces.
- The headline CLI Codex finding is real. `extractUsage` reads `payload.info.total_token_usage` (`codex-output-parser.ts:508`), which is a running session total, and `agent-spawner.ts:1022` feeds it into `mergeUsageStats`, which *adds* into the accumulator on every `token_count` event. That is the quadratic overcount CLAUDE.md already documents, reached by a different route.
- `CueRunStatus` is five-valued at `src/shared/cue/contracts.ts:390`, so the correction to the first pass was right.
- `StdoutHandler.handleLegacyMessage` exists and does bypass the delta correction.
- The moved files are import-path rewrites only. I diffed `ssh-spawn-wrapper`, `agent-args`, `codex-output-parser`, `usage-aggregator`, and `capabilities` against their `rc` originals and nothing but the specifiers changed.

One thing I would like fixed before this merges, plus two small ones.

**1. The base branch (the main ask)**

This PR targets `rc`, but it carries all five commits from #1607 on top of the single doc commit. As it stands, merging this into `rc` merges Part One with it, and the 68-file diff buries the one file the PR is actually about.

Could you retarget the base to `feat/shared-agent-spawn-library-part-one`? GitHub will recompute the diff down to just `Plans/maestro-lib-turn-contract.md`, which is what the description promises and what makes this reviewable on its own. Merging #1607 first works equally well if you would rather land the stack in order. Same applies to #1611 and #1612, which look to be stacked the same way.

While you are in there, the Test Plan's "Documentation-only PR with no code changes" is accurate about your intent but not about what would currently land, so it is worth a line either way.

Good news: `mergeable` is clean, so there are no conflicts to rebase away. This is purely a base-pointer change.

**2. Checklist contradicts itself on macOS**

CodeRabbit is right on this one. `Plans/maestro-lib-part-one-checklist.md:18` marks "verify the no-desktop-framework smoke test, Windows behavior, and macOS binary lookup" complete, while line 113 says macOS binary lookup was explicitly *not* verified. Splitting that action item so the macOS half stays unchecked would settle it.

**3. The "none of these import Electron" claim is a little too strong**

Checklist line 68 says the residual `src/main/utils/{logger,sentry,terminalFilter}` reach-backs do not import Electron. `path-prober.ts` calls `captureException`, and `src/main/utils/sentry.ts` does `await import('@sentry/electron/main')` at lines 57, 82, and 116. It is a dynamic import inside a try/catch, so the smoke test still passes and standalone execution still will not crash, but the sentence as written is not quite true. A clause noting the deferred, caught Electron import would make it exact.

**Two bot findings you can safely disregard:**

- Both bots flagged that `IMPORT_SPECIFIER_PATTERN` misses bare `import 'electron'`. You already fixed that in `089f793c7`; the current head has the `|import\s+` branch with a comment explaining why it cannot double-match `import(`. Both reviews ran against a stale SHA. Nothing to do.
- CodeRabbit's Major finding on the Cue SSH paths, and the "fix before merge" merge-risk verdict attached to it, is not yours. `src/main/cue/` is untouched by this PR, and the unresolved-remote-falls-back-to-local behavior in `cue-spawn-builder.ts:245` and `cue-shell-executor.ts:169` is pre-existing on `rc`. It is a real bug (it contradicts the fail-loudly rule in CLAUDE.md, and the CLI's `sshUnresolvedFailure()` is the pattern it should follow), so it deserves its own issue, but it should not gate a refactor that did not touch it.

**On CI:** the red `Test (windows-latest, shard 1/2)` is not yours either. The same four files fail on the `rc` base commit `6b06f5d1d` (`claude-usage-startup`, `codex-usage-startup`, `codexUsageStore`, `agents.test`). Identical set, identical count. Ignore it.

Retarget the base and fix the two checklist lines and I am happy to approve.

PR #1611, opened 2026-09-18
Maestro-lib Part Two: streaming layer (resolveTurnOutcome, BufferedLineReader, UsageAccumulator)

reachrazamair:
## Summary

Second of three stacked PRs, based on `feat/maestro-lib-turn-contract`. This PR adds the three shared primitives defined by the turn contract under `src/shared/maestro-lib/streaming/`. This is pure new library code with no production call sites wired up yet. That integration happens in PR 3.

* **resolveTurnOutcome**: A single pure function that maps `TurnFacts` to one of four outcomes: `completed`, `completed-with-warning`, `interrupted`, or `crashed`. It replaces the divergent completion rules currently duplicated across desktop's `ExitHandler`, the CLI's agent spawner, and Cue's `cue-process-lifecycle`. Behavior is preserved by default. In particular, the `omp`-only empty-answer override remains scoped to `omp` unless explicitly generalized.

* **BufferedLineReader**: Generalizes the inline chunk-to-line buffering currently handled by `StdoutHandler` into a transport-agnostic, pluggable primitive. It supports custom frame extractors for non-newline framing, such as Copilot, plus an opt-in maximum buffer length as a safety net.

* **UsageAccumulator**: A faithful port of `StdoutHandler`'s `normalizeUsageToDelta` logic into an instance-scoped class. It preserves the existing monotonic-delta detection and absolute-usage attachment behavior without redesigning the underlying usage model.

None of these primitives are wired into production in this PR. PR 3 consumes them as part of the desktop chat migration.

## Test Plan

* [x] 33 unit tests across all three primitives, covering every documented divergent case, including Factory Droid's no-done-event flush, `omp`'s silent clean exit and session exclusions, interrupted short-circuiting before the provider is queried, split-across-chunks reassembly, usage delta computation, and non-cumulative detection.
* [x] TypeScript checks clean for main, renderer, and CLI configurations.
* [x] ESLint clean.

## Summary by CodeRabbit

* **New Features**
  * Added shared support for provider-neutral agent definitions, capabilities, output parsing, usage tracking, shell-path discovery, executable detection, and SSH-based launches.
  * Added normalized streaming event handling, turn outcomes, buffered output processing, and usage accumulation for supported providers.

* **Documentation**
  * Added migration checklist and streaming turn-contract documentation.

* **Tests**
  * Expanded coverage for streaming behavior, usage tracking, provider parsing, platform handling, and shared-library boundaries.


pedramamini:
@reachrazamair Thanks for this one, and thanks for splitting the maestro-lib work into a reviewable stack rather than one giant PR. The turn contract doc plus three pure primitives with 38 tests is exactly the right shape for this, and the doc comments carrying the original call sites (`StdoutHandler.ts:431-432`, `ExitHandler.ts:324-353`, `resetOversizedCopilotJsonBuffer`) make the "faithful port, not a redesign" claim checkable instead of asking us to take it on faith. `UsageAccumulator` reads line for line against `normalizeUsageToDelta`, including the cumulative-flag latch and the conditional `absoluteUsage` re-attachment.

Status check before the review notes:

- Both Greptile P1s are already resolved in `35fb8c2`. The cap now runs after extraction on the remainder alone, and the signal-kill fallthrough is closed. The bot comments are stale against current HEAD.
- CodeRabbit's incremental review bailed out and never ran a full pass, so treat its "no actionable comments" as no signal rather than a clean bill.
- No merge conflicts, `rc` merges cleanly.
- The red Windows shard is pre-existing on `rc`. The failing set (`claude-usage-startup`, `codex-usage-startup`, `codexUsageStore`, `agents.test.ts`) is byte-identical to the last `rc` baseline run, so nothing here is yours.

One thing I would like fixed before this goes in, plus a note to carry into PR 3.

**1. `signal !== null` treats `undefined` and `0` as a signal kill (`turn-outcome.ts:148`)**

`TurnFacts.signal` is documented as loosely typed because PTY reports a number and child_process reports a string. The problem is what a PTY reports on a *normal* exit. node-pty types it `signal?: number`, and `PtySpawner.ts:252` forwards that value straight through to `emitter.emit('exit', sessionId, exitCode, signal)`. So on a clean PTY exit the value arriving at a PR 3 adapter is `undefined`, not `null`, and some platforms report `0`.

Either one makes `killedBySignal` true, and since that flag now bypasses the omp gate for every provider, a clean PTY turn with no captured answer and no result event gets classified `crashed` for everybody. That is the opposite of the "behavior preserved by default" promise, and it would land as a user-visible false crash on ordinary turns.

The type signature says an adapter should normalize, but this resolver is meant to be the one place the rule lives, so I would rather it not depend on every future adapter remembering. Suggest:

```ts
const killedBySignal = facts.signal != null && facts.signal !== 0;
```

Loose `!=` catches `undefined`, and the `!== 0` catches the platform that reports zero for a normal exit. A test for each of those two shapes would lock it in alongside the `SIGKILL` case you already cover.

**2. `isStreamJsonMode` is not modeled in `TurnFacts` (note for PR 3, not a blocker here)**

The original omp condition at `ExitHandler.ts:313` gates on `isStreamJsonMode` as well as `!resultEmitted`. `TurnFacts` has no equivalent field, so rule 4 is slightly wider than production: an omp turn that is not in stream-json mode would resolve `crashed` where today it does not. The other clauses all have homes (`!errorEmitted` is covered by the early returns, `!interrupted` by rule 1, the session exclusions are ported), so this is the one that has nowhere to live. Fine to leave if PR 3 supplies the gate at the call site, but worth a line in the contract doc saying so rather than discovering it during the chat migration.

**Merge order**

1607, 1610, 1611 and 1612 all target `rc` directly rather than stacking their bases, so whichever merges first drags its ancestors in with it. Worth landing them in order (1607, then 1610, 1611, 1612) so each diff stays reviewable in isolation.

Fix the signal check and I am happy to approve. Nice work.

PR #1612, opened 2026-09-18
Maestro-lib Part Two: migrate desktop chat, add turn recordings

reachrazamair:
## Summary

Third of three stacked PRs, based on `feat/maestro-lib-streaming-layer`. This wires desktop chat into the new streaming primitives and adds end-to-end coverage proving the integration works.

**ExitHandler:** Error classification now routes through `resolveTurnOutcome`. The PR includes one approved behavior change: `interrupted` now suppresses the entire `tErrorFromExit` path and SSH pattern matching, rather than only the `omp` override and provisional-error hold. A stopped turn can therefore no longer surface as a crash.

A bug from the first pass of the migration involving the existing `outputParser` was caught by three existing tests and fixed before this PR.

**StdoutHandler:** Usage delta normalization now delegates to `UsageAccumulator`, with one instance per process cached on `ManagedProcess`. `lastUsageTotals` and `usageIsCumulative` remain mirrored on `ManagedProcess` because `plugin-event-listener.ts` and the existing test suite read them directly.

**Turn recordings:** Added 9 end-to-end recordings covering chunked output, interleaved events, cut streams, bad exits with answers, silent resume, stop vs. crash, and related turn-outcome cases. These run the real handlers against hand-authored synthetic transcripts.

One genuine finding came from this coverage: the existing desktop path has an unconditional "no match → generic crash" fallback for any non-zero exit. This means desktop chat has never had the CLI's "captured answer overrides a bad exit code" behavior. This is a CLI-vs-desktop behavior gap identified by the tests, rather than a behavior silently changed by this PR.

## Test Plan

* [x] All 44 existing ExitHandler tests pass unedited, plus 1 new regression test for interrupted precedence.
* [x] All 9 turn recordings pass.
* [x] Full process-manager test directory passes: 341 tests.
* [x] Top-level `process-manager.test.ts` passes: 48 tests.
* [x] Full repository suite passes: 11,415 tests. Initial timeout flakes were unrelated to this branch and passed when run in isolation.
* [x] TypeScript checks are clean for main, renderer, CLI, and lint configurations.

## Summary by CodeRabbit

- **Improvements**
  - Improved support for multiple AI agent providers and their streamed output, including tool activity, sessions, errors, and usage details.
  - Improved command discovery across packaged environments and operating systems.
  - Improved remote SSH execution, including long prompts and custom environment handling.
  - Usage reporting now more accurately tracks cumulative and per-turn token totals.
  - Interrupted runs are distinguished from crashes, reducing misleading error messages.

- **Documentation**
  - Added planning documentation for shared streaming behavior and future conversation-resume support.

- **Tests**
  - Added coverage for streamed output, interruptions, resumed sessions, usage tracking, and cross-platform library behavior.


pedramamini:
@reachrazamair Thanks for this - the three-PR stack is a genuinely careful piece of work, and the turn recordings are the right call. Running real handlers against hand-authored transcripts is what makes the `resolveTurnOutcome` migration reviewable at all, and the inline comments explaining *why* each rule is shaped the way it is (the `!resultMessageSeen` note, the `generalizeEmptyAnswerRule` scope fence, the `absoluteUsage` provenance) are doing real work for whoever touches this next.

I went through both bot reviews and verified each finding against the head commit rather than taking them at face value. A few land, several do not. Details below so you don't spend time chasing the false ones.

## Please address

**1. `turn-recordings.test.ts:103` - the generation stamp is a no-op.**

```ts
nextSpawnGeneration(sessionId, managedProcess);
```

`nextSpawnGeneration(sessionId: string): number` takes one argument (`src/main/process-manager/generation.ts:36`). The second argument is silently dropped, and the returned generation is never stored, so `managedProcess.spawnGeneration` stays `undefined` for every recording. That is benign today only because `isSupersededGeneration` treats `undefined` as current by design - which means the line reads as though it arms the supersession guard while actually arranging for the guard to be bypassed. CodeRabbit's fix is right:

```ts
managedProcess.spawnGeneration = nextSpawnGeneration(sessionId);
```

Worth noting this is invisible to `npm run lint`: no tsconfig includes `src/__tests__` (`tsconfig.lint.json` covers renderer/shared/web/types only), so the excess argument never reaches `tsc`. Not something you could have caught from a green CI.

**2. The SSH branch in `StdoutHandler` still escapes the interrupted rule.**

This PR's headline behavior change is that `interrupted` is authoritative and a stopped turn can no longer surface as a crash. The parser-error branch honors that (`StdoutHandler.ts:425`, `!managedProcess.interrupted`, with a good comment about grok's `stopReason: "cancelled"`). The SSH-error branch 70 lines below does not:

```ts
if (!managedProcess.errorEmitted && managedProcess.sshRemoteId && parsed === null) {
    const sshError = matchSshErrorPattern(line);
```

So on an SSH remote, a non-JSON stdout line matching an SSH pattern after a Stop still emits `agent-error` and arms recovery for a turn the user deliberately abandoned. Narrow (stdout only, non-JSON only), and it is pre-existing rather than introduced here - but this is the PR that establishes the invariant, and an invariant with a hole in it is harder to reason about later than one that was never claimed. Either add the `!managedProcess.interrupted` guard alongside the one above, or add a line to the contract doc saying the pre-resolver stdout paths are explicitly out of scope for Part Two and name the follow-up. `StderrHandler`'s two `agent-error` emits have the same shape if you want to sweep them together.

**3. Precedence between `detectErrorFromExit` and a captured answer.**

`resolveTurnOutcome` checks the provider's exit classification (step 3) before `capturedAnswerText` (step 6), so a classified exit failure returns `crashed` even when a usable answer was captured. Your PR description already names the adjacent gap - desktop's unconditional "no match, generic crash" fallback means desktop chat never had the CLI's "captured answer overrides a bad exit code" behavior - so this is clearly on your radar. It would be worth stating the chosen precedence explicitly in the contract doc's section 1 and adding a recording for the answer-plus-classified-error case, since that is exactly the pair a future reader will assume was an oversight.

## Findings I checked that do not hold

Flagging these so they don't cost you a round trip:

- **Greptile P1, "OMP Results Misclassified"** - does not reproduce. The rule at `turn-outcome.ts:147` requires `!hasAnswer && !facts.resultMessageSeen`, and `ExitHandler` passes `resultMessageSeen: Boolean(managedProcess.resultEmitted)`. The old `!resultEmitted` guard is preserved exactly, and your comment at lines 108-112 already anticipates this case by name.
- **Greptile P2 / CodeRabbit on `buffered-line-reader.ts`** - already fixed at head. `push()` extracts frames first and applies the cap to the remainder alone, with a comment citing the Copilot precedent. Both bots appear to have reviewed an earlier commit.
- **CodeRabbit on the contract doc's Rule 4 session identity** - the resolver does take `context: { providerId, sessionId }` and applies `OMP_EMPTY_ANSWER_SESSION_EXCLUSIONS` against it. The exemptions are implementable as written.

## Scope and CI notes, no action needed

- The 13 `src/main` imports inside `src/shared/maestro-lib` (logger, sentry, execFile, `terminalFilter`, `capability-snapshot`) are a real layering problem for a library whose point is to be desktop-free, and the `sentry` one drags `@sentry/electron/main` onto a shared error path. But they arrived with Part One and the streaming-layer PR - this PR's own delta is 8 files - and your checklist already tracks them as known gaps. Not a blocker here; just make sure they don't survive the stack.
- **Windows shard 1/2 is red, and it is not you.** Run 35389405893 fails 7 tests across `claude-usage-startup`, `codex-usage-startup`, `codexUsageStore`, and `agents.test`. The `rc` baseline (run 35322274949) fails the same 7. Please don't rebase chasing it.

Locally, `src/__tests__/main/process-manager/` is 333 passing and `src/shared/maestro-lib` is 51 passing. No merge conflicts - the branch is clean against `rc`.

Happy to take another look once the test fix is in and you've made a call on items 2 and 3.

